PR #2, opened 2015-05-07
Samples from bigquery-samples-python

elibixby:
PTAL @GoogleCloudPlatform/devrel-samples


bigquery/tox.ini
```diff
@@ -0,0 +1,15 @@
+[tox]
```
tmatsuo: I will merge necessary bit to the top level tox.ini, so maybe you can remove it?

bigquery/dump/load-data-by-POST.py
```diff
@@ -0,0 +1,104 @@
+import sys
```
tmatsuo: Isn't the 'dump' dir used for just copying something outside to the repo?
Do we need this file?

elibixby: I was using dump as a place to put files that were included in docs but I hadn't had time to clean up yet. I can delete it.

proppy:
It's probably a good idea to remove `-` (hyphen) from the name, it's not common in python and make itu un-importable.

tmatsuo:
Ok, LGTM
Just a nit. I would like anyone to wait for the reviewers explicitly say LGTM before merging

elibixby:
Yup. Sorry got ahead of myself.

PR #3, opened 2015-05-07
Added examples from appengine-ndb-snippets

elibixby:
@lesv PTAL 

Ok to delete appengine-ndb-snippets after this is merged?


lesv:
I don't think I had anything to do with appengine-ndb-snippets, so I've assigned to @tmatsuo

tmatsuo:
Yep, let's move them to here and delete the old repos

elibixby:
@tmatsuo Do we need the app.yaml and stuff from the root of that repo? perhaps should throw it in a testing folder for this sample?

tmatsuo:
I'm ok as long as the tests run!

elibixby:
@tmatsuo Merging with note to delete unnecessary files once #5 is complete

PR #4, opened 2015-05-07
Localtesting

elibixby:
Adding code from appengine-local-testing-samples.


elibixby:
Note, appengine-local-testing-samples is safe to delete after this is merged.

jonparrott:
This seems to include the commits for datastore and ndb, but otherwise LGTM.

What about the docs?

elibixby:
Yeah, messed up my git tree. =/

Anyway, did you end up including into the docs from that repo? If so I'll merge this, and then you can open an issue on the old repo to delete and assign me once you have updated the docs. (Think this is a good workflow).

jonparrott:
Yes it's included from the docs. Will you open the issue? I'm swamped atm.

elibixby:
kk

PR #12, opened 2015-05-14
Updated tox.ini for passing PYTHONPATH env var

tmatsuo:


tmatsuo:
@elibixby all green again. I'm merging this since it's a small change. Please add comments if you have opinions. I'll fix them later.

PR #15, opened 2015-05-14
Added TRAVIS* envvar for coveralls

tmatsuo:


tmatsuo:
The first bigquery test often times out (10 minutes). Maybe we should make the test faster.

tmatsuo:
Coverall started working again. @elibixby mind taking a look?

tmatsuo:
@elibixby Do you think of any other envvars we should pass to the commands within tox?

elibixby:
The gcloud environment variables should be set up in a private way so other than that I can't think of anything.

tmatsuo:
Per offline chat, added PROJECT_ID.
PTAL

elibixby:
LGTM.

tmatsuo:
Thx much

elibixby:
Could also consider adding CLOUDSDK_COMPUTE_ZONE, CLOUDSDK_COMPUTE_REGION.

tmatsuo:
Nice, I'll add an issue

PR #16, opened 2015-05-14
Create a unified README for ndb samples

elibixby:
overview/ and transactions/  READMEs' are nearly identical, we should consolidate similar information, and consider adding more detailed readmes to the subdirectories.


datastore/ndb/README.md
```diff
@@ -1,5 +1,58 @@
...
+   gcloud auth login
+   gcloud config set project <your-app-id>
+   ```
+   You don't need a valid app-id to run locally, but will need a valid id to deploy below.
+
+1. Clone this repo.
+
+   ```
+   git clone https://github.com/GoogleCloudPlatform/python-docs-samples.git
+   cd python-docs-samples/datastore/ndb/<sub-directory>
```
tmatsuo: Actually you can not run datastore/ndb/modeling. You can only run the tests.

elibixby: Should we make it runnable, or update the README to make it clear that not all sub-directories can be run (leaning towards the former).

tmatsuo:
The reason why the travis build is failing is that the pull request is coming from another repo. I recommend that you create a branch in this repo and create a pull request from there.

datastore/ndb/README.md
tmatsuo: Making the modeling subdir runnable in a meaningful way is somewhat cumbersome. Personally I'm OK only with unittests.

elibixby:
ahhh travis being weird again... will do

PR #29, opened 2015-05-21
Slightly more readable tox.ini

jonparrott:
This also has the bonus of making the pep8 environment more lightweight.


tmatsuo:
LGTM

PR #31, opened 2015-05-22
Update Discovery Doc Caching

elibixby:


jonparrott:
@elibixby for build_and_update, it's specific to bigquery scopes. Shouldn't we make the scopes an argument?

elibixby:
@jonparrott +1 didn't catch that. Will do that.

jonparrott:
Also, create_scoped isn't required when using build as per the docs [here](https://developers.google.com/identity/protocols/application-default-credentials)

> The build() method takes care of injecting the proper scopes for the given service, although the method create_scoped can be used to do this explicitly.

And per the source code for ``build`` [here](https://github.com/google/google-api-python-client/blob/master/googleapiclient/discovery.py#L273)

jonparrott:
LGTM w/ small nit.

jonparrott:
LGTM, merging

PR #32, opened 2015-05-22
Adds Regression Test By Mocking Input

waprin:
If someone checked out origin/master and ran the samples, they immediately hit a string concatenation error. While an easy fix, I generally think it's a good idea, if possible, to write a failing test that then passes along with the bug fix. This is slightly awkward when getting user input is involved, but here's a pass at it, let me know what you think.

### DIFF
--- a/bigquery/samples/utils.py
+++ b/bigquery/samples/utils.py
@@ -48,3 +48,6 @@ def paging(service, request_func, num_retries=5, **kwargs):
             has_next = False
         yield response
 # [END paging]
+
+def get_input(text):
+    return input(text)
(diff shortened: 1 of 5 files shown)


bigquery/samples/utils.py
```diff
@@ -48,3 +48,6 @@ def paging(service, request_func, num_retries=5, **kwargs):
             has_next = False
         yield response
 # [END paging]
+
+def get_input(text):
+    return input(text)
```
tmatsuo: @jonparrott has an opinion on this!

jonparrott: @waprin is fixing it, and removing get_input all together in favor of directly mocking raw_input.

PR #39, opened 2015-06-18
Move sample using appengine images to this repository, add tests.

jeffmendoza:
Currently here: https://github.com/GoogleCloudPlatform/appengine-images-guestbook-python
From here: https://cloud.google.com/appengine/docs/python/images/usingimages


tmatsuo:
Ah, the secret key files are not available for pull requests from other repos, thus the travis build fails.
You need to create a PR from another branch in the same repo.

jeffmendoza:
ah ok, will do

jeffmendoza:
see #40

PR #44, opened 2015-06-19
Added doc strings to functions with ambiguous args

elibixby:
Please review for style


bigquery/samples/export_data_to_cloud_storage.py
```diff
@@ -21,6 +21,15 @@
 def export_table(service, cloud_storage_path,
                  projectId, datasetId, tableId,
                  num_retries=5):
+    """
+    service: initialized and authorized bigquery
```
jonparrott: Something like this instead:

```
"""Exports a table from bigquery.

Arguments: (or Args:)
    service: ..
    cloud_storage_path: ...

Returns: A x instance representing..

"""
```

jonparrott:
LGTM

Issue #43, opened 2015-06-19
Add 3LO Auth examples for python clients


elibixby:

jerjou:
Note: integration tests can be tricky, since 3LO requires user intervention to authorize tokens.

elibixby:
Is there an easy way to mock 3LO Flow? @jonparrott maybe has an idea?

jonparrott:
The oauth2 decorators in oauth2client are well tested by that library specifically. No need to re-tread that ground, just test that the following:

1. No credentials -> redirect to auth screen
2. Has credentials -> gets through to page

jerjou:
Yeah - how do you set up the has-credentials scenario? Do we just generate a refresh token for one of our accounts and put that in the repo? That sounds all kinds of sketch :)

jonparrott:
Fake one, and never *actually* use it.

jerjou:
Trouble is, if we want the test to be an integration test, we'll want the api call to actually be authenticated, so we can verify we're getting the expected (successful) response.
For unit tests, of course, we can mock to our heart's content.

jonparrott:
:/ well, hmm, we should either see if we can obtain a test account specifically for this, or only do the int test for 2lo.

elibixby:
Internally we have a way to get dummy accounts that don't last. Which I have been using to test my 3LO stuff. But this is a larger problem with Travis being on the other side of SSO.

jonparrott:
We'll have a good 3lo sample once the [Flask 3LO Helper](https://github.com/google/oauth2client/pull/226) is merged.

jonparrott:
Closing this as obsolete.

PR #53, opened 2015-07-17
Move bigquery 3lo sample on GAE to github.

jerjou:
Also, update it to use webapp2.

### DIFF
--- a/bigquery/samples/appengine_auth/README.md
+++ b/bigquery/samples/appengine_auth/README.md
@@ -1,0 +1,19 @@
+## Google App Engine accessing BigQuery using OAuth2
+
+This sample demonstrates [authenticating to BigQuery in App Engine using OAuth2](https://cloud.google.com/bigquery/authorization).
+
+### Setup
+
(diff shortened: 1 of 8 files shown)


elibixby:
Can you replace the client secrets with an encrypted version that talks to a real project for integration tests? 

Also once we have an oauth2 decorator for flask we should switch this to flask (or at least add a flask version)

Otherwise LGTM.

jerjou:
Yeah - I thought about doing that, but not sure if it would work. ie the client secrets need a 3LO flow to work, and I'm not sure how I would get an access token from the confirmation url within the context of the test, and provide it to the oauth decorator.

I _could_ mock the decorator such that it provides an ADC-authenticated http object to do the actual api calls, but it seems like it'd be a lot of trouble, for not much value.

Open to suggestions..

elibixby:
Yeah. I suggest that we merge this and roll this discussion into issue #43

jerjou:
Merging...

PR #55, opened 2015-07-22
Add storagetransfer samples

shun-fan:


jonparrott:
Travis failure seems to be due to a different test. This LGTM, as I reviewed it before you pushed it here.

PR #60, opened 2015-07-28
Adding datastore blog samples and tests.

jonparrott:
More importantly, I've modified how tests are run:

* Separate environments for GAE and regular python samples.
* Coverage is now done at the travis level instead of at the tox level.


elibixby:
LGTM

jonparrott:
Thanks, I'm still not 100% sure what I did with coveralls is correct. But I'll go ahead and merge and we can figure it out later.

PR #64, opened 2015-08-02
Misplaced datastore_example_test comment

Anish2:
I noticed that the datastore_example_test end comment seems to be accidentally misplaced in the file.


jonparrott:
Thanks for the contribution, but I believe this is correct as is. The tags are used to embed the code in [this page](https://cloud.google.com/appengine/docs/python/tools/localunittesting?hl=en#Python_Writing_Datastore_and_memcache_tests), and it seems they are correct in context.

PR #66, opened 2015-08-13
Make authn sample more uniform with the java one.

jerjou:


bigquery/samples/utils.py
```diff
@@ -21,9 +21,16 @@ def get_service():
     from oauth2client.client import GoogleCredentials
 
     credentials = GoogleCredentials.get_application_default()
+
```
jonparrott: iirc, this is handled automatically by the library. https://github.com/google/google-api-python-client/blob/master/googleapiclient/discovery.py#L272

jerjou: Good point.

waprin:
Thanks for pointing out that out Jon :) LGTM

PR #67, opened 2015-08-14
Fill in copyright information.

jerjou:


jonparrott:
I don't think we're supposed to modify that part of the license. I remember getting yelled at for the same thing.

jerjou:
Well, that's silly.

jonparrott:
I think the argument is that this is the full license stuff from Apache,
but to *apply* the license we have to actually put the header in our source
code.

PR #69, opened 2015-08-25
Compute Autoscaler demo

kgrygiel:
This application is a very simple web server, which responds to get requests by consuming some CPU cycles to demonstrate how Autoscaler can scale web frontends based on CPU utilization.
It is part of a tutorial video we've recently recorded. Link to the video will follow.

### DIFF
--- a/compute/autoscaler/demo/frontend.py
+++ b/compute/autoscaler/demo/frontend.py
@@ -1,0 +1,107 @@
+# Copyright 2015, Google, Inc.
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+"""A simple web server which responds to HTTP GET requests by consuming CPU.
+This binary runs in a GCE VM. It serves HTTP requests on port 80. Every request
+with path '/service' consumes 1 core-second of CPU time, with the timeout of
+5 (walltime) seconds. The purpose of this application is to demonstrate how
+Google Compute Engine Autoscaler can scale a web frontend server based on CPU
+utilization.
+The original version of this file is available here:
+https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/compute/
+    autoscaler/demo/tests/test_frontend.py
+"""
+
+import BaseHTTPServer
+from multiprocessing import Process
+import os
+import SocketServer
+import sys
+import time
+
+REQUEST_CPUTIME_SEC = 1.0
+REQUEST_TIMEOUT_SEC = 5.0
+
+
+class CpuBurner(object):
+    def get_walltime(self):
+        return time.time()
+
+    def get_user_cputime(self):
+        return os.times()[0]
+
+    def busy_wait(self):
+        for _ in xrange(100000):
+            pass
+
+    def burn_cpu(self):
+        """Consume REQUEST_CPUTIME_SEC core seconds.
+        This method consumes REQUEST_CPUTIME_SEC core seconds. If unable to
+        complete within REQUEST_TIMEOUT_SEC walltime seconds, it times out and
+        terminates the process.
+        """
+        start_walltime_sec = self.get_walltime()
+        start_cputime_sec = self.get_user_cputime()
+        while (self.get_user_cputime() <
+               start_cputime_sec + REQUEST_CPUTIME_SEC):
+            self.busy_wait()
+            if (self.get_walltime() >
+                    start_walltime_sec + REQUEST_TIMEOUT_SEC):
+                sys.exit(1)
+
+    def handle_http_request(self):
+        """Process a request to consume CPU and produce an HTTP response."""
+        start_time = self.get_walltime()
+        p = Process(target=self.burn_cpu)  # Run in a separate process.
+        p.start()
+        # Force kill after timeout + 1 sec.
+        p.join(timeout=REQUEST_TIMEOUT_SEC + 1)
+        if p.is_alive():
+            p.terminate()
+        if p.exitcode != 0:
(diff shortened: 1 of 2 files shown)


compute/autoscaler/demo/frontend.py
```diff
@@ -0,0 +1,107 @@
...
+import sys
+import thread
+import time
+
+
+class CpuBurner(object):
+  REQUEST_CPUTIME_SEC = 1.0
+  REQUEST_TIMEOUT_SEC = 5.0
+
+  def GetWallTime(self):
```
jonparrott: please follow pep8 standards and use ``snake_case`` for functions and methods.

compute/autoscaler/demo/frontend.py
```diff
@@ -0,0 +1,107 @@
...
+    self.wfile.close()
+
+
+class DemoHttpServer(SocketServer.ThreadingMixIn,
+                     BaseHTTPServer.HTTPServer):
+  pass
+
+
+if __name__ == "__main__":
+  httpd = DemoHttpServer(("", 80), DemoRequestHandler)
```
jonparrott: Any reason to use BaseHTTPServer instead of a framework like Flask? Your sample app could be quite a bit smaller.

jonparrott:
Don't sweat travis at the moment, it fails for outside pull requests because of our encrypted key.

compute/autoscaler/demo/frontend.py
kgrygiel: done

compute/autoscaler/demo/frontend.py
kgrygiel: We prefer to use only what's available on the standard gce image. Otherwise we would either need to create our own image (which makes the demo much more complex) or install this extra stuff in the start-up script on every VM: in this case I guess first pip and then Flask. This also make things more complex because it slows down VM spin-up and interferes with utlization signal which may make the autoscaling demo less deterministic. So we figured it's probably best to just use standard tools.

jonparrott: What instructions do you have now for deploying the instance?

It seems you're making at tradeoff between readability of this sample and readability of the instructions. This sample is very surprising- it's quite unusual to see BaseHTTPServer used directly.

jonparrott:
Please squash commits.

compute/autoscaler/demo/frontend.py
tmatsuo: > We prefer to use only what's available on the standard gce image.

Everyone needs to install something on your VMs. Samples for autoscaler should reflect what majority of people are doing.

How about to create a simple docker image with a lightweight OS and only flask installed, then upload that image to gcr.io? Is that still too slow for you?

kgrygiel: The instructions we have for deploying the instance is basically to set the following startup script: "wget https://storage.googleapis.com/autoscaler-demo/frontend.py; sudo python frontend&" (see slide 23 of the video script: https://docs.google.com/presentation/d/1BfKmGQAuV7Qq0vj4wHntrHcZxECoaJaEHTuW5COdMFM/edit#slide=id.gb796329e8_1_0).

I have to confess that we've already done the video recording and we're at the post-processing stage, which gives us little flexibility with the actual setup of the demo. We're still free to change the code sample, but we can't add additional steps such as using custom images.

Our goal was to keep the demo as short and self contained as possible to convince users that the basic setup can be done in a very few steps. You're totally right there is a trade-off between the simplicity of the code and the setup.

As for the code: I'd love to use something higher level than BaseHTTPServer if it lived in a standard module..

tmatsuo: I got it. Although I'd still love to have another sample with using gcr.io, I'm OK with proceeding.

Here is a result of my quick sanity check for pep8.
tmatsuo@tmatsuo:~/work/python-docs-samples$ tox -e pep8
pep8 runtests: PYTHONHASHSEED='3745109524'
pep8 runtests: commands[0] | flake8 --max-complexity=10 --import-order-style=google
./compute/__init__.py:1:1: W391 blank line at end of file
./compute/autoscaler/__init__.py:1:1: W391 blank line at end of file
./compute/autoscaler/demo/__init__.py:1:1: W391 blank line at end of file
./compute/autoscaler/demo/tests/__init__.py:1:1: W391 blank line at end of file
./compute/autoscaler/demo/frontend.py:14:80: E501 line too long (80 > 79 characters)
./compute/autoscaler/demo/frontend.py:30:1: F401 'thread' imported but unused
./compute/autoscaler/demo/frontend.py:59:80: E501 line too long (80 > 79 characters)

Can you address them, and squash the commits?

jonparrott: If you're going to do

```
wget https://storage.googleapis.com/autoscaler-demo/frontend.py; sudo python frontend&
```

Why not do:

```
curl https://raw.githubusercontent.com/python-docs-samples/master/compute/autoscaler/install.sh | sudo bash
```

Or use a shortened URL:

```
curl https://goo.gl/xxxxx | sudo bash
```

The startup.sh could contain:

```
easy_install flask
wget ... # python script
python main.py
```

It's one extra layer of indirection - it'll make you slides cleaner and your sample more clear.

kgrygiel: Well, I would need to install easy_install too..Maybe not a big deal, but every such dependency on an external site makes the demo less deterministic (plus, not sure if that matters, puts load on these sites on every VM spin-up). Indeed, the proper way of doing this would be to download the dependencies once and create an image.
I certainly agree that your suggestions would make the code cleaner and I wish we had this discussion before, as this is something we can't change after recording was done (the speaker explicitly describes what the startup script does).

kgrygiel:
Done.

compute/autoscaler/demo/frontend.py
jonparrott: easy_install comes with python, I'm almost certain, but yes, you would have to reach out to another package manager.

Videos have a very short lifespan where code samples tend to stick around a lot longer, so I'm not overjoyed that the video is blocking us from improving the sample.

As long as it passed pep8 and has tests, I'll merge.

kgrygiel: Thanks, just let me fix one thing I missed and I'll let you know when it's ready.

kgrygiel: OK, pep8 & unit test pass for me.

PR #70, opened 2015-08-31
Test in both python 2 and 3

jerjou:


jonparrott:
You add the runtimes, but you'll need to modify tox.ini to run 3.4 tests as well.

jerjou:
Oh thanks.
Also - turns out the gcloud sdk [doesn't support python 3](https://travis-ci.org/GoogleCloudPlatform/python-docs-samples/jobs/78102352) anyway..

jonparrott:
The pure-python tests can and should be run under python 3. I'd intended to do this eventually.

If you want, you can give it a shot, if not, can you file an issue for it?

jerjou:
[Issue filed](https://github.com/GoogleCloudPlatform/python-docs-samples/issues/71)

PR #72, opened 2015-09-02
Move annotation for greater brevity.

jerjou:
Also add comments for more explanation of what's going on with the service
construction.


jerjou:
Actually, I think I want to do something different... closing for now.

PR #73, opened 2015-09-02
Add annotation for just the list_bucket part

jerjou:
Also add comments for more explanation of what's going on with the service
construction.


jonparrott:
LGTM.

PR #78, opened 2015-09-04
Move bigquery 'getting started' sample into github.

jerjou:
This sample queries the shakespeare dataset via bigquery.

### DIFF
--- a/bigquery/samples/getting_started.py
+++ b/bigquery/samples/getting_started.py
@@ -1,0 +1,66 @@
+#!/usr/bin/env python
+
+# Copyright 2015, Google, Inc.
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Sample for making BigQuery queries using the python sdk.
+
+This is a command-line script that queries a public shakespeare dataset, and
+displays the 10 of Shakespeare's works with the greatest number of distinct
+words.
+"""
+# [START all]
+from apiclient.discovery import build
+from apiclient.errors import HttpError
+
+from oauth2client.client import GoogleCredentials
+
+
+def main(project_id):
+    # [START build_service]
+    # Grab the application's default credentials from the environment.
+    credentials = GoogleCredentials.get_application_default()
+    # Construct the service object for interacting with the BigQuery API.
+    bigquery_service = build('bigquery', 'v2', credentials=credentials)
+    # [END build_service]
+
+    try:
+        # [START run_query]
+        query_request = bigquery_service.jobs()
+        query_data = {
+            'query': ('SELECT TOP(corpus, 10) as title, '
+                      'COUNT(*) as unique_words '
+                      'FROM [publicdata:samples.shakespeare];')
+            }
+
+        query_response = query_request.query(
+            projectId=project_id,
+            body=query_data).execute()
+        # [END run_query]
+
+        # [START print_results]
+        print('Query Results:')
+        for row in query_response['rows']:
+            print('\t'.join(field['v'] for field in row['f']))
+        # [END print_results]
+
+    except HttpError as err:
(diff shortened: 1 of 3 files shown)


jonparrott:
Squash commits?

jerjou:
I figure the two of them are unrelated enough that they deserve separate commits

jonparrott:
Fair enough.

jonparrott:
Where is this sample used in the docs? It seems like the other samples already cover a lot of this?

bigquery/samples/getting_started.py
```diff
@@ -0,0 +1,65 @@
...
+from pprint import pprint
+
+from apiclient.discovery import build
+from apiclient.errors import HttpError
+
+from oauth2client.client import GoogleCredentials
+
+
+# The id of the project to run queries under.
+PROJECT_ID = 'bigquery-devrel-samples'
```
jonparrott: prompt the user and pass this in as an argument to main(), like the [other samples](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/bigquery/samples/sync_query.py#L50).

jerjou:
This sample is from here:

https://cloud.google.com/bigquery/bigquery-api-quickstart

I also realize that I have to update the docstring, since I changed the dataset queried from the wikipedia revisions to shakespeare.

bigquery/samples/getting_started.py
```diff
@@ -0,0 +1,65 @@
...
+        # [END run_query]
+
+        # [START print_results]
+        print('Query Results:')
+        for row in query_response['rows']:
+            print('\t'.join(field['v'] for field in row['f']))
+        # [END print_results]
+
+    except HttpError as err:
+        print 'Error:', pprint(err.content)
```
jonparrott: This is causing a python 3 incompatibility, change it to:

```
print('Error: {}'.format(err.content))
raise err
```

bigquery/samples/getting_started.py
```diff
@@ -0,0 +1,65 @@
...
+from oauth2client.client import GoogleCredentials
+
+
+# The id of the project to run queries under.
+PROJECT_ID = 'bigquery-devrel-samples'
+
+
+def main():
+    # [START build_service]
+    # Grab the application's default credentials from the environment.
```
jonparrott: Add a link to the ADC page.

bigquery/samples/getting_started.py
```diff
@@ -0,0 +1,65 @@
...
+
+from oauth2client.client import GoogleCredentials
+
+
+# The id of the project to run queries under.
+PROJECT_ID = 'bigquery-devrel-samples'
+
+
+def main():
+    # [START build_service]
```
jonparrott: Any reason not to use [bigquery.samples.utils.get_service()](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/bigquery/samples/utils.py#L16)?

bigquery/tests/test_getting_started.py
```diff
@@ -0,0 +1,43 @@
+# Copyright 2015, Google, Inc.
```
jonparrott: Thanks for adding tests. :)

bigquery/samples/getting_started.py
jerjou: I feel like that's better left in the doc that uses this sample - urls take up a lot of room, and warrant some text around it explaining why you'd want to go there. eg if I were to add a url in this comment, I'd also want to say something like, "For more information on application default credentials, go to: http...", and that's another two lines of code. If I could just add a footnote to ADC, or link it, that'd work, but since I can't, I'd rather keep the code brief and to the point.

bigquery/samples/getting_started.py
jerjou: Mainly just keeping the sample self-contained.
If I were writing an app, factoring into utils would make sense; but for the purposes of explanation, it's nice to have everything relevant in one place.

bigquery/samples/getting_started.py
```diff
@@ -0,0 +1,64 @@
...
+    try:
+        # [START run_query]
+        query_request = bigquery_service.jobs()
+        query_data = {
+            'query': ('SELECT TOP(corpus, 10) as title, '
+                      'COUNT(*) as unique_words '
+                      'FROM [publicdata:samples.shakespeare];')
+            }
+
+        query_response = query_request.query(projectId=project_id,
```
jonparrott: nit: I consider this poor style, I prefer:

```
query_response = query_request.query(
    projectId=project_id,
    body=query_data).execute()
```

jonparrott:
LGTM with tiny nit. You can merge yourself or ping me when you're ready.

PR #82, opened 2015-09-14
Update README To Reflect Current GAE Download Instructions

waprin:


jonparrott:
LGTM, thanks for this.

PR #83, opened 2015-09-14
Switching app engine tests to webtest and consolidating testing utils…

jonparrott:
… into AppEngineTestbedCase

### DIFF
--- a/appengine/images/tests/test_guestbook.py
+++ b/appengine/images/tests/test_guestbook.py
@@ -15,17 +15,18 @@
 # from the app main.py
 from appengine.images import main
 import mock
-from tests import DatastoreTestbedCase
+from tests import AppEngineTestbedCase
+import webtest
 
-import webapp2
 
-
-class TestHandlers(DatastoreTestbedCase):
+class TestHandlers(AppEngineTestbedCase):
     def setUp(self):
         super(TestHandlers, self).setUp()
 
         # Workaround for other tests clobbering our Greeting model.
         reload(main)
+
+        self.app = webtest.TestApp(main.app)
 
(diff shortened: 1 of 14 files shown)


jonparrott:
Travis died on pep8, should be fixed. (man, import orders are annoying).

appengine/images/tests/test_guestbook.py
```diff
@@ -66,30 +60,19 @@ def test_img(self):
...
 
         self.assertEqual(response.status_int, 200)
 
     def test_img_missing(self):
         # Bogus image id, should get error
-        request = webapp2.Request.blank('/img?img_id=123')
-        response = request.get_response(main.app)
-
-        self.assertEqual(response.status_int, 500)
+        self.app.get('/img?img_id=123', status=500)
```
waprin: Nothing to do with your commit, but we might want to use named constants for HTTP codes.

waprin:
LGTM, one extraneous comment. I don't understand the benefits of switching to web request, maybe throw a blurb into the commit message?

appengine/images/tests/test_guestbook.py
jonparrott: I do not feel strongly about this, but if you do, feel free to file a bug and we'll follow up with it.

jonparrott:
Reduces the amount of code needed to test these samples, and will make it easier for us to reach full coverage later.

appengine/images/tests/test_guestbook.py
waprin: Fair, oogleCloudPlatform/python-docs-samples/issues/85

PR #84, opened 2015-09-14
Removed an example for caching discovery doc.

### LINKED ISSUE #75: Can discoverydoccaching be deprecated?
@tmatsuo in light of your contribution [here](https://github.com/google/google-api-python-client/pull/127), can we remove the `discoverydoccaching` sample?

tmatsuo:
This is in favour of the google-api-python-client 1.4.2 release. The new
version of that client library automatically cache the discovery docs,
so there is no need to have this example any more.

Fixes #75.


jonparrott:
LGTM, merging.

PR #89, opened 2015-09-14
Rewriting bigquery samples to use proper argparse and main execution

### LINKED ISSUE #80: All console-runnable samples should use better main pattern.
Instead of:

```
def main(argv):
    args = parser.parse_args(argv[1:])

if __name__ == '__main__':
    main(sys.argv)
```

It should be

```
def descriptive_function_name(arg1, arg2, arg3):
    ...

if __name__ == '__main__':
    args = parser.parse_args()
    descriptive_function_name(args.arg1, args.arg2, args.arg3)
```

jonparrott:
Resolves #80.


bigquery/samples/export_data_to_cloud_storage.py
```diff
@@ -11,16 +11,15 @@
...
-from bigquery.samples.utils import get_service
-from bigquery.samples.utils import poll_job
-from six.moves import input
+from utils import get_service, poll_job
 
 
 # [START export_table]
 def export_table(service, cloud_storage_path,
-                 projectId, datasetId, tableId,
+                 project_id, dataset_id, table_id,
```
elibixby: Just to give my thinking behind this. Camel case matches the field names in the json objects. So you could make these kwargs and then do something like:
```
tableRef = fetch_some_bigquery_object()['tableReference']
export_table(service, cloud_storage, **tableRef)
```
Probably not worth inconsistent/camel case naming, but something to consider.

jonparrott: True, but I'd rather follow python convention where possible:

> Names that are visible to the user as public parts of the API should follow conventions that reflect usage rather than implementation.
> ...
> mixedCase is allowed only in contexts where that's already the prevailing style (e.g. threading.py), to retain backwards compatibility.

jonparrott:
gotta figure out what's up with the tests.

jerjou:
I like.
LGTM after tests pass, and Eli gives his stamp of approval.

jonparrott:
@elibixby, @jerjou so the issue with tests is murky.

We're expecting users to run these scripts as such:

```
python bigquery/samples/async_query.py [...]
```

However, this causes the file to have the `__name__` of `"__main__"` which means that relative imports do not work, and importing sibling modules only works because of some slightly sneaky behavior. So, right now I have

```
import utils
```

in the bigquery samples, which works if you execute the script directly (because it's a sibling module), but breaks when the tests import this script because it's expected a top-level module named ``utils``.

The correct way to fix this is to change the way users execute the file to:

```
python -m bigquery.samples.async_query [args]
```

Is this acceptable for you guys? I can't come up with very many other clever ways of doing things without black magic.

jonparrott:
This is a recurring problem in a couple of samples. It appears we've been making it impossible for users to execute these the normal way. We have two options here:

1. Tell users to use ``python -m`` to run the samples. We can do this in top-level readmes.
2. Remove dependent modules and make all sample free-standing. So ``bigquery.utils`` would go away and ``get_service``, ``poll_job``, etc. would be moved into individual files.

I don't have a horse in this race. But I have two conflicting feelings:
* I don't like code duplication.
* These are samples and each sample should be self-contained.

/cc @jerjou @elibixby @tmatsuo @SurferJeffAtGoogle for voting.

elibixby:
My vote goes to `python -m`. Looking at the docs I don't think this flag is terribly opaque, and I think it's also good for users to have one place to look for helpful `utils`

jonparrott:
Will wait for the remaining cc'd folks to cast their vote before deciding.

tmatsuo:
Does the following work:

```python
if __name__ == '__main__' and __package__ is None:
    import bigquery.samples
    __package__ = 'bigquery.samples'

from . import utils
```
? Although it seems hacky.

I'm fine with `-m` option.

jerjou:
I am conflicted as well. It does seem the `python -m` flag is the more maintainable of the two, but it seems we'd have to audit our docs to make sure our instructions include it.

...OTOH, these are samples and they'd probably have nominally more utility as such if they were self-contained. And the functions in question - at least here - aren't that many lines of code, so duplication wouldn't be _that_ bad..

I think I'm leaning more towards self-contained samples, but I'd really be fine either way (well, maybe not _fine_, but about the same degree of resigned)

jonparrott:
Ha, @jerjou make a decisive vote!

@tmatsuo I'd rather not have hacky code at the top of our samples just for our benefit.

These samples are for our users, and tests are there to ensure things work for our users - which is why I'm okay with code duplication as long as our tests are good. If duplicating code makes testing easier and makes it easier for users to run our samples, then I think that's a win.

tmatsuo:
I'm fine with either of them, but I'm still thinking there should be a way to handle this (correctly and not a hacky way). Maybe I'm wrong.

jonparrott:
Afaik, there isn't. This is just a quirk of python and it hurts in our case. Most "executable" scripts should just add a simple bootstrap script to their distribution:

```
from somepackage import run

if __name__=='__main__':
    run()
```

See [psq](https://github.com/GoogleCloudPlatform/psq/blob/master/psq/psqworker.py) for a slightly more advanced example, but same principle.

I'm personally slightly more in favor of dropping ``utils`` and improving test coverage to ensure that they still work.

jerjou:
I vote for self-contained - it makes for better samples, when you can have confidence that what you're seeing is all you need to write.

tmatsuo:
If I had to choose one of them, I choose dups.

jonparrott:
dups: jerjou, tmatsuo
m: elibixb

waprin:
I vote for -m. It seems designed for this exact purpose and so is the most Pythonic, even if Python is occasionally super hacky. I don't like code duplication from a maintenance standpoint. I don't think it's critical that samples are self-contained, as long as they are relatively easy to copy/paste and get working.  I don't mind either way, but my decisive vote is -m.

jonparrott:
My vote is for duplicating these handful of functions. Our users should not have to understand our repository or workflow to use our samples.

For duplication: jerjou, tmatsuo, jowayne
For -m: elibixby, waprin

I'll start down the dup path, but if anyone else votes I can undo it.

jonparrott:
Alright tests are green, give it another review @jerjou and @elibixby.

bigquery/samples/utils.py
```diff
@@ -1,66 +0,0 @@
...
-    #   https://developers.google.com/api-client-library/python/apis/
-    #
-    # The build() method also takes care of injecting the necessary scopes
-    # to access the given service, when using application default credentials.
-    return build('bigquery', 'v2', credentials=credentials)
-# [END get_service]
-
-
-# [START poll_job]
-def poll_job(service, projectId, jobId, interval=5.0, num_retries=5):
```
elibixby: I believe this region is actually pulled into the docs somewhere, we should double check to make sure this doesn't break any imports.

jonparrott: It will, we'll need to update any docs that reference ``utils.py`` to point to the actual thing. @jerjou volunteered to do that.

jonparrott:
@elibixby waiting for your final LGTM.

jerjou:
LGTM
Thanks for all the little fixes :)

PR #97, opened 2015-09-17
Renaming 'samples' -> 'api'.

jonparrott:


tmatsuo:
LGTM.

BTW tests are failing, maybe because of concurrent run. Maybe you can do either of
1. suppressing concurrent builds on travis UI
2. using random datastore keys
3. introducing retry

jonparrott:
Probably best to disable concurrent builds. Thanks, will merge.

PR #101, opened 2015-09-18
More consistency and standardization improvements.

jonparrott:
* Adding argparse/docstring to cloud storage and monitoring samples.
* Updating readmes for BigQuery, Monitoring, Cloud Storage, and the Blog.
* Removing all non-appengine requirements.txt files and consolidating into a single top-level requirements.txt. This makes it easier for us to verify that the samples work with the latest version of all dependent libraries.
* Moved test dependences out of tox.ini and into requirements-dev.txt, as per standard practice.


monitoring/api/auth.py
```diff
@@ -53,42 +35,39 @@
...
 
 def main(project_name):
-    # Create and return the CloudMonitoring service object.
-    service = build('cloudmonitoring', 'v2beta2',
-                    credentials=GoogleCredentials.get_application_default())
+    credentials = GoogleCredentials.get_application_default()
+    service = build('cloudmonitoring', 'v2beta2', credentials=credentials)
 
-    ListTimeseries(project_name, service)
+    list_timeseries(project_name, service)
```
tmatsuo: Apparently you need to swap project_name and service here.

tmatsuo:
LGTM except for the simple fix

monitoring/api/auth.py
jonparrott: Good catch, will do.

jonparrott:
Fixed, merging.

PR #103, opened 2015-09-18
Making travis only test master branch to reduce thrashing.

jonparrott:
/cc @jerjou @tmatsuo


jonparrott:
Merging now to prevent thrashing on our new samples migrations.

PR #104, opened 2015-09-18
Adding standard module docstrings and argparse description to transfe…

jonparrott:
…r samples, and removing usage of logging for consistency with other samples


shun-fan:
LGTM

PR #114, opened 2015-09-23
Remove datastore namespaces

waprin:
There is confusion. Namespace is a specific App Engine API, NOT using datastore to put things in separate entity groups. This sample should have just been deleted, and all references to it point to the multitenancy example.


jonparrott:
Datastore *does* have namespaces separate from app engine's multitenancy. Multitenancy just handles setting the namespace for you.

PR #128, opened 2015-10-15
Update python docstring.

jerjou:


jonparrott:
Typically with small changes like this you can go ahead and merge, just add me as FYI.

PR #130, opened 2015-10-16
Specify source format, so json uploads work.

### LINKED ISSUE #129: "Cannot load CSV data with a nested schema" with a JSON file in load_data_by_post
Hello

With the new load_data_by_post, I'm not able to upload a JSON file and I have this error "Cannot load CSV data with a nested schema".

Sounds like the job expect a CSV not a JSON file so do I need to do something special ?

And I tested on the web client with my data/schema and it works fine, here the [schema](https://gist.github.com/Falydoor/200477dee839c84afc2a) and the [data](https://gist.github.com/Falydoor/cc4fc2a28a91e86cefb5) in case you wanna test.

Regards

jerjou:
Should fix #129

### DIFF
--- a/bigquery/api/load_data_by_post.py
+++ b/bigquery/api/load_data_by_post.py
@@ -50,26 +50,35 @@ def load_data(schema_path, data_path, project_id, dataset_id, table_id):
     credentials = GoogleCredentials.get_application_default()
     bigquery = discovery.build('bigquery', 'v2', credentials=credentials)
 
+    # Infer the data format from the name of the data file.
+    source_format = 'CSV'
+    if data_path[-5:].lower() == '.json':
+        source_format = 'NEWLINE_DELIMITED_JSON'
+
+    # Post to the jobs resource using the client's media upload interface. See:
+    # http://developers.google.com/api-client-library/python/guide/media_upload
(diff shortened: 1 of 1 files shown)


bigquery/api/load_data_by_post.py
```diff
@@ -78,12 +87,14 @@ def load_data(schema_path, data_path, project_id, dataset_id, table_id):
         projectId=job['jobReference']['projectId'],
         jobId=job['jobReference']['jobId'])
 
+    # Poll the job until it finishes.
     while True:
         result = status_request.execute(num_retries=2)
 
         if result['status']['state'] == 'DONE':
-            if 'errorResult' in result['status']:
-                raise RuntimeError(result['status']['errorResult'])
```
jonparrott: Did this change?

jonparrott:
LGTM.

bigquery/api/load_data_by_post.py
jerjou: Yeah - when I was repro'ing, the errorResult just gave the last error, which was pretty uninformative ("too many errors encountered" or some such). It was more helpful to print out all the errors.

PR #131, opened 2015-10-16
Add test to load_data_by_post.

jerjou:


bigquery/api/load_data_by_post_test.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+@attr('slow')
+class TestLoadDataByPost(tests.CloudBaseTest):
+    dataset_id = 'ephemeral_dataset'
+    table_id = 'load_data_by_post'
+
+    def test_load_csv_data(self):
+        timestamp = int(time.time())
+        with tempfile.NamedTemporaryFile(suffix=".json") as schema_file, \
```
jonparrott: we have fixtures in `/tests/data`, please use those.

jonparrott:
So, how is this sample different from the load_data_from_csv, load_data_from_json samples?

jerjou:
What load_data_from_json sample? :)
It's different from the load_data_from_csv because it loads it from a local file, rather than from a file in cloud storage.

jonparrott:
gotcha. okay, I think perhaps we should give these better names in the future. Anyway, this LGTM.

jerjou:
Updated to use the fixtures. Merging..

PR #132, opened 2015-10-21
Adding mailgun sample

jonparrott:


appengine/mailgun/appengine_config.py
```diff
@@ -0,0 +1,4 @@
+from google.appengine.ext import vendor
```
waprin: license? technically code so I think it needs it.

appengine/mailgun/main_test.py
```diff
@@ -0,0 +1,55 @@
...
+        with http:
+            response = self.app.post('/', {
+                'recipient': 'jonwayne@google.com',
+                'submit': 'Send simple email'})
+
+            self.assertEqual(response.status_int, 200)
+
+        http = Http2Mock(responses=[{}])
+
+        with http:
```
waprin: could assert http was called, check the data it was called with matches what's expected. That's probably more than necessary, just food for thought.

tests/utils.py
```diff
@@ -21,6 +21,7 @@
 import tempfile
 import unittest
```
waprin: license?

waprin:
LGTM, think you're missing one license file.

appengine/mailgun/main_test.py
jonparrott: Coverage handles that.

waprin: Gotcha

jonparrott:
Fixed missing license. Merging.

PR #133, opened 2015-10-22
Make create_instance_test fail on bad output.

jerjou:


waprin:
LGTM, but how does this fix the tests?

jerjou:
It doesn't. The test was failing because the instance it was trying to create already existed. My hypothesis is that it existed because this test failed in the middle, at some point, and left the instance it created lying around.

This change will at least alert us when the first test run fails (the actual failure), instead of when subsequent test runs fail because of the first failure.

PR #141, opened 2015-11-11
Temoprarily Pin Tox Until Bug Is Fixed

waprin:
https://bitbucket.org/hpk42/tox/issues/285/tox-220-breaks-some-toxini-config-files


jonparrott:
Lgtm

PR #144, opened 2015-11-20
Restoring missing requirements.txt files

jonparrott:
Originally, I had consolidated to one top-level requirements.txt because it was difficult to keep them all updated. I've remedied this with two new scripts in our repositories:

* check_requirements.py which will return non-zero if any requirements are out
* update_requirements.py which will automatically update all requirements in requirements.txt to the latest available version.

Tox now runs check_requirements.py over all requirements.txt files in the repository during testing. There's also an additional, manually-run eviroment that does the same with update_requirements.

Additionally, I updated readmes to reflect the new requirements.txt and also updated gcloud auth login -> gcloud init.

Phew.


appengine/bigquery/requirements.txt
```diff
@@ -1 +1 @@
-google-api-python-client
+google-api-python-client==1.4.2
```
waprin: Do we think it's better to pin versions? I feel like given we have CI tests its better to let them auto upgrade and only pin if they break.

waprin: Ah, now I understand that's part of the point of this patch...

waprin: But I guess I'm still confused, why not keep the dependencies unpinned?

jonparrott: By pinning the requirements, we can be certain that our users get the versions that have been tested and vetted by us, and aren't surprised by breaking changes in new versions.

The scripts to check and update are for us, so that we can be made aware of new releases and get them tested and released to users.

waprin:
lgtm

appengine/bigquery/requirements.txt
waprin: alright fine with me...its an age-old debate in other contexts too (static vs dynamic linkins is similar), advantages and disadvantages each way but if you want to pin, fine by me

jonparrott: Yeah, in this case I just want to have the confidence that what we're giving to our users works.

PR #149, opened 2015-12-15
Updating requirements.

jonparrott:


waprin:
LGTM, think these can just be fyis?

PR #150, opened 2015-12-17
Small discrepancy between argument and pydoc

drincruz:
Changing pydoc argument from `source_csv` to `source_path` to align with the argument name.


drincruz:
@googlebot I signed it!

jonparrott:
LGTM, thanks for your contribution!

PR #158, opened 2016-01-29
Adding datastore samples.

jonparrott:
@waprin this is missing tests, but ready for initial review. I will add tests. It's just a ton to write. :|

@pcostell Please verify that all of the relevant snippets are here. These are based on the node.js snippets, not the ndb ones.

@dhermes if you have a chance, please take a quick look and make sure I'm not using `gcloud-python` like a moron.

### DIFF
--- a/datastore/api/README.md
+++ b/datastore/api/README.md
@@ -1,0 +1,4 @@
+# Cloud Datastore API Samples
+
+<!-- auto-doc-link -->
+<!-- end-auto-doc-link -->
--- a/datastore/api/snippets.py
+++ b/datastore/api/snippets.py
@@ -1,0 +1,809 @@
+# Copyright 2016, Google, Inc.
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import argparse
+from collections import defaultdict
+import datetime
+from pprint import pprint
+
+import gcloud
+from gcloud import datastore
+
+
+def incomplete_key(client):
+    # [START incomplete_key]
+    key = client.key('Task')
+    # [END incomplete_key]
+
+    return key
+
+
+def named_key(client):
+    # [START named_key]
+    key = client.key('Task', 'sample_task')
+    # [END named_key]
+
+    return key
+
+
+def key_with_parent(client):
+    # [START key_with_parent]
+    key = client.key('TaskList', 'default', 'Task', 'sample_task')
+    # Alternatively
+    parent_key = client.key('TaskList', 'default')
+    key = client.key('Task', 'sample_task', parent=parent_key)
+    # [END key_with_parent]
+
+    return key
+
+
+def key_with_multilevel_parent(client):
+    # [START key_with_multilevel_parent]
+    key = client.key(
+        'User', 'alice',
+        'TaskList', 'default',
+        'Task', 'sample_task')
+    # [END key_with_multilevel_parent]
+
+    return key
+
+
+def basic_entity(client):
+    # [START basic_entity]
+    task = datastore.Entity(client.key('Task'))
+    task.update({
+        'type': 'Personal',
+        'done': False,
+        'priority': 4,
+        'description': 'Learn Cloud Datastore'
+    })
+    # [END basic_entity]
+
+    return task
+
+
+def entity_with_parent(client):
+    # [START entity_with_parent]
+    key_with_parent = client.key(
+        'TaskList', 'default', 'Task', 'sample_task')
+
+    task = datastore.Entity(key=key_with_parent)
+
+    task.update({
+        'type': 'Personal',
+        'done': False,
+        'priority': 4,
+        'description': 'Learn Cloud Datastore'
+    })
+    # [END entity_with_parent]
+
+    return task
+
+
+def properties(client):
+    key = client.key('Task')
+    # [START properties]
+    task = datastore.Entity(
+        key,
+        exclude_from_indexes=['description'])
+    task.update({
+        'type': 'Personal',
+        'description': 'Learn Cloud Datastore',
+        'created': datetime.datetime.utcnow(),
+        'done': False,
+        'priority': 4,
+        'percent_complete': 10.5,
+    })
+    # [END properties]
+
+    return task
+
+
+def array_value(client):
+    key = client.key('Task')
+    # [START array_value]
+    task = datastore.Entity(key)
+    task.update({
+        'tags': [
+            'fun',
+            'programming'
+        ],
+        'collaborators': [
+            'alice',
+            'bob'
+        ]
+    })
+    # [END array_value]
+
+    return task
+
+
+def upsert(client):
+    # [START upsert]
+    complete_key = client.key('Task', 'sample_task')
+
+    task = datastore.Entity(key=complete_key)
+
+    task.update({
+        'type': 'Personal',
+        'done': False,
+        'priority': 4,
+        'description': 'Learn Cloud Datastore'
+    })
+
+    client.put(task)
+    # [END upsert]
+
+    return task
+
+
+def insert(client):
+    # [START insert]
+    with client.transaction():
+        incomplete_key = client.key('Task')
+
+        task = datastore.Entity(key=incomplete_key)
+
+        task.update({
+            'type': 'Personal',
+            'done': False,
+            'priority': 4,
+            'description': 'Learn Cloud Datastore'
+        })
+
+        client.put(task)
+    # [END insert]
+
+    return task
+
+
+def update(client):
+    # Create the entity we're going to update.
+    upsert(client)
+
+    # [START update]
+    with client.transaction():
+        key = client.key('Task', 'sample_task')
+        task = client.get(key)
+
+        task['done'] = True
+
+        client.put(task)
+    # [END update]
+
+    return task
+
+
+def lookup(client):
+    # Create the entity that we're going to look up.
+    upsert(client)
+
+    # [START lookup]
+    key = client.key('Task', 'sample_task')
+    task = client.get(key)
+    # [END lookup]
+
+    return task
+
+
+def delete(client):
+    # Create the entity we're going to delete.
+    upsert(client)
+
+    # [START delete]
+    key = client.key('Task', 'sample_task')
+    client.delete(key)
+    # [END delete]
+
+    return key
+
+
+def batch_upsert(client):
+    task1 = datastore.Entity(client.key('Task', 1))
+
+    task1.update({
+        'type': 'Personal',
+        'done': False,
+        'priority': 4,
+        'description': 'Learn Cloud Datastore'
+    })
+
+    task2 = datastore.Entity(client.key('Task', 2))
+
+    task2.update({
+        'type': 'Work',
+        'done': False,
+        'priority': 8,
+        'description': 'Integrate Cloud Datastore'
+    })
+
+    # [START batch_upsert]
+    client.put_multi([task1, task2])
+    # [END batch_upsert]
+
+    return task1, task2
+
+
+def batch_lookup(client):
+    # Create the entities we will lookup.
+    batch_upsert(client)
+
+    keys = [
+        client.key('Task', 1),
+        client.key('Task', 2)
+    ]
+
+    # [START batch_lookup]
+    tasks = client.get_multi(keys)
+    # [END batch_lookup]
+
+    return tasks
+
+
+def batch_delete(client):
+    # Create the entities we will delete.
+    batch_upsert(client)
+
+    keys = [
+        client.key('Task', 1),
+        client.key('Task', 2)
+    ]
+
+    # [START batch_delete]
+    client.delete_multi(keys)
+    # [END batch_delete]
+
+    return keys
+
+
+def unindexed_property_query(client):
+    # Create the entity that we're going to query.
+    upsert(client)
+
+    # [START unindexed_property_query]
+    query = client.query(kind='Task')
+    query.add_filter('description', '=', 'Learn Cloud Datastore')
+    # [END unindexed_property_query]
+
+    return list(query.fetch())
+
+
+def basic_query(client):
+    # Create the entity that we're going to query.
+    upsert(client)
+
+    # [START basic_query]
+    query = client.query(kind='Task')
+    query.add_filter('done', '=', False)
+    query.add_filter('priority', '>=', 4)
+    query.order = ['-priority']
+    # [END basic_query]
+
+    return list(query.fetch())
+
+
+def projection_query(client):
+    # Create the entity that we're going to query.
+    task = datastore.Entity(client.key('Task'))
+    task.update({
+        'type': 'Personal',
+        'done': False,
+        'priority': 4,
+        'description': 'Learn Cloud Datastore',
+        'percent_complete': 0.5
+    })
+    client.put(task)
+
+    # [START projection_query]
+    query = client.query(kind='Task')
+    query.projection = ['priority', 'percent_complete']
+    # [END projection_query]
+
+    # [START run_query_projection]
+    priorities = []
+    percent_completes = []
+
+    for task in query.fetch():
+        priorities.append(task['priority'])
+        percent_completes.append(task['priority'])
+    # [END run_query_projection]
+
+    return priorities, percent_completes
+
+
+def ancestor_query(client):
+    task = datastore.Entity(
+        client.key('TaskList', 'default', 'Task'))
+    task.update({
+        'type': 'Personal',
+        'description': 'Learn Cloud Datastore',
+    })
+    client.put(task)
+
+    # [START ancestor_query]
+    ancestor = client.key('TaskList', 'default')
+    query = client.query(kind='Task', ancestor=ancestor)
+    # [END ancestor_query]
+
+    return list(query.fetch())
+
+
+def run_query(client):
+    # [START run_query]
+    query = client.query()
+    results = list(query.fetch())
+    # [END run_query]
+
+    return results
+
+
+def limit(client):
+    # [START limit]
+    query = client.query()
+    tasks = list(query.fetch(limit=5))
+    # [END limit]
+
+    return tasks
+
+
+def cursor_paging(client):
+    # [START cursor_paging]
+
+    def get_one_page_of_tasks(cursor=None):
+        query = client.query(kind='Task')
+        query_iter = query.fetch(start_cursor=cursor, limit=5)
+        tasks, _, cursor = query_iter.next_page()
+
+        return tasks, cursor
+    # [END cursor_paging]
+
+    page_one, cursor_one = get_one_page_of_tasks()
+    page_two, cursor_two = get_one_page_of_tasks(cursor=cursor_one)
+    return page_one, cursor_one, page_two, cursor_two
+
+
+def property_filter(client):
+    # Create the entity that we're going to query.
+    upsert(client)
+
+    # [START property_filter]
+    query = client.query(kind='Task')
+    query.add_filter('done', '=', False)
+    # [END property_filter]
+
+    return list(query.fetch())
+
+
+def composite_filter(client):
+    # Create the entity that we're going to query.
+    upsert(client)
+
+    # [START composite_filter]
+    query = client.query(kind='Task')
+    query.add_filter('done', '=', False)
+    query.add_filter('priority', '=', 4)
+    # [END composite_filter]
+
+    return list(query.fetch())
+
+
+def key_filter(client):
+    # Create the entity that we're going to query.
+    upsert(client)
+
+    # [START key_filter]
+    query = client.query(kind='Task')
+    first_key = client.key('Task', 'first_task')
+    query.add_filter('__key__', '>', first_key)
+    # [END key_filter]
+
+    return list(query.fetch())
+
+
+def ascending_sort(client):
+    # Create the entity that we're going to query.
+    task = upsert(client)
+    task['created'] = datetime.datetime.utcnow()
+    client.put(task)
+
+    # [START ascending_sort]
+    query = client.query(kind='Task')
+    query.order = ['created']
+    # [END ascending_sort]
+
+    return list(query.fetch())
+
+
+def descending_sort(client):
+    # Create the entity that we're going to query.
+    task = upsert(client)
+    task['created'] = datetime.datetime.utcnow()
+    client.put(task)
+
+    # [START descending_sort]
+    query = client.query(kind='Task')
+    query.order = ['-created']
+    # [END descending_sort]
+
+    return list(query.fetch())
+
+
+def multi_sort(client):
+    # Create the entity that we're going to query.
+    task = upsert(client)
+    task['created'] = datetime.datetime.utcnow()
+    client.put(task)
+
+    # [START multi_sort]
+    query = client.query(kind='Task')
+    query.order = [
+        '-priority',
+        'created'
+    ]
+    # [END multi_sort]
+
+    return list(query.fetch())
+
+
+def keys_only_query(client):
+    # Create the entity that we're going to query.
+    upsert(client)
+
+    # [START keys_only_query]
+    query = client.query()
+    query.keys_only()
+    # [END keys_only_query]
+
+    # [START run_keys_only_query]
+    keys = list([entity.key for entity in query.fetch(limit=10)])
+    # [END run_keys_only_query]
+
+    return keys
+
+
+def distinct_query(client):
+    # Create the entity that we're going to query.
+    upsert(client)
+
+    # [START distinct_query]
+    query = client.query(kind='Task')
+    query.group_by = ['type', 'priority']
+    query.order = ['type', 'priority']
+    query.projection = ['type', 'priority']
+    # [END distinct_query]
+
+    return list(query.fetch())
+
+
+def distinct_on_query(client):
+    # Create the entity that we're going to query.
+    upsert(client)
+
+    # [START distinct_on_query]
+    query = client.query(kind='Task')
+    query.group_by = ['type']
+    query.order = ['type', 'priority']
+    # [END distinct_on_query]
+
+    return list(query.fetch())
+
+
+def kindless_query(client):
+    # Create the entity that we're going to query.
+    upsert(client)
+
+    last_seen_key = client.key('Task', 'a')
+
+    # [START kindless_query]
+    query = client.query()
+    query.add_filter('__key__', '>', last_seen_key)
+    # [END kindless_query]
+
+    return list(query.fetch())
+
+
+def inequality_range(client):
+    # [START inequality_range]
+    start_date = datetime.datetime(1990, 1, 1)
+    end_date = datetime.datetime(2000, 1, 1)
+    query = client.query(kind='Task')
+    query.add_filter(
+        'created', '>', start_date)
+    query.add_filter(
+        'created', '<', end_date)
+    # [END inequality_range]
+
+    return list(query.fetch())
+
+
+def inequality_invalid(client):
+    try:
+        # [START inequality_invalid]
+        start_date = datetime.datetime(1990, 1, 1)
+        query = client.query(kind='Task')
+        query.add_filter(
+            'created', '>', start_date)
+        query.add_filter(
+            'priority', '>', 3)
+        # [END inequality_invalid]
+
+        return list(query.fetch())
+
+    except gcloud.exceptions.BadRequest:
+        pass
+
+
+def equal_and_inequality_range(client):
+    # [START equal_and_inequality_range]
+    start_date = datetime.datetime(1990, 1, 1)
+    end_date = datetime.datetime(2000, 12, 31, 23, 59, 59)
+    query = client.query(kind='Task')
+    query.add_filter('priority', '=', 4)
+    query.add_filter('done', '=', False)
+    query.add_filter(
+        'created', '>', start_date)
+    query.add_filter(
+        'created', '<', end_date)
+    # [END equal_and_inequality_range]
+
+    return list(query.fetch())
+
+
+def inequality_sort(client):
+    # [START inequality_sort]
+    query = client.query(kind='Task')
+    query.add_filter('priority', '>', 3)
+    query.order = ['priority', 'created']
+    # [END inequality_sort]
+
+    return list(query.fetch())
+
+
+def inequality_sort_invalid_not_same(client):
+    try:
+        # [START inequality_sort_invalid_not_same]
+        query = client.query(kind='Task')
+        query.add_filter('priority', '>', 3)
+        query.order = ['created']
+        # [END inequality_sort_invalid_not_same]
+
+        return list(query.fetch())
+
+    except gcloud.exceptions.BadRequest:
+        pass
+
+
+def inequality_sort_invalid_not_first(client):
+    try:
+        # [START inequality_sort_invalid_not_first]
+        query = client.query(kind='Task')
+        query.add_filter('priority', '>', 3)
+        query.order = ['created', 'priority']
+        # [END inequality_sort_invalid_not_first]
+
+        return list(query.fetch())
+
+    except gcloud.exceptions.BadRequest:
+        pass
+
+
+def array_value_inequality_range(client):
+    # [START array_value_inequality_range]
+    query = client.query(kind='Task')
+    query.add_filter('tag', '>', 'learn')
+    query.add_filter('tag', '<', 'math')
+    # [END array_value_inequality_range]
+
+    return list(query.fetch())
+
+
+def array_value_equality(client):
+    # [START array_value_equality]
+    query = client.query(kind='Task')
+    query.add_filter('tag', '=', 'fun')
+    query.add_filter('tag', '=', 'programming')
+    # [END array_value_equality]
+
+    return list(query.fetch())
+
+
+def exploding_properties(client):
+    # [START exploding_properties]
+    task = datastore.Entity(client.key('Task'))
+    task.update({
+        'tags': [
+            'fun',
+            'programming',
+            'learn'
+        ],
+        'collaborators': [
+            'alice',
+            'bob',
+            'charlie'
+        ],
+        'created': datetime.datetime.utcnow()
+    })
+    # [END exploding_properties]
+
+    return task
+
+
+def transactional_update(client):
+    # Create the entities we're going to manipulate
+    account1 = datastore.Entity(client.key('Account'))
+    account1['balance'] = 100
+    account2 = datastore.Entity(client.key('Account'))
+    account2['balance'] = 100
+    client.put_multi([account1, account2])
+
+    # [START transactional_update]
+    def transfer_funds(client, from_key, to_key, amount):
+        with client.transaction():
+            from_account, to_account = client.get_multi([from_key, to_key])
+
+            from_account['balance'] -= amount
+            to_account['balance'] += amount
+
+            client.put_multi([from_account, to_account])
+    # [END transactional_update]
+
+    # [START transactional_retry]
+    for _ in range(5):
+        try:
+            transfer_funds(client, account1.key, account2.key, 50)
+        except gcloud.exceptions.Conflict:
+            continue
+    # [END transaction_retry]
+
+    return account1.key, account2.key
+
+
+def transactional_get_or_create(client):
+    # [START transactional_get_or_create]
+    with client.transaction():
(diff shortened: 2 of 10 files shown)


datastore/api/README.md
```diff
@@ -0,0 +1,4 @@
+# Cloud Datastore API Samples
+
+<!-- auto-doc-link -->
+<!-- end-auto-doc-link -->
```
dhermes: Is this for Sphinx? Never seen it before. Neat.

datastore/api/snippets.py
```diff
@@ -0,0 +1,820 @@
...
+    # [START named_key]
+    key = client.key('Task', 'sample_task')
+    # [END named_key]
+
+    return key
+
+
+def key_with_parent(client):
+    # [START key_with_parent]
+    key = client.key('TaskList', 'default', 'Task', 'sample_task')
```
dhermes: You might also consider doing

```python
k1 = client.key('TaskList', 'default')
k2 = client.key('Task', 'sample_task1', parent=k1)
k3 = client.key('Task', 'sample_task2', parent=k1)

File: datastore/api/snippets.py [hunk 3]
```diff
@@ -0,0 +1,820 @@
...
+        'TaskList', 'default',
+        'Task', 'sample_task')
+    # [END key_with_multilevel_parent]
+
+    return key
+
+
+def basic_entity(client):
+    # [START basic_entity]
+    task = datastore.Entity()
```
dhermes: No key?

datastore/api/README.md
jonparrott: It's for a [tool](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/scripts/auto_link_to_docs.py) that takes a list of links to this project in https://cloud.google.com and sticks them into the readme. The part of the tool that generates the list of links is internal. :(

datastore/api/snippets.py
```diff
@@ -0,0 +1,820 @@
...
+    })
+    # [END basic_entity]
+
+    return task
+
+
+def entity_with_parent(client):
+    # [START entity_with_parent]
+    key_with_parent = client.key(
+        'TaskList', 'default', 'Task', 'sample_task')
```
dhermes: Is the goal a named key or auto ID?

datastore/api/snippets.py
jonparrott: The node.js sample didn't have one either. Though I agree, it should have a key.

datastore/api/snippets.py
jonparrott: In this sample it's irrelevant, it's just to show how to use a multilevel key.

datastore/api/snippets.py
```diff
@@ -0,0 +1,820 @@
...
+    })
+
+    client.put(task)
+    # [END upsert]
+
+    return task
+
+
+def insert(client):
+    # [START upsert]
```
dhermes: This **also** says `START upsert`.

FWIW we don't even implement `insert` ([yet][1]).

[1]: https://github.com/GoogleCloudPlatform/gcloud-python/issues/599

datastore/api/snippets.py
```diff
@@ -0,0 +1,820 @@
...
+        'description': 'Learn Cloud Datastore'
+    })
+
+    client.put(task)
+    # [END upsert]
+
+    return task
+
+
+def update(client):
```
dhermes: Ditto here on https://github.com/GoogleCloudPlatform/gcloud-python/issues/599 being pending

datastore/api/snippets.py
```diff
@@ -0,0 +1,820 @@
...
+
+
+def lookup(client):
+    # Create the entity that we're going to look up.
+    upsert(client)
+
+    # [START lookup]
+    key = client.key('Task', 'sample_task')
+    task = client.get(key)
+    # <Entity
```
dhermes: Is this comment supposed to be here? Generated result?

datastore/api/snippets.py
```diff
@@ -0,0 +1,820 @@
...
+    # [END ancestor_query]
+
+    return tasks
+
+
+def run_query(client):
+
+    # [START run_query]
+    query = client.query()
+    results = list(query.fetch())
```
dhermes: Seems this is a bit repetitive.

datastore/api/snippets.py
```diff
@@ -0,0 +1,820 @@
...
+
+    return tasks
+
+
+def cursor_paging(client):
+
+    # [START cursor_paging]
+    tasks = []
+    query = client.query(kind='Task')
+    it = query.fetch(limit=5)
```
dhermes: Suggest: `query_iter` would be more descriptive

datastore/api/snippets.py
```diff
@@ -0,0 +1,820 @@
...
+    return tasks
+
+
+def key_filter(client):
+    # Create the entity that we're going to query.
+    upsert(client)
+
+    # [START key_filter]
+    query = client.query(kind='Task')
+    query.add_filter('__key__', '>', client.key('Task', 'first_task'))
```
dhermes: Might be more instructive to do

```python
compare_key = client.key('Task', 'first_task')
query.add_filter('__key__', '>', compare_key)
```

datastore/api/snippets.py
```diff
@@ -0,0 +1,820 @@
...
+
+
+def keys_only_query(client):
+    # Create the entity that we're going to query.
+    upsert(client)
+
+    # [START keys_only_query]
+    query = client.query()
+    query.keys_only()
+    keys = list([x.key for x in query.fetch()])
```
dhermes: s/`x`/`entity`

datastore/api/snippets.py
```diff
@@ -0,0 +1,820 @@
...
+    return keys
+
+
+def distinct_query(client):
+    # Create the entity that we're going to query.
+    upsert(client)
+
+    # [START distinct_query]
+    query = client.query(kind='Task')
+    query.group_by = ['type', 'priority']
```
dhermes: FYI this will change once `v1beta3` is in

datastore/api/snippets.py
```diff
@@ -0,0 +1,820 @@
...
+    query.group_by = ['type']
+    query.order = ['type', 'priority']
+    tasks = list(query.fetch())
+    # [END distinct_on_query]
+
+    return tasks
+
+
+def kindless_query(client):
+    # [START kindless_query]
```
dhermes: Just learned a thing! Didn't know kindless was possible!

datastore/api/snippets.py
```diff
@@ -0,0 +1,820 @@
...
+            'created', '>', datetime.datetime(1990, 1, 1))
+        query.add_filter(
+            'priority', '>', 3)
+        tasks = list(query.fetch())
+        # [END inequality_invalid]
+
+        return tasks
+
+    except gcloud.exceptions.BadRequest:
+        pass
```
dhermes: Wouldn't it be nice to print something to confirm the error occurred?

dhermes: **UPDATE**: NVM this isn't in the snippet

datastore/api/snippets.py
```diff
@@ -0,0 +1,820 @@
...
+    query.order = ['priority', 'created']
+    tasks = list(query.fetch())
+    # [END inequality_sort]
+
+    return tasks
+
+
+def inequality_sort_invalid_not_same(client):
+    try:
+        # [START inequality_sort_invalid_not_same]
```
dhermes: What's the point of this?

datastore/api/snippets.py
```diff
@@ -0,0 +1,820 @@
...
+
+        return tasks
+
+    except gcloud.exceptions.BadRequest:
+        pass
+
+
+def inequality_sort_invalid_not_first(client):
+    try:
+        # [START inequality_sort_invalid_not_first]
```
dhermes: Ditto here, what makes it invalid?

datastore/api/snippets.py
```diff
@@ -0,0 +1,820 @@
...
+    # [END array_value_inequality_range]
+
+    return tasks
+
+
+def array_value_equality(client):
+    # [START array_value_equality]
+    query = client.query(kind='Task')
+    query.add_filter('tag', '=', 'fun')
+    query.add_filter('tag', '=', 'programming')
```
dhermes: Yay! Math is fun too!

datastore/api/snippets.py
```diff
@@ -0,0 +1,820 @@
...
+    tasks = list(query.fetch())
+    # [END array_value_equality]
+
+    return tasks
+
+
+def exploding_properties(client):
+    # [START exploding_queries]
```
dhermes: There is no query here. Also what makes it explode?

datastore/api/snippets.py
jonparrott: @pcostell thoughts?

datastore/api/snippets.py
jonparrott: Yeah, the node.js sample also shows what the returned entity's data is.

datastore/api/snippets.py
jonparrott: Welcome to documentation snippets.

datastore/api/snippets.py
jonparrott: What will it change to?

datastore/api/snippets.py
```diff
@@ -0,0 +1,820 @@
...
+            from_account['balance'] -= amount
+            to_account['balance'] += amount
+
+            client.put_multi([from_account, to_account])
+
+            return from_account, to_account
+    # [END transactional_update]
+
+    # [START transactional_retry]
+    for _ in range(5):
```
dhermes: `xrange`?

datastore/api/snippets.py
jonparrott: To show that you can't do this kind of query (you can't use an inequality filter on one property and sort on a different one)

datastore/api/snippets.py
```diff
@@ -0,0 +1,820 @@
...
+
+def transactional_update(client):
+    # Create the entities we're going to manipulate
+    account1 = datastore.Entity(client.key('Account'))
+    account1['balance'] = 100
+    account2 = datastore.Entity(client.key('Account'))
+    account2['balance'] = 100
+    client.put_multi([account1, account2])
+
+    # [START transactional_update]
```
dhermes: Why does this need to be define inline?

datastore/api/snippets.py
jonparrott: Same reason as above - if you're doing an inequality filter the property you're filtering must be the first item in the sort list.

datastore/api/snippets.py
jonparrott: Haha I just blindly ported this from the Node samples.

datastore/api/snippets.py
jonparrott: Yeah I'm not worried about the memory overhead of a list of 5 integers.

datastore/api/snippets.py
```diff
@@ -0,0 +1,820 @@
...
+    def transfer_funds(client, from_key, to_key, amount):
+        with client.transaction():
+            from_account, to_account = client.get_multi([from_key, to_key])
+
+            from_account['balance'] -= amount
+            to_account['balance'] += amount
+
+            client.put_multi([from_account, to_account])
+
+            return from_account, to_account
```
dhermes: Might be "better to `return` after the transaction completes

datastore/api/snippets.py
jonparrott: The documentation explains it (e.g., these snippets are not intended to be consumed without the documentation). Essentially having a indexed on two properties that are arrays leads to the index containing the combination of all elements of the two arrays.

datastore/api/snippets.py
jonparrott: I always feel like I'm abusing scope if I return variables defined within a context block.

datastore/api/snippets.py
```diff
@@ -0,0 +1,820 @@
...
+
+def namespace_run_query(client):
+    # Create an entity in another namespace.
+    task = datastore.Entity(
+        client.key('Task', namespace='namespace-b'))
+    client.put(task)
+
+    # [START namespace_run_query]
+    # All namespaces
+    query = client.query(kind='__namespace__')
```
dhermes: :+1: I've got an [issue][1] to make these easier (i.e. you don't need to know the esoteric `__foo__` kind to query).

[1]: https://github.com/GoogleCloudPlatform/gcloud-python/issues/1067

datastore/api/snippets.py
```diff
@@ -0,0 +1,820 @@
...
+    query.keys_only()
+
+    all_namespaces = [x.key.id_or_name for x in query.fetch()]
```
dhermes: s/`x`/`entity`

datastore/api/snippets.py
```diff
@@ -0,0 +1,820 @@
...
+
+    # Filtered namespaces
+    query = client.query(kind='__namespace__')
+    query.add_filter(
+        '__key__', '>=', client.key('__namespace__', 'namespace-a'))
+    query.add_filter(
+        '__key__', '<', client.key('__namespace__', 'namespace-z'))
```
dhermes: Do you think we should add a `key_filter` helper?

datastore/api/snippets.py
```diff
@@ -0,0 +1,820 @@
...
+
+    filtered_namespaces = [x.key.id_or_name for x in query.fetch()]
```
dhermes: Ditto on `x`

datastore/api/snippets.py
jonparrott: If you mean the function, just so that I can have separate snippets for `transactional_update` and `retry_transaction`.

datastore/api/snippets.py
```diff
@@ -0,0 +1,820 @@
...
+
+def kind_run_query(client):
+    # Create the entity that we're going to query.
+    upsert(client)
+
+    # [START kind_run_query]
+    query = client.query(kind='__kind__')
+    query.keys_only()
+
+    kinds = [x.key.id_or_name for x in query.fetch()]
```
dhermes: `x`

datastore/api/snippets.py
```diff
@@ -0,0 +1,820 @@
...
+
+    # [START kind_run_query]
+    query = client.query(kind='__property__')
+    query.keys_only()
+
+    properties_by_kind = defaultdict(list)
+
+    for entity in query.fetch():
+        kind = entity.key.parent.name
+        property = entity.key.name
```
dhermes: s/`property`/`property_`

datastore/api/snippets.py
dhermes: `disctinct_on`

datastore/api/snippets.py
dhermes: Ah I see

datastore/api/snippets.py
dhermes: Could you define `transfer_funds` at the top level and put the `# [START` at the top level?

datastore/api/snippets.py
dhermes: Well this function doesn't even need a returned value.

datastore/api/snippets.py
dhermes: Sinful!

datastore/api/snippets.py
jonparrott: done.

datastore/api/snippets.py
jonparrott: done.

datastore/api/snippets.py
jonparrott: done.

datastore/api/snippets.py
jonparrott: That would actually be nice.

datastore/api/snippets.py
jonparrott: done.

datastore/api/snippets.py
jonparrott: Fair enough, removed.

datastore/api/snippets.py
jonparrott: Gotcha. I'll keep my eyes out for that.

datastore/api/snippets.py
jonparrott: done.

datastore/api/snippets.py
jonparrott: done.

datastore/api/snippets.py
jonparrott: done

datastore/api/snippets.py
jonparrott: Done.

datastore/api/snippets.py
jonparrott: I decided to show both.

dhermes: :+1:

datastore/api/snippets.py
dhermes: Filed issue. Cheers.

datastore/api/snippets.py
```diff
@@ -0,0 +1,823 @@
...
+    query.projection = ['priority', 'percent_complete']
+    # [END projection_query]
+
+    # [START run_query_projection]
+    priorities = []
+    percent_completes = []
+
+    for task in query.fetch():
+        priorities.append(task['priority'])
+        percent_completes.append(task['priority'])
```
waprin: feels weird these aren't list comprehensions. Are query results not cached (and if not, shouldn't they be)? Or just wanted to keep it simpler?

jonparrott: I feel like a list comprehension would've been too hard to follow in this case (more clever than clear).

datastore/api/snippets.py
```diff
@@ -0,0 +1,823 @@
...
+    })
+    # [END entity_with_parent]
+
+    return task
+
+
+def properties(client):
+    # [START properties]
+    task = datastore.Entity()
+    task.update({
```
pcostell: Description should be marked with `excludeFromIndexes`

datastore/api/snippets.py
```diff
@@ -0,0 +1,823 @@
...
+        # String
```
pcostell: I don't think we need these comments -- we can include it directly in the docs if we want the info.

datastore/api/snippets.py
```diff
@@ -0,0 +1,820 @@
...
+from collections import defaultdict
+import datetime
+from pprint import pprint
+import time
+
+import gcloud
+from gcloud import datastore
+
+
+def incomplete_key(client):
```
waprin: Seems kind of silly to pass the same client to every single method. But I also understand the aversion to a module level variable or classes. idk, just a thought.

datastore/api/snippets.py
```diff
@@ -0,0 +1,820 @@
...
+
+    # [START ancestor_query]
+    ancestor = client.key('TaskList', 'default')
+    query = client.query(kind='Task', ancestor=ancestor)
+    tasks = list(query.fetch())
+    # [END ancestor_query]
+
+    return tasks
+
+
```
waprin: spurious newline at the start of the next 3 methods.

datastore/api/snippets.py
```diff
@@ -0,0 +1,823 @@
...
+    # Create the entity we're going to update.
+    upsert(client)
+
+    # [START update]
+    key = client.key('Task', 'sample_task')
+    task = client.get(key)
+
+    task['done'] = True
+
+    client.put(task)
```
pcostell: This should all be in a transaction until GoogleCloudPlatform/gcloud-python#599 goes through (this is a bug in the gcloud-node snippets, I'll get those fixed).

datastore/api/snippets.py
```diff
@@ -0,0 +1,823 @@
...
+    return task
+
+
+def insert(client):
+    # [START insert]
+    incomplete_key = client.key('Task')
+
+    task = datastore.Entity(key=incomplete_key)
+
+    task.update({
```
pcostell: See comment below -- this should be done in a transaction until the API supports native insert.

pcostell: (Also can't be an incomplete key -- otherwise the transactional lookup / throw doesn't make sense)

datastore/api/snippets.py
```diff
@@ -0,0 +1,823 @@
...
+
+    task2.update({
+        'type': 'Work',
+        'done': False,
+        'priority': 8,
+        'description': 'Integrate Cloud Datastore'
+    })
+
+    client.put_multi([task1, task2])
+    # [END batch_upsert]
```
pcostell: I think this snippet should only have the put_multi call (based on other snippets).

datastore/api/snippets.py
```diff
@@ -0,0 +1,823 @@
...
+    # Create the entities we will lookup.
+    batch_upsert(client)
+
+    # [START batch_lookup]
+    keys = [
+        client.key('Task', 1),
+        client.key('Task', 2)
+    ]
+
+    tasks = client.get_multi(keys)
```
pcostell: Same here, remove the keys definition from the snippet.

datastore/api/snippets.py
```diff
@@ -0,0 +1,823 @@
...
+    # Create the entities we will delete.
+    batch_upsert(client)
+
+    # [START batch_delete]
+    keys = [
+        client.key('Task', 1),
+        client.key('Task', 2)
+    ]
+
+    client.delete_multi(keys)
```
pcostell: And here.

datastore/api/snippets.py
```diff
@@ -0,0 +1,823 @@
...
+    # Create the entity that we're going to query.
+    upsert(client)
+
+    # [START basic_query]
+    query = client.query(kind='Task')
+    query.add_filter('done', '=', False)
+    query.add_filter('priority', '>=', 4)
+    query.order = ['-priority']
+
+    tasks = list(query.fetch())
```
pcostell: basic_query should only define the query, this piece should be in a separate run_query snippet.

datastore/api/snippets.py
```diff
@@ -0,0 +1,823 @@
...
+    task.update({
+        'type': 'Personal',
+        'description': 'Learn Cloud Datastore',
+    })
+    client.put(task)
+
+    # [START ancestor_query]
+    ancestor = client.key('TaskList', 'default')
+    query = client.query(kind='Task', ancestor=ancestor)
+    tasks = list(query.fetch())
```
pcostell: no need to run the query here

datastore/api/snippets.py
pcostell: Commented a couple of times, but *most* of the query snippets should only define the query, not run it.

datastore/api/snippets.py
```diff
@@ -0,0 +1,823 @@
...
+
+def cursor_paging(client):
+
+    # [START cursor_paging]
+    tasks = []
+    query = client.query(kind='Task')
+    query_iter = query.fetch(limit=5)
+
+    while True:
+        results, _, cursor = query_iter.next_page()
```
pcostell: This should show how to use cursor with pages (i.e. search results). You should only get one page then return the results, setting a "next_page_cursor" variable that would actually get displayed on your website as part of your next page url.

datastore/api/snippets.py
pcostell: This query should also have a projection on type, priority

datastore/api/snippets.py
```diff
@@ -0,0 +1,823 @@
...
+
+def distinct_on_query(client):
+    # Create the entity that we're going to query.
+    upsert(client)
+
+    # [START distinct_on_query]
+    query = client.query(kind='Task')
+    query.group_by = ['type']
+    query.order = ['type', 'priority']
+    tasks = list(query.fetch())
```
pcostell: This should also have a projection on type, priority.

datastore/api/snippets.py
```diff
@@ -0,0 +1,823 @@
...
+    # [END distinct_on_query]
+
+    return tasks
+
+
+def kindless_query(client):
+    # [START kindless_query]
+    an_hour_ago = str(time.time() - 3600)
```
pcostell: Can we remove the computation of this and just use last_seen_key?

datastore/api/snippets.py
```diff
@@ -0,0 +1,823 @@
...
+        pass
+
+
+def equal_and_inequality_range(client):
+    # [START equal_and_inequality_range]
+    query = client.query(kind='Task')
+    query.add_filter('priority', '=', 4)
+    query.add_filter('done', '=', False)
+    query.add_filter(
+        'created', '>', datetime.datetime(1990, 1, 1))
```
pcostell: Just start_date and end_date here for the snippet.

datastore/api/snippets.py
pcostell: I think this snippet should be exploding_properties

datastore/api/snippets.py
```diff
@@ -0,0 +1,823 @@
...
+    # All namespaces
+    query = client.query(kind='__namespace__')
+    query.keys_only()
+
+    all_namespaces = [entity.key.id_or_name for entity in query.fetch()]
+
+    # Filtered namespaces
+    query = client.query(kind='__namespace__')
+    query.add_filter(
+        '__key__', '>=', client.key('__namespace__', 'namespace-a'))
```
pcostell: use stand in variables start_namespace and end_namespace

datastore/api/snippets.py
jonparrott: I heavily prefer not to have global state. It'll make this much less clear and much harder to test.

datastore/api/snippets.py
jonparrott: I think they're helpful but if you feel strongly I can remove.

datastore/api/snippets.py
jonparrott: Should the transaction be within or outside of the snippet?

datastore/api/snippets.py
jonparrott: Same as a above: Should the transaction be within or outside of the snippet?

datastore/api/snippets.py
jonparrott: Seems weird to hide that detail from the user, but if that's how it should be I can move the fence.

datastore/api/snippets.py
jonparrott: I disagree here. I think it's important to show them that it's expecting a list of keys. If you feel more strongly in the other direction we can discuss.

datastore/api/snippets.py
jonparrott: It's up to you, but considering running the query is one line I'm okay with leaving it in the snippet. Your call though since we do actually show how to run it in another snippet.

datastore/api/snippets.py
jonparrott: Ah okay, so you want it to actually just return the results and the cursor (and accept a start cursor as an argument) instead of draining the query in one go? Do you think it's useful to show both methods?

datastore/api/snippets.py
jonparrott: I'm okay with removing this line from the snippet, but I'd like to keep the actual call to create the key.

datastore/api/snippets.py
jonparrott: I'd prefer to show the users what type is expected here. I can make them variables at the top of the snippet if you'd like.

datastore/api/snippets.py
jonparrott: Again, I'd rather show them the expected type. I can make them variables defined at the start of the snippet.

pcostell: We should be consistent between the snippets. When snippets start to look different you get weird things where the docs might given an example but then the snippet doesn't actually match that example.

datastore/api/snippets.py
pcostell: Correct. No I don't think we should show it. If you want all the results you shouldn't be using next_page -- let the gcloud library handle batching.

datastore/api/snippets.py
pcostell: I think the consistency is important here. The docs can say "here you define a query:" and "then here you run it:". If the snippets include running it in the first one it looks weird.

datastore/api/snippets.py
pcostell: Again, we need to be consistent with the other APIs so that the docs can also have a consistent feel. I think this was one of the problems with the existing docs -- each language had slightly different code so if the docs made references to code pieces they would be incorrect in certain languages.

datastore/api/snippets.py
pcostell: I don't think it's useful here. The node ones showed it because the properties are hidden under data.

datastore/api/snippets.py
pcostell: In the snippet -- a little bit unreadable but technically correct because that is how the native insert/update work.

datastore/api/snippets.py
pcostell: I would rather have the docs state (but better worded): This snippet creates a string property called 'type', a date property called 'created', etc).

datastore/api/snippets.py
pcostell: Again, I think we need to be consistent across our snippets. In particular here using a timestamp as your primary key is a known datastore anti-pattern.

datastore/api/snippets.py
jonparrott: I would say in that case that the *other* snippets are incorrect. What use is showing `tasks = client.get_multi(keys)` to a user? You might as well just link them to gcloud-python's library documentation.

datastore/api/snippets.py
jonparrott: It's still consistent, it just shows how to create this weird key. Consider the key *is weird* and non-trivial I think it's important that we show how to make it.

datastore/api/snippets.py
jonparrott: > known datastore anti-pattern

Hey I was just copying the Node.js sample. Want to make a better recommendation for the key name?

datastore/api/snippets.py
jonparrott: done (in incoming push)

datastore/api/snippets.py
jonparrott: done (in incoming push)

datastore/api/snippets.py
pcostell: Probably fair but we should probably fix them all together.

datastore/api/snippets.py
jonparrott: done (in incoming push)

datastore/api/snippets.py
jonparrott: done (in incoming push)

datastore/api/snippets.py
jonparrott: done (in incoming push)

datastore/api/snippets.py
jonparrott: done (in incoming push)

datastore/api/snippets.py
jonparrott: done (in incoming push)

datastore/api/snippets.py
jonparrott: done (in incoming push)

datastore/api/snippets.py
jonparrott: Done.

datastore/api/snippets.py
pcostell: But it's not consistent -- If we decide to change the docs to go into more depth and describe what the start and end are to get specific results: i.e. setting the start and end namespaces to "f" and "h" would result in the namespaces including "google", all of a sudden these snippets are no longer correct. I know it's a contrived example in this case, but I've seen this a lot with the existing app engine documentation and want to have something consistent for our new docs. Once the docs are actually usable, I'm happy to go through (or even better have people start sending feedback) and update our spec for the snippets. However we are looking to have snippets in enough languages that subtle differences between them will lead to documentation and snippets that are not compatible.

datastore/api/snippets.py
jonparrott: done

datastore/api/snippets.py
jonparrott: done

datastore/api/snippets.py
jonparrott: moved to top of snippet.

datastore/api/snippets.py
jonparrott: Fixed.

datastore/api/snippets.py
jonparrott: Bleh, I feel like that's a poor excuse for hiding things from the user. I'd rather err on the side of showing too much to the user than too little. In this particular example (and from the documentation page I just read) I'd have *no* idea how to make the keys for the filtered namespace query.

jonparrott:
@pcostell github is terrible at long reviews, but most of your comments should be addressed. PTAL.

jonparrott:
@pcostell In general, I understand your concerns about being inconsistent between samples. However, I feel like in Python (and probably Node.js to an extent) because of the lack of strong typing it's extremely important to demonstrate to users how to create the objects we're using as parameters. As a user it's extremely frustrating than seeing an isolated sample like this:

```
library.do_thing_i_need(something_i_know, something_i_dont_know, argument_i_dont_know=another_thing_that_appeared_out_of_nowhere)
```

Especially when it's only 2-5 more lines of code for us to demonstrate where those objects come from.

jonparrott:
@pcostell please take another look. This should conform to the spec with these exceptions:

* `insert`, `upsert` because they are currently weird in gcloud-python. (Happy to change these to whatever you want them to be).
* `batch_lookup` and `batch_delete` show how to create a list of keys. I believe this is important as we do not otherwise show how to do this.
* `keys_only_query` shows how to extract the keys from the result. 
* `kindless_query` shows how to create a key.
* `equal_and_inequality_range` shows how to create the proper datetime object.
* `namespace_run_query` shows how to construct namespace keys (this is not documented elsewhere).

If you feel really strongly on any of these, I can change them to what you want. The only one I'm rather bullish on right now is the last one.

pcostell:
I've updated the spec (and added comments where it has changed) with
explicit instructions for these.

In particular:
* batch_lookup and batch_delete don't show how to get the key because we
should encourage auto keys and not key creation. Instead, batch_upsert
shows how to create explicit entities and store those keys, then the
batch_lookup and delete will use those keys. We will make the docs explicit
about where those came from.
* keys_only_query shouldn't show how to get the keys, just the query.
However there is now a run_keys_only_query which should show how to get the
keys.
* Please leave kindless_query as is for now (with a fake variable). Using a
time as the key doesn't match any of the other examples (which use auto gen
keys or the key "sampleTask"). This makes it look like our auto ids are
time based, which they are not. My current plan is to separate these out
and talk about how to do your own batch processing (the main use case
here).
* equal_and_inequal_range I defined a range in the spec. I made it >=
1990/1/1 and < 2000/1/1 just to reduce the amount of information in the
snippet which is irrelevant (in this case the hr/min setup). I don't care
what the range is though if you'd prefer something different.
* namespace_run_query has been changed to >= 'g' and < 'h'. We'll build
into the docs a description of why you might choose that range (i.e.
listing all your customers that start with 'g'). I want to avoid
'namespace-*' as a name because you shouldn't have to prefix your namespace
by namespace and I want to make that clear.

jonparrott:
Excellent. Update the snippets as follows:

* "batch_lookup and batch_delete don't show how to get the key because we
should encourage auto keys and not key creation." - Done.
* "keys_only_query shouldn't show how to get the keys, just the query.
However there is now a run_keys_only_query which should show how to get the
keys." - Done.
* "Please leave kindless_query as is for now (with a fake variable)." - Kindless query still shows `last_seen_key = client.key('Task', an_hour_ago)`, it does not show where 'an_hour_ago' comes from. I'm willing to move that outside of the fence if you want.
* "equal_and_inequal_range" - Updated the dates to reflect that.
* "namespace_run_query has been changed to >= 'g' and < 'h'" - Done.

I've also added the getting started samples (tasks.py) please take a look at that when you get a chance.

Once you're good with everything, I just need to write tests and we can merge this in. The tech writer can go ahead and start embedding these using the `datastore` branch.

pcostell:
Yes, can we pull the `an_hour_ago` out of the snippet? It just doesn't make sense in the context of everything. I know that last_seen_key is very vague, but for now that is intentional until we come up with an example that actually shows proper usage here.

(Also LGTM in general, will take a look at task_list now).

jonparrott:
> Yes, can we pull the an_hour_ago out of the snippet? 

Done.

pcostell:
LGTM.

jonparrott:
Great. Give me until the end of week or so to write tests. There's a lot to test here.

This shouldn't block the Walter from integrating the samples into the docs.

jonparrott:
@waprin FYI tests added. Will merge once tests pass if you do not object.

waprin:
sgtm

jonparrott:
Sigh. Testing datastore is like playing whack-a-mole. :( I'm marking all of these tests as slow and we can monitor them via the nightly builds.

dhermes:
@jonparrott RE: slow testing, you could always run them with the emulator. Try out `tox -e datastore-emulator` in `gcloud-python` to see how slick and easy it is to run production-ready code with no internet connection at all.

PR #167, opened 2016-02-03
Trying to improve datastore flaky tests.

jonparrott:
@waprin PTAL, there's some stuff in here I'm not proud of.

### DIFF
--- a/datastore/api/snippets_test.py
+++ b/datastore/api/snippets_test.py
@@ -11,21 +11,39 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-from flaky import flaky
-import gcloud
+from functools import wraps
+import time
+
 from gcloud import datastore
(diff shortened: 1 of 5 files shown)


waprin:
lgtm overall. I do think it might be slightly better to replace the client in setUp with a proxy client type thing that knows to intercept put. That way you don't have to put the same decorator on lots of methods and possibly miss some. but just a nitpick and debatable anyway.

Does 1s consistently work? Usually with these types of things I would do some sort of exponential backoff (say up to 8 seconds) on the read side of thing but if this works then sgtm.

jonparrott:
> I do think it might be slightly better to replace the client in setUp with a proxy client type thing that knows to intercept put. 

I did that at first but realized only a few tests were impacted by eventual consistency.

> Does 1s consistently work?

Typically, yeah. We can monitor it though and do exponential backoff later if needed.

PR #170, opened 2016-02-04
Provides eventual_consistent_query region tag with comment saying rea…

WalterHub:
…d consisteny cannot be specified in gcloud-python.


jonparrott:
LGTM, merging. In the future, try to keep the first line of the message < 80 characters. :)

PR #175, opened 2016-02-05
Add requirements-dev to reqcheck

waprin:


jonparrott:
LGTM. we should eventually split this out into its own library.

PR #184, opened 2016-02-15
BQ: Empty array when no results

hraban:
Better than KeyError

### DIFF
--- a/bigquery/api/sync_query.py
+++ b/bigquery/api/sync_query.py
@@ -66,7 +66,7 @@ def main(project_id, query, timeout, num_retries):
             pageToken=page_token,
             **query_job['jobReference']).execute(num_retries=2)
 
-        print(json.dumps(page['rows']))
+        print(json.dumps(page.get('rows', [])))
 
         page_token = page.get('pageToken')
         if not page_token:
--- a/bigquery/api/sync_query.py
+++ b/bigquery/api/sync_query.py
@@ -60,17 +60,21 @@ def main(project_id, query, timeout, num_retries):
 
     # [START paging]
     # Page through the result set and print all results.
+    results = []
     page_token = None
+
     while True:
         page = bigquery.jobs().getQueryResults(
             pageToken=page_token,
             **query_job['jobReference']).execute(num_retries=2)
 
-        print(json.dumps(page.get('rows', [])))
+        results.extend(page.get('rows', []))
 
         page_token = page.get('pageToken')
         if not page_token:
             break
+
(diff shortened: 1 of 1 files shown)


bigquery/api/sync_query.py
```diff
@@ -66,7 +66,7 @@ def main(project_id, query, timeout, num_retries):
             pageToken=page_token,
             **query_job['jobReference']).execute(num_retries=2)
 
-        print(json.dumps(page['rows']))
+        print(json.dumps(page.get('rows', [])))
```
jonparrott: I would prefer  `None` over an empty list. How does `json.dumps()` respond when passed `None`?

hraban:
```
$ python3 -c 'import json ; print(json.dumps(None))'
null
```
I went with [] because the return value is an array containing all the result rows, which feels like it should be an empty array when there are no rows. But I am almost apathetic towards this :P as long as it doesn't KeyError, I'm fine.

jonparrott: Heh, fair enough... How about we do even better and do this:

```python
rows = page.get('rows')

if rows:
    print json.dumps(rows)
else:
    print 'No results found.'
```

hraban: Not a fan. The appeal of this script is its composability through valid JSON output on stdout; you can pipe it to `jq` or other JSON tools. Incidentally this is also why I chose for []; by always returning an array you can just loop over the result no matter what.

jonparrott: Interesting, I don't think we considered that a primary use case here. How about:

```python
    # [START paging]
    # Page through the result set and print all results.
    results = []
    page_token = None

    while True:
        page = bigquery.jobs().getQueryResults(
            pageToken=page_token,
            **query_job['jobReference']).execute(num_retries=2)

        results.extend(page.get('rows', []))

        page_token = page.get('pageToken')
        if not page_token:
            break

    print(json.dumps(results))
    # [END paging]
```

hraban: Ah! so you're saying paging would break the original script's JSON output?

This is a nice solution, good idea.

The valid JSON output is very nice in combination with jq because the raw output can be a bit.. raw :)

+1 !

jonparrott: Cool, if you'll update your PR I'll get it merged. Thank you for your contribution!

hraban: Done.

jonparrott:
LGTM, merging. Thank you for your contribution!

PR #185, opened 2016-02-17
Moving to py.test test runner

jonparrott:
This is the first step in moving to py.test.

* Created `scripts/run-tests.py` that handles invoking `py.test` for all samples.
* Updated `tox.ini` to use `scripts/run-tests.py` for the `{py27,py34}{-all}` environments.
* Removed all unneeded `__init__.py` files. The repository is now a loose collection of independent packages instead of being one monster package.

Outstanding questions/issues:
* Currently pytest arguments are specified in `tox.ini`. Not sure if we want to move those into `run-tests.py` or not.
* App Engine is still using `nose` because of `nosegae`. It appears I need to consider creating a `py.test` plugin for app engine.
* App Engine can't use coverage because apparently nose-cover can't append to the coverage file generated by pytest-cov. But, once we switch appengine over to pytest we'll be fine.

After this is merged, I can look into converting our tests from `unittest` style to `py.test` style, which is generally cleaner.


jonparrott:
/cc @jerjou @waprin 

Note: this has #178 as a diffbase.

jonparrott:
App engine concerns have been fixed. I can now run the app engine tests with `py.test`. :)

jonparrott:
@waprin and/or @jerjou ready for your review

Most of this is me removing `__init__.py` files and switching relative imports to absolute imports.

The key files to review are:

* `scripts/run-tests.py` - which will probably eventually be moved to `gcp-python-repo-tools`.
* `testing/appengine.py`.
* `appengine/conftest.py`.
* `tox.ini`.

scripts/run-tests.py
```diff
@@ -0,0 +1,58 @@
...
+    parser.add_argument('--junit', action='store_true', help='Output junit test report.')
+    parser.add_argument('--run-slow', action='store_true', help='Run slow tests.')
+    parser.add_argument(
+        'directories', nargs='+', help='Directories to run py.test on.')
+    args = parser.parse_args()
+
+    extra_args = []
+
+    if not args.run_slow:
+        extra_args.append('-m not slow and not flaky')
```
jerjou: Does this work? I would expect `-m` and `not slow and not flaky` to have to be separate elements in the list..

jonparrott: Yep, it works either way.

tox.ini
```diff
@@ -19,111 +19,133 @@ passenv =
...
+    managed_vms/disk
+    managed_vms/extending_runtime
+    managed_vms/hello_world
+    managed_vms/hello_world_compat
+    managed_vms/memcache
+    managed_vms/pubsub
+    managed_vms/static_files
+    managed_vms/storage
+    monitoring/api
+    storage/api
```
jerjou: There isn't a way to auto-detect these? I can see myself adding a new example and forgetting that this list exists, running the tests, seeing that it passes and assuming that it included my tests when it didn't..

jonparrott: Not easily. There's no unique property for those folders. Any ideas?

jerjou: What if instead of an 'examples' list, you had either an `exclude` or `include` list, where `run-tests.py` has corresponding flags. If `--exclude`, then look for tests everywhere but the excluded dirs. If `--include`, only look for tests in the included dirs?

jerjou:
LGTM - you (or I) can make the `run-tests.py` change in a subsequent CL.

tox.ini
jonparrott: I'm not sure I understand; but feel free to make a PR so we can discuss.

jonparrott:
Thank you, I'll rebase to resolve conflict then merge.

PR #186, opened 2016-02-18
Passing all environment variables in tox.ini

jonparrott:


waprin:
lgtm

jonparrott:
For future context: this fixes jenkins because something is weird with jenkin's shell environment and tox's passenv.

jonparrott:
Travis failed because of reqcheck, which will be removed from travis with #185.

PR #187, opened 2016-02-18
Turning off character echoing in encrypt-secrets.sh

jonparrott:


scripts/encrypt-secrets.sh
```diff
@@ -15,7 +15,7 @@
 # limitations under the License.
 
 echo "Enter password for encryption: "
-read password
+read -s -p password
```
jerjou: Try:

    read -s -p "Enter password for encryption: " password
    echo

:-) The `-p` sets the prompt. (The `echo` afterwards is 'cuz apparently suppressing echo also suppresses the trailing endline...)

jonparrott:
@jerjou updated

jerjou:
LGTM

PR #188, opened 2016-02-18
Removing reqcheck from travis tests

jonparrott:
This is to fix travis in the interim between and and when we merge #185.


jonparrott:
or @jerjou.

jonparrott:
FYI merging, #185 has been LGTMed, so this will be rolled in as well.

jerjou:
LGTM

PR #189, opened 2016-02-18
Moving mvm samples to use py.test

jonparrott:


jonparrott:
Travis failed due to a flaky storage test - we can ignore for now and keep an eye on it - it may be due to how much we're hitting the service right now.

jerjou:
LGTM

PR #190, opened 2016-02-18
Moving storage samples to py.test

jonparrott:
@jerjou - there was also a slight refactor to the resource thingy in `cloud.py`. Once this repository stabilizes, I'll like move a lot of this over to `gcp-python-repo-tools`.

### DIFF
--- a/testing/cloud.py
+++ b/testing/cloud.py
@@ -47,6 +47,20 @@ class Config(object):
         return os.environ[name]
 
 
+def get_resource_path(resource, local_path):
+    global_resource_path = os.path.join(GLOBAL_RESOURCE_PATH, *resource)
+    local_resource_path = os.path.join(local_path, 'resources', *resource)
+
+    if os.path.exists(local_resource_path):
+        return local_resource_path
+
+    if os.path.exists(global_resource_path):
+        return global_resource_path
+
+    raise EnvironmentError('Resource {} not found.'.format(
+        os.path.join(*resource)))
+
+
 class CloudTest(unittest.TestCase):
     """Common base class for cloud tests."""
     def __init__(self, *args, **kwargs):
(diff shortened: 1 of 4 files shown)


testing/cloud.py
```diff
@@ -47,6 +47,20 @@ def __getattr__(self, name):
         return os.environ[name]
 
 
+def get_resource_path(resource, local_path):
```
jerjou: Any particular reason you didn't stick with the same method signature as `resource_path`? eg with the variable-length `*resource`?

jonparrott: `local_path` makes things awkward.

jerjou: Howso?

jonparrott: Hmm. I guess I could make it the first argument if you want - that would more or less fix it.

jerjou: Oh I see what the difference is. I didn't realize that you weren't *necessarily* getting the resource from the local path. I see where the awkwardness lies now.

Okay - it's fine the way it is. Suggestion if you like it:

    def get_resource_path(*resource, local_path=None):

and add a check for `local_path` before using it. That feels marginally more semantic for me, but whatever.

jerjou:
LGTM

testing/cloud.py
jonparrott: You can do that with python 3, but not 2. It's invalid syntax to have vargs before kwargs.

jonparrott: I'll likely refactor this a bit anyway when I lift it into python-repo-tools.

TRIMIXING1AI: TWEAKING 🎼BEATS #TRIXMIXING1AI ♥️

TRIMIXING1AI: TRIXMI@trixmixing1ai.com

PR #192, opened 2016-02-19
Moving datastore samples to py.test

jonparrott:


datastore/api/snippets_test.py
```diff
@@ -11,291 +11,269 @@
...
-    @wraps(f)
-    def inner(self, *args, **kwargs):
-        # This is pretty hacky, but make datastore wait 1s after any
-        # put operation to in order to account for eventual consistency.
-        original_put_multi = self.client.put_multi
+class CleanupClient(datastore.Client):
+    def __init__(self, *args, **kwargs):
+        super(CleanupClient, self).__init__(*args, **kwargs)
+        self.entities_to_delete = []
+        self.keys_to_delete = []
```
jerjou: Could you wrap the relevant methods to automatically add to these lists? I'm thinking something like:

    def record_entity(method, *args, **kwargs):
        entity = method(*args, **kwargs)
        self.entities_to_delete.append(entity)
        return entity
    entity_methods_to_wrap = ('insert', 'lookup', ...)
    for method in entity_methods_to_wrap:
        setattr(self, method, record_entity(getattr(self, method)))

That way if someone ever needs to add a test to this, they don't have to remember to cleanup.

Granted: this is just how the file was doing it before, so don't feel obligated to do this - just a comment.

jonparrott: The problem is that not everything needs to be deleted. Some of the queries do thing with stats, namespaces, and other meta objects. The tests themselves add stuff to the lists, so there shouldn't be any issue for someone else running these tests.

datastore/api/snippets_test.py
```diff
@@ -11,291 +11,269 @@
...
         return result
-    return inner
 
 
-@mark_flaky
-class DatastoreSnippetsTest(CloudTest):
+@pytest.yield_fixture
+def client(cloud_config):
+    client = CleanupClient(cloud_config.GCLOUD_PROJECT)
+    yield client
```
jerjou: Might want to put this in a try..finally block

datastore/api/snippets_test.py
```diff
@@ -11,291 +11,269 @@
...
 
-    def setUp(self):
-        super(DatastoreSnippetsTest, self).setUp()
-        self.client = datastore.Client(self.config.GCLOUD_PROJECT)
-        self.to_delete_entities = []
-        self.to_delete_keys = []
+@pytest.yield_fixture
+def waiting_client(cloud_config):
+    client = WaitingClient(cloud_config.GCLOUD_PROJECT)
+    yield client
```
jerjou: ditto

datastore/api/snippets_test.py
jonparrott: py.test handles it.

> While the yield syntax is similar to what contextlib.contextmanager() decorated functions provide, with pytest fixture functions the part after the “yield” will always be invoked, independently from the exception status of the test function which uses the fixture.

datastore/api/snippets_test.py
jonparrott: see above.

jerjou:
LGTM

PR #214, opened 2016-03-02
Fix up Django URLs

waprin:
Managed VMs tutorial should serve the polls app from the index.

Container Engine url should also have a dollar sign (not technically necessary but good for consistency).


jonparrott:
LGTM, merge whenever.

PR #218, opened 2016-03-09
Typo fix in samples/datastore/api/tasks.py

relip:
In https://cloud.google.com/datastore/docs/datastore-api-tutorial#storing_data, `build_service` block starts with `import gcloud` so calling create_client() throws an error that `datastore` is undefined.


relip:
I signed it!

jonparrott:
Thank you for your contribution!

PR #219, opened 2016-03-11
Fix pep8 style errors.

jerjou:


jonparrott:
so `container_engine/django_tutorial/polls/migrations/0001_initial.py` is a generated file. Is it not being ignored?

jerjou:
Oh - not when I run it on my local box, although I noticed that travis didn't seem to care. What mechanism ignores it?

waprin:
It's probably easiest just to delete the file. Even though you might check them into a real project and setup all the ignore stuff, for our cases probably easiest just to let users generate them and add them to .gitignore

jonparrott:
The `tox -e lint` environment should automatically ignore it. There's a `[flake8]` section in the `tox.ini`. (More magic that I'm looking forward to removing)

jerjou:
Do you mean `tox -e pep8`? It doesn't seem to be ignoring it on my local box :(

jonparrott:
LGTM.

PR #220, opened 2016-03-12
Get rid of Django URL warnings

waprin:
Sorry, last PR for this wasn't quite this right, this should be right.


jonparrott:
LGTM

PR #222, opened 2016-03-16
Restoring region tags

feczo:
commit/5207559f92013ebd3253c8aaa45a568800dd31d8 broke autogenerated docs
Restoring region tags


feczo:
I signed it!

jerjou:
FYI this shouldn't be necessary, per other email thread. Let me know if I'm missing something.

PR #226, opened 2016-03-17
Fix metric.name

waprin:
All the filters should be metric.type (API
change, name is now the fully qualified
REST resource).


waprin:
Fixes #227

peaton:
LGTM

PR #228, opened 2016-03-22
Monitoring Doc Fixups

waprin:
Just a README change so I think we should merge.

This recommends the new gcloud ADC command.


jonparrott:
Yep, LGTM.

PR #234, opened 2016-03-25
Noxify

jonparrott:


waprin:
LGTM. One thing I'm having trouble following that might be worth some extra docstrings is the argument passing. Like what is the difference between session.posargs and extra_pytest args in terms of where they come from?

jonparrott:
> Like what is the difference between session.posargs and extra_pytest args in terms of where they come from?

That's fair, I'm just trying to prevent repeating myself too much. posargs comes from the command line, extra_pytest_args can come from a calling function. I'll update those before merge.

PR #235, opened 2016-03-28
Use Main Project (now whitelisted)

waprin:
Now use main integration test project because it can be whitelisted.

Fix all variables names of metric name to metric type. 

Even though I was experimenting with flaky, the custom metric tests now seem less flaky so I think for now we can run them as default.


jonparrott:
LGTM, merging.

PR #237, opened 2016-03-31
Auto-detect directories to test.

jerjou:
This has the advantage of discovering dirs that weren't being tested - eg, apparently `./managed_vms/extending_runtime_compat` wasn't being tested (in addition to speech/)

Also, add speech/ to .coveragerc

Addresses #236


nox.py
```diff
@@ -94,6 +58,22 @@ def session_reqcheck(session):
...
 
+def collect_sample_dirs(start_dir, blacklist=set()):
+    """Recursively collects a list of dirs that contain tests."""
+    # Collect all the directories that have tests in them.
+    samples = []
+    for parent, subdirs, files in os.walk(start_dir):
+        if any(f for f in files if f[-8:] == '_test.py'):
+            # This dir has tests in it. Add it, and stop recursion
+            samples.append(parent)
+            del subdirs[:]
```
waprin: can just yield parent and get rid of the samples list

nox.py
```diff
@@ -94,6 +58,22 @@ def session_reqcheck(session):
         session.run('gcprepotools', command, reqfile)
 
 
+def collect_sample_dirs(start_dir, blacklist=set()):
+    """Recursively collects a list of dirs that contain tests."""
+    # Collect all the directories that have tests in them.
+    samples = []
+    for parent, subdirs, files in os.walk(start_dir):
+        if any(f for f in files if f[-8:] == '_test.py'):
+            # This dir has tests in it. Add it, and stop recursion
```
waprin: why stop recursion? I don't think we need recursion but also don't see any reason to disable it in case someone wants multiple levels of tests to organize them better.

nox.py
```diff
@@ -94,6 +58,22 @@ def session_reqcheck(session):
...
+            samples.append(parent)
+            del subdirs[:]
+        else:
+            # Filter out dirs we don't want to recurse into
+            subdirs[:] = [s for s in subdirs
+                          if re.match(r'[a-z]', s) and s not in blacklist]
```
waprin: more of an opinion thing but I think s[0].isAlpha() is better since it accomplishes the same thing without adding re as a dependency.

nox.py
jerjou: py.test will recurse down the subdirs when it's invoked on a dir, so recursing down a dir we've already collected is redundant.

You might ask - why manually collect subdirs here if py.test does it already? I believe this is because we want to run each of our manually-collected subdirs in their own virtualenvs, which pytest doesn't do, but you can ask Jon for the definitive answer :-)

nox.py
jerjou: Oh right. I always forget about yield...

nox.py
jerjou: I agree.

waprin:
lgtm

jonparrott:
Clever. Let's see if this returns to hurt us in the future. :)

jerjou:
I look forward to the day we rant about how past-jerjou dug us into this mess.

PR #239, opened 2016-04-04
Add speech api streaming sample.

jerjou:
/cc @jonparrott @waprin


jerjou:
Oh wait I should write a test shouldn't I...

jerjou:
Okay, added tests.
However, the sample doesn't seem to work in python3 (it segfaults). After digging around a bunch, I _think_ the problem is in the generated grpc lib. I skipped the test for python 3 for now..

PTAL

nox.py
```diff
@@ -86,6 +86,11 @@ def session_tests(session, interpreter, extra_pytest_args=None):
     # allows users to run a particular test instead of all of them.
     for sample in (session.posargs or
                    collect_sample_dirs('.', SESSION_TESTS_BLACKLIST)):
+        # Install additional dependencies if they exist
```
jonparrott: This could cause some weirdness with clashing dependency versions. I'm okay with this for now, but we should be careful.

speech/api/README.md
```diff
@@ -37,10 +37,36 @@ See the
 [Cloud Platform Auth Guide](https://cloud.google.com/docs/authentication#developer_workflow)
 for more information.
 
+### Install the dependencies
+
+* If you're running the `speechrest.py` sample:
+
+    ```sh
+    $ pip install speechrest-requirements.txt
```
jonparrott: These are backwards (should be `requirements-speechrest.txt`)

speech/api/README.md
```diff
@@ -37,10 +37,36 @@ See the
...
+
 ## Run the example
 
-```sh
-$ python speechrest.py resources/audio.raw
-```
+* To run the `speechrest.py` sample:
+
+    ```sh
+    $ python speechrest.py resources/audio.raw
```
jonparrott: nit: speechrest has no underscore, but speach_streaming does.

speech/api/speech_streaming.py
```diff
@@ -0,0 +1,123 @@
+#!/usr/bin/python
+
+import contextlib
+import threading
+
+from gcloud.credentials import get_credentials
+
+from google.cloud.speech.v1.cloud_speech_pb2 import *  # noqa
+from google.rpc import code_pb2
```
jonparrott: why the weird newlines?

speech/api/speech_streaming.py
```diff
@@ -0,0 +1,123 @@
...
+RATE = 16000
+CHANNELS = 1
+CHUNK = RATE // 10  # 100ms
+
+# Keep the request alive for this many seconds
+DEADLINE_SECS = 8 * 60 * 60
+SPEECH_SCOPE = 'https://www.googleapis.com/auth/cloud-platform'
+
+
+def _make_channel(host, port):
```
jonparrott: This functions aren't part of a module, so there's no need to make them private.

speech/api/speech_streaming.py
```diff
@@ -0,0 +1,123 @@
...
+    composite_channel = implementations.composite_channel_credentials(
+            ssl_channel, auth_plugin)
+
+    return implementations.secure_channel(host, port, composite_channel)
+
+
+@contextlib.contextmanager
+def _record_audio(channels, rate, chunk):
+    """Opens a recording stream in a context manager."""
+    p = pyaudio.PyAudio()
```
jonparrott: Please use a variable name more descriptive than `p`.

speech/api/speech_streaming.py
```diff
@@ -0,0 +1,123 @@
...
+            # Subsequent requests can all just have the content
+            audio_request = AudioRequest(content=audio_stream.read(chunk))
+
+            yield RecognizeRequest(audio_request=audio_request)
+
+
+def listen_print_loop(recognize_stream):
+    for resp in recognize_stream:
+        if resp.error.code != code_pb2.OK:
+            raise Exception('Server error: ' + resp.error.message)
```
jonparrott: Please raise an appropriate subclass, such as `RuntimeError` or `ValueError`.

speech/api/speech_streaming.py
```diff
@@ -0,0 +1,123 @@
...
+
+        # Display the transcriptions & their alternatives
+        for result in resp.results:
+            print(result.alternatives)
+
+        # Exit recognition if any of the transcribed phrases could be
```
jonparrott: Nice.

jonparrott:
Finished first review pass. In general, I like this but I'm unsure about the control flow. It seems weird to me, but most threaded things are. hmm..

nox.py
jerjou: Yeah - I think we should move to a world where, instead of having a global requirements-py[23].txt at the top level, each sample module would have its own smaller requirements.txt, and the global one would be either empty, or only include the stuff needed for pytest.

speech/api/README.md
jerjou: Oh yup - vestige from an earlier iteration. Will fix.

speech/api/README.md
jerjou: Yeah - I was considering adding an underscore to speechrest, but got lazy 'cuz then I'd have to change the references to it -_-

Will update in a later PR

speech/api/speech_streaming.py
jerjou: I blame flake8!
Oh.. okay. After some consulting with the internets, I added a config to my flake8 to have the import-order-style be 'google', and it's happy without all the newlines now.

Will fix.

speech/api/speech_streaming.py
jerjou: But that's what [pyaudio](https://people.csail.mit.edu/hubert/pyaudio/#examples) does!

...will fix.

jerjou:
Updated.

Re: control flow - yeah, I had an iteration where I made the request stream a class that was an iterator, with a `stop` method. It's probably the Right Way to do it, but it looked uglier / harder to explain, because then I had to deal with the `record_audio` context around the iteration.. so I figured this was better just because it's easier to see what's going on, even if the control flow is a bit icky.

(something like:

    class RequestStream:
        def __init__(self):
            self.stop_audio = threading.Event()
            self.audio_stream = record_audio(..)
            self.is_first_request = True
        def __iter__(self):
            self.audio_stream.__enter__()
        def __next__(self):
            if self.is_first_request:
                self.is_first_request = False
                # ... do the special thing
            else if not self.stop_audio.is_set():
                # do the normal thing
            else:
                self.audio_stream.__exit__()
                raise StopIteration()

Just seemed like a lot of different pieces were interleaved.
I dunno - open to suggestions.

Travis is having trouble pulling down grpcio... will look into that.

jonparrott:
Please rebase, but otherwise I suppose this LGTM. I would appreciate some more time put into the control flow. Perhaps we can chat about it in person next week?

nox.py
jerjou: Oh I see what you mean - the virtualenvs are split by session, not by subdir.
Hrm... that's a puzzle... perhaps we could have a separate virtualenv for each subdir that contains a requirements.txt?

Pros:
* Also tests that requirements.txt are valid
Cons:
* Adds overhead to testing

The con might be mitigated if we could run the tests in parallel..

jonparrott: I tried that originally, and it's dreadfully slow. :/

jerjou:
Rebased.

What do you have in mind for the control flow? At this point it's not a matter of time so much as lack of better ideas.

jonparrott:
> What do you have in mind for the control flow? At this point it's not a matter of time so much as lack of better ideas.

Not sure, I'd have to more deeply understand a few things. Much better to discuss in person. Feel free to merge.

PR #241, opened 2016-04-06
Fix property name in projection_query

quartzmo:
Replace `priority` with `percent_complete` in `projection_query` example.


jonparrott:
Thanks for the change!

quartzmo:
@jonparrott You bet! Thanks for the examples! I'm adapting them to Ruby.

PR #248, opened 2016-04-11
App Engine Task Queue Sample.

jeffmendoza:
Counter app. Copied from: https://github.com/GoogleCloudPlatform/appengine-taskqueue-counter
Added tests.

### DIFF
--- a/appengine/taskqueue/counter/app.yaml
+++ b/appengine/taskqueue/counter/app.yaml
@@ -1,0 +1,11 @@
+runtime: python27
+api_version: 1
+threadsafe: true
+
+handlers:
+- url: /.*
+  script: main.app
+
+libraries:
+- name: jinja2
+  version: 2.6
--- a/appengine/taskqueue/counter/main_test.py
+++ b/appengine/taskqueue/counter/main_test.py
@@ -1,0 +1,29 @@
+# Copyright 2016 Google Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from google.appengine.ext import ndb
+import main
+import webtest
+
+
+def test_app(testbed, run_tasks):
+    key_name = 'foo'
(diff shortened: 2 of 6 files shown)


appengine/taskqueue/counter/app.yaml
```diff
@@ -0,0 +1,13 @@
+application: taskqueue-counter
```
jonparrott: Remove application and version from this.

appengine/taskqueue/counter/app.yaml
```diff
@@ -0,0 +1,13 @@
...
+version: 1
+threadsafe: true
+
+handlers:
+- url: /.*
+  script: main.app
+
+libraries:
+- name: jinja2
+  version: latest
```
jonparrott: please use an explicit version here.

appengine/taskqueue/counter/queue.yaml
```diff
@@ -0,0 +1,3 @@
+queue:
```
jonparrott: Is this used in the doc? If not, it's not necessary.

appengine/taskqueue/counter/main_test.py
```diff
@@ -0,0 +1,35 @@
...
+import main
+import webtest
+from google.appengine.ext import ndb
+
+
+def test_app(testbed, run_tasks):
+    key_name = 'foo'
+    key = ndb.Key('Counter', key_name)
+    counter = key.get()
+    if counter:
```
jonparrott: counter should never be set at this point in the test.

jeffmendoza: ok, I didn't know if the local datastore persisted state, I'll refactor

appengine/taskqueue/counter/queue.yaml
jeffmendoza: Looks like the first snippit in https://cloud.google.com/appengine/docs/python/config/queue uses it

jonparrott: Can you add the comment that's there here?

jeffmendoza:
Addressed comments, fixed linter errors.

appengine/taskqueue/counter/app.yaml
```diff
@@ -0,0 +1,11 @@
...
+api_version: 1
+threadsafe: true
+
+handlers:
+- url: /.*
+  script: main.app
+
+libraries:
+- name: jinja2
+  version: 2.8
```
jonparrott: Seems like only "2.6" is available on GAE.

jonparrott:
One last comment to address, and this is good to merge.

appengine/taskqueue/counter/app.yaml
jeffmendoza: Updated, is there a way to check?

PR #249, opened 2016-04-12
Use gcloud 0.12.0 for Datastore samples so they use v1beta3.

pcostell:
@dhermes -- will you take a look?


jonparrott:
LGTM, we need to run our update requirements script anyway.

PR #257, opened 2016-04-13
Adding app engine request handling samples.

jonparrott:


appengine/requests/README.md
```diff
@@ -0,0 +1,10 @@
+## App Engine Requests Docs Snippets
+
+These snippets demonstrate various aspects of App Engine Python request handling.
+
+<!-- auto-doc-link -->
```
jerjou: What are these auto-doc-link comments I'm seeing?

appengine/requests/main.py
```diff
@@ -0,0 +1,66 @@
...
+"""
+
+import os
+import time
+
+import webapp2
+
+
+# [START request_timer]
+class TimerHandler(webapp2.RequestHandler):
```
jerjou: FYI you can do indented_block="class TimerHandler" to pull in the entire class

jerjou:
LGTM

The region tags shouldn't be necessary, but I won't block on it.

appengine/requests/README.md
jonparrott: there's a script that can automatically populate these. I need to re-run it once we're done moving all samples.

PR #268, opened 2016-04-20
Adding remote api sample

jonparrott:


waprin:
lgtm agree it looks close to impossible to really test

jonparrott:
Not without deploying and/or import black magic. Oh well.

PR #270, opened 2016-04-21
Adding metadata sample

jonparrott:


compute/metadata/main.py
```diff
@@ -0,0 +1,77 @@
...
+            time.sleep(1)
+            continue
+        # [END hanging_get]
+
+        last_etag = r.headers['etag']
+
+        if r.text == 'MIGRATE_ON_HOST_MAINTENANCE':
+            in_maintenance = True
+        else:
+            in_maintenance = False
```
waprin: in_maintenance =  (r.text == 'MIGRATE_ON_HOST_MAINTENANCE')

jonparrott: Meh, too clever. This a bit more clear.

compute/metadata/main.py
```diff
@@ -0,0 +1,77 @@
...
+METADATA_HEADERS = {'Metadata-Flavor': 'Google'}
+
+
+def wait_for_maintenance(callback):
+    url = METADATA_URL + 'instance/maintenance-event'
+    last_in_maintenance = False
+    # [START hanging_get]
+    last_etag = 0
+
+    while True:
```
waprin: probably missing something...looks like this code should terminate when status is false but it doesn't look like logic exists to do so?

jonparrott: It never terminates. It constantly watches for migration events and calls the callback. The callback can decide to terminate or whatever.

compute/metadata/main.py
waprin: Well I disagree, it's not nested in any way, it's not clever as much as what I consider to be standard Pythonic syntax. BUT /bikeshed, don't want to block code purple work etc so whatever.

waprin:
lgtm

PR #272, opened 2016-04-21
Add Error Reporting Code

waprin:
I know the testing is a little lazy but it's similar to what Jason did for Node.

I did manually test the startup script and it did work. 

I can write an e2e using the API but it's obviously going to be more involved/flaky, so give CDCP and the fact that I'm OOO till Tuesday I thought this is a reasonable starting point.


error_reporting/main.py
```diff
@@ -0,0 +1,39 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# [START error_reporting]
+from fluent import sender
+from fluent import event
+import traceback
```
jonparrott: traceback is a standard library import, and thus should go in the section above fluent

error_reporting/main.py
```diff
@@ -0,0 +1,39 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# [START error_reporting]
+from fluent import sender
+from fluent import event
```
jonparrott: Prefer just `import fluent` and use `fluent.sender` and `fluent.event`.

jonparrott:
Looks mostly good, make travis happy and merge.

error_reporting/main.py
waprin: can't just import fluent since sender and event are modules. can import fluent.event and fluent.sender and make everything more explit but still need the two imports

jonparrott: that's fine then.

PR #274, opened 2016-04-21
moved image guestbook, added basic image example

thesandlord:
I moved the current contents of "appengine/images" to "appengine/images/guestbook" and added "appengine/images/basic" that will host the code currently on https://cloud.google.com/appengine/docs/python/images/usingimages


jonparrott:
Can you move /basic to /api, to be consistent with other samples?

appengine/images/guestbook/app.yaml
```diff
@@ -0,0 +1,17 @@
+# This file specifies your Python application's runtime configuration
```
jonparrott: You can strip all of the comments out of this file. They're only needed for hello world.

jonparrott:
LGTM with two minor nits. 👍 Thanks, @thesandlord.

thesandlord:
done and done

PR #279, opened 2016-04-25
Adding compute engine private key generation sample.

jonparrott:


compute/encryption/generate_wrapped_rsa_key.py
```diff
@@ -0,0 +1,80 @@
...
+        private_key_bytes,
+        padding.OAEP(
+            mgf=padding.MGF1(algorithm=hashes.SHA1()),
+            algorithm=hashes.SHA1(),
+            label=None))
+    encoded_wrapped_key = base64.b64encode(wrapped_key)
+    return encoded_wrapped_key
+
+
+def main():
```
tmatsuo: How about to make this script a real utility by just printing out the `wrapperd_rsa_key`, and receiving the csek filename as an argument? That way, customers can use this script like following:

```
cd compute/encryption/
pip install -r requirements.txt
touch /somewhere/my_wrapped_key
chmod 0600 /somewhere/my_wrapped_key
generate_wrapped_rsa_key.py /somewhere/my_private_key > /somewhere/my_wrapped_key
gcloud beta compute instances create example-instance --csek-key-file /somewhere/my_wrapped_key
```

Of course you can write a python script that uses functions in this file, but I think it's easier for some users to just use shell scripting.

compute/encryption/generate_wrapped_rsa_key_test.py
```diff
@@ -0,0 +1,18 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import generate_wrapped_rsa_key
+
+
+def test_main():
+    generate_wrapped_rsa_key.main()
```
tmatsuo: Good for initial test. How about to write an e2e test that performs one of the operations listed on:
https://cloud.google.com/compute/docs/disks/customer-supplied-encryption
and runs it on jenkins?

tmatsuo:
The code LGTM with few minor suggestions.

compute/encryption/generate_wrapped_rsa_key_test.py
jonparrott: You can file a bug for that, but that's a really large test. It'd have to create an instance with an encrypted disk and then run tests from there. Yikes.

compute/encryption/generate_wrapped_rsa_key.py
jonparrott: Fair enough, I'll do that.

jonparrott: Done.

compute/encryption/generate_wrapped_rsa_key.py
```diff
@@ -0,0 +1,93 @@
...
+    if not key_file:
+        customer_key_bytes = os.urandom(32)
+    else:
+        with open(key_file, 'rb') as f:
+            customer_key_bytes = f.read()
+
+    google_public_key = get_google_public_cert_key()
+    wrapped_rsa_key = wrap_rsa_key(google_public_key, customer_key_bytes)
+
+    print('Base-64 encoded private key: {}'.format(
```
tmatsuo: It will print out:

```
Base-64 encoded private key: xxxxxxxxxx
Wrapped RSA key: xxxxxxxx
```
so it can not be used for producing the file which you can use with gcloud. Does it make sense to only print out the wrapped key so that you can use this script in your shell scripting?

Maybe I'm wrong, and this use case should be covered by `gcloud` sdk (gcloud should provide wrap-private-key subcommand, or something) though.

compute/encryption/generate_wrapped_rsa_key_test.py
tmatsuo: That's a tough test, but IMO only checking if one of the operations succeeds is a good first step. I will file an issue.

compute/encryption/generate_wrapped_rsa_key.py
jonparrott: sed can extract it if someone is hard-headed enough to use this for shell scripting but too adverse to actually modify the file. :P

PR #282, opened 2016-04-26
Add snippets for ndb cache doc

jerjou:


appengine/ndb/cache/snippets.py
```diff
@@ -0,0 +1,37 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+def set_in_process_cache_policy(ctx, func):
+    ctx.set_cache_policy(func)
```
jonparrott: I'd prefer if you actually showed `ctx = ndb.get_context()` here - and for all other snippets.

Alternatively you can one-line it with `ndb.get_context().set_cache_policy(func)`.

jonparrott: Can you actually demonstrate an example func here (and elsewhere) as well?

jerjou: Added explicit context.

jerjou: I agree that demonstrating an example func is good - but the flow of the doc page uses these snippets to build up to that. ie it first just shows an invocation with the 'func' placeholder. Then it describes what the func is, and then shows an example of it.

jonparrott: Gotcha. SGTM.

jerjou:
LGTY?

jonparrott:
Sure.

PR #289, opened 2016-04-27
Adding endpoints multiapi sample.

jonparrott:


jonparrott:
@waprin FYI merging so I can continue with large docs update. I'm happy to address any issues with a follow-up.

PR #306, opened 2016-04-28
Add Missing XMPP Snippet 

waprin:


appengine/xmpp/xmpp.py
```diff
@@ -68,10 +68,25 @@ def post(self):
         # [END error]
 
 
+# [START send-chat-to-user]
+class SendChatHandler(webapp2.RequestHandler):
+    def post(self):
+        user_address = 'example@gmail.com'
+        msg = ('Someone has sent you a gift on Example.com.'
```
jerjou: Probably want to add a space or two after the period. Otherwise, it'll say:

    ...on Example.com.To view: ...

jerjou:
LGTM

appengine/xmpp/xmpp.py
waprin: done

PR #311, opened 2016-04-29
Add samples for ndb async ops.

jerjou:


appengine/ndb/async/app_async.py
```diff
@@ -0,0 +1,44 @@
...
+    def get(self):
+        acct = Account.get_by_id(users.get_current_user().user_id())
+        acct.view_counter += 1
+        future = acct.put_async()
+
+        # ...read something else from Datastore...
+        template_values = {}
+
+        self.response.out.write(template.render(template_values))
+        future.get_result()
```
jonparrott: If you're not going to eat that future, can I have it?

(you should do something with the result, log it, maybe?)

appengine/ndb/async/app_async.py
```diff
@@ -0,0 +1,44 @@
...
+
+app = webapp2.WSGIApplication([('/', MyRequestHandler)])
+
+# Set up a generic template just for demonstration
```
jonparrott: Can we just remove this? It just needlessly complicates the sample.

appengine/ndb/async/app_async_test.py
```diff
@@ -0,0 +1,36 @@
...
+@pytest.fixture
+def app(testbed):
+    return webtest.TestApp(app_async.app)
+
+
+def test_main(app, testbed):
+    app_async.Account(id='123', view_counter=4).put()
+
+    # Log the user in
+    testbed.setup_env(
```
jonparrott: We have a login fixture: https://github.com/GoogleCloudPlatform/python-repo-tools/blob/master/gcp/testing/appengine.py#L113

appengine/ndb/async/app_toplevel_test.py
```diff
@@ -0,0 +1,39 @@
...
+import pytest
+import webtest
+
+
+@pytest.fixture
+def app(testbed):
+    return webtest.TestApp(app_toplevel.app)
+
+
+@pytest.skip('This succeeds when run alone, but fails when run with the other '
```
jonparrott: This is because py.test imports these all together. You can separate them into folders to get things to pass.

appengine/ndb/async/guestbook.py
```diff
@@ -0,0 +1,101 @@
+# Copyright 2016 Google Inc. All rights reserved.
```
jonparrott: If I see another guestbook I'm going to lose it.

appengine/ndb/async/guestbook.py
```diff
@@ -0,0 +1,101 @@
...
+        self.response.out.write('<html><body>{}</body></html>'.format(''.join(
+            '<p>{}</p>'.format(entry.content) for entry in recent_entries)))
+
+        return acct, recent_entries
+
+    def get_messages_sync(self):
+        qry = Message.query().order(-Message.when)
+        for msg in qry.fetch(20):
+            acct = msg.author.get()
+            self.response.out.write(
```
jonparrott: please use format.

appengine/ndb/async/app_async.py
jerjou: The text around the sample explains this. Basically it says - you have to call it, so the request will block until the value actually gets written.

appengine/ndb/async/app_async.py
jerjou: What would you suggest? The original sample has:

    # ...read something else from Datastore...
    template.render(...)

which still requires a `template` variable...

jonparrott: Don't render the template? Just write out. You're already handwaving a lot in this sample.

appengine/ndb/async/app_async_test.py
jerjou: Oh neat. Should have checked for that first -_-

jonparrott: It's fine. That's what I'm here for. :)

jerjou:
Merged per async lgtm from Jon

PR #320, opened 2016-05-03
Adding sample of how to use requests with GAE standard

jonparrott:


appengine/urlfetch/requests/main.py
```diff
@@ -0,0 +1,48 @@
...
+
+app = Flask(__name__)
+
+
+@app.route('/')
+def index():
+    # [START requests_get]
+    # Use the App Engine Requests adapter. This makes sure that Requests uses
+    # URLFetch.
+    requests_toolbelt.adapters.appengine.monkeypatch()
```
jerjou: Might be nice to monkeypatch this globally instead of when the function is called

jerjou:
LGTM

appengine/urlfetch/requests/main.py
jonparrott: I tried, but the logic hits the urlfetch service which does not have a stub until the tests initialize them. 

Actually I have a solution. Will fix.

PR #321, opened 2016-05-03
Add a port flag to the endpoints example

elibixby:
I need to be able to specify the port for use in the GCE example.

### DIFF
--- a/managed_vms/endpoints/main.py
+++ b/managed_vms/endpoints/main.py
@@ -18,6 +18,7 @@ Demonstrates how to create a simple echo API as well as how to deal with
 various authentication methods.
 """
 
+import argparse
 import base64
 import json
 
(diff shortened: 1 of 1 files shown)


jonparrott:
Why not just make the GCE sample use 8080?

jonparrott:
Also if you're not running this locally you should never do `python main.py`.

You should execute it with gunicorn `gunicorn -b :$PORT main:app`.

elibixby:
(1) The default config that comes with the debian install of endpoints uses 8080 as the serve port and 8081 as the proxy port, and Seph thinks we should stick with the defaults. Also I tried to modify it to serve on 80 and proxy to 8080 but the nginx conf is very sticky to modify. 

(2) I'm running it directly on a GCE VM... should I still use gunicorn?

jonparrott:
(2) absolutely you should if you're responding to non-local traffic

jonparrott:
@elibixby I'm closing this, we can sync via chat if you have questions on how to get this running on GCE.

PR #323, opened 2016-05-04
json -> data arg

elibixby:


jonparrott:
Hmm, why is this needed? The server accepts json, not POST params.

elibixby:
If data is a dict isn't it serialized by default to json as the response body (and the mimetype is set to application/json). In any case `json` isn't a recognized kwarg.

jonparrott:
> If data is a dict isn't it serialized by default to json as the response body (and the mimetype is set to application/json).

Data is for form encoded (`application/x-www-form-urlencoded)` data.

> In any case json isn't a recognized kwarg.

Yes, it [is](http://docs.python-requests.org/en/master/user/quickstart/#more-complicated-post-requests). What version of requests are you using?

elibixby:
arrrrg another brain fart for me.

PR #325, opened 2016-05-05
Remove beta settings from memcache app.yam

waprin:


jonparrott:
LGTM. Is this the only instance of this?

waprin:
Good point, looks like we should take another look at cloudssql and endpoints, which have different beta settings.

PR #326, opened 2016-05-05
Change app to listen on 8081

oaktowner:
the proxy listens on 8080 and forwards to 8081


jonparrott:
As I mentioned to @elibixby in #321, you should not use `python main.py` to run this application in production. It's only for running locally as the werkzeug server is single-threaded and has all kinds of debugging stuff enabled that makes it unusable and dangerous to use in production. Gunicorn is listed in `requirements.txt`, you should start the app with `gunicorn -b :8081 main:app`.

PR #327, opened 2016-05-05
Making nox skip unchanged samples.

jonparrott:


nox.py
```diff
@@ -73,37 +53,37 @@ def collect_sample_dirs(start_dir, blacklist=set()):
...
+        changed = subprocess.check_output(
+            ['git', 'show', '--pretty=format:', '--name-only',
+                os.environ.get('TRAVIS_COMMIT_RANGE')])
+    elif pr is not None:
+        changed = subprocess.check_output(
+            ['git', 'diff', '--name-only',
+                os.environ.get('TRAVIS_COMMIT'),
+                os.environ.get('TRAVIS_BRANCH')])
+    else:
+        changed = ''
```
waprin: maybe raise since this should never happen?

jonparrott: Not yet, I want to keep and eye on this first.

waprin:
code lgtm...i feel like some of the nox.py could be refactored to be a bit more declarative but im still thinking out how.

jonparrott:
Yeah, I cleaned it up a bit this round, but I'm wondering how much more we could clean it and how much of this we could shove in to repo-tools.

PR #328, opened 2016-05-05
Don’t send name in create custom metric

### LINKED ISSUE #269: create_custom_metric() shouldn't send "name"
The sample code for creating a custom metric ([create_custom_metric()](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/monitoring/api/v3/custom_metric.py#L62)) includes the `name` field in the request body. It shouldn't. The service ignores this field in the request (but includes it in the response).

waprin:
Fixes #269


jonparrott:
LGTM, let's see how travis works. :)

PR #330, opened 2016-05-05
create dockerfile to run echo service

wlu2016:
endpoints alpha doc will need to link to this dockerfile


jonparrott:
This will actually break the flexible samples, as the presence of a `Dockerfile` will cause `gcloud app deploy` to complain about `runtime` not being set to `custom`.

We have two options:
* We could put the Dockerfile elsewhere or by a different name and tell the user to copy it.
* we can duplicate this sample to `container_engine/endpoints`.

wlu2016:
if I rename this to echo.Dockerfile, will this solve that problem?

By the way, like the idea of having a sample in container_engine/endpoints. But I am hesitating on causing code duplicacy. Let me know what you would prefer :)

wlu2016:
they don't need to copy it. we document in the user doc and ask them to run:
docker build -t image_path ./echo.Dockerfile

jonparrott:
I'd prefer `Dockerfile.container-engine`.

wlu2016:
done. in the same directory? not sure if this will pass the CI.

By the way, this sample will be used for both gce and gke.

jonparrott:
For the record @wlu2016 is a Googler and no CLA is needed.

managed_vms/endpoints/Dockerfile.container-engine
```diff
@@ -0,0 +1,13 @@
+FROM debian:jessie
+
+RUN apt-get update && \
+    apt-get install -y python2.7 python-pip python-gevent && \
```
jonparrott: gevent shouldn't be needed.

managed_vms/endpoints/Dockerfile.container-engine
```diff
@@ -0,0 +1,13 @@
...
+    apt-get clean && \
+    rm /var/lib/apt/lists/*_*
+
+COPY requirements.txt /requirements.txt
```
jonparrott: Why not just `ADD . /app` ?

wlu2016:
done ptal

jonparrott:
LGTM.

PR #332, opened 2016-05-06
Update compute metadata hanging get example to use "wait_for_change".

Galabar001:
We need to include "wait_for_change" in order to perform a hanging get.  In
addition, we should deal with possible errors (like too many hanging gets).
Finally, this change allows for any of the maintenance event supported by
the metadata server.

### DIFF
--- a/compute/metadata/main.py
+++ b/compute/metadata/main.py
@@ -26,7 +26,7 @@ import time
 import requests
 
 
-METADATA_URL = "http://metadata.google.internal/computeMetadata/v1/"
+METADATA_URL = 'http://metadata.google.internal/computeMetadata/v1/'
 METADATA_HEADERS = {'Metadata-Flavor': 'Google'}
 
 
--- a/compute/metadata/main.py
+++ b/compute/metadata/main.py
@@ -32,14 +32,14 @@ METADATA_HEADERS = {'Metadata-Flavor': 'Google'}
 
 def wait_for_maintenance(callback):
     url = METADATA_URL + 'instance/maintenance-event'
-    last_in_maintenance = False
+    last_maintenance_event = None
     # [START hanging_get]
-    last_etag = 0
+    last_etag = "0"
 
     while True:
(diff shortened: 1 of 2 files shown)


compute/metadata/main.py
```diff
@@ -32,38 +32,46 @@
...
-            params={'last_etag': last_etag},
+            params={'last_etag': last_etag, 'wait_for_change': True},
             headers=METADATA_HEADERS)
 
         # During maintenance the service can return a 503, so these should
         # be retried.
         if r.status_code == 503:
             time.sleep(1)
             continue
+        elif r.status_code != requests.codes.ok:
```
jonparrott: The only status the docs mention that we should explicitly check for is 503. Any other error is better surfaced through the started exception infrastructure rather than just sleeping and trying again.

Why not remove this and just use `r.raise_for_status()`?

compute/metadata/main.py
```diff
@@ -32,38 +32,46 @@
 
 def wait_for_maintenance(callback):
     url = METADATA_URL + 'instance/maintenance-event'
-    last_in_maintenance = False
+    last_maintenance_event = None
     # [START hanging_get]
-    last_etag = 0
+    last_etag = "0"
```
jonparrott: Use single quotes everywhere, please.

compute/metadata/main.py
```diff
@@ -32,38 +32,46 @@
...
+                  .format(r.status_code, r.reason))
+            time.sleep(1)
+            continue
 
         last_etag = r.headers['etag']
         # [END hanging_get]
 
-        if r.text == 'MIGRATE_ON_HOST_MAINTENANCE':
-            in_maintenance = True
+        if r.text != 'NONE':
```
jonparrott: Since you've added an else, please switch the logic here.

jonparrott:
@Galabar001 please register your github username internally to tell CLAbot about it and reply "I signed it" to get it to check again.

PR #333, opened 2016-05-06
Update compute metadata hanging get example to use "wait_for_change".

Galabar001:
We need to include "wait_for_change" in order to perform a hanging get. In
addition, we should deal with possible errors (like too many hanging gets).
Finally, this change allows for any of the maintenance event supported by
the metadata server.


Galabar001:
Had some git issues.  Recreated this.

PR #342, opened 2016-05-10
Adding test for customer-supplied disk encryption key. Fixes #281

### LINKED ISSUE #281: Write a real test for encryption sample code
[The current test](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/compute/encryption/generate_wrapped_rsa_key_test.py#L17) just invokes the function, but it's very nice if the test actually check if one of the operations described in [the doc](https://cloud.google.com/compute/docs/disks/customer-supplied-encryption).

For example, you can invoke:
```
$ gcloud beta compute disks create example-disk --csek-key-file example-file
```
with the file created by the current code in the example. When I run the command above with an arbitrary file, it threw the following error:
```
InvalidKeyFileException: No JSON object could be decoded
```
IMHO, ensuring it successfully creates a disk by such a test will be a big win.

jonparrott:
/cc @waprin

### DIFF
--- a/compute/encryption/generate_wrapped_rsa_key_test.py
+++ b/compute/encryption/generate_wrapped_rsa_key_test.py
@@ -11,8 +11,40 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import os
+
 import generate_wrapped_rsa_key
+from googleapiclient import discovery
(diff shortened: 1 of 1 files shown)


tmatsuo:
LGTM

PR #343, opened 2016-05-10
Fix speech streaming sample & test

jerjou:


jonparrott:
LGTM once travis is satisfied.

PR #345, opened 2016-05-11
Update README.md

waprin:
Need to provide an actual string to echo back


jonparrott:
@waprin googlebot forgot you. :(

PR #352, opened 2016-05-17
Add sample for speech api on GCS file via grpc.

jerjou:
Note that grpc doesn't yet work on python3, so the test is skipped.


speech/api/speech_gcs.py
```diff
@@ -0,0 +1,92 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Sample that transcribes a FLAC audio file stored in Google Cloud Storage,
+using GRPC."""
+
+import argparse
+
+from gcloud.credentials import get_credentials
+from google.cloud.speech.v1.cloud_speech_pb2 import *  # noqa
```
jonparrott: Have you done this elsewhere? please never do this.

jonparrott:
Please make travis happy.

speech/api/speech_gcs.py
jerjou: Yeah - I don't usually, but the auto-generated protobuf libs are pretty opaque about what exactly they declared -_-;

Actually, I realize that I can just import the module. I'll do that... LMKWYT

speech/api/speech_streaming.py
```diff
@@ -1,11 +1,25 @@
...
+# limitations under the License.
+"""Sample that streams audio to the Google Cloud Speech API via GRPC."""
 
 import contextlib
 import re
 import threading
 
 from gcloud.credentials import get_credentials
-from google.cloud.speech.v1.cloud_speech_pb2 import *  # noqa
+from google.cloud.speech.v1 import cloud_speech_pb2 as pb
```
jonparrott: I would rather you not alias it. If you are going to alias it, use `cloud_speech`.

jerjou: Done

jonparrott:
LGTM once travis is happy.

jerjou:
Me and travis are tight.

PR #353, opened 2016-05-17
No need for speech discovery doc any longer.

jerjou:


speech/api/speech_rest.py
```diff
@@ -19,34 +19,29 @@
...
         ['https://www.googleapis.com/auth/cloud-platform'])
-    with open(API_DISCOVERY_FILE, 'r') as f:
-        doc = f.read()
+    http = httplib2.Http()
+    credentials.authorize(http)
 
-    return discovery.build_from_document(
-        doc, credentials=credentials, http=httplib2.Http())
-    # [END authenticating]
+    return discovery.build('speech', 'v1', http=http,
```
jonparrott: No action needed, but in general I prefer to break after the parenthesis instead of hanging indents, e.g.:

```python
return discovery.build(
    'speech', 'v1', http=http, discoveryServiceUrl=DISCOVERY_URL)
```

Also, I thought that google-api-python-client will automatically check this new endpoint as well?

jonparrott:
LGTM, merge at will.

PR #358, opened 2016-05-23
updating requirements

jonparrott:
Hopefully nox/travis will do its magic and test all of the samples that were touched.


waprin:
lgtm

PR #367, opened 2016-06-07
Rename model usage to entity

waprin:
Models are the class, entities are the datastore
instantiations of this class, so this change
is more conceptually accurate.


jerjou:
LGTM after travis is happy

PR #368, opened 2016-06-07
Add missing instructions for test project setup.

tswast:
Adds missing sections to script and environment variables needed when running `nox -s tests` on a new project.

### DIFF
--- a/scripts/prepare-testing-project.sh
+++ b/scripts/prepare-testing-project.sh
@@ -25,11 +25,11 @@ gsutil defacl set public-read gs://$GCLOUD_PROJECT
 echo "Creating bigquery resources."
 gcloud alpha bigquery datasets create test_dataset
 gcloud alpha bigquery datasets create ephemeral_test_dataset
-gsutil cp tests/resources/data.csv gs://$GCLOUD_PROJECT/data.csv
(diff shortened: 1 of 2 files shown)


scripts/prepare-testing-project.sh
```diff
@@ -38,4 +38,4 @@ echo "Creating pubsub resources."
 gcloud alpha pubsub topics create gae-mvm-pubsub-topic
 
 echo "To finish setup, follow this link to enable APIs."
-echo "https://console.cloud.google.com/flows/enableapi?apiid=datastore,pubsub,storage_api,logging,plus,bigquery,cloudmonitoring,compute_component"
+echo "https://console.cloud.google.com/flows/enableapi?project=${GCLOUD_PROJECT}&apiid=bigquery,cloudmonitoring,compute_component,datastore,dataproc,dns,plus,pubsub,logging,storage_api,vision.googleapis.com"
```
jonparrott: probably need to add `datastore.googleapis.com` here (datastore v1beta3)

tswast: Done in latest commit.

I don't remember having to enable that API. It could be on by default, but explicit is better than implicit.

PR #371, opened 2016-06-08
Add Bigtable hello world sample.

tswast:
Replaces https://github.com/GoogleCloudPlatform/cloud-bigtable-examples/pull/173.


bigtable/hello/hello.py
```diff
@@ -0,0 +1,99 @@
+#!/usr/bin/env python
```
jonparrott: I perfer `main.py`

bigtable/hello/hello.py
```diff
@@ -0,0 +1,99 @@
...
+  https://support.google.com/cloud/answer/6158840
+"""
+
+import argparse
+
+from gcloud import bigtable
+from gcloud.bigtable import happybase
+
+
+TABLE_NAME = 'Hello-Bigtable'
```
jonparrott: The constants can all be variables in the function below.

bigtable/hello/e2etest.sh
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env bash
```
jonparrott: Is this necessary anymore? Please remove it.

testing/resources/test-env.tmpl.sh
```diff
@@ -1,6 +1,8 @@
 # Environment variables for system tests.
```
jonparrott: You'll need to update the encrypted version of this in secrets.tar.gz, the password is in valentine.

lesv:
LGTM - but it looks like you still need to do what Jon asked for.

lesv:
@GoogleCloudPlatform/anviltop FYI - python Hello World

bigtable/hello/main.py
```diff
@@ -0,0 +1,93 @@
...
+import argparse
+
+from gcloud import bigtable
+from gcloud.bigtable import happybase
+
+
+def main(project, cluster_id, zone, table_name):
+    column_family_name = 'cf1'
+    column_name = '{fam}:greeting'.format(fam=column_family_name)
+    greetings = [
```
jonparrott: Please define variables close to where they are used, in this case, it seems like you can push the `greetings` down to before line 62.

testing/resources/test-env.tmpl.sh
tswast: Done.

bigtable/hello/main.py
tswast: Variables moved.

Note: greetings is used only once, but I kept it defined since I think it'd look weird to loop over a naked list. Though with only three items I wonder if just unrolling the loop would be clearer?

PR #375, opened 2016-06-14
Logging Fixups

waprin:
1) Bug fix, wasn’t passing in token to pagination
2) Rearrange region tags to include more
3) Have v1 samples recommend v2 samples

### DIFF
--- a/logging/api/logs_api.py
+++ b/logging/api/logs_api.py
@@ -21,22 +21,24 @@ from oauth2client.client import GoogleCredentials
 
 
 def write_entry(client, args):
+    # [START write]
     print('Writing log entry for logger '.format(args.logger_name))
     mylogger = client.logger(args.logger_name)
-    # [START write]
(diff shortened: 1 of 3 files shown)


logging/api/logs_api.py
```diff
@@ -21,22 +21,25 @@
...
-    # [START list]
     entries = []
+    token = None
     while True:
-        new_entries, token = client.list_entries(filter_='logName="{}"'.format(
-           logger.full_name))
+        new_entries, token = client.list_entries(
+            filter_='logName="{}"'.format(logger.full_name),
+            page_token=token
+            )
```
jonparrott: nit: prefer this to be at the end of the preceding line.

jonparrott:
LGTM.

logging/api/logs_api.py
waprin: done

PR #376, opened 2016-06-14
Add region tags to bigtable/hello sample.

tswast:
Also, change the sample to use sequential keys (with a disclaimer) to
match the Java sample. I had forgotten to add a sample usage to get a
specific row, so add that, too.

### DIFF
--- a/bigtable/hello/main.py
+++ b/bigtable/hello/main.py
@@ -27,22 +27,24 @@ Prerequisites:
 """
 
 import argparse
-import uuid
 
 from gcloud import bigtable
 from gcloud.bigtable import happybase
 
 
-def main(project, cluster_id, zone, table_name):
+def main(project_id, cluster_id, zone, table_name):
+    # [START connecting_to_bigtable]
     # The client must be created with admin=True because it will create a
     # table.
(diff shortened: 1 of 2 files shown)


bigtable/hello/main.py
```diff
@@ -27,13 +27,17 @@
...
 import argparse
-import uuid
 
 from gcloud import bigtable
 from gcloud.bigtable import happybase
 
 
 def main(project, cluster_id, zone, table_name):
+    # [START connecting_to_bigtable]
+    # project = "my-project-id"
```
jonparrott: These comments are unnecessary.

jonparrott:
LGTM with small nit, but @waprin should review.

bigtable/hello/main.py
lesv: I agree w/ Jon - rename project => projectId (or the Python convention), same for the others.

lesv:
LGTM - w/ Jon's nit.

bigtable/hello/main.py
tswast: Unnecessary comments removed and `project` renamed to `project_id`.

PR #377, opened 2016-06-16
Close HappyBase connection in bigtable/hello sample.

tswast:
I noticed that the `bigtable/hello` sample was not quitting when I sent
a `Ctrl-C` this should fix that problem.

### DIFF
--- a/bigtable/hello/main.py
+++ b/bigtable/hello/main.py
@@ -37,13 +37,11 @@ def main(project_id, cluster_id, zone, table_name):
     # The client must be created with admin=True because it will create a
     # table.
     client = bigtable.Client(project=project_id, admin=True)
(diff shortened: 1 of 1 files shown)


waprin:
wow not sure why travis is suddenly unhappy about those appengine snippets, will take a look.

code lgtm, though a bit surprised the change is necessary. do you think we need an issue on the client lib?

waprin:
Once #378 is merged just rebase on that.

tswast:
Rebased. I'm not sure it warrants an issue on the client. I started using the "raw" apis without this problem and then encountered it after converting to the happybase package. I think the HappyBase layer starts some background threads or something that I needed to explicitly close that connection, not just the underlying client.

PR #383, opened 2016-06-24
bigtable: Move hello to hello_happybase.

tswast:
I will be adding a "raw" (gcloud-python) hello world sample for Cloud
Bigtable soon, so this moves the HappyBase version to its own directory.
No docs reference this sample yet.

FYI: @lesv


tswast:
@jonparrott Looks like Travis is failing for having no changed files?

Maybe because we exclude the bigtable dir from Travis? Nox does pass when I run locally.

jonparrott:
LGTM.

jonparrott:
Yeah it's probably just whack with poo brain. I'll keep an eye on it.

PR #386, opened 2016-06-27
Update Reqs

waprin:


jonparrott:
Should we wait for [gcloud-python#1879](https://github.com/GoogleCloudPlatform/gcloud-python/issues/1879) first?

waprin:
Yes we should. But 0.16.0 contains bugfixes for logging so Jenkins will just be broken until the mess gets sorted out.

Also looks like sendgrid was a major version bump and changed their API so we will need to fix that too.

jonparrott:
If you can track the gcloud stuff, I'll work on getting sendgrid updated independently.

tswast:
gcloud-python is at 0.17.0 now. https://pypi.python.org/pypi/gcloud

waprin:
ty, ill just run reqcheck before i amend this :)

PR #388, opened 2016-06-28
New code samples for Endpoints v1.1 for Python (Alpha)

bradfriedman:
These are code samples which will support the alpha release of Endpoints v1.1 for Python.

### DIFF
--- a/appengine/standard/endpoints/v1.1/backend/app.yaml
+++ b/appengine/standard/endpoints/v1.1/backend/app.yaml
@@ -1,0 +1,14 @@
+runtime: python27
+threadsafe: true
+api_version: 1
+
+handlers:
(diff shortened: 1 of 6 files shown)


waprin:
Travis is broken because our tests look misconfigured to try to use credentials for system tests on PRs even though PRs can't access them. Will fix...

If this is alpha, do people need to be whitelisted?

Brief README would be good with at least a link to the docs on what this is about.

jonparrott:
@bradfriedman what has changed significantly from v1.0 to v1.1? Basically, why wouldn't we just update the existing samples to use 1.1?

bradfriedman:
Per discussion with @jonparrott, I'll close this PR in favor of an in-place change.

PR #397, opened 2016-06-30
Add google cloud storage samples using gcloud.

hohaichi:
Simple upload and download samples using python gcloud.

### DIFF
--- a/storage/gcloud/customer_supplied_keys.py
+++ b/storage/gcloud/customer_supplied_keys.py
@@ -34,54 +34,65 @@ import tempfile
 
 from gcloud import storage
 
-# You can (and should) generate your own encryption key. os.urandom(32) is a
-# good way to accomplish this with Python.
+# An AES256 encryption key.  It must be exactly 256 bits (32 bytes).  You can
+# (and should) generate your own encryption key.  os.urandom(32) is a good way
+# to accomplish this with Python.
 #
 # Although these keys are provided here for simplicity, please remember
 # that it is a bad idea to store your encryption keys in your source code.
 ENCRYPTION_KEY = os.urandom(32)
 
 
-def upload_object(client, bucket_name, filename, object_name, encryption_key):
+def upload_object(storage_client,
+                  bucket_name,
+                  filename,
+                  object_name,
+                  encryption_key):
     """Uploads an object, specifying a custom encryption key.
 
     Args:
-        client: gcloud client
+        storage_client: gcloud client to access cloud storage
         bucket_name: name of the destination bucket
         filename: name of file to be uploaded
         object_name: name of resulting object
-        encryption_key: encryption key to encrypt the object.
+        encryption_key: encryption key to encrypt the object,
+                        either 32 raw bytes or a string of 32 bytes.
     """
-    bucket = client.get_bucket(bucket_name)
+    bucket = storage_client.get_bucket(bucket_name)
     blob = bucket.blob(object_name)
     with open(filename, 'rb') as f:
       blob.upload_from_file(f, encryption_key=encryption_key)
 
 
-def download_object(client, bucket_name, object_name, filename, encryption_key):
+def download_object(storage_client,
+                    bucket_name,
+                    object_name,
+                    filename,
+                    encryption_key):
     """Downloads an object protected by a custom encryption key.
 
     Args:
-        client: gcloud client
+        storage_client: gcloud client to access cloud storage
         bucket_name: name of the source bucket
         object_name: name of the object to be downloaded
         filename: name of the resulting file
-        encryption_key: the encryption key that the object is encrypted by.
+        encryption_key: the encryption key that the object is encrypted by,
+                        either 32 raw bytes or a string of 32 bytes.
     """
-    bucket = client.get_bucket(bucket_name)
+    bucket = storage_client.get_bucket(bucket_name)
     blob = bucket.blob(object_name)
     with open(filename, 'wb') as f:
         blob.download_to_file(f, encryption_key=encryption_key)
 
 
 def main(bucket, filename):
-    client = storage.Client()
+    storage_client = storage.Client()
     print('Uploading object gs://{}/{} using encryption key (base64 formatted)'
           ' {}'.format(bucket, filename, base64.encodestring(ENCRYPTION_KEY)))
-    upload_object(client, bucket, filename, filename, ENCRYPTION_KEY)
+    upload_object(storage_client, bucket, filename, filename, ENCRYPTION_KEY)
     print('Downloading it back')
     with tempfile.NamedTemporaryFile(mode='w+b') as tmpfile:
-        download_object(client, bucket, object_name=filename,
+        download_object(storage_client, bucket, object_name=filename,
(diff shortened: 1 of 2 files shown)


storage/gcloud/customer_supplied_keys.py
```diff
@@ -0,0 +1,99 @@
...
+import filecmp
+import os
+import tempfile
+
+from gcloud import storage
+from gcloud.storage import Blob
+
+# You can (and should) generate your own encryption key. os.urandom(32) is a
+# good way to accomplish this with Python.
+#     Although these keys are provided here for simplicity, please remember
```
jonparrott: Fix weird indentation?

storage/gcloud/customer_supplied_keys.py
```diff
@@ -0,0 +1,99 @@
...
+"""
+
+import argparse
+import base64
+import filecmp
+import os
+import tempfile
+
+from gcloud import storage
+from gcloud.storage import Blob
```
jonparrott: Don't do this, just use `storage.Blob`.

storage/gcloud/customer_supplied_keys.py
```diff
@@ -0,0 +1,99 @@
...
+
+    Args:
+        bucket_name: name of the destination bucket
+        filename: name of file to be uploaded
+        object_name: name of resulting object
+        encryption_key: encryption key to encrypt the object.
+    """
+    client = storage.Client()
+    bucket = client.get_bucket(bucket_name)
+    blob = Blob(object_name, bucket)
```
jonparrott: `bucket.blob` should work.

storage/gcloud/customer_supplied_keys.py
```diff
@@ -0,0 +1,99 @@
...
+def upload_object(bucket_name, filename, object_name, encryption_key):
+    """Uploads an object, specifying a custom encryption key.
+
+    Args:
+        bucket_name: name of the destination bucket
+        filename: name of file to be uploaded
+        object_name: name of resulting object
+        encryption_key: encryption key to encrypt the object.
+    """
+    client = storage.Client()
```
jonparrott: Construct the client in `main()`, pass it into these functions.

jonparrott:
I would prefer that we update all of the cloud storage samples at once to use gcloud-python, however, this would need better coordination.

storage/gcloud/customer_supplied_keys.py
hohaichi: Done.

storage/gcloud/customer_supplied_keys.py
hohaichi: Removed.  Don't need it anymore after changing to bucket.blob.

storage/gcloud/customer_supplied_keys.py
hohaichi: Done.

storage/gcloud/customer_supplied_keys.py
hohaichi: Done.

hohaichi:
Thank you for the super quick turnaround.  I incorporated your suggestions to the change.  PTAL.

storage/gcloud/customer_supplied_keys.py
```diff
@@ -0,0 +1,100 @@
...
+import tempfile
+
+from gcloud import storage
+
+# You can (and should) generate your own encryption key. os.urandom(32) is a
+# good way to accomplish this with Python.
+#
+# Although these keys are provided here for simplicity, please remember
+# that it is a bad idea to store your encryption keys in your source code.
+ENCRYPTION_KEY = os.urandom(32)
```
BrandonY: Might be worth pointing out explicitly that it MUST be exactly 256 bits.

storage/gcloud/customer_supplied_keys.py
```diff
@@ -0,0 +1,100 @@
...
+
+def upload_object(client, bucket_name, filename, object_name, encryption_key):
+    """Uploads an object, specifying a custom encryption key.
+
+    Args:
+        client: gcloud client
+        bucket_name: name of the destination bucket
+        filename: name of file to be uploaded
+        object_name: name of resulting object
+        encryption_key: encryption key to encrypt the object.
```
BrandonY: Does encryption_key need to be represented as a number instead of, say, a byte array or a string? Probably worth noting.

storage/gcloud/customer_supplied_keys.py
```diff
@@ -0,0 +1,100 @@
...
+        bucket_name: name of the source bucket
+        object_name: name of the object to be downloaded
+        filename: name of the resulting file
+        encryption_key: the encryption key that the object is encrypted by.
+    """
+    bucket = client.get_bucket(bucket_name)
+    blob = bucket.blob(object_name)
+    with open(filename, 'wb') as f:
+        blob.download_to_file(f, encryption_key=encryption_key)
+
```
BrandonY: A key rotation example would also be nice.

jonparrott: File a bug, so we can track that. No need to tie up this PR.

storage/gcloud/customer_supplied_keys.py
jonparrott: Needs to be bytes.

storage/gcloud/customer_supplied_keys.py
```diff
@@ -0,0 +1,100 @@
...
+# You can (and should) generate your own encryption key. os.urandom(32) is a
+# good way to accomplish this with Python.
+#
+# Although these keys are provided here for simplicity, please remember
+# that it is a bad idea to store your encryption keys in your source code.
+ENCRYPTION_KEY = os.urandom(32)
+
+
+def upload_object(client, bucket_name, filename, object_name, encryption_key):
```
jonparrott: `storage_client` is a better name than just `client`.

storage/gcloud/customer_supplied_keys.py
hohaichi: Done.

storage/gcloud/customer_supplied_keys.py
hohaichi: String also works.  Updated.

storage/gcloud/customer_supplied_keys.py
hohaichi: It has to be a future change as gcloud is not currently supporting rewrite afaik.

storage/gcloud/customer_supplied_keys.py
hohaichi: Done.

hohaichi:
Updated with improvements per feedback.  PTAL.

jonparrott:
LGTM

hohaichi:
Thanks.  Do you know what the Travis-ci error means?

Quoted:
    The command "openssl aes-256-cbc -k "$secrets_password" -in secrets.tar.enc -out secrets.tar -d" failed and exited with 1 during .

jonparrott:
Travis doesn't succeed for outside pull requests, it's expected.

PR #399, opened 2016-07-01
Changing cloud speech code samples to work with v1beta1

xinjiez:
This is adapted from alpha samples in speech/api.

### DIFF
--- a/speech/api/speech_gcs.py
+++ b/speech/api/speech_gcs.py
@@ -18,7 +18,7 @@ using GRPC."""
 import argparse
 
 from gcloud.credentials import get_credentials
-from google.cloud.speech.v1 import cloud_speech_pb2 as cloud_speech
+from google.cloud.speech.v1beta1 import cloud_speech_pb2 as cloud_speech
 from grpc.beta import implementations
 
 # Keep the request alive for this many seconds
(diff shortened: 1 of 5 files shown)


jonparrott:
This LGTM. I made some small changes for style, can you confirm it still works?

speech/api/speech_gcs.py
```diff
@@ -77,16 +75,15 @@ def _gcs_uri(text):
 
 
 PROTO_URL = ('https://github.com/googleapis/googleapis/blob/master/'
-             'google/cloud/speech/v1/cloud_speech.proto')
+             'google/cloud/speech/v1beta1/cloud_speech.proto')
 if __name__ == '__main__':
     parser = argparse.ArgumentParser()
     parser.add_argument('input_uri', type=_gcs_uri)
-    parser.add_argument('output_uri', type=_gcs_uri)
```
jerjou: The main point of this sample was to demonstrate having an output uri on gcs (thus '_gcs' in the filename). Is that not supported anymore?

xinjiez:
That's not supported any more in v1beta1. I think I also changed the file name to speech_grpc to repurpose it to demo grpc instead.

PR #409, opened 2016-07-07
Update requirements

waprin:
Also testing Jenkins PR stuff


jonparrott:
LGTM as long as the test god is pleased.

waprin:
Just going to merge this and use dummy commits instead

PR #415, opened 2016-07-13
Readme fixes

puneith:


danaharon:
can you remove "The sample uses the PyAudio library to stream audio from your computer's microphone. PyAudio depends on PortAudio, which may need to be compiled when you install PyAudio." and instead add PortAudio + PyAudio to the section next to pip, gcloud?

puneith:
@jonparrott please take a look. Need help to make sure this looks appropriate. Please, don't merge though.

speech/api/README.md
```diff
@@ -38,25 +38,42 @@ See the
...
+
+  ```
+  $ sudo apt-get install portaudio19-dev
+  ```
 
 * If you're running the `speech_rest.py` sample:
 
     ```sh
-    $ sudo pip install -r requirements-speech_rest.txt
+    $ sudo pip install -r requirements-speech_rest.txt --ignore-installed six
```
jonparrott: Don't use sudo, don't ignore installed six, just tell them to use virtualenv.

puneith: I have been thinking about that as well. virtual env or conda. But, the problem is what if they want to do something outside of virtual?

puneith: virtual / conda makes life so much easier.

jonparrott: Then that's up to them. Virtualenv isn't a controversial recommendation. It's broadly used, and recommended by every other sample here.

PR #417, opened 2016-07-13
Fixing speech samples lost in merge.

jonparrott:


jonparrott:
@waprin FYI this is just fixing a bad merge.

PR #421, opened 2016-07-14
Refactor cloud client storage samples.

jonparrott:
Please give these a fair amount of scrutiny, as I will codify this pattern into the standard for our samples.

/r @jerjou and @waprin
/cc @puneith and @pfritzsche please feel free to also chime in.


jonparrott:
I'm also unsure of the naming of the files `encryption.py` and `manage_blobs.py`, open to suggestions.

storage/cloud-client/encryption_test.py
```diff
@@ -0,0 +1,67 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import base64
+import tempfile
+
+import encryption
+from gcloud import storage
+import pytest
```
pfritzsche: Should this match [PEP8](https://www.python.org/dev/peps/pep-0008/#imports)?

Specifically, encryption should be in a third section if so.

jonparrott: Due to this project's organization this is actually in the correct section.

storage/cloud-client/encryption_test.py
```diff
@@ -0,0 +1,67 @@
...
+
+def test_upload_encrypted_blob(cloud_config):
+    with tempfile.NamedTemporaryFile() as source_file:
+        source_file.write(b'test')
+
+        encryption.upload_encrypted_blob(
+            cloud_config.storage_bucket,
+            source_file.name,
+            'test_encrypted_upload_blob',
+            TEST_ENCRYPTION_KEY)
```
pfritzsche: What is this testing?

storage/cloud-client/encryption_test.py
pfritzsche: Interesting. How so?

jonparrott: quirk of flake8-import-order, because I run flake8 from the root directory and this lives in a subfolder, it considers encryption part of the third-party section instead of the local section. I have an idea for a fix now that we're using nox, I'll see if I can get that to work in another PR.

storage/cloud-client/encryption_test.py
jonparrott: There are no asserts, but the sample is simple enough that there are only two possible branches: success or exception. If there is any exception, the case will fail.

pfritzsche: Makes sense. I was being silly and didn't realize it was a system test (as opposed to a mocked out call)

storage/cloud-client/encryption_test.py
pfritzsche: Sounds good.

storage/cloud-client/encryption.py
```diff
@@ -0,0 +1,161 @@
...
+        'bucket_name', help='Your cloud storage bucket.')
+    download_parser.add_argument('blob_name')
+    download_parser.add_argument('base64_encryption_key')
+    download_parser.add_argument('base64_new_encryption_key')
+
+    args = parser.parse_args()
+
+    if args.command == 'generate-encryption-key':
+        generate_encryption_key()
+    elif args.command == 'upload':
```
waprin: How do you feel about making doing download_parser.set_default(func=encrypted_blog) ? Advantage is we kill this ugly big else, disadvantage is that our functions just take args and have to parse out the params themselves. Guessing you prefer this way?

jonparrott: Yeah, so I explicitly picked this pattern over that one.

I don't like our functions just taking args. It's less clear, it makes testing slightly harder, and it's just doesn't feel right. We've worked around this in the past by having a '_command' function, eg:

```
list_objects(bucket_name):
    ...

list_objects_command(args):
    list_objects(args.bucket_name)
```

but I don't like having to write those either. This is more direct and simple, but I'm open to arguments against it or for something else.

storage/cloud-client/encryption.py
```diff
@@ -0,0 +1,161 @@
...
+
+    The file will be encrypted by Google Cloud Storage and only
+    retrievable using the provided encryption key.
+    """
+    storage_client = storage.Client()
+    bucket = storage_client.get_bucket(bucket_name)
+    blob = bucket.blob(destination_blob_name)
+
+    # Encryption key must be an AES256 key represented as a bytestring with
+    # 32 bytes. Since it's passed in as a base64 encoded string, it needs
```
pfritzsche: Per our chat - I'm debating if the fact that you use base64 could throw people off. It's a side-effect of the way you produced the key, and not a requirement of the API.

Maybe update this comment as follows? "Since this sample application uses a base64 encoded string to store our encryption key, it needs..."

That way you can make it clear it's just a design choice, and not a functional requirement.

jonparrott: So it's also a side-effect of the way this function is called (from the command-line), but let me think on this.

storage/cloud-client/encryption.py
```diff
@@ -0,0 +1,161 @@
...
+    comprehensive security policy.
+    """
+    key = os.urandom(32)
+    encoded_key = base64.b64encode(key).decode('utf-8')
+    print('Base 64 encoded encryption key: {}'.format(encoded_key))
+
+
+def upload_encrypted_blob(bucket_name, source_file_name,
+                          destination_blob_name, base64_encryption_key):
+    """Uploads a file to a Google Cloud Storage bucket using a custom
```
waprin: I know we discussed this offline but...still think describing what the params are and their type may be valuable standard to set. Even if they are all strings, because it's good to document that and it's good standard to set.

jonparrott: I want to strike a balance between simplicity and clarity. I think that the idea is that the name of the variable and its usage should imply its type well enough - I feel it's overt that all of these parameters should be strings. @pfritzsche what are your thoughts?

Overall, I want to avoid having a completely overbearing docstring here. I can be talked into them, but I tend towards less is more for samples. (For libraries, I'm the other way - super docstrings everywhere).

Also, keep in mind that the main section also provides us a chance to give help strings for command-line arguments. Although only the bucket name has a help string in these samples.

pfritzsche: I tend to prefer self-documenting parameters, without explicit arg definitions in doc strings because 1) the doc strings become a hassle to maintain, and 2) they're often unnecessary. For example, I try to avoid things like:

```
def upload_blob(bucket_name):
    """Keyword arguments:
        bucket_name: the name of the bucket to upload to
    """
```

When you have more complex input params (objects, dictionaries with variable keys), etc., then I think it becomes more worth it.

tl;dr: In this case, I don't think it's necessary, but in some more complex cases I do.

waprin: Sure, sounds like you two are leaning against it which is fine. To me, value of the doc string is 1) the type expected 2) setting the standard for all those cases where an extra sentence can help clarify thing. But don't feel strongly about it, just wanted to make sure we discussed/decided on it.

pfritzsche: It's a fair argument. I'm not opposed to defining types in docstrings

waprin:
Overall lgtm

pfritzsche:
+1

- Phil

storage/cloud-client/encryption.py
```diff
@@ -0,0 +1,161 @@
...
+    """Generates a 256 bit (32 byte) AES encryption key and prints the
+    base64 representation.
+
+    This is included for demonstration purposes. You should generate your own
+    key. Please remember that encryption keys should be handled with a
+    comprehensive security policy.
+    """
+    key = os.urandom(32)
+    encoded_key = base64.b64encode(key).decode('utf-8')
+    print('Base 64 encoded encryption key: {}'.format(encoded_key))
```
jerjou: What do you think about returning the value, rather than printing it?
I'm of two minds about it:
* I think in a real-world scenario / reading this script directly, it makes more sense to return the value and have the calling function do the printing
* As a code snippet, though, it's cleaner to just print the value - ie if you're just including this method as a snippet in the docs. In this case, however, I usually name the function with a `print_` prefix to make this behavior clear

storage/cloud-client/encryption.py
jerjou: It might make sense to do this using argparse - ie

    upload_parser.add_argument('base64_encryption_key', type=base64.b64decode)

which is nice because it localizes the side-effect to the command-line interface, and this function only has to care about what it cares about.

storage/cloud-client/encryption.py
jerjou: +1 use your judgment :)

In the canonical template sample, though, we should definitely include a method where the arguments are described (including its type), so codify the format we want it in.

storage/cloud-client/encryption.py
puneith: @jonparrott remove the space in 'utf-8'

storage/cloud-client/encryption_test.py
```diff
@@ -0,0 +1,67 @@
...
+def test_download_blob(test_blob, cloud_config):
+    test_blob_name, test_blob_content = test_blob
+    with tempfile.NamedTemporaryFile() as dest_file:
+        encryption.download_encrypted_blob(
+            cloud_config.storage_bucket,
+            test_blob_name,
+            dest_file.name,
+            TEST_ENCRYPTION_KEY)
+
+        downloaded_content = dest_file.read().decode('utf-8')
```
puneith: same here remove space in utf-8

jonparrott: Space? You mean dash? The dash? The dash is present in the standard library
codecs module and is generally accepted as the right alias for codecs.UTF8

storage/cloud-client/encryption.py
jonparrott: That seems a bit magical. So ostensibly gcloud-python will support base64 keys soon, so I won't have to decode them anyway. But even without that I'd rather do the decoding in the function do it doesn't seem magical.

storage/cloud-client/encryption.py
jonparrott: > In the canonical template sample, though, we should definitely include a method where the arguments are described (including its type), so codify the format we want it in.

For sure, I'll make sure that the sample guide shows an example of a more complex function that needs to be documented.

storage/cloud-client/encryption.py
jonparrott: pytests' `capsys` does a good job of making the return or not return not much of an issue for tests.

> it makes more sense to return the value and have the calling function do the printing

There is no calling function, just the main clause. I'd prefer not to have the main clause do any printing.

> if you're just including this method as a snippet in the docs.

All of these should be in the docs. :)

I think I'm going to say "Use your judgement on whether to return or print the value. If this function is the top of the stack, it likely makes sense to just print."

storage/cloud-client/encryption.py
jerjou: But it's so cooooolll :)

jerjou:
LGTM

jonparrott:
Thanks all for the review!

PR #425, opened 2016-07-19
Add samples for natural language api.

jerjou:


conftest.py
```diff
@@ -15,6 +15,7 @@
 import os
 
 import pytest
+import requests
 
 
 class Namespace:
```
waprin: unrelated to PR but shouldn't Namespace inherit from object?

conftest.py
```diff
@@ -48,3 +49,22 @@ def resource(request):
     testing resource"""
     local_path = os.path.dirname(request.module.__file__)
     return lambda *args: get_resource_path(args, local_path)
+
+
+def fetch_gcs_resource(resource, tmpdir):
+    resp = requests.get(resource, stream=True)
+    dest_file = str(tmpdir.join(os.path.basename(resource)))
+    with open(dest_file, 'wb') as f:
+        for chunk in resp.iter_content(1024):
```
waprin: give magic number a name

conftest.py
```diff
@@ -48,3 +49,22 @@ def resource(request):
...
+            f.write(chunk)
+
+    return dest_file
+
+
+@pytest.fixture(scope='module')
+def remote_resource(cloud_config):
```
waprin: think the return type is sufficiently complicated to errant a doc string

language/movie_nl/main.py
```diff
@@ -0,0 +1,345 @@
...
+            rank_entities(reader, sentiment, sample, reverse_bool)
+    else:
+        # Create logger settings
+        logging.basicConfig(filename=log_file, level=logging.DEBUG)
+
+        # Create a Google Service object
+        service = get_service()
+
+        # Create a sentiment output writer
+        swriter = open(sent_out, 'w')
```
waprin: just call it sentiment_writer

language/ocr_nl/main.py
```diff
@@ -0,0 +1,363 @@
...
+            request = {
+                'image': {},
+                'features': [{
+                    'type': 'TEXT_DETECTION',
+                    'maxResults': max_results,
+                }]
+            }
+
+            # Accept both files in cloud storage, as well as local files.
+            if re.match(r'gs://', filename):
```
waprin: is this only use of re? think it could just be filename.startsWith?

conftest.py
jerjou: Oh - good point.

language/movie_nl/main.py
jerjou: Also changed ewriter to entity_writer

language/ocr_nl/main.py
jerjou: Good point.
I think the `re` is vestigial for when I was using `https?://` as the protocol to check for. But then I realized I could just pass `gs://` uris directly..

jerjou:
Updated.

language/api/README.md
```diff
@@ -0,0 +1,87 @@
...
+
+## Setup
+
+Please follow the [Set Up Your Project](https://cloud.google.com/natural-language/docs/getting-started#set_up_your_project)
+steps in the Quickstart doc to create a project and enable the
+Cloud Natural Language API. Following those steps, make sure that you
+[Set Up a Service Account](https://cloud.google.com/natural-language/docs/common/auth#set_up_a_service_account),
+and export the following environment variable:
+
+```
```
waprin: does gcloud beta auth application-default login not work btw?

language/api/analyze_test.py
```diff
@@ -0,0 +1,258 @@
...
+
+
+def test_analyze_syntax_utf16():
+    """Demonstrate the interpretation of the offsets when encoding=utf16.
+
+    UTF16 is a variable-length encoding, where each character is at least 16
+    bits. The returned offsets will be the index of the first 2-byte character
+    of the token.
+    """
+    test_string = u'a \u00e3 \u0201 \U0001f636 b'
```
waprin: super confused what you are doing here ...what is the point of the \u stuff?

language/api/README.md
jerjou: Not sure. See my comment [here](https://github.com/GoogleCloudPlatform/java-docs-samples/pull/275#discussion_r71449047), though..

language/movie_nl/main.py
```diff
@@ -0,0 +1,345 @@
...
+        if sentiment:
+            if sentiment == 'pos' and sent > 0:
+                items.append(entity_item)
+            elif sentiment == 'neg' and sent < 0:
+                items.append(entity_item)
+        else:
+            items.append(entity_item)
+
+    items.sort()
+    items.reverse()
```
waprin: believe you can sort with reversed=True

language/movie_nl/README.md
```diff
@@ -0,0 +1,146 @@
...
+### Set Up to Authenticate With Your Project's Credentials
+
+Please follow the [Set Up Your Project](https://cloud.google.com/natural-language/docs/getting-started#set_up_your_project)
+steps in the Quickstart doc to create a project and enable the
+Cloud Natural Language API. Following those steps, make sure that you
+[Set Up a Service Account](https://cloud.google.com/natural-language/docs/common/auth#set_up_a_service_account),
+and export the following environment variable:
+
+```
+export GOOGLE_APPLICATION_CREDENTIALS=/path/to/your-project-credentials.json
```
waprin: I would like to seea  big fat warning about an issue that monitoring also had. If you screw this step up (and environment variables are one of the top things people screw up) you get a misleading error message that the API still isn't enabled and you should wait a few minutes. We need to warn people that they might also need to double check this is set correctly.

language/api/analyze_test.py
jerjou: Haha - yeah. It's super confusing.
So - the NL api calculates the offsets into the string you send it, so you know where entities appear and stuff. The problem is, the offset will be different depending on how your string is encoded. Everything is serialized over the wire in UTF8, so the NL api doesn't know a priori what encoding your string is in, so you need to tell it when you make the request.

For example, if your string is in UTF8, the smallest unit of character is a byte, but some characters take up two bytes. So if you want the index of a character that comes after a two-byte character, your index will be one more than you might expect.

If your string is in UTF16, the smallest unit of character is two bytes, but some characters take up four bytes. So when you're indexing into a UTF16 string, a four-byte character will take up two character slots, so characters appearing after that four-byte character will have an index one more than you might expect.

If your string is UTF32, everything is encoded in the same number of bytes, so your character should be at the index you expect.

Anyway, these tests are meant to try to illustrate that explanation. The \u characters are characters that take 2, 4, and uh... 8? bytes each, so that the offsets (and my understanding of things) can be properly tested.

That make sense?

language/movie_nl/main.py
```diff
@@ -0,0 +1,345 @@
...
+
+def get_sentiment_entities(service, document):
+    """Compute the overall sentiment volume in the document"""
+    sentiments, entities = analyze_document(service, document)
+
+    sentiments = [sent for sent in sentiments if sent[0] is not None]
+    negative_sentiments = [
+        sentiment for sentiment in sentiments if sentiment[0] < 0.0]
+    positive_sentiments = [
+        sentiment for sentiment in sentiments if sentiment[0] >= 0.0]
```
waprin: Is 0 positive?

language/api/analyze_test.py
waprin: It does but I wonder if it's more or less clear if you actually use a printable character and then decode it into bytes. Not sure, food for thought mostly.

language/movie_nl/README.md
```diff
@@ -0,0 +1,151 @@
...
+    "sentiment": 3.099
+  }
+```
+
+### Entity Output Sorting
+In order to sort and rank the entities generated, use the same `main.py` script. For example,
+this will print the top 5 actors with negative sentiment:
+
+```
+$ python main.py --inp entity.json
```
waprin: first of all, better to put \ at end of each line to make easier copy paste.

secondly I get

TypeError: coercing to Unicode: need string or buffer, NoneType found

language/api/analyze_test.py
jerjou: My thinking was it'd be more clear using the unicode escapes, to make it clear that we're testing unicode here. As opposed to, say, a letter with an accent that you might not see, or a smiley that you might just think is there for fun :)

language/movie_nl/README.md
jerjou: Added a semi-fat warning. Let me know what you think (open to suggestions)

language/ocr_nl/README.md
```diff
@@ -0,0 +1,224 @@
...
+The example requires that sqlite3 be installed. Most likely, sqlite3 is already
+installed for you on your machine, but if not, you can find it
+[here](https://www.sqlite.org/download.html).
+
+### Set Up to Authenticate With Your Project's Credentials
+
+Please follow the [Set Up Your Project](https://cloud.google.com/natural-language/docs/getting-started#set_up_your_project)
+steps in the Quickstart doc to create a project and enable the
+Cloud Natural Language API. Following those steps, make sure that you
+[Set Up a Service Account](https://cloud.google.com/natural-language/docs/common/auth#set_up_a_service_account),
```
waprin: This one also requires you to enable Visino API.

language/movie_nl/main.py
waprin: whatev though

waprin:
Ok we can nitpick code later but I'm fine to merge once TypeError: coercing to Unicode: need string or buffer, NoneType found issue is fixed

waprin:
And note about enabling Vision API

language/movie_nl/main.py
jerjou: Well, it's not negative? ;)
Probably doesn't matter either way, since it's summed up later.
Actually, broke out the tuple to make things clear, and simplified the sum later on. PTAL

jerjou:
Merging - we'll add the movie_nl sample in a separate PR

PR #426, opened 2016-07-20
Nl

puneith:
@jerjou @waprin please review.


language/movie_nl/main.py
```diff
@@ -0,0 +1,367 @@
...
+                items.append(entity_item)
+            elif sentiment == 'neg' and sent < 0:
+                items.append(entity_item)
+        else:
+            items.append(entity_item)
+
+    items.sort(reverse=True)
+    items = [json.dumps(item[1]) for item in items]
+
+    if reverse_bool:
```
waprin: sorry if I'm misunderstanding this, but it looks like you are reversing then condtionally reverse again. any way to write to just conditionally reverse once?

language/movie_nl/main.py
```diff
@@ -0,0 +1,367 @@
...
+    service = get_service()
+
+    reader = document_generator(input_dir, sample)
+
+    # Process the movie documents
+    process_movie_reviews(service, reader, sentiment_writer, entity_writer)
+
+    # close reader and writers
+    sentiment_writer.close()
+    entity_writer.close()
```
waprin: if there's no way to do a with: clause with files from argparse, i would at least put this in a try: finally clause.

language/movie_nl/main.py
```diff
@@ -0,0 +1,366 @@
...
+def analyze_document(service, document):
+    """Analyze the document and get the distribution of sentiments and
+    the movie name."""
+    logging.info('Analyzing {}'.format(document.doc_id))
+
+    sentences, entities = document.extract_all_sentences(service)
+
+    try:
+        sentiments = [get_sentiment(service, sentence) for sentence in sentences]
+    except HttpError as e:
```
jerjou: This try/catch is unnecessary.

language/movie_nl/main.py
```diff
@@ -0,0 +1,366 @@
...
+    }
+
+    return body
+
+def get_response_with_retry(request, tries, retry=3):
+    """Get the response using re-try"""
+    try:
+        response = request.execute()
+        return response
+    except HttpError as e:
```
jerjou: This except block is unnecessary

language/movie_nl/main.py
```diff
@@ -0,0 +1,366 @@
...
+        },
+        'encoding_type': 'UTF32'
+    }
+
+    return body
+
+def get_response_with_retry(request, tries, retry=3):
+    """Get the response using re-try"""
+    try:
+        response = request.execute()
```
jerjou: FYI you can just pass in a `num_retries` parameter to `execute`, which will do nice things like exponential backoff, and also won't retry on user error.
[reference](https://google.github.io/google-api-python-client/docs/epy/googleapiclient.http.HttpRequest-class.html)

language/movie_nl/main.py
```diff
@@ -0,0 +1,366 @@
...
+
+    sentiment = response.get("documentSentiment")
+
+    if sentiment is None:
+        return (None, None)
+    else:
+        pol = sentiment.get("polarity")
+        mag = sentiment.get("magnitude")
+
+    if pol is None and mag is not None:
```
jerjou: Have you seen this happen? If so, can you add a comment on the situations in which this can happen?

language/movie_nl/main.py
puneith: I was looking for that exact thing when I asked Mona. Should have asked you. Thanks!

language/movie_nl/main.py
```diff
@@ -0,0 +1,366 @@
...
+    """Get the sentence-level sentiment."""
+    body = get_request_body(
+        sentence, syntax=False, entities=True, sentiment=True)
+
+    docs = service.documents()
+    request = docs.annotateText(body=body)
+
+    response = get_response_with_retry(request, 1)
+
+    sentiment = response.get("documentSentiment")
```
jerjou: Prefer single quotes to double quotes (except for docstrings)

language/movie_nl/main.py
```diff
@@ -0,0 +1,366 @@
...
+
+
+class Document(object):
+    """Document class captures a single document of movie reviews."""
+
+    def __init__(self, text, doc_id, doc_path):
+        self.text = text
+        self.doc_id = doc_id
+        self.doc_path = doc_path
+        self.sentent_pair = None
```
jerjou: Can you either give this property a more descriptive name (it's hard to know out-of-context what `sentent` is supposed to mean), or split it into two properties with descriptive names?

language/movie_nl/main.py
puneith: Yes I have. This is a bug in API which might have been fixed now. I didn't want to explicitly say "fixing because of a bug" or maybe I should?

language/movie_nl/main.py
```diff
@@ -0,0 +1,366 @@
...
+        self.label = None
+
+    def extract_all_sentences(self, service):
+        """Extract the sentences in a document."""
+
+        if self.sentent_pair is None:
```
jerjou: nit: I'd just check if it's `not None` and return it immediately, which lets you dedent the rest of the function.

language/movie_nl/main.py
```diff
@@ -0,0 +1,366 @@
...
+            request = docs.annotateText(body=request_body)
+
+            ent_list = []
+
+            response = request.execute()
+            entities = response.get('entities', [])
+            sentences = response.get('sentences', [])
+
+            sent_list = [
+                sentence.get('text').get('content') for sentence in sentences
```
jerjou: sentence.get('text', {}).get...

If you're using `.get`, it means you might be getting `None`, which will error if you try to `.get` it.

language/movie_nl/main.py
```diff
@@ -0,0 +1,366 @@
...
+def get_wiki_title(wiki_url):
+    """Get the wikipedia page title for a given wikipedia URL."""
+    try:
+        content = requests.get(wiki_url).text
+        return content.split('title')[1].split('-')[0].split('>')[1].strip()
+    except:
+        return os.path.basename(wiki_url).replace('_', ' ')
+
+
+def to_entity_json(entity, e_tuple):
```
jerjou: Can you give `e_tuple` a more descriptive name? It's hard to know what it is.

language/movie_nl/main.py
```diff
@@ -0,0 +1,366 @@
...
+
+    return json.dumps(json_doc)
+
+
+def get_sentiment_entities(service, document):
+    """Compute the overall sentiment volume in the document"""
+
+    try:
+        sentiments, entities = analyze_document(service, document)
+    except HttpError as e:
```
jerjou: unnecessary try/catch

language/movie_nl/main.py
```diff
@@ -0,0 +1,366 @@
...
+
+    sentiment_writer.flush()
+    entity_writer.flush()
+
+
+def document_generator(dir_path_pattern, count=None):
+    """Generator for the input movie documents."""
+    for running_count, item in enumerate(glob.iglob(dir_path_pattern)):
+        if count and running_count >= count:
+            raise StopIteration()
```
jerjou: FYI the more pythonic way to do this is to slice the list being iterated over:

    for item in glob.iglob(dir_name_pattern)[:count]:

If `count` is `None`, it will correctly give you the entire list.

language/movie_nl/main.py
```diff
@@ -0,0 +1,366 @@
...
+        else:
+            items.append(entity_item)
+
+    items.sort(reverse=reverse_bool)
+    items = [json.dumps(item[1]) for item in items]
+
+    if topn:
+        print('\n'.join(items[:topn]))
+    else:
+        print('\n'.join(items))
```
jerjou: Similarly here - if you default `--sample` to `None` below (when adding the `--sample` argument to `rank_parser`), you can omit the conditionals here and simply do:

    print('\n'.join(items[:topn]))

language/movie_nl/main.py
```diff
@@ -0,0 +1,366 @@
...
+
+def get_service():
+    """Build a client to the Google Cloud Natural Language API."""
+
+    credentials = GoogleCredentials.get_application_default()
+    scoped_credentials = credentials.create_scoped(
+        ['https://www.googleapis.com/auth/cloud-platform'])
+    http = httplib2.Http()
+    scoped_credentials.authorize(http)
+    return discovery.build('language', 'v1beta1', http=http)
```
jerjou: FYI now that the api is public, you can just do:

    return discovery.build('language', 'v1beta1' ,credentials=GoogleCredentials.get_application_default())

language/movie_nl/main.py
```diff
@@ -0,0 +1,366 @@
...
+
+
+def analyze(input_dir, sentiment_writer, entity_writer, sample, log_file):
+    """Movie demo main program"""
```
jerjou: Oh - can you update this doc string?

language/movie_nl/main.py
jerjou: Actually, I think all three (well, four including the comment) of these lines can be removed. The files will be closed once the script executes, no? And the reader is a generator, so doesn't need to be closed.

language/movie_nl/requirements.txt
```diff
@@ -0,0 +1,3 @@
+urlparse2==1.1.1
```
jerjou: This seems vestigial

language/movie_nl/main_test.py
```diff
@@ -0,0 +1,82 @@
...
+
+    assert sentiments[1].get('sentiment') == 1.0
+    assert sentiments[1].get('label') == 1
+
+    # assert entities
+    assert len(entities) == 1
+    assert entities[0].get('name') == 'Tom Cruise'
+    assert (entities[0].get('wiki_url') ==
+            'http://en.wikipedia.org/wiki/Tom_Cruise')
+    assert entities[0].get('sentiment') == 2.0
```
jerjou: Can you add a test for `rank_entities`?
Can you also either add a test for `analyze`, or modify the `process_movie_reviews` test to test `analyze` instead?

language/movie_nl/main.py
jerjou: Haha - nah. Good call :)

language/movie_nl/main.py
puneith: Yes. Cool

jerjou:
Mostly some clean-up and readability stuff. Also make sure to run

    nox -s lint -- language/movie_nl

and fix the lint errors to keep travis happy.

language/movie_nl/main_test.py
puneith: I will add the test. But for the test for analyze that's why I am tetsing process_movie_reviews because thats where the end to end work is done. DIP for process_movie_reviews since analyze creates the reader and service inside.

jerjou: (What's "DIP"?)

Yeah - I was thinking that `process_movie_reviews` was most of the test anyway. I did want to try to address the concern last night that the tests didn't catch a bunch of the issues we had, though, and figured more thoroughness in testing might help. I dunno - use your judgment I guess.

language/movie_nl/main.py
puneith: This is a genrator

language/movie_nl/main_test.py
puneith: Dependency Injection Principle

puneith:
@jerjou PTAL

language/movie_nl/main.py
waprin: I don't think Python interpreters guarantee files are closed at exit, maybe in practice they are but I think it's best practice to close them.

puneith: No guarantee and so fd's can pile up.

puneith:
Damn testing against some different nox perhaps.

language/movie_nl/main.py
jonparrott: I would much prefer you to just pass in filenames form argparse instead of complex types.

puneith: Ok I never noticed this change Jerjou made. I agree with you on that. Jerjou did you have other thoughts?

language/movie_nl/main.py
jerjou: Oh - do you mean that `glob.iglob` is a generator? In that case, you could use `itertools.islice(glob..., count)`, but I guess that doesn't make it much simpler (though still more pythonic).

Anyway, not a big deal.

language/movie_nl/main.py
```diff
@@ -0,0 +1,342 @@
...
+    try:
+        content = requests.get(wiki_url).text
+        return content.split('title')[1].split('-')[0].split('>')[1].strip()
+    except:
+        return os.path.basename(wiki_url).replace('_', ' ')
+
+
+def to_entity_json(entity, entity_sentiment):
+    """Convert the entity info to json."""
+    json_doc = {}
```
waprin: Think some more explanation of args would make it more readable

language/movie_nl/main.py
jerjou: I tend to prefer the `type`s that argparse provides (they keep the type casting to the part of the program that should care about it a la Single Responsibility Principle), but I'll defer to Jon.

language/movie_nl/main_test.py
```diff
@@ -80,3 +81,31 @@ def test_process_movie_reviews():
...
+
+def test_rank_positive_entities():
+    reader = [
+        '{"avg_sentiment": -12.0, "name": "Patrick Macnee", "sentiment": -12.0}',
+        '{"avg_sentiment": 5.0, "name": "Paul Rudd", "sentiment": 5.0}',
+        '{"avg_sentiment": -5.0, "name": "Martha Plimpton", "sentiment": -5.0}',
+        '{"avg_sentiment": 7.0, "name": "Lucy (2014 film)", "sentiment": 7.0}'
+    ]
+
+    sys.stdout = writer = StringIO.StringIO()
```
jerjou: Use `capsys`: [example](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/storage/api/crud_object_test.py#L19)

language/movie_nl/main.py
```diff
@@ -0,0 +1,342 @@
...
+def process_movie_reviews(service, reader, sentiment_writer, entity_writer):
+    """Perform some sentiment math and come up with movie review."""
+    collected_entities = {}
+
+    for document in reader:
+        try:
+            sentiment_total, entities = get_sentiment_entities(
+                service, document)
+        except HttpError as e:
+            logging.error("Error process_movie_reviews {}".format(e.content))
```
jerjou: Prefer single quotes to double quotes

jerjou:
After `capsys` fix, LGTM if waprin, jonparrott, and travis are happy :)

puneith:
@jonparrott @waprin PTAL

waprin:
I know you are trying to get this merged and I am fine to LGTM it but maybe a followup commit with more doc strings? I know people don't want spurious docstrings and think params are "obvious" but I think we overestimate how obvious things are. For example at one point you take `entity_sentiment` as a param and then do `avg_sentiment = float(entity_sentiment[0]) / float(entity_sentiment[1]`...so entity sentiment is a tuple with two items that need to be cast to floats? Especially given Python is an untyped language, a short docstring explaining the types of the params would make the code easier to follow.

jonparrott:
> a short docstring explaining the types of the params would make the code easier to follow.

In general, all non-primitive parameters should be explained. You can either do this within the long description or in an args section- whichever makes the most sense.

Also I'm curious as to why it's necessary to pass a fixed-length list as an argument when two separate parameters would do just as well?

language/movie_nl/main.py
```diff
@@ -0,0 +1,342 @@
...
+    """Get the wikipedia page title for a given wikipedia URL."""
+    try:
+        content = requests.get(wiki_url).text
+        return content.split('title')[1].split('-')[0].split('>')[1].strip()
+    except:
+        return os.path.basename(wiki_url).replace('_', ' ')
+
+
+def to_entity_json(entity, entity_sentiment):
+    """Convert entities and their associated sentiment to json."""
```
waprin: This is one place I think doc strings will help.

As Jon says I think for self-explanatory string types or int types, we can skip doc strings, otherwise we should have them

language/movie_nl/main.py
```diff
@@ -0,0 +1,389 @@
...
+        default=None
+        )
+
+    analyze_parser = subparsers.add_parser('analyze')
+
+    analyze_parser.add_argument(
+        '--inp', help='location of the input', required=True)
+    analyze_parser.add_argument(
+        '--sout', help='location of the sentiment output', required=True,
+        type=argparse.FileType('w'))
```
waprin: I thought conversation above ended up with us saying we weren't doing FileType argparse stuff?

puneith: Actually it never ended with that conclusion. The debate was still on. To actually think of it this does abstract away the file opening which is nice :)

jonparrott: Nice for who? The person reading this for the first time will likely have
to go to the argparse docs to figure out what exactly is going on here.

puneith: Well thats true of any abstraction. This is saving us an extra step.

waprin: Well at a minimum the file closing needs to be  in a try: finally to close it, and the thing I like least about this is lack of clear semantics for closing the file. But I'm not a huge fan of this argparse feature and neither is Jon so while it's a small thing that could go either way, I think that's enough consensus to get rid of it.

puneith: @waprin @jonparrott Ok then we will get rid. Should we get rid before merge or can we do that after merge?

waprin: It's of course better to do before merge. If you think there's a launch-related urgency then we can merge.

puneith: Ok PTAL at the fix.

waprin:
LGTM

waprin:
Given Traivs passes I'm sure it's fine but is this a nox issue?

```
import file mismatch:
imported module 'main_test' has this __file__ attribute:
  /Users/waprin/code/python-docs-samples/language/movie_nl/main_test.py
which is not the same as the test file we want to collect:
  /Users/waprin/code/python-docs-samples/language/ocr_nl/main_test.py
```

waprin:
That's when I just run 

` nox --session tests -- language/`

puneith:
So when I do nox it is successful. I see there is movie_nl and ocr_nl in your output above? Is this something with ocr? Donno?

waprin:
Are you doing just nox or the command I ran that just does the language directory? It passes on Travis so likely an issue on my laptop ,just mentioning here....

puneith:
@waprin I had to update the README and also get rid of the FileType for the rank feature.

waprin:
alright i just ran commands again and worked for me. meeeerge away

waprin:
I still have that test issue but again pretty sure it's nox related. out of curiosity can you try that? `nox --session tests -- language`

jonparrott:
You need to specify the individual sample -- language/movie_nl

jerjou:
Next time you merge, can you please squash your commits?

jonparrott:
Please let Jerjou, Bill, or Myself merge in the future, just for consistency. I'm going to manually squash these commits.

Issue #424, opened 2016-07-18
Speech API not working for some files


BalaExotel:
Though speech_rest API worked correctly for the given resources, it returned null JSON everytime whenever I passed other audio files in raw format (16-bit, little endian, 16KHz)?

I've attached screenshot of the error. I've also attached that audio file as ZIP (Though the conversation is in another language, it has a few English words like "Hello" which I want to capture.).

![googlecloudasr](https://cloud.githubusercontent.com/assets/15904944/16906406/fffc28ae-4cd4-11e6-9dac-0b16a5c943d0.png)
[Hello1.raw.zip](https://github.com/GoogleCloudPlatform/python-docs-samples/files/368360/Hello1.raw.zip)

jonparrott:
@puneith or @jerjou can you take a look?

edesaintchamas:
Same issue:

The provided .raw files are working:
`$ python speech_rest.py resources/audio2.raw`
`{"results": [{"alternatives": [{"confidence": 0.97258663, "transcript": "the rain in Spain stays mainly on the plain"}]}]}`

But the provided FLAC does not
`$ python speech_rest.py resources/audio.flac`
`{}`

I also tried with voice samples I found online (https://archive.org/details/VoiceSample) I also just get `{}`, no results or error are provided

puneith:
Can you please edit the speech_rest.py file (in line 63) and change the encoding to "FLAC". I will open a ticket to make this configurable.

BalaExotel:
Hey,

I did NOT pass a FLAC filr.

I passed in a "raw" format file. It worked for the raw files in the
resources directory. When I passed our own 16-bit little endian 16 KHz
sampled raw audio file, it printed null set.

PFA the raw file. Thanks.



Balaraman Lakshmanan | Research Engineer | +91 9900531463
Website <http://exotel.in/> | We’re Hiring <http://exotel.in/about/careers/> |
Twitter <https://twitter.com/balaramanl> | LinkedIn
<https://in.linkedin.com/in/balaramanl>

jerjou:
To be fair, I didn't understand the audio file either :-)

I think at least one of the main issues here is that it wasn't obvious that the api looks for english (US) by default if you don't specify the language. I'm updating the sample to explicitly state the language, and add a link to the config docs so you can look up other languages you can specify.

[...]

PR #429, opened 2016-07-22
Add comments for config options.

jerjou:
Some folks were tripping up on this.

Addresses #424

### DIFF
--- a/speech/api/speech_async_grpc.py
+++ b/speech/api/speech_async_grpc.py
@@ -59,8 +59,14 @@ def main(input_uri, encoding, sample_rate):
     # https://github.com/googleapis/googleapis/blob/master/google/cloud/speech/v1beta1/cloud_speech.proto
     response = service.AsyncRecognize(cloud_speech_pb2.AsyncRecognizeRequest(
         config=cloud_speech_pb2.RecognitionConfig(
-            encoding=encoding,
-            sample_rate=sample_rate,
+            # There are a bunch of config options you can specify. See
+            # https://goo.gl/A6xv5G for the full list.
+            encoding=encoding,  # one of LINEAR16, FLAC, MULAW, AMR, AMR_WB
+            sample_rate=sample_rate,  # the rate in hertz
+            # See
+            # https://g.co/cloud/speech/docs/best-practices#language_support
+            # for a list of supported languages.
+            language_code='en-US',  # a BCP-47 language tag
         ),
         audio=cloud_speech_pb2.RecognitionAudio(
             uri=input_uri,
--- a/speech/api/speech_async_rest.py
+++ b/speech/api/speech_async_rest.py
@@ -57,8 +57,12 @@ def main(speech_file):
     service_request = service.speech().asyncrecognize(
         body={
             'config': {
-                'encoding': 'LINEAR16',
-                'sampleRate': 16000
+                # There are a bunch of config options you can specify. See
+                # https://goo.gl/EPjAup for the full list.
+                'encoding': 'LINEAR16',  # raw 16-bit signed LE samples
+                'sampleRate': 16000,  # 16 khz
+                # See https://goo.gl/DPeVFW for a list of supported languages.
+                'languageCode': 'en-US',  # a BCP-47 language tag
             },
             'audio': {
                 'content': speech_content.decode('UTF-8')
--- a/speech/api/speech_grpc.py
+++ b/speech/api/speech_grpc.py
@@ -56,8 +56,14 @@ def main(input_uri, encoding, sample_rate):
     # https://github.com/googleapis/googleapis/blob/master/google/cloud/speech/v1beta1/cloud_speech.proto
     response = service.SyncRecognize(cloud_speech.SyncRecognizeRequest(
         config=cloud_speech.RecognitionConfig(
-            encoding=encoding,
-            sample_rate=sample_rate,
+            # There are a bunch of config options you can specify. See
+            # https://goo.gl/A6xv5G for the full list.
+            encoding=encoding,  # one of LINEAR16, FLAC, MULAW, AMR, AMR_WB
+            sample_rate=sample_rate,  # the rate in hertz
+            # See
+            # https://g.co/cloud/speech/docs/best-practices#language_support
+            # for a list of supported languages.
+            language_code='en-US',  # a BCP-47 language tag
         ),
         audio=cloud_speech.RecognitionAudio(
             uri=input_uri,
--- a/speech/api/speech_rest.py
+++ b/speech/api/speech_rest.py
@@ -60,8 +60,12 @@ def main(speech_file):
     service_request = service.speech().syncrecognize(
         body={
             'config': {
-                'encoding': 'LINEAR16',
-                'sampleRate': 16000
+                # There are a bunch of config options you can specify. See
+                # https://goo.gl/EPjAup for the full list.
+                'encoding': 'LINEAR16',  # raw 16-bit signed LE samples
+                'sampleRate': 16000,  # 16 khz
+                # See https://goo.gl/DPeVFW for a list of supported languages.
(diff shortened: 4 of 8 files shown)


puneith:
Do we want to make the language a parameter to the code or just the way it is so they need to edit the py file?

jonparrott:
Lgtm

jerjou:
Updated PR to update all the samples in this dir with similar comments. PTAL.

jonparrott:
Please do those as separate PRs in the future. :)

speech/api/speech_async_grpc.py
```diff
@@ -59,8 +59,12 @@ def main(input_uri, encoding, sample_rate):
     # https://github.com/googleapis/googleapis/blob/master/google/cloud/speech/v1beta1/cloud_speech.proto
     response = service.AsyncRecognize(cloud_speech_pb2.AsyncRecognizeRequest(
         config=cloud_speech_pb2.RecognitionConfig(
-            encoding=encoding,
-            sample_rate=sample_rate,
+            # There are a bunch of config options you can specify. See
+            # https://g.co/cloud/speech/reference/rpc/google.cloud.speech.v1beta1#google.cloud.speech.v1beta1.RecognitionConfig  # noqa
+            # for the full list.
+            encoding=encoding,  # one of LINEAR16, FLAC, MULAW, AMR, AMR_WB
+            sample_rate=sample_rate,  # the rate in Hz
```
jonparrott: Nit: only use the abbreviation after a figure, use the full term elsewhere

jerjou: Hrm... is it capitalized when spelled out? Makes it look like the car company...

jonparrott: "hertz" is fine.

jerjou: Done

speech/api/speech_async_rest.py
```diff
@@ -57,8 +57,12 @@ def main(speech_file):
     service_request = service.speech().asyncrecognize(
         body={
             'config': {
-                'encoding': 'LINEAR16',
-                'sampleRate': 16000
+                # There are a bunch of config options you can specify. See
+                # https://cloud.google.com/speech/reference/rest/Shared.Types/RecognitionConfig  # noqa
```
ace-n: what's the #noqa thing doing here?

speech/api/speech_grpc.py
```diff
@@ -56,8 +56,12 @@ def main(input_uri, encoding, sample_rate):
     # https://github.com/googleapis/googleapis/blob/master/google/cloud/speech/v1beta1/cloud_speech.proto
     response = service.SyncRecognize(cloud_speech.SyncRecognizeRequest(
         config=cloud_speech.RecognitionConfig(
-            encoding=encoding,
-            sample_rate=sample_rate,
+            # There are a bunch of config options you can specify. See
+            # https://g.co/cloud/speech/reference/rpc/google.cloud.speech.v1beta1#google.cloud.speech.v1beta1.RecognitionConfig  # noqa
```
ace-n: Another #noqa

speech/api/speech_rest.py
```diff
@@ -60,8 +60,12 @@ def main(speech_file):
     service_request = service.speech().syncrecognize(
         body={
             'config': {
-                'encoding': 'LINEAR16',
-                'sampleRate': 16000
+                # There are a bunch of config options you can specify. See
+                # https://cloud.google.com/speech/reference/rest/Shared.Types/RecognitionConfig  # noqa
```
ace-n: #noqa

speech/api/speech_streaming.py
```diff
@@ -83,13 +83,19 @@ def request_stream(stop_audio, channels=CHANNELS, rate=RATE, chunk=CHUNK):
...
-        rate: The sampling rate.
+        rate: The sampling rate in hertz.
         chunk: Buffer audio into chunks of this size before sending to the api.
     """
     # The initial request must contain metadata about the stream, so the
     # server knows how to interpret it.
     recognition_config = cloud_speech.RecognitionConfig(
-        encoding='LINEAR16', sample_rate=rate)
+        # There are a bunch of config options you can specify. See
+        # https://g.co/cloud/speech/reference/rpc/google.cloud.speech.v1beta1#google.cloud.speech.v1beta1.RecognitionConfig  # noqa
```
ace-n: #noqa

ace-n:
LGTM, other than the #noqa thing.

speech/api/speech_async_rest.py
jerjou: It's 'cuz it's a long line - if I don't put the # noqa comment, the linter will complain that the line exceeds 79 characters in length - which normally we don't want, but you can't really break up a url in a comment..

ace-n: Can you do it with a goo.gl URL maybe?

ace-n: (Or better yet: submit this as an issue for the linter team.)

puneith:
@jerjou I am thinking of adding this somewhere https://cloud.google.com/speech/docs/best-practices#language_support appropriately.

speech/api/speech_async_rest.py
jerjou: We are the linter team :)
What's the desired behavior? We still want to catch long lines, and we still want a way to tell the linter to ignore long lines that for one reason or another we wish to allow..

jonparrott: I'm generally not a fan of our tools causing us to add things into sample code that wouldn't otherwise be there. (Principle of least surprise).

We should either:
1) leave the line.
2) use a shorturl.

jerjou:
@puneith: I think it's sufficient in the sample to mention that there's stuff you can configure and point to the config documentation, and let the config docs point to the list of supported languages, etc. What do you think?

puneith:
@jerjou In general agree, but its just that this encoding is bit tedious thing to grok that I am inclined towards more sympathy towards the user, hence the idea of adding the link.

speech/api/speech_async_rest.py
jerjou: Done - used a shorturl.

jerjou:
Updated to add link to supported languages. PTAL

puneith:
@jerjou I think we should be consistent in either using underscore or camel case. E.g., either "languageCode" or "language_code"

jonparrott:
Underscore, please.

jerjou:
This is dictated by the API. We have no control over this from the client.

jerjou:
Added a commit to fix the lint error.

jerjou:
Oh cool - you're right. You can use snake_case for the REST stuff too.

I wonder, though, whether it's best to do so... since the parameters are passed in as a dict, the keys are quoted, which makes it more okay in my head to use camelCase.  And the [documentation](https://cloud.google.com/speech/reference/rest/Shared.Types/RecognitionConfig) that we point to uses camelCase, since it's JSON documentation. Also - [other apis](https://developers.google.com/resources/api-libraries/documentation/vision/v1/python/latest/vision_v1.images.html#annotate) use camelCase in their documentation as well...

puneith:
My comment was just about consistency. @jonparrott said snake_case.

jonparrott:
Use camelCase inside of the API request body for consistency with the API documentation. It sucks, but that's how it is.

Jerjou: LGTM once you're happy.

jerjou:
Yeah - I agree with you that consistency is a good thing.

What I'm saying, though, is that there are multiple ways to be consistent. We can be consistent with the documentation that we point to as well as historical usage of the google-api-python-client (which specifies camelCase), or we can be consistent between our samples and with the python style guide (ie snake_case. The grpc client doesn't give us a choice of camelCase, since it uses named parameters which are snake_case).

I think the decision isn't clear-cut, but I'm leaning towards camelCase, since the key value is quoted.

PR #434, opened 2016-07-30
Streaming bug fixes

vishay:
* Exit script upon returning streaming response, if single_utterance = True
* Accept single_utterance as a command line arg ("-s").


jerjou:
Thanks for the PR! This would definitely be a useful addition to the script.
However, our goal for this repository is to have scripts that are relatively simple for the purposes of education, that folks can copy and add to according to their needs. So - while this is a useful feature, my sense is that it falls more in the category of a feature to add on after understanding the basics of the API, rather than something you need to understand in order to use the API.

PR #436, opened 2016-08-03
Update requirements.

jonparrott:


jonparrott:
@waprin @jerjou FYI, will merge when Travis passes.

PR #438, opened 2016-08-03
Change the way nox handles installing requirements.

jonparrott:
Instead of running pip like 100 times to install dependencies for all samples,
there's a new `nox -s reqrollup` that collates everything into one file. Pip
is now only called once. This should somewhat speed up builds.


jonparrott:
@waprin @jerjou FYI, merging once Travis is green.

PR #439, opened 2016-08-04
Add Endpoints 1.1 sample

saicheems:
Updated the sample to use the Endpoints 1.1 Python client with
configuration options for service management and service control.


saicheems:
This is an in progress sample for the 1.1 client - The functions include /echo and /echo/getUserEmail. We're also trying to show off integration with service control and service management via swagger.

Any feedback on the design and how to improve the surface is welcome.

appengine/standard/endpoints/backend/app.yaml
```diff
@@ -3,14 +3,25 @@ threadsafe: true
...
 
 libraries:
 - name: pycrypto
   version: 2.6
-- name: endpoints
-  version: 1.0
+
+# Endpoints 1.1 uses background threads for caching and reporting to service
+# management and service control.
+manual_scaling:
```
jonparrott: Ouch. Is there any plans to make this work with auto-scaled modules?

appengine/standard/endpoints/backend/app.yaml
```diff
@@ -3,14 +3,25 @@ threadsafe: true
...
+  instances: 1
+
+beta_settings:
+  use_endpoints_api_management: true
+  endpoints_swagger_spec_file: echo-v1_swagger.json
+
+env_variables:
+  #TODO: The Endpoints service name.
```
jonparrott: Don't use #TODO, just say "Replace with your endpoints service name, for example your-service.appspot.com"

appengine/standard/endpoints/backend/main.py
```diff
@@ -22,128 +22,46 @@
 from protorpc import remote
 # [END imports]
 
-
 # [START messages]
-class Greeting(messages.Message):
-    """Greeting that stores a message."""
+class Echo(messages.Message):
+    """An echo that stores a message."""
```
jonparrott: How about `A proto Message that contains a simple string field.`

appengine/standard/endpoints/backend/main.py
```diff
@@ -22,128 +22,46 @@
...
     message = messages.StringField(1)
```
jonparrott: how about for disambiguation we call this 'content' instead?

appengine/standard/endpoints/backend/main_test.py
```diff
@@ -12,42 +12,43 @@
...
 # limitations under the License.
 
+import endpoints
 import main
 import mock
+import unittest
+from google.appengine.ext import testbed
 from protorpc import message_types
 
+class EchoTestCase(unittest.TestCase):
```
jonparrott: please use pytest style tests. Refer to the test you overwrote here for an example.

appengine/standard/endpoints/backend/app.yaml
saicheems: There are, I don't have all the details around it, but once I do I'll be sure to test/update this pull.

Maybe @tbetbetbe can offer some more information.

jonparrott: SG.

appengine/standard/endpoints/backend/app.yaml
saicheems: Done

appengine/standard/endpoints/backend/main.py
saicheems: Done

appengine/standard/endpoints/backend/main.py
saicheems: Done

appengine/standard/endpoints/backend/main_test.py
saicheems: Done

appengine/standard/endpoints/backend/echo-v1_swagger.json
```diff
@@ -0,0 +1,64 @@
+{
```
jonparrott: In the flexible endpoints sample, we use yaml for the definition. Is it possible to use yaml here as well or not?

jonparrott:
I know this goes back on what I said initial, but since this will break a lot of the current docs pages can I recommend that you place this in `/appengine/standard/endpoints-v1.1/backend` instead of replacing the existing sample?

appengine/standard/endpoints/backend/echo-v1_swagger.json
saicheems: It is possible, I think. This swagger is autogenerated from the endpoints service by a new tool. I don't think it has an option to generate yaml, but I can convert it if that's preferred.

jonparrott: If it's autogenerated don't sweat it.

saicheems:
Done

saicheems:
Redone against master.

PR #463, opened 2016-08-18
New code samples for Python quickstart for Endpoints Frameworks v2

bradfriedman:
These are for the quickstarts for the Endpoints Frameworks v2 for Python beta.


jonparrott:
@bradfriedman help me understand how this differs from @saicheems' sample [here](https://github.com/GoogleCloudPlatform/python-docs-samples/tree/master/appengine/standard/endpoints-frameworks-v2/backend)?

bradfriedman:
The reason we're migrating the old samples to v2 is so that we don't have to write a whole new tutorial for the Echo sample that @saicheems already wrote. We just wanted to bring the old sample up to v2 for this purpose.

jonparrott:
Gotcha - so this is a separate sample. We'll have two - an echo sample and this sample?

If that's the case, can you make sure both of these are in the `endpoints-framework-v2` folder? I guess @saicheems should be under `echo` and yours can be `quickstart`.

jonparrott:
(be sure to rebase your branch on master before you do the rename, @saicheems has a README change in before this PR)

bradfriedman:
Sure, PTAL

jonparrott:
Can we drop the `backend` part from the path too?

(sorry! other than that this sample looks good)

bradfriedman:
PTAL :)

jonparrott:

PR #468, opened 2016-08-19
Make nox lint each sample individually

jonparrott:
This makes import order checking work correctly.

/fyi @waprin @pfritzsche


waprin:
lgtm

PR #474, opened 2016-08-23
Add Scipy example

waprin:
I tried to keep this pretty simple just to demonstrate it works on Flex.

Could use GCS to make it a little less clumsy but leaned towards simplicity.


waprin:
I love scipy!

waprin:
test3

.gitignore
```diff
@@ -19,3 +19,4 @@ secrets.tar
 junit.xml
 credentials.dat
 .nox
+appengine/flexible/scipy/assets/resized_google_logo.jpg
```
jonparrott: Just add this to the .gitignore in `/appengine/flexible/scipy`.

appengine/flexible/scipy/main.py
```diff
@@ -0,0 +1,52 @@
...
+    """Demonstrates using scipy to resize an image."""
+    app_path = os.path.dirname(os.path.realpath(__file__))
+    image_path = os.path.join(app_path, 'assets/google_logo.jpg')
+    img = scipy.misc.imread(image_path)
+    img_tinted = scipy.misc.imresize(img, (300, 300))
+    output_image_path = os.path.join(
+        app_path, 'assets/resized_google_logo.jpg')
+    # Write the tinted image back to disk
+    scipy.misc.imsave(output_image_path, img_tinted)
+    return "Image resized."
```
jonparrott: Why not return the image itself?

requirements-dev.txt
```diff
@@ -40,6 +40,7 @@ requests==2.11.1
 requests[security]==2.11.1
 responses==0.5.1
 rsa==3.4.2
+scipy==0.18.0
```
jonparrott: Did you do `nox -s reqrollup`?

waprin: I did not but I just did and there's no changes.

jonparrott: Cool, just wanted to verify.

appengine/flexible/scipy/main.py
waprin: I don't think we should really be serving static images from the filesystem like that and I'd rather keep it simple than demonstrate a bad practice. I'm more inclined to actually upload to a bucket but it just seems like overkill considering the only point is to say "hey, scipy will work if you add it to the requirements.txt"

.gitignore
waprin: done

appengine/flexible/scipy/main.py
jonparrott: I understand, but I'd still like the user to see the result. It's one more line of code, why not?

waprin: ¯\_(ツ)_/¯

appengine/flexible/scipy/main.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+# [START scipy]
+@app.route('/')
+def resize():
+    """Demonstrates using scipy to resize an image."""
+    app_path = os.path.dirname(os.path.realpath(__file__))
+    image_path = os.path.join(app_path, 'assets/google_logo.jpg')
+    img = scipy.misc.imread(image_path)
+    img_tinted = scipy.misc.imresize(img, (300, 300))
```
jerjou: perhaps `img_resized` instead? Also, update the comment in line 34

appengine/flexible/scipy/main.py
```diff
@@ -0,0 +1,52 @@
...
+    output_image_path = os.path.join(
+        app_path, 'assets/resized_google_logo.jpg')
```
jerjou: Probably unnecessary to have the resized file checked in, since that's what you're generating..

appengine/flexible/scipy/main.py
jonparrott: It's been merged and Bill's at burning man, *man*. Send a PR if you feel strongly.

appengine/flexible/scipy/requirements.txt
```diff
@@ -0,0 +1,4 @@
+Flask==0.11.1
+gunicorn==19.6.0
+scipy==0.18.0
+Pillow==3.3.1
```
jerjou: nit: Is Pillow necessary? I assume it's just a dependency of scipy, since I don't see it imported itself?

appengine/flexible/scipy/main.py
jerjou: Just catching up on email 😅

PR #477, opened 2016-08-24
Add pubsub publisher and subscriber samples

jonparrott:
/cc @tmatsuo you are an optional reviewer.


pubsub/cloud-client/subscriber.py
```diff
@@ -0,0 +1,127 @@
...
+    # Change return_immediately=False to block until messages are
+    # received.
+    results = subscription.pull(return_immediately=True)
+
+    print('Received {} messages.'.format(len(results)))
+
+    for ack_id, message in results:
+        print('* {}: {}, {}'.format(
+            message.message_id, message.data, message.attributes))
+
```
waprin: seems like if you're iterating through the list twice you could just ack after printing.

waprin:
lgtm

pubsub/cloud-client/subscriber.py
jonparrott: I explicitly wanted to do this separately.

waprin: yep, seems reasonable as well, just pointing out in unlikely case you overlooked.

jonparrott:
@dpebot merge when travis passes.

PR #479, opened 2016-08-25
Ignore bad file in endpoints lib

saicheems:
Causes an error in `gcloud beta app deploy`


jonparrott:
Clever. FYI, there's an internal bug to get this resolved as it block the usage of gcloud-python on GAE standard.

saicheems:
Ah, good to know! Thanks!

PR #480, opened 2016-08-25
Add pubsub iam samples

jonparrott:


waprin:
code lgtm, but i feel weird about this whole pub/sub iam thing given what @elibixby  is doing which seems more general. guessing you have a reason not to just hold off for that?

jonparrott:
> guessing you have a reason not to just hold off for that?

We need samples for this ASAP, as it's part of the veneer launch. We can update them when Eli's general thing is in, they're trivial samples.

jonparrott:

PR #481, opened 2016-08-25
Add style guide information to CONTRIBUTING.md.

tswast:
Rather than use https://github.com/GoogleCloudPlatform/Template/wiki/style.html I want to have that page link to the appropriate repos.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -33,7 +33,8 @@ be able to accept your pull requests.
 
 ## Style
 
-Samples in this repository follow the [Google Python Style
+Samples in this repository follow [PEP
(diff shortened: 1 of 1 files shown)


jonparrott:
Might as well go ahead and mention that we follow the Google Python style guide: https://google.github.io/styleguide/pyguide.html

tswast:
Should I change the PEP 8 link to the Google Python Style guide or do we follow the intersection of those two?

tswast:
I updated to point at the Google Python Style Guide instead of PEP 8. I see that the external one doesn't allow CamelCaseFunctionName, so that's good.

jonparrott:
Yeah, we follow both.

jonparrott:
Thanks, @tswast.

PR #482, opened 2016-08-25
Add needed exclude tags for docs.

halgrimur:


jonparrott:
I'd rather a tag around the whole app than an exclusion tag around the license. You can also use a regexp for `include.*`

PR #490, opened 2016-08-31
Remove local deployment from README

bradfriedman:
Local running does not yet work with dev_appserver.py. Removing those instructions from the README.


jonparrott:
I'd prefer a warning rather than just removing these instructions, how would you feel about that @bradfriedman?

jonparrott:
@bradfriedman bump?

bradfriedman:
Sorry for delay, just getting back from vacation. Please take another look here.

PR #494, opened 2016-09-01
Correct the usage of the echo-client.py command

lookuptable:
The command requires three arguments while the instruction only
specified two.


jonparrott:
Thanks!

Issue #503, opened 2016-09-07
App Engine testrunner should exit with a non-zero exit code when tests fail


maltem-za:
This doesn't seem significant enough for me to want to fill out a CLA... / please add this.

I'm trying out [Bitbucket Pipelines](https://confluence.atlassian.com/bitbucket/bitbucket-pipelines-beta-792496469.html) and noticed that my builds were passing even though tests were failing. This is due to the fact that the testrunner always exits with code 0 ([someone else had a similar problem with PHP tests](https://bitbucket.org/site/master/issues/12810/pipeline-was-marked-as-passed-even-though)).

I suspect the same is true for other build agents. The following patch takes care of it:

```
--- a/runner.py    Tue Sep 06 17:49:32 2016 +0200
+++ b/runner.py    Wed Sep 07 10:56:25 2016 +0200
@@ -76,7 +76,7 @@
 
     # Discover and run tests.
     suite = unittest.loader.TestLoader().discover(test_path, "*_test.py")
-    unittest.TextTestRunner(verbosity=2).run(suite)
+    return unittest.TextTestRunner(verbosity=2).run(suite)
 
 
 if __name__ == '__main__':
@@ -88,6 +88,8 @@
         sys.exit(1)
     SDK_PATH = args[0]
     TEST_PATH = args[1]
-    main(SDK_PATH, TEST_PATH)
+    result = main(SDK_PATH, TEST_PATH)
+    if not result.wasSuccessful():
+        sys.exit(1)
 
 # [END runner]
```

jonparrott:
Thanks! I'll address this at the same time as #500.

PR #504, opened 2016-09-07
Fixup GAE standard test runner sample

### LINKED ISSUE #500: App Engine testrunner doesn't behave like docs suggest
This is relatively minor - just something I noticed relating to [/appengine/standard/localtesting/runner.py](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/appengine/standard/localtesting/runner.py)

The App Engine [Setting up a testing framework](https://cloud.google.com/appengine/docs/python/tools/localunittesting#Python_Setting_up_a_testing_framework) documentation states: "Test files follow the convention of having **test** prefixed to their name."

However, 1fe6ae4 changed the behaviour to look for files **ending** with **_test.py**

This makes sense if you're trying to run the sample tests, but not if you've used the testrunner on another project and named your test modules according to what has up until now been the default expectation, being test files with test **prefixed**. If you were to update your testrunner code to the latest and greatest you'll now find none of your tests being run. That is **if** you notice - assuming most people don't set up their builds to fail on:

```
----------------------------------------------------------------------
Ran 0 tests in 0.000s

OK
```

So, I would argue that it would be better to stick with the default (discover test files prefixed with test) and instead rename the sample tests.

TL;DR;
minor mismatch with documentation, has potential to cause havoc for people that have used the testrunner and who update it

jonparrott:
Fixes #500 and #503.

### DIFF
--- a/appengine/standard/localtesting/runner.py
+++ b/appengine/standard/localtesting/runner.py
@@ -1,4 +1,5 @@
-#!/usr/bin/python
+#!/usr/bin/env python2
+
 # Copyright 2015 Google Inc
 #
 #    Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


jerjou:
LGTM

jonparrott:
Merging w/o travis, as this file does not get tested (it's a test runner, go figure).

PR #518, opened 2016-09-13
endpoints-frameworks-v2: fix echo sample

tangiel:
* Tweak README
* Add namespace declaration to appengine_config.py


appengine/standard/endpoints-frameworks-v2/echo/appengine_config.py
```diff
@@ -1,4 +1,10 @@
 from google.appengine.ext import vendor
+import pkg_resources
 
 # Add any libraries installed in the `lib` folder.
 vendor.add('lib')
+
+# The Python runtime declares a "google" package, but so does
```
jonparrott: This should already be the case.

Issue #515, opened 2016-09-12
grpc streaming example crashes and hangs


thecloudist:
Just cloned the latest repo for grpc -
running this on Raspbian (Raspberry Pi)

Cannot even break out of the program at all.
The sample rate of my usb mic is limited to 44100 and 48000.  I tried changing those in the example but I get the same results.

I have pyaudio installed up to date.
Sound recording and playback both work with pyaudio.

The  ~/.asoundrc script is here:

```
pcm.!default {
    type asym
    playback.pcm "hw:0,0"
    capture.pcm "hw:1,1"
}

# pcm.!default {
#        type hw
#        card 0
# }

ctl.!default {
        type hw
        card 0
}

#pcm.record {
#    type plug;
#    slave {
#        pcm "hw 1,1"
#    }
# }
```

Here is the output of the program.

```
(grpc) pi@raspberrypi:~/python-docs-samples/speech/grpc $ python transcribe_streaming.py
ALSA lib pcm_dmix.c:1022:(snd_pcm_dmix_open) unable to open slave
ALSA lib pcm.c:2239:(snd_pcm_open_noupdate) Unknown PCM cards.pcm.rear
ALSA lib pcm.c:2239:(snd_pcm_open_noupdate) Unknown PCM cards.pcm.center_lfe
ALSA lib pcm.c:2239:(snd_pcm_open_noupdate) Unknown PCM cards.pcm.side
ALSA lib pcm.c:2239:(snd_pcm_open_noupdate) Unknown PCM cards.pcm.hdmi
ALSA lib pcm.c:2239:(snd_pcm_open_noupdate) Unknown PCM cards.pcm.hdmi
ALSA lib pcm.c:2239:(snd_pcm_open_noupdate) Unknown PCM cards.pcm.modem
ALSA lib pcm.c:2239:(snd_pcm_open_noupdate) Unknown PCM cards.pcm.modem
ALSA lib pcm.c:2239:(snd_pcm_open_noupdate) Unknown PCM cards.pcm.phoneline
ALSA lib pcm.c:2239:(snd_pcm_open_noupdate) Unknown PCM cards.pcm.phoneline
ALSA lib pulse.c:243:(pulse_connect) PulseAudio: Unable to connect: Connection refused

ALSA lib pulse.c:243:(pulse_connect) PulseAudio: Unable to connect: Connection refused

ALSA lib pcm_dmix.c:1022:(snd_pcm_dmix_open) unable to open slave
[...]

PR #527, opened 2016-09-15
Speech api: Fill audio buffer in a separate thread.

jerjou:
This is to avoid timing issues where the request thread doesn't poll the
generator fast enough to consume all the incoming audio data from the input
device. In that case, the audio device buffer overflows, leading to lost data
and exceptions and other nastiness.

Address #515

### DIFF
--- a/speech/grpc/transcribe_streaming.py
+++ b/speech/grpc/transcribe_streaming.py
@@ -18,21 +18,26 @@ from __future__ import division
 
 import contextlib
 import re
+import signal
 import threading
 
-from gcloud.credentials import get_credentials
+from gcloud import credentials
 from google.cloud.speech.v1beta1 import cloud_speech_pb2 as cloud_speech
 from google.rpc import code_pb2
 from grpc.beta import implementations
+from grpc.framework.interfaces.face import face
 import pyaudio
+from six.moves import queue
 
 # Audio recording parameters
 RATE = 16000
-CHANNELS = 1
 CHUNK = int(RATE / 10)  # 100ms
 
-# Keep the request alive for this many seconds
-DEADLINE_SECS = 8 * 60 * 60
+# The Speech API has a streaming limit of 60 seconds of audio*, so keep the
+# connection alive for that long, plus some more to give the API time to figure
+# out the transcription.
+# * https://g.co/cloud/speech/limits#content
+DEADLINE_SECS = 60 * 3 + 5
 SPEECH_SCOPE = 'https://www.googleapis.com/auth/cloud-platform'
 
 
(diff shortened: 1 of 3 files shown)


jonparrott:
This seems to add a lot of complexity. Is this an edge case or a common case?

jerjou:
Mm... so I think the way I was doing it before was actually incorrect. Here's my current understanding: before, there were two threads - one handling sending requests, and one handling receiving responses. However, there are actually three asynchronous things happening - requests, responses, and input from the microphone. Two threads works as long as your request thread consumes audio faster than realtime, because input devices have a buffer somewhere. But if your request thread stalls - because of network latency, or resource constraints, or what have you - then the audio buffer can overflow.

So I think this is actually the Right Way to do it. If you have any ideas on how to express it more simply, I'm all ears :-)

jerjou:
Actually, I have an idea... gimme a sec.

jonparrott:
Okay - let me know.

puneith:
@jonparrott I am afraid this may not be edge case. I think @jerjou is heading in the right direction, but I am thinking that with the same logic this code should be simplified by using futures.

jerjou:
Okay - modified to make the three threads more explicit.
@puneith - can you be more specific? How can we use futures to simplify things?

puneith:
@jerjou https://docs.python.org/3/library/concurrent.futures.html using Executors.

jonparrott:
This looks good to me. We can switch to `concurrent` in another PR if you two agree to do so.

jonparrott:

jerjou:
@puneith - can you be more specific? How can we use Executors to simplify things?

I think a major refactor is outside the scope of this PR. If you'd like to refactor to use Executors, please submit a separate PR, which will also help the specificity.

puneith:
Sure, you can merge this and I will do that as a separate PR.

puneith:
@jerjou Other than travis this is merge eligible?

jerjou:
Okay, yeah. Now it's ready.
I had an idea when I was on the way home on how to make things better / more correct :-)

jerjou:
FYI the changes I made were somewhat non-trivial, so I'd appreciate another look.

speech/grpc/transcribe_streaming.py
```diff
@@ -126,25 +174,34 @@ def listen_print_loop(recognize_stream):
 
         # Exit recognition if any of the transcribed phrases could be
         # one of our keywords.
-        if any(re.search(r'\b(exit|quit)\b', alt.transcript)
+        if any(re.search(r'\b(exit|quit)\b', alt.transcript, re.I)
                for result in resp.results
                for alt in result.alternatives):
             print('Exiting..')
```
puneith: The program doesn't exit even when you say exit.

jerjou: It should - are you seeing that it doesn't? It should break out of the listen-print loop and return from the function, which should exit all the context handlers in the calling function. What behavior are you seeing?

puneith: It prints "Exiting" but hangs there.

jerjou: Ah - probably one of the threads isn't exiting correctly -_-;
Just to make sure - you're sync'ed to the latest changes, right?

jerjou: Updated to explicitly cancel the request stream. See if that helps?

puneith: Ok that works!

PR #529, opened 2016-09-15
Verify issuer claim

tristangoogle:


jonparrott:

jonparrott:
@tristangoogle there seems to be a minor lint issue, can you fix?

PR #534, opened 2016-09-19
Make nox use Python 3.5 instead of 3.4

jonparrott:


jonparrott:

PR #535, opened 2016-09-19
Taskqueue bug appyaml

ryanmats:
Fixed bug named 'Incorrect filename in taskqueue example.' Just renamed application.yaml to app.yaml.

At some point I will have to figure out how to remove those old commits so I don't have to keep deleting the unnecessary files from the older schema_update commit.


jonparrott:
@ryanmats can you also update the README.md in the sample directory? It mentions the old file name.

PR #539, opened 2016-09-20
Remove vestigial discovery url.

jerjou:


jerjou:
@dpebot tswast has granted his lgtm - would you be so kind as to oblige us, once the requisite checks have completed?

jerjou:
@dpebot? Why you lie to me?

PR #549, opened 2016-09-29
Update shortlink with even-shorter link.

jerjou:


jerjou:
Oh wait. There are more changes here than intended... gimme a sec.

jerjou:
Okay, all fixed.

PR #560, opened 2016-10-06
Add firebase-based tic-tac-toe sample.

jerjou:
Mostly a refactor of stuff from @mogar1980 #558

### DIFF
--- a/appengine/standard/firebase/firetactoe/firetactoe.py
+++ b/appengine/standard/firebase/firetactoe/firetactoe.py
@@ -1,0 +1,247 @@
+# Copyright 2016 Google Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Tic Tac Toe with the Firebase API"""
+
+import datetime
+import json
+import os
+import re
+import urllib
+
+from Crypto.PublicKey import RSA
+import flask
+from flask import request
+from google.appengine.api import users
+from google.appengine.ext import ndb
+import httplib2
+import jwt
+from oauth2client.service_account import ServiceAccountCredentials
+
+
+_FIREBASE_CONFIG = '_firebase_config.html'
+_SERVICE_ACCOUNT_FILENAME = 'credentials.json'
+
+_CWD = os.path.dirname(__file__)
+_IDENTITY_ENDPOINT = ('https://identitytoolkit.googleapis.com/'
+                      'google.identity.identitytoolkit.v1.IdentityToolkit')
+_FIREBASE_SCOPES = [
+    'https://www.googleapis.com/auth/firebase.database',
+    'https://www.googleapis.com/auth/userinfo.email']
+
+_X_WIN_PATTERNS = [
+    'XXX......', '...XXX...', '......XXX', 'X..X..X..', '.X..X..X.',
+    '..X..X..X', 'X...X...X', '..X.X.X..']
+_O_WIN_PATTERNS = map(lambda s: s.replace('X', 'O'), _X_WIN_PATTERNS)
+
+X_WINS = map(lambda s: re.compile(s), _X_WIN_PATTERNS)
+O_WINS = map(lambda s: re.compile(s), _O_WIN_PATTERNS)
+
+
+app = flask.Flask(__name__)
+
+
+def _get_firebase_db_url(_memo={}):
+    """Grabs the databaseURL from the Firebase config snippet."""
+    # Memoize the value, to avoid parsing the code snippet every time
+    if 'dburl' not in _memo:
+        regex = re.compile(r'\bdatabaseURL\b.*?["\']([^"\']+)')
+        cwd = os.path.dirname(__file__)
+        with open(os.path.join(cwd, 'templates', _FIREBASE_CONFIG)) as f:
+            url = next(regex.search(line) for line in f if regex.search(line))
+        _memo['dburl'] = url.group(1)
+    return _memo['dburl']
+
+
+def _get_http(_memo={}):
+    """Provides an authed http object."""
+    if 'http' not in _memo:
+        # Memoize the authorized http, to avoid fetching new access tokens
+        http = httplib2.Http()
+        # Use service account credentials to make the Firebase calls
+        # https://firebase.google.com/docs/reference/rest/database/user-auth
+        creds = ServiceAccountCredentials.from_json_keyfile_name(
+            os.path.join(_CWD, _SERVICE_ACCOUNT_FILENAME), _FIREBASE_SCOPES)
+        creds.authorize(http)
+        _memo['http'] = http
+    return _memo['http']
+
+
+def _send_firebase_message(u_id, message=None):
+    url = '{}/channels/{}.json'.format(_get_firebase_db_url(), u_id)
+
+    if message:
+        return _get_http().request(url, 'PATCH', body=message)
+    else:
+        return _get_http().request(url, 'DELETE')
+
+
+def create_custom_token(uid):
+    """Create a secure token for the given id.
+
+    This method is used to create secure custom tokens to be passed to clients
+    it takes a unique id (uid) that will be used by Firebase's security rules
+    to prevent unauthorized access. In this case, the uid will be the channel
+    id which is a combination of user_id and game_key
+    """
+    with open(os.path.join(_CWD, _SERVICE_ACCOUNT_FILENAME), 'r') as f:
+        credentials = json.load(f)
+
+    payload = {
+        'iss': credentials['client_email'],
+        'sub': credentials['client_email'],
+        'aud': _IDENTITY_ENDPOINT,
+        'uid': uid,
+    }
+    exp = datetime.timedelta(minutes=60)
+    return jwt.generate_jwt(
+        payload, RSA.importKey(credentials['private_key']), 'RS256', exp)
+
+
+class Game(ndb.Model):
+    """All the data we store for a game"""
+    userX = ndb.UserProperty()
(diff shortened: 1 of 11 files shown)


jerjou:
FYI I still intend to write tests for this, but I figured I wanted some eyes on it first to make sure there're no objections.

appengine/standard/firebase/firetactoe/firetactoe.py
```diff
@@ -0,0 +1,270 @@
...
+
+
+# Firebase database server authorization key
+# This is used to authenticate the server and authorize write access
+#
+# Replace this value with your Firebase db key from:
+#
+#     https://console.firebase.google.com/project/_/settings/database
+#
+FIRE_DB_KEY = 'REPLACE_WITH_YOUR_DB_KEY'
```
jonparrott: use `env_variables` in app.yaml?

appengine/standard/firebase/firetactoe/firetactoe.py
```diff
@@ -0,0 +1,270 @@
...
+_IDENTITY_ENDPOINT = ('https://identitytoolkit.googleapis.com/'
+                      'google.identity.identitytoolkit.v1.IdentityToolkit')
+# Firebase database url/channels memoization
+_FIRE_URL = None
+
+
+app = flask.Flask(__name__)
+
+
+def _get_firebase_config():
```
jonparrott: This is pretty gnarly, is there another way to get the info you need?

appengine/standard/firebase/firetactoe/firetactoe.py
```diff
@@ -0,0 +1,270 @@
...
+    """Parses the client-side Firebase config into a dict."""
+    with open(os.path.join(_CWD, 'templates', _FIREBASE_CONFIG_TEMPLATE)) as f:
+        html = f.read()
+    obj = re.search(r'{([^}]+)}', html, re.MULTILINE)
+    params = [p.strip().split(':', 1) for p in obj.group(1).split(',')]
+    return dict((k.strip('" '), v.strip('" ')) for k, v in params)
+
+
+def _send_firebase_message(u_id, message=None):
+    global _FIRE_URL
```
jonparrott: Don't modify globals if you can help it, and definitely don't modify global constants. Can this just be pulled into a function?

appengine/standard/firebase/firetactoe/firetactoe.py
```diff
@@ -0,0 +1,270 @@
...
+
+    payload = {
+        'iss': credentials['client_email'],
+        'sub': credentials['client_email'],
+        'aud': _IDENTITY_ENDPOINT,
+        'uid': uid,
+    }
+    exp = datetime.timedelta(minutes=60)
+    return jwt.generate_jwt(
+        payload, RSA.importKey(credentials['private_key']), 'RS256', exp)
```
jonparrott: So it seems the only reason we require them to give the private key is so that pyjwt can sign the jwt. However, App engine can actually sign blob directly using app_identity. We'd have to forgo the jwt library, but do you think that's doable?

appengine/standard/firebase/firetactoe/firetactoe.py
```diff
@@ -0,0 +1,270 @@
...
+
+
+class Wins():
```
jonparrott: Does this need to be a class, or could it just be some constants?

appengine/standard/firebase/firetactoe/firetactoe.py
```diff
@@ -0,0 +1,270 @@
...
+                      '..X..X..X',
+                      'X...X...X',
+                      '..X.X.X..']
+    o_win_patterns = map(lambda s: s.replace('X', 'O'), x_win_patterns)
+
+    x_wins = map(lambda s: re.compile(s), x_win_patterns)
+    o_wins = map(lambda s: re.compile(s), o_win_patterns)
+
+
+class Game(db.Model):
```
jonparrott: ndb instead of db?

appengine/standard/firebase/firetactoe/firetactoe.py
```diff
@@ -0,0 +1,270 @@
...
+                    boardList[position] = 'X' if self.moveX else 'O'
+                    self.board = ''.join(boardList)
+                    self.moveX = not self.moveX
+                    self._check_win()
+                    self.put()
+                    self.send_update()
+                    return
+
+
+def login_required(f):
```
jonparrott: s/f/view_func

appengine/standard/firebase/firetactoe/static/main.js
```diff
@@ -0,0 +1,219 @@
...
+  // This is our Firebase realtime DB path that we'll listen to for updates
+  // We'll initialize this later in openChannel()
+  var channel = null;
+
+  /**
+   * Updates the displayed game board.
+   */
+  function updateGame() {
+    for (var i = 0; i < 9; i++) {
+      var square = document.getElementById(i);
```
jonparrott: jquery could make a lot of this dom manipulation much easier to read.

appengine/standard/firebase/firetactoe/templates/fire_index.html
```diff
@@ -0,0 +1,48 @@
+<html>
+  <head>
+    {% include "_firebase_config.html" %}
+    <link rel="stylesheet" type="text/css" href="static/main.css" />
+    <script src="static/main.js"></script>
+    <script>
+      initGame('{{ game_key }}', '{{ me }}', '{{ token }}', '{{ channel_id }}',
```
jonparrott: I would prefer you to store these as [`data-`](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Using_data_attributes) attributes on some well-known element (`body` would be fine).

appengine/standard/firebase/firetactoe/static/main.js
jerjou: (not to say there might not be wins to be gained with a larger refactor... (/me eyes the manual style setting below) but it's non-obvious and the clock is ticking :-/)

jerjou:
Updated to eliminate need for `@login_required` decorator, as well as separate Firebase api key step

jerjou:
Removed dependency on service accounts, in favor of ADC all the way.

appengine/standard/firebase/firetactoe/firetactoe.py
jonparrott: That's fine, but can you pull all of this caching stuff into it's own function? `_get_fire_url`?

appengine/standard/firebase/firetactoe/firetactoe.py
jonparrott: Not a strong enough argument for class. :)

appengine/standard/firebase/firetactoe/static/main.js
jonparrott: It won't decrease the line count, but it'll make things easier. (Also, jquery is hardly heavyweight).

appengine/standard/firebase/firetactoe/templates/fire_index.html
jonparrott: You can install invoke the function:
```
var $config = $('something),
    game_key = $config.data('game-key'),
    ...;

initGame(game_key, ...);
```

It just decouples the JS from the template engine - I really, really abhor mixing server-side template interpolation in javascript.

appengine/standard/firebase/firetactoe/firetactoe.py
```diff
@@ -0,0 +1,254 @@
...
+def _get_firebase_config(key=re.compile(r'\bdatabaseURL\b.*?["\']([^"\']+)')):
+    """Grabs a key from the Firebase config, defaulting to databaseURL."""
+    cwd = os.path.dirname(__file__)
+    with open(os.path.join(cwd, 'templates', _FIREBASE_CONFIG_TEMPLATE)) as f:
+        match = next(key.search(line) for line in f if key.search(line))
+    return match.group(1)
+
+
+def _send_firebase_message(u_id, message=None, _memo={}):
+    if 'http' not in _memo:
```
jonparrott: Pull memoization into it's own function, `_get_http`.

appengine/standard/firebase/firetactoe/firetactoe.py
jonparrott: Just noticed that you've done this. 👍

appengine/standard/firebase/firetactoe/firetactoe.py
jerjou:
```html
<script src="https://www.gstatic.com/firebasejs/3.4.1/firebase.js"></script>
<script>
  // Initialize Firebase
  var config = {
    apiKey: "ABACADABA",
    authDomain: "project-id.firebaseapp.com",
    databaseURL: "https://project-id.firebaseio.com",
    storageBucket: "project-id.appspot.com",
    messagingSenderId: "1234567889"
  };
  firebase.initializeApp(config);
</script>
```

appengine/standard/firebase/firetactoe/firetactoe.py
jerjou: I ran into a problem with this which I fixed, but then realized that using the gcloud way of providing ADC doesn't provide app_identity with everything it needs; so modified the README to use a service account instead.

jonparrott: SG.

appengine/standard/firebase/firetactoe/firetactoe.py
jonparrott: Just

```
X_WIN_PATTERNS = ...
```

It's fine to have these as constants because, well, they are constants. There's no argument for having them in a class. Classes are for collective state or behavior. You could maybe argue for having them in a separate module.

appengine/standard/firebase/firetactoe/static/main.js
jerjou: I disagree, but whatever. Refactored.

appengine/standard/firebase/firetactoe/templates/fire_index.html
jerjou: Why?
I like it - it puts the information precisely where it's needed, and the templating system makes sure it's safe.

jerjou:
Updated. PTAL

appengine/standard/firebase/firetactoe/firetactoe.py
jerjou: I'm unconvinced, but want to get this in sooner rather than later.
Done.

PR #570, opened 2016-10-11
Merges the contents of index.html into the README for the Endpoints sample.

wmwong:
Removes index.html and adds two new sections in the README for viewing
the Endpoints graphs and Swagger UI.


jonparrott:
Should probably remove the handlers from main.py as well?

PR #587, opened 2016-10-17
fixed template bugs, added regions for docs writers, and moved creden…

mogar1980:
…tials into json to make code more obvious to readers


appengine/standard/firebase/firetactoe/firetactoe.py
```diff
@@ -75,17 +77,24 @@ def _get_http(_memo={}):
         creds.authorize(http)
         _memo['http'] = http
     return _memo['http']
+# [END authed_http]
 
-
+# [START send_msg]
```
jerjou: For region tags that are just for including an existing method, you can just use the [`indented_block`](https://developers.google.com/internal/publishing/reference/templatetags/includecode#indented_block) attribute to include the entire method.

appengine/standard/firebase/firetactoe/firetactoe.py
```diff
@@ -53,20 +56,19 @@
 
 def _get_firebase_db_url(_memo={}):
     """Grabs the databaseURL from the Firebase config snippet."""
-    # Memoize the value, to avoid parsing the code snippet every time
+    # Memorize the value, to avoid parsing the code snippet every time
```
jerjou: [memoize](https://en.wikipedia.org/wiki/Memoization) is correct.

appengine/standard/firebase/firetactoe/README.md
```diff
@@ -15,7 +15,7 @@ installed. You'll need this to test and deploy your App Engine app.
   console](https://firebase.google.com/console)
 * In the Overview section, click 'Add Firebase to your web app' and replace the
   contents of the file
-  [`templates/_firebase_config.html`](templates/_firebase_config.html) with the
+  [`templates/_firebase_config.json`](templates/_firebase_config.json) with the
```
jerjou: I kept this as html for simplicity reasons. This way, they can just copy/paste the code snippet from the firebase console into the template. If you change it to json, then they have to individually select each field, find the corresponding field in the json template, and paste each one. This leaves more room for error, and requires them to ignore the instructions and convenient "Copy" button in the firebase console.

I agree in principle that reading from json is clearer for someone reading the code, but I think there's a trade-off here between ease of setup and code readability. In this case, I figured the friction of figuring out what the regex line in the `get_firebase_db_url` did could be mitigated with a clear function name, docstring, and comments. (and was less than the alternative)

appengine/standard/firebase/firetactoe/firetactoe.py
```diff
@@ -94,25 +103,30 @@ def create_custom_token(uid, valid_minutes=60):
     security rules to prevent unauthorized access. In this case, the uid will
     be the channel id which is a combination of user_id and game_key
     """
-    header = base64.b64encode(json.dumps({'typ': 'JWT', 'alg': 'RS256'}))
```
jerjou: I kept the header above and separate from the payload intentionally, because it is unrelated and doesn't need to be thought of in conjunction with the payload. Moving it down below `client_email` interleaves the payload with the unrelated header, giving the impression that the header will be included in the payload, which it isn't.

appengine/standard/firebase/firetactoe/rest_api.py
```diff
@@ -0,0 +1,108 @@
...
+
+_FIREBASE_SCOPES = [
+    'https://www.googleapis.com/auth/firebase.database',
+    'https://www.googleapis.com/auth/userinfo.email']
+
+
+def _get_http(_memo={}):
+    """Provides an authed http object."""
+    if 'http' not in _memo:
+        # Memorize the authorized http, to avoid fetching new access tokens
```
jerjou: [memoize](https://en.wikipedia.org/wiki/Memoization)

appengine/standard/firebase/firetactoe/rest_api.py
```diff
@@ -0,0 +1,108 @@
...
+        http = httplib2.Http()
+        # Use application default credentials to make the Firebase calls
+        # https://firebase.google.com/docs/reference/rest/database/user-auth
+        creds = GoogleCredentials.get_application_default().create_scoped(
+            _FIREBASE_SCOPES)
+        creds.authorize(http)
+        _memo['http'] = http
+    return _memo['http']
+    
+def fb_put(path, value=None):
```
jerjou: Prefer avoiding abbreviations, for clarity.
Also - I always read `fb` as `facebook`

appengine/standard/firebase/firetactoe/rest_api.py
```diff
@@ -0,0 +1,108 @@
...
+    """Writes data to Firebase. Value should be a valid json object.
+     Put writes an entire object at the given database path. Updates to 
+     fields cannot be performed without overwriting the entire object
+     """
+    response, content = _get_http().request(path, method='PUT', body=value)
+    if content != "null":
```
jerjou: Is "null" ever returned? Isn't that invalid json?

appengine/standard/firebase/firetactoe/rest_api.py
```diff
@@ -0,0 +1,108 @@
...
+        return json.loads(content)
+    else:
+        return None
+
+def fb_patch(path, value=None):
```
jerjou: This module doesn't seem to provide much value - it adds a layer of indirection to what's otherwise a straightforward REST call. Is there a particular reason you want these to be function calls rather than just making the REST call directly? Sample code is meant to educate, so we should err on the side of transparency when practical.

appengine/standard/firebase/firetactoe/static/main.js
```diff
@@ -29,7 +29,9 @@
 function initGame(gameKey, me, token, channelId, initialMessage) {
   var state = {
     gameKey: gameKey,
-    me: me
+    me: me,
+    channelId: channelId,
+    initialMessage: initialMessage
```
jerjou: Is there a reason to save the initialMessage? Its contents are going to be incorporated into the state almost immediately anyway.

appengine/standard/firebase/firetactoe/static/main.js
```diff
@@ -29,7 +29,9 @@
 function initGame(gameKey, me, token, channelId, initialMessage) {
   var state = {
     gameKey: gameKey,
-    me: me
+    me: me,
+    channelId: channelId,
```
jerjou: Any particular reason to save the channelId into the state? It doesn't seem to be used..

appengine/standard/firebase/firetactoe/firetactoe.py
```diff
@@ -31,7 +31,10 @@
 from oauth2client.client import GoogleCredentials
 
 
-_FIREBASE_CONFIG = '_firebase_config.html'
+import rest_api
```
jerjou: If you're breaking this into a separate module, then a bunch of the code below needs to be updated to use it.

appengine/standard/firebase/firetactoe/templates/_firebase_config.json
```diff
@@ -0,0 +1,7 @@
+{
```
jerjou: If you change this file into json, then you also need to change the way that the firebase config is included and initialized in the `fire_index.html` template.

appengine/standard/firebase/firetactoe/firetactoe.py
```diff
@@ -94,25 +103,30 @@ def create_custom_token(uid, valid_minutes=60):
...
+    header = base64.b64encode(json.dumps({'typ': 'JWT', 'alg': 'RS256'}))
     now = int(time.time())
+    # encode the required claims 
+    # per https://firebase.google.com/docs/auth/server/create-custom-tokens
     payload = base64.b64encode(json.dumps({
         'iss': client_email,
         'sub': client_email,
         'aud': _IDENTITY_ENDPOINT,
-        'uid': uid,
+        'uid': uid, # this is the important parameter as it will be the channel id
```
jerjou: Thanks for adding all these comments ^_^

appengine/standard/firebase/firetactoe/firetactoe.py
mogar1980: Feel free to just kill this line. Was only using for testing. Don't actually want to break into a separate module. I created the file solely to allow the code for the REST API to be surfaced in the article.

appengine/standard/firebase/firetactoe/firetactoe.py
mogar1980: Leave it then. Never used the term myself and it's got an edit distance of 1 from a statement that would also be true :)

appengine/standard/firebase/firetactoe/firetactoe.py
mogar1980: Let's let Jim decide. As you noted in email, we probably can't get rid of them all.

appengine/standard/firebase/firetactoe/firetactoe.py
mogar1980: I'd prefer to leave this where it is. Using the app identity service is a google-specific feature that I wanted the user to grok before getting lost in the guts of generating a JWT.

appengine/standard/firebase/firetactoe/firetactoe.py
mogar1980: sure. np

appengine/standard/firebase/firetactoe/README.md
mogar1980: I took one look at the regex and my head nearly exploded :). Would prefer json, but just noticed that you inject the snippet into the html, which is simpler than having to copy values twice. And I broke the sample when I removed the snippet file for the pull request. Let's go with what you had. Perhaps you could just add a really good comment about what you're doing. The regex really does make it look unnecessarily complex.

appengine/standard/firebase/firetactoe/rest_api.py
mogar1980: sure

appengine/standard/firebase/firetactoe/rest_api.py
mogar1980: sure. works for me

appengine/standard/firebase/firetactoe/rest_api.py
mogar1980: The module is really only here so that we can include the REST API in the document. I don't disagree that it's not very useful. In fact, it shouldn't appear anywhere else in the code. It was only imported into firetactoe.py by mistake. An artifact of testing.

appengine/standard/firebase/firetactoe/templates/_firebase_config.json
mogar1980: You're absolutely right. I missed this in dev b/c I had both files in there. I'm tempted to just use your technique given simpler setup.

appengine/standard/firebase/firetactoe/firetactoe.py
```diff
@@ -31,6 +31,7 @@
 from oauth2client.client import GoogleCredentials
 
 
+
```
jerjou: Can you remove this space? It won't pass our linter.
You can check by running:

    nox -s lint -- appengine/standard/firebase/firetactoe

from the root of the github repo. ([details](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/CONTRIBUTING.md#style))

appengine/standard/firebase/firetactoe/templates/_firebase_config.html
```diff
@@ -1,3 +1,3 @@
 REPLACE ME WITH YOUR FIREBASE WEBAPP CODE SNIPPET:
 
-https://console.firebase.google.com/project/_/overview
+https://console.firebase.google.com/project/_/overview
```
jerjou: could you just revert this file?

    git checkout a76c3dc7437b53bbc4b4c5a4e3e4dd9c056e7322 appengine/standard/firebase/firetactoe/templates/_firebase_config.html

mogar1980:
Sorry. Looks like I might not have generated a pull request for the final
changes I made yesterday. Let me know if you see them now.

Thanks!
Morgan

PR #594, opened 2016-10-19
Add test for firetactoe.

jerjou:


appengine/standard/firebase/firetactoe/firetactoe_test.py
```diff
@@ -0,0 +1,162 @@
...
+
+class MockHttp(object):
+    """Mock the Http object, so we can set what the response will be."""
+    def __init__(self, status, content=''):
+        self.content = content
+        self.status = status
+        self.request_url = None
+
+    def __call__(self, *args, **kwargs):
+        self.constructor_args = (args, kwargs)
```
jonparrott: This doesn't seem to be used, remove it?

jerjou:
@dpebot would you mind merging when travis approves?

jerjou:
FYI added a commit to fix the style errors that are breaking the build.

jerjou:
FYI removed a file that also had style errors (and a dubious null check). Also, a rewrite of the doc that uses it may make the file obsolete (so not worth trying to fix it). If it turns out to still be needed, we can open another PR to re-add it.

PR #595, opened 2016-10-19
Manually add required padding before base64.b64decode.

lookuptable:
Currently the value of the "X-Endpoint-API-UserInfo" header does not
contain paddings. This causes incorrect padding error when doing
base64.b64decode(). This commit fixes the behavior by manually adding
paddings when necessary.

### DIFF
--- a/appengine/flexible/endpoints/main.py
+++ b/appengine/flexible/endpoints/main.py
@@ -42,6 +42,11 @@ def auth_info():
     encoded_info = request.headers.get('X-Endpoint-API-UserInfo', None)
 
     if encoded_info:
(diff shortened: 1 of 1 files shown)


appengine/flexible/endpoints/main.py
```diff
@@ -42,6 +42,11 @@ def auth_info():
     encoded_info = request.headers.get('X-Endpoint-API-UserInfo', None)
 
     if encoded_info:
+        # Add paddings manually if necessary.
```
lookuptable: Done

jonparrott:
@dpebot merge when travis is green

PR #598, opened 2016-10-20
Use decorator to memoize.

jerjou:
This is to avoid adding complexity unrelated to what's being explained, when the
snippet of code is included in a doc.


jerjou:
@dpebot merge when green, if you please.

PR #600, opened 2016-10-20
Add support for third-party memcache

jonparrott:
R: @justinbeckwith

/cc @jmdobry @bshaffer @broady @tmatsuo @lesv @remi

### DIFF
--- a/appengine/flexible/memcache/README.md
+++ b/appengine/flexible/memcache/README.md
@@ -20,6 +20,6 @@ Start your application:
 
 Deploy using `gcloud`:
 
-    gcloud app deploy app.yaml
+    gcloud app deploy
 
 You can now access the application at `https://your-app-id.appspot.com`.
--- a/appengine/flexible/memcache/README.md
+++ b/appengine/flexible/memcache/README.md
@@ -20,6 +20,6 @@ Start your application:
 
 Deploy using `gcloud`:
 
-    gcloud beta app deploy app.yaml
+    gcloud app deploy app.yaml
 
 You can now access the application at `https://your-app-id.appspot.com`.
--- a/appengine/flexible/memcache/main.py
+++ b/appengine/flexible/memcache/main.py
@@ -16,16 +16,26 @@ import logging
 import os
 
 from flask import Flask
-from pymemcache.client.base import Client as MemcacheClient
-
+import pylibmc
 
 app = Flask(__name__)
 
 
 # [START client]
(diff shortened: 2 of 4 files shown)


appengine/flexible/memcache/README.md
```diff
@@ -20,6 +20,6 @@ Start your application:
 
 Deploy using `gcloud`:
 
-    gcloud beta app deploy app.yaml
+    gcloud app deploy app.yaml
```
JustinBeckwith: Just `gcloud app deploy` here is fine

appengine/flexible/memcache/main.py
```diff
@@ -16,16 +16,26 @@
...
 
 app = Flask(__name__)
 
 
 # [START client]
-memcache_addr = os.environ.get('GAE_MEMCACHE_HOST', 'localhost')
-memcache_port = os.environ.get('GAE_MEMCACHE_PORT', 11211)
-memcache_client = MemcacheClient((memcache_addr, int(memcache_port)))
+# Environment variables are defined in app.yaml.
+if os.environ.get('USE_GAE_MEMCACHE'):
```
broady: USE_APPENGINE_MEMCACHE

appengine/flexible/memcache/main.py
```diff
@@ -16,16 +16,26 @@
...
+    MEMCACHE_SERVER = ':'.join(
+        os.environ.get('GAE_MEMCACHE_HOST', 'localhost'),
+        os.environ.get('GAE_MEMCACHE_PORT', 11211))
```
broady: why not just check these env vars instead of having a separate env var as a flag?

appengine/flexible/memcache/main.py
jonparrott: Why prefer that when the GAE env vars all use the "GAE"  initialism?

appengine/flexible/memcache/main.py
jonparrott: Because eventually these will always be set, and I don't want code that's using third-party memcache to suddenly start using GAE memcache.

appengine/flexible/memcache/main.py
broady: they're inconsistent. change this one or the one in app.yaml

jonparrott: Oh, gotcha. Sorry! Fixed.

PR #610, opened 2016-10-24
Double quotes for HTML attributes

broady:


jerjou:
Oh gross. How did I do this.

PR #614, opened 2016-10-24
Use unittest.TestCase assertions

docent-net:
While inside a class that inherits from unittest.TestCase it makes sense to use provided by TestCase assertion methods instead of pure Python's assertion()


jonparrott:
@docent-net thanks for the change, but we'll need you to sign the CLA before we can review and merge. Let me know if you have any questions or concerns.

docent-net:
@jonparrott ah again - I already signed CLA but pushed changes using non-gmail email address set in repository config. I just amended it and should be fine

jonparrott:
LGTM, thank you!

PR #618, opened 2016-10-26
xmpp wikibot for GCE

jerjou:
@mogar1980 fixed some style issues, and cleaned some stuff up. PTAL


compute/xmpp_wikibot/wikibot.py
```diff
@@ -0,0 +1,165 @@
...
+    if chat_client and recipient and message:
+        chat_client.send_message(mto=recipient, mbody=message)
+        return 'message sent to {} with body: {}'.format(recipient, message)
+    else:
+        logging.info('chat client or recipient or message does not exist!')
+        return 'message failed to send', 400
+
+
+def run_server(host='0.0.0.0'):
+    app.run(threaded=False, use_reloader=False, host=host)
```
jonparrott: debug=False here, please. We shouldn't be using Flask's dev server to serve public traffic *at all*, but at least disable the debug flag.

jerjou:
Whoops. My push errored out without my noticing, so I merged before adding `debug=False`. Will fix in subsequent CL.

compute/xmpp_wikibot/wikibot.py
mogar1980: Yeah. I know <shakes head>. The setup gets increasingly more complex if I try to do the right thing with a real web server. Jerjou and I discussed possibly removing the server component altogether as the chat interactions are more interesting anyway. Will refine tomorrow. Thanks for all the help!

PR #620, opened 2016-10-27
Simplify xmpp wikibot a bit

jerjou:
@mogar1980

### DIFF
--- a/compute/xmpp_wikibot/wikibot.py
+++ b/compute/xmpp_wikibot/wikibot.py
@@ -17,16 +17,17 @@
 import argparse
 import getpass
 import json
-import logging
-import threading
+import re
+import sys
 import urllib
 
 from flask import Flask, request
 import requests
+from six.moves import html_parser
 import sleekxmpp
 
-
 app = Flask(__name__)
+
 
 @app.route('/send_message', methods=['GET'])
 def send_message():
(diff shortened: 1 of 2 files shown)


compute/xmpp_wikibot/wikibot.py
```diff
@@ -94,22 +97,27 @@ def message(self, msg):
         """
         if msg['type'] in ('chat', 'normal'):
             msg_body = msg['body']
-            logging.info('Message sent was: {}'.format(msg_body))
             encoded_body = urllib.quote_plus(msg_body)
             svrResponse = requests.get(
```
jonparrott: Go ahead and make this `snake_case` while you're here?

compute/xmpp_wikibot/wikibot.py
```diff
@@ -94,22 +97,27 @@ def message(self, msg):
...
-            except KeyError as e:
-                logging.info('key error: {0}'.format(e))
+
+            results = doc.get('query', {}).get('search')
+            if not results:
                 msg.reply('I wasn\'t able to locate info on "{}" Sorry'.format(
                     msg_body)).send()
+                return
+
+            snippet = unicode(results[0]['snippet'])
```
jonparrott: `unicode` doesn't exist in python 3, try `six.text_type`.

compute/xmpp_wikibot/wikibot.py
```diff
@@ -118,48 +126,28 @@ def message(self, msg):
...
+        if not xmpp.connect():
+            print('Unable to connect.')
+            sys.exit(1)
+
+        xmpp.process(block=False)
 
-    # Connect to the XMPP server and start processing XMPP stanzas.
-    if xmpp.connect():
-        xmpp.process(block=True)
+        app.run(threaded=True, use_reloader=False, host='0.0.0.0')
```
jonparrott: debug=False?

compute/xmpp_wikibot/wikibot.py
jerjou: Hrm... I don't even remember why I was casting it to unicode anyway (especially since the encoding in wikipedia is fickle, I think). I'm just going to remove the unicode cast.

PR #621, opened 2016-10-27
Revert & fix "Revert "Lint session should check all changed files. (#…

jerjou:
…619)""

This reverts commit ae8dccbab095f09caddb96d6d304c41cdebad232.


jerjou:
(tested manually by running `nox -s lint .` in a virtualenv)

jerjou:
(actually - hold off - I realize my test was a little suspicious.. investigating..)

jonparrott:
I don't think you should run `nox -s lint .`, nox is supposed to run a separate lint for each sample because of local import names.

jerjou:
Okay - fixed. Also fixed a bunch of lint errors. PTAL.

PR #631, opened 2016-11-02
Copy Endpoints Python Echo sample to a new location.

kdeus:
This will be its new home, once we clean up the docs.

This sample is used for multiple platforms, and hiding it under appengine/flex is confusing.


endpoints/echo/Dockerfile.container-engine
```diff
@@ -0,0 +1,12 @@
+FROM debian:jessie
```
jonparrott: Why not just use `gcr.io/google_appengine/python` - it's designed to work on GKE as well.

kdeus: Turns out we don't actually need this Dockerfile.  Removed, in a following commit.

jonparrott:
Why remove the dockerfile? Isn't that needed?

kdeus:
Turns out this isn't needed by any of our current samples.  It'd only be
needed if we were showing users how to build this example code into a new
docker container.  Currently, all our samples use pre-built docker
containers.

I'm double checking that with our team, to make sure no upcoming changes
require it.  As is, though, it's unused.

jonparrott:
@kdeus would you not want to tell users how to create their own docker image?

kdeus:
Yes, we'll want to do that.  But it sounds like we're going to use a
different set of sample code to do that, not this one.

jonparrott:
It seems weird to have that separate from the source code that'll live in the container, no?

kdeus:
We won't be using this source code to build a docker image.  We'll use a
completely different example.

endpoints/echo/clients/service_to_service_gae_default/main.py
```diff
@@ -0,0 +1,83 @@
+# Copyright 2016 Google Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0(the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http:  // www.apache.org/licenses/LICENSE-2.0
```
tangiel: Dunno what happened here but this URL is borked.

endpoints/echo/clients/service_to_service_google_id_token/main.py
```diff
@@ -0,0 +1,98 @@
+# Copyright 2016 Google Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0(the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http:  // www.apache.org/licenses/LICENSE-2.0
```
tangiel: Here too.

endpoints/echo/clients/service_to_service_non_default/main.py
```diff
@@ -0,0 +1,94 @@
+# Copyright 2016 Google Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0(the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http:  // www.apache.org/licenses/LICENSE-2.0
```
tangiel: And here.

jonparrott:
Then what's the point of having any GKE stuff here?

endpoints/echo/clients/service_to_service_gae_default/main.py
kdeus: Fixed

endpoints/echo/clients/service_to_service_google_id_token/main.py
kdeus: Done

endpoints/echo/clients/service_to_service_non_default/main.py
kdeus: Done

jonparrott:
My point is that why have that stuff here if it does not pertain to this
specific sample? What's the relevance to this sample if the GKE config
doesn't use this sample code and there's no obvious way to build the GKE
sample's image using this code?

On Wed, Nov 2, 2016, 4:59 PM kdeus <notifications@github.com> wrote:

*@kdeus* commented on this pull request.

The GKE stuff is used by examples showing how Google Cloud Endpoints works
for different environments (GKE, GCE + Docker). It's similar to why
app.yaml is there: to show how this works for GAE.
------------------------------

In endpoints/echo/clients/service_to_service_gae_default/main.py
<https://github.com/GoogleCloudPlatform/python-docs-samples/pull/631#pullrequestreview-6933885>
:

> @@ -0,0 +1,83 @@
+# Copyright 2016 Google Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0(the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http:  // www.apache.org/licenses/LICENSE-2.0

Fixed
------------------------------

In endpoints/echo/clients/service_to_service_google_id_token/main.py
<https://github.com/GoogleCloudPlatform/python-docs-samples/pull/631#pullrequestreview-6933885>
:

> @@ -0,0 +1,98 @@
+# Copyright 2016 Google Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0(the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http:  // www.apache.org/licenses/LICENSE-2.0

Done
------------------------------

In endpoints/echo/clients/service_to_service_non_default/main.py
<https://github.com/GoogleCloudPlatform/python-docs-samples/pull/631#pullrequestreview-6933885>
:

> @@ -0,0 +1,94 @@
+# Copyright 2016 Google Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0(the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http:  // www.apache.org/licenses/LICENSE-2.0

Done

—
You are receiving this because you commented.


Reply to this email directly, view it on GitHub
<https://github.com/GoogleCloudPlatform/python-docs-samples/pull/631#pullrequestreview-6933885>,
or mute the thread

kdeus:
We've been debating that.  :)

The intent here is that if someone comes in wanting to know how to use
Google Cloud Endpoints for a specific language, they can look at a
language-specific example and see the various pieces of how it works.  The
Quickstarts will reference different pieces of this example to get the user
up and running quickly, so they can see what happens when you send traffic
to an Endpoints backend, but they aren't intended to be in-depth tutorials
showing how this would work in your own APIs.  In some environments, the
server runs this sample code (GCE, GAE); in others, it uses the pre-built
Docker images (GKE, GCE + Docker).

There will be more in-depth tutorials that include steps such as building a
docker image, but those are still being written.  It's possible the
Dockerfile will come back, if we choose to base one of those tutorials on
this Echo example.  But for the moment, this is only referenced by the
Quickstarts, and none of the Quickstarts show how to go from source code to
a deployed Docker image.

This PR isn't really changing anything about the content of the sample.
It's just copying files from /appengine/flex/endpoints to
/endpoints/getting-started.  And removing the unused Dockerfile, copying
gke.yaml from the endpoints-sample repository, and doing a little clean-up.

jonparrott:
So my question is this: What's the point of *duplicating* the gke.yaml config from the endpoints-samples repository here if this source code is not used for that sample? In fact, *no* source code is used for that sample.

In my opinion adding the `gke.yaml` here made this previously standalone, app engine-specific sample confusing. The readme is very focused on deploying this to app engine.

Why can't we just leave the `gke.yaml` where it is in endpoint-samples? Why can't we augment the readme here and say "If you want to use GKE, check out the example config [here] and [here's] instructions on building a docker container.".

/cc @jeffmendoza

jeffmendoza:
Agree with Jon here.

We need the Dockerfile, even if the quickstart does not document how to build the code into a docker container.

Also, the gke and gce+docker quickstarts should use the docker image built from the code in this repo. Same for all the languages.

endpoints/getting-started/gke.yaml
```diff
@@ -0,0 +1,68 @@
...
+        ]
+        ports:
+          - containerPort: 8080
+          - containerPort: 443
+        volumeMounts:
+        - mountPath: /etc/nginx/ssl
+          name: nginx-ssl
+          readOnly: true
+      - name: echo
+        image: b.gcr.io/endpoints/echo:latest
```
jeffmendoza: Needs to point to an image of this code.

jeffmendoza: built and pushed `gcr.io/google-samples/echo-python:1.0`

kdeus:
The Dockerfile is back (as Dockerfile.custom).
Jeff will build a new language-specific image based on this code.  Jeff, let me know what gke.yaml should point to, for the language-specific image.

Also, I've partially stripped the Readme, removing references to GAE and anything that's duplicated in our Quickstarts.  We don't yet have a tutorial describing how to use the client libraries in this sample, so I'm leaving that information in until we create those tutorials.

kdeus:
Updated gke.yaml to point to the new image.

jonparrott:
Please use `container-engine.yaml`, abbreviations aren't as fun to
newcomers. :)

kdeus:
Renamed to container-engine.yaml.

On Thu, Nov 3, 2016 at 10:48 AM, Jon Wayne Parrott <notifications@github.com
> wrote:

> Please use `container-engine.yaml`, abbreviations aren't as fun to
> newcomers. :)
>

PR #635, opened 2016-11-04
streaming code used to hang when it was asked to "exit" - fixed now

puneith:
@jerjou @jonparrott PTAL

### DIFF
--- a/speech/grpc/transcribe_streaming.py
+++ b/speech/grpc/transcribe_streaming.py
@@ -86,7 +86,8 @@ def _audio_data_generator(buff):
             except queue.Empty:
                 break
 
-        # If the data contains None then set stop = True.
+        # If `_fill_buffer` adds `None` to the buffer, the audio stream is closed.
+        # Yield the final bit of the buffer and exit the loop.
         if None in data:
             stop = True
             data.remove(None)
@@ -101,6 +102,8 @@ def _fill_buffer(audio_stream, buff, chunk, stoprequest):
     except IOError:
         pass
     finally:
+        # Add `None` to the buff, indicating that a stop request is made.
+        # This will signal `_audio_data_generator` to exit.
         buff.put(None)
 
 
@@ -208,12 +211,14 @@ def listen_print_loop(recognize_stream, stoprequest):
 
 
 def main():
+    # For streaming audio from the microphone, there are three threads.
+    # First, a thread that collects audio data as it comes in
     with cloud_speech.beta_create_Speech_stub(
             make_channel('speech.googleapis.com', 443)) as service:
-        # For streaming audio from the microphone, there are three threads.
-        # First, a thread that collects audio data as it comes in
         
-        # stop request
+        # stoprequest is event object which is set in `listen_print_loop`.
+        # To indicate that the trancsription should be stopped.
+        # `_fill_buffer` checks and stops collecting data from audio_stream.
         stoprequest = threading.Event()
 
         with record_audio(RATE, CHUNK, stoprequest) as buffered_audio_data:
--- a/speech/grpc/transcribe_streaming.py
+++ b/speech/grpc/transcribe_streaming.py
@@ -75,7 +75,7 @@ def _audio_data_generator(buff):
     """
     stop = False
     while not stop:
-        # Use a blocking get() to ensure there's at least one chunk of data. 
+        # Use a blocking get() to ensure there's at least one chunk of data.
         chunk = buff.get()
         data = [chunk]
 
@@ -86,7 +86,7 @@ def _audio_data_generator(buff):
             except queue.Empty:
                 break
 
-        # If `_fill_buffer` adds `None` to the buffer, the audio stream is closed.
+        # If `_fill_buffer` adds `None` to the buffer, audio stream is closed.
         # Yield the final bit of the buffer and exit the loop.
         if None in data:
             stop = True
--- a/speech/grpc/transcribe_streaming.py
+++ b/speech/grpc/transcribe_streaming.py
@@ -73,7 +73,7 @@ def _audio_data_generator(buff, stoprequest):
         A chunk of data that is the aggregate of all chunks of data in `buff`.
         The function will block until at least one data chunk is available.
     """
-    while not stoprequest.isSet():
+    while not stoprequest.is_set():
         # Use a blocking get() to ensure there's at least one chunk of data    
         chunk = buff.get()
         if not chunk:
@@ -93,7 +93,7 @@ def _audio_data_generator(buff, stoprequest):
 def _fill_buffer(audio_stream, buff, chunk, stoprequest):
     """Continuously collect data from the audio stream, into the buffer."""
     try:
-        while not stoprequest.isSet():
+        while not stoprequest.is_set():
             buff.put(audio_stream.read(chunk))
     except IOError:
         # This happens when the stream is closed. Signal that we're done.
@@ -129,6 +129,7 @@ def record_audio(rate, chunk, stoprequest):
     yield _audio_data_generator(buff, stoprequest)
 
     fill_buffer_thread.join()
+    audio_interface.terminate()
 # [END audio_stream]
 
 
--- a/speech/grpc/transcribe_streaming.py
+++ b/speech/grpc/transcribe_streaming.py
@@ -64,7 +64,7 @@ def make_channel(host, port):
     return implementations.secure_channel(host, port, composite_channel)
 
 
-def _audio_data_generator(buff):
+def _audio_data_generator(buff, stoprequest):
     """A generator that yields all available data in the given buffer.
 
     Args:
@@ -73,8 +73,8 @@ def _audio_data_generator(buff):
         A chunk of data that is the aggregate of all chunks of data in `buff`.
         The function will block until at least one data chunk is available.
     """
-    while True:
-        # Use a blocking get() to ensure there's at least one chunk of data
+    while not stoprequest.isSet():
+        # Use a blocking get() to ensure there's at least one chunk of data    
         chunk = buff.get()
         if not chunk:
             # A falsey value indicates the stream is closed.
@@ -90,19 +90,22 @@ def _audio_data_generator(buff):
         yield b''.join(data)
 
 
-def _fill_buffer(audio_stream, buff, chunk):
+def _fill_buffer(audio_stream, buff, chunk, stoprequest):
     """Continuously collect data from the audio stream, into the buffer."""
     try:
-        while True:
+        while not stoprequest.isSet():
             buff.put(audio_stream.read(chunk))
     except IOError:
         # This happens when the stream is closed. Signal that we're done.
         buff.put(None)
 
+    audio_stream.stop_stream()
+    audio_stream.close()
+
 
 # [START audio_stream]
 @contextlib.contextmanager
-def record_audio(rate, chunk):
+def record_audio(rate, chunk, stoprequest):
     """Opens a recording stream in a context manager."""
     audio_interface = pyaudio.PyAudio()
     audio_stream = audio_interface.open(
@@ -120,15 +123,12 @@ def record_audio(rate, chunk):
     # This is necessary so that the input device's buffer doesn't overflow
     # while the calling thread makes network requests, etc.
     fill_buffer_thread = threading.Thread(
-        target=_fill_buffer, args=(audio_stream, buff, chunk))
+        target=_fill_buffer, args=(audio_stream, buff, chunk, stoprequest))
     fill_buffer_thread.start()
 
-    yield _audio_data_generator(buff)
+    yield _audio_data_generator(buff, stoprequest)
 
-    audio_stream.stop_stream()
-    audio_stream.close()
     fill_buffer_thread.join()
-    audio_interface.terminate()
 # [END audio_stream]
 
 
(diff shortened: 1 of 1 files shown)


speech/grpc/transcribe_streaming.py
```diff
@@ -90,19 +90,22 @@ def _audio_data_generator(buff):
...
-        while True:
+        while not stoprequest.isSet():
             buff.put(audio_stream.read(chunk))
     except IOError:
         # This happens when the stream is closed. Signal that we're done.
         buff.put(None)
 
+    audio_stream.stop_stream()
+    audio_stream.close()
+
```
jerjou: Was this move necessary? I'd prefer to have the function that opened the stream be responsible for closing it.

speech/grpc/transcribe_streaming.py
```diff
@@ -120,15 +123,12 @@ def record_audio(rate, chunk):
...
+        target=_fill_buffer, args=(audio_stream, buff, chunk, stoprequest))
     fill_buffer_thread.start()
 
-    yield _audio_data_generator(buff)
+    yield _audio_data_generator(buff, stoprequest)
 
-    audio_stream.stop_stream()
-    audio_stream.close()
     fill_buffer_thread.join()
-    audio_interface.terminate()
```
jerjou: Any particular reason to remove this line? It seems like good hygiene..

speech/grpc/transcribe_streaming.py
```diff
@@ -73,8 +73,8 @@ def _audio_data_generator(buff):
         A chunk of data that is the aggregate of all chunks of data in `buff`.
         The function will block until at least one data chunk is available.
     """
-    while True:
-        # Use a blocking get() to ensure there's at least one chunk of data
+    while not stoprequest.isSet():
```
jerjou: prefer `.is_set`, for python style conventions. Ditto below.

speech/grpc/transcribe_streaming.py
```diff
@@ -64,7 +64,7 @@ def make_channel(host, port):
     return implementations.secure_channel(host, port, composite_channel)
 
 
-def _audio_data_generator(buff):
+def _audio_data_generator(buff, stoprequest):
```
jerjou: `_audio_data_generator` shouldn't need this. Once `_fill_buffer` gets the signal, it'll add `None` to the buffer, which will cause this generator to exit.

puneith:
Yes I was able to reproduce this. It did hang on my mac.

speech/grpc/transcribe_streaming.py
puneith: done

speech/grpc/transcribe_streaming.py
puneith: done

speech/grpc/transcribe_streaming.py
puneith: But, fill_buffer doesn't add the None unless it receives IOError which it doesn't. Ok let me make one change to _fill_buffer.

puneith: Actually to think of it making both threads stop from the same signal is better.

jerjou: Can you explain your reasoning?

puneith: The reasoning: Both reader and writer threads consider the stopsignal. If reader looks at the indirect signal from the writer, we don't know if it needed to stop because there is really a stop signal or an error in the writer's output.

speech/grpc/transcribe_streaming.py
```diff
@@ -73,8 +73,8 @@ def _audio_data_generator(buff):
         A chunk of data that is the aggregate of all chunks of data in `buff`.
         The function will block until at least one data chunk is available.
     """
-    while True:
-        # Use a blocking get() to ensure there's at least one chunk of data
+    while not stoprequest.is_set():
+        # Use a blocking get() to ensure there's at least one chunk of data
```
jerjou: Please remove trailing whitespace

speech/grpc/transcribe_streaming.py
jerjou: Hm... sure, okay, I can see that. I think there's a case for just letting `_fill_buffer` propagate the way it is, but I won't block on it.

speech/grpc/transcribe_streaming.py
jerjou: Ah - this is because `buff.put(None)` is never executed if the above loop is exited because `stoprequest.is_set()` is `True`. (this means the blocking `buff.get()` in `_audio_data_generator` blocks forever, since it's never given the `None` to exit from its loop).

If you change the above loop to:

```py
try:
    ...
except IOError:
    pass
finally:
    buff.put(None)
```

then you should be able to put this back.

speech/grpc/transcribe_streaming.py
puneith: I actually had thought about both cases, but can now also see the case for other option. And now as I am thinking more I think the other option makes more sense :) Let me go back to it.

speech/grpc/transcribe_streaming.py
puneith: No, this happens because the stream is stopped and the thread hangs when this thread is still in the loop.

speech/grpc/transcribe_streaming.py
```diff
@@ -74,11 +74,8 @@ def _audio_data_generator(buff):
         The function will block until at least one data chunk is available.
     """
     while True:
-        # Use a blocking get() to ensure there's at least one chunk of data
+        # Use a blocking get() to ensure there's at least one chunk of data
```
jerjou: Trailing whitespace

speech/grpc/transcribe_streaming.py
```diff
@@ -120,14 +116,13 @@ def record_audio(rate, chunk):
     # This is necessary so that the input device's buffer doesn't overflow
     # while the calling thread makes network requests, etc.
     fill_buffer_thread = threading.Thread(
-        target=_fill_buffer, args=(audio_stream, buff, chunk))
+        target=_fill_buffer, args=(audio_stream, buff, chunk, stoprequest))
     fill_buffer_thread.start()
 
     yield _audio_data_generator(buff)
 
-    audio_stream.stop_stream()
```
jerjou: Any particular reason not to stop the stream before closing? While I'm not certain it's necessary, [all the examples](https://people.csail.mit.edu/hubert/pyaudio/docs/) stop the stream before closing it.

speech/grpc/transcribe_streaming.py
```diff
@@ -90,19 +87,18 @@ def _audio_data_generator(buff):
...
+def _fill_buffer(audio_stream, buff, chunk, stoprequest):
     """Continuously collect data from the audio stream, into the buffer."""
     try:
-        while True:
+        while not stoprequest.is_set():
             buff.put(audio_stream.read(chunk))
     except IOError:
-        # This happens when the stream is closed. Signal that we're done.
-        buff.put(None)
+        pass
```
jerjou: Without this, you'll have a race condition. ie if `_audio_data_generator`'s `buff.get()` is waiting for data in the buffer when `stoprequest` is triggered, it could potentially wait forever and never yield a value.

speech/grpc/transcribe_streaming.py
puneith: Once the stream is stopped, one may not call write or read. And thread _fill_buffer gets stuck.

speech/grpc/transcribe_streaming.py
puneith: done

speech/grpc/transcribe_streaming.py
```diff
@@ -208,7 +212,11 @@ def main():
             make_channel('speech.googleapis.com', 443)) as service:
         # For streaming audio from the microphone, there are three threads.
         # First, a thread that collects audio data as it comes in
```
jerjou: This comment should be right above the `with` block, since that's what it's describing.

speech/grpc/transcribe_streaming.py
```diff
@@ -87,22 +85,28 @@ def _audio_data_generator(buff):
                 data.append(buff.get(block=False))
             except queue.Empty:
                 break
+
+        # If the data contains None then set stop = True.
```
jerjou: Comments should describe the meaning and purpose of the code, not just narrate what the code is doing. Maybe something like:

```
# If `_fill_buffer` adds `None` to the buffer, the audio stream is closed.
# Yield the final bit of the buffer and exit the loop.
```

speech/grpc/transcribe_streaming.py
```diff
@@ -87,22 +85,28 @@ def _audio_data_generator(buff):
...
+        if None in data:
+            stop = True
+            data.remove(None)
```
jerjou: Ah - nice catch.

speech/grpc/transcribe_streaming.py
puneith: done

speech/grpc/transcribe_streaming.py
```diff
@@ -208,7 +212,11 @@ def main():
             make_channel('speech.googleapis.com', 443)) as service:
         # For streaming audio from the microphone, there are three threads.
         # First, a thread that collects audio data as it comes in
-        with record_audio(RATE, CHUNK) as buffered_audio_data:
+        
+        # stop request
```
puneith: done

speech/grpc/transcribe_streaming.py
```diff
@@ -87,22 +85,28 @@ def _audio_data_generator(buff):
...
     """Continuously collect data from the audio stream, into the buffer."""
     try:
-        while True:
+        while not stoprequest.is_set():
             buff.put(audio_stream.read(chunk))
     except IOError:
-        # This happens when the stream is closed. Signal that we're done.
+        pass
+    finally:
         buff.put(None)
```
puneith: done

speech/grpc/transcribe_streaming.py
```diff
@@ -198,17 +204,24 @@ def listen_print_loop(recognize_stream):
...
+    # First, a thread that collects audio data as it comes in
     with cloud_speech.beta_create_Speech_stub(
             make_channel('speech.googleapis.com', 443)) as service:
-        # For streaming audio from the microphone, there are three threads.
-        # First, a thread that collects audio data as it comes in
-        with record_audio(RATE, CHUNK) as buffered_audio_data:
+        
+        # stoprequest is event object which is set in `listen_print_loop`.
+        # To indicate that the trancsription should be stopped.
+        # `_fill_buffer` checks and stops collecting data from audio_stream.
```
jerjou: I had a bunch of trouble parsing this sentence. Perhaps:

```
# The `_fill_buffer` thread checks this object, and closes the `audio_stream` once it's set.
```

speech/grpc/transcribe_streaming.py
```diff
@@ -198,17 +204,24 @@ def listen_print_loop(recognize_stream):
...
+    # For streaming audio from the microphone, there are three threads.
+    # First, a thread that collects audio data as it comes in
     with cloud_speech.beta_create_Speech_stub(
             make_channel('speech.googleapis.com', 443)) as service:
-        # For streaming audio from the microphone, there are three threads.
-        # First, a thread that collects audio data as it comes in
-        with record_audio(RATE, CHUNK) as buffered_audio_data:
+        
+        # stoprequest is event object which is set in `listen_print_loop`.
+        # To indicate that the trancsription should be stopped.
```
puneith: done

speech/grpc/transcribe_streaming.py
puneith: done

puneith:
@dpebot merge when green.

puneith:
Does dpebot know he can also merge when someone approves the PR?

jerjou:
I'm not sure what you're asking. Are you suggesting that we should merge PRs before the automated tests pass?

puneith:
I am asking can dpebot take over when a) someone has approved the PR b) green status. Both a and b are true.

jerjou:
I don't understand what you're saying. Are you stating that - currently - someone has approved this PR AND the PR's status is green? Or are you asking whether dpebot can submit the PR if a OR b are true? Or are you asking whether dpebot can submit the PR when a AND b are true?

jerjou:
If it helps - IIUC, dpebot just waits for travis to turn green before submitting. Currently travis is still waiting for test results, so dpebot is not merging the PR yet.

puneith:
I thought my question was simple :)

In other words I am asking. Do we have to ask dpebot explicitly "merge when everything is green"? Can dpebot not figure that out automatically and do the needful?

jerjou:
I think your question is simple, but you're not expressing the full question, or you're not giving enough context for the question, so it's hard to figure out what you're asking. Perhaps it'd be helpful to structure your questions into three steps:

1. What the current situation is
1. What you expect the situation to be
1. The question

For example:

1. We ask dpebot to merge when green
1. dpebot should be able to do this without us asking
1. Do we need to ask dpebot, when we want to merge?

It was ambiguous in your original question what you thought the current situation was, and what you expected the situation to be. For instance - what did you mean by 'also'? That implies that there's another situation wherein dpebot will merge, besides when it's approved + green; but you don't describe this other situation you allude to, so there's no way for me to know what you're contrasting.

Does that make sense?

To answer your question - dpebot listens for travis to complete, and only merges if it has `status:success` AND the `automerge` label on it. Telling dpebot to 'merge when green' simply adds the `automerge` label. We don't want dpebot to just merge all green PRs because eg the user might still be working on the PR (for example, addressing comments from the reviewer), even if the tests pass.

puneith:
Wow! Travis is sloooow!

jerjou:
Yeah - it's usually one slow-running project in our org that backs everything else up :-/ We might consider changing to circleci, like some of the other teams are doing..

PR #645, opened 2016-11-09
Use pyaudio's built-in async api.

jerjou:
I discovered, while reviewing #635, that pyaudio has an async api.
This lets us avoid spinning up our own thread, and saves some LOC.
Also - since the callback thread will automatically be closed by pyaudio when the stream is closed, we don't have to check `stopaudio` in it any longer, which saves us a control object being passed around.

Tested in OSX and linux.

### DIFF
--- a/speech/grpc/transcribe_streaming.py
+++ b/speech/grpc/transcribe_streaming.py
@@ -17,10 +17,10 @@
 from __future__ import division
 
 import contextlib
+import functools
 import re
 import signal
 import sys
-import threading
 
 from google.cloud import credentials
 from google.cloud.speech.v1beta1 import cloud_speech_pb2 as cloud_speech
@@ -76,8 +76,7 @@ def _audio_data_generator(buff):
     stop = False
     while not stop:
         # Use a blocking get() to ensure there's at least one chunk of data.
-        chunk = buff.get()
-        data = [chunk]
+        data = [buff.get()]
 
         # Now consume whatever other data's still buffered.
         while True:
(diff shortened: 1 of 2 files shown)


speech/grpc/transcribe_streaming.py
```diff
@@ -86,54 +84,47 @@ def _audio_data_generator(buff):
...
-    # This is necessary so that the input device's buffer doesn't overflow
-    # while the calling thread makes network requests, etc.
-    fill_buffer_thread = threading.Thread(
-        target=_fill_buffer, args=(audio_stream, buff, chunk, stoprequest))
-    fill_buffer_thread.start()
-
     yield _audio_data_generator(buff)
 
-    fill_buffer_thread.join()
     audio_stream.close()
```
puneith: You can add the audio_stream.stop_stream() back.

puneith:
LGTM other than the single comment I had.

speech/grpc/transcribe_streaming.py
```diff
@@ -86,54 +84,48 @@ def _audio_data_generator(buff):
...
                 break
 
-        # If `_fill_buffer` adds `None` to the buffer, the audio stream is
-        # closed. Yield the final bit of the buffer and exit the loop.
+        # `None` in the buffer signals that the audio stream is closed. Yield
+        # the final bit of the buffer and exit the loop.
         if None in data:
             stop = True
             data.remove(None)
         yield b''.join(data)
```
jonparrott: nit: blank line above yield.

speech/grpc/transcribe_streaming.py
```diff
@@ -86,54 +84,48 @@ def _audio_data_generator(buff):
...
 
 
-def _fill_buffer(audio_stream, buff, chunk, stoprequest):
+def _fill_buffer(buff):
```
jonparrott: Instead of a closure, just use `functools.partial`:

```python
def audio_stream_callback(buffer, in_data, frame_count, time_info, status_flags):
    ...

stream_callback=functools.partial(audio_stream_callback, buffer)
```

speech/grpc/transcribe_streaming.py
```diff
@@ -86,54 +84,48 @@ def _audio_data_generator(buff):
...
+    return callback
 
 
 # [START audio_stream]
 @contextlib.contextmanager
-def record_audio(rate, chunk, stoprequest):
+def record_audio(rate, chunk):
     """Opens a recording stream in a context manager."""
+    # Create a thread-safe buffer of audio data
+    buff = queue.Queue()
```
jonparrott: s/buff/buffer? (throughout)

speech/grpc/transcribe_streaming.py
```diff
@@ -213,17 +214,9 @@ def listen_print_loop(recognize_stream, stoprequest):
...
             make_channel('speech.googleapis.com', 443)) as service:
-
-        # stoprequest is event object which is set in `listen_print_loop`
-        # to indicate that the trancsription should be stopped.
-        #
-        # The `_fill_buffer` thread checks this object, and closes
-        # the `audio_stream` once it's set.
-        stoprequest = threading.Event()
-
         # For streaming audio from the microphone, there are three threads.
```
jonparrott: are these comments about the number of threads still accurate?

speech/grpc/transcribe_streaming.py
jerjou: Oh neat.

speech/grpc/transcribe_streaming.py
jerjou: I avoided this since `buffer` [is a keyword](https://docs.python.org/2/library/functions.html#buffer)

speech/grpc/transcribe_streaming.py
jerjou: Yup! It's just that before we were spinning up the buffer-the-audio thread manually, and now pyaudio is doing it for us.

speech/grpc/transcribe_streaming.py
jonparrott: TIL.

(It's a built-in function which means it *can* be shadowed, like `file`, but I'm fine if you leave it as `buff`).

jerjou: Yeah, me too. I just avoided it because vim bolded it after I spelled it out, which made me wary :-)

jerjou:
@dpebot would you please merge when travis becomes green (presumably with envy)?

jonparrott:
@jerjou travis failed here.

jerjou:
Oh right. Tests.
Should be fixed now.

PR #658, opened 2016-11-15
Add quickstart sample for the Cloud Speech API.

jmdobry:


speech/cloud-client/quickstart.py
```diff
@@ -0,0 +1,47 @@
...
+    import os
+
+    # Imports the Google Cloud client library
+    from google.cloud import speech
+
+    # Instantiates a client
+    speech_client = speech.Client()
+
+    # The name of the audio file to transcribe
+    file_name = os.path.join(os.path.dirname(__file__), 'resources/audio.raw')
```
jonparrott: don't use `/` in join, just pass more arguments:

```
os.path.join(os.path.dirname(__file__), 'resources', 'audio.raw')
```

speech/cloud-client/quickstart.py
```diff
@@ -0,0 +1,47 @@
...
+
+    # Loads the audio into memory
+    with io.open(file_name, 'rb') as audio_file:
+        content = audio_file.read()
+        audio_sample = speech_client.sample(
+            content, None, 'LINEAR16', 16000)
```
jmdobry: Done.

speech/cloud-client/quickstart.py
jmdobry: Done.

PR #662, opened 2016-11-16
Add comment about python runtime to Dockerfile

jeffmendoza:
Need to link to source of Docker image


endpoints/getting-started/Dockerfile.custom
```diff
@@ -1,3 +1,7 @@
+# The Google App Engine python runtime is Debian Jessie with Python installed
```
waprin: Probably better not to say it's Jessie in case it changes underneath us and the comment becomes outdated, better just to link to it. Also while we're at it ,probably best to consistently capitalize Python and Github in prose. Finally, I think we still put license headers on Dockerfiles?

PR #663, opened 2016-11-16
Switch Error Reporting to Google Cloud Python

waprin:


error_reporting/main_test.py
```diff
@@ -17,7 +17,9 @@
 import main
 
 
-@mock.patch("fluent.event")
-def test_error_sends(event_mock):
+@mock.patch("main.error_reporting")
+def test_error_sends(error_reporting_mock):
+    client_mock = mock.Mock()
```
jonparrott: This isn't a system test, it's a unit test. Why not forgo mocking?

waprin: Not sure exactly what you mean? Without mocks there's not much to assert, seemed better to throw in one mock to assert something.

jonparrott: We have several tests that don't assert, they implicitly assert the absence of failure.

jonparrott: with the mock we actually test *nothing* here. The client library can change and break the sample and tests would never fail, which is why we don't write unit tests here. Everything in this repo is a system test so I do not care about the cost of making an RPC.

If you want to assert that the call is being made, use [`wraps`](http://wesmckinney.com/blog/spying-with-python-mocks/).

waprin: It doesn't test nothing, it tests the code path, which it doesn't without the mock. `wraps` seems like it cleans up the mocking code. However, can see both sides and not worth arguing about so made change.

waprin:
Also renamed files and added the README.rst.in

PR #675, opened 2016-11-19
Add gRPC api config files.

qiwzhang:


qiwzhang:
@wlu2016, @kyessenov, @lizan  for review

qiwzhang:
@jeffmendoza Can you review?

kyessenov:
Can we rename YAML files to bookstore_config.yaml and so on?
I think we need to be careful not to make it easy to confuse these files with Kubernetes manifests.

qiwzhang:
OK, I will rename them to bookstore_api_config.yaml and bookstore_http_api_config.yaml

PR #685, opened 2016-11-28
Change the listening port from 9000 to 80

lookuptable:
This makes our Quickstarts consistent such that port 80 is used in both
examples on GCE and GKE.


lookuptable:
@kyessenov for SA

PR #688, opened 2016-11-29
Speech transcribe async bug

ryanmats:
Fixed bugs for Speech API GRPC sample code:
- Issues with import statements
- Changed default encoding choice from FLAC to LINEAR16. Added code to print error message if there was an error with the transcribe_async long running operation. (As per Jerjou's suggestion)
- Added tests for GRPC transcribe and transcribe_async samples. Code uploads the audio.raw blob to the default CLOUD_STORAGE_BUCKET, tests sample code using this blob's URI, and deletes the blob at the end.


jonparrott:
@ryanmats all of speeching's resources should be pre-uploaded, so there should be no need to upload/delete the blob every time (see `scripts/prepare-testing-project.sh` L47.

jonparrott:
Other than the note about the tests, this LGTM. @jerjou can you take a quick look before we merge?

speech/grpc/transcribe_async.py
```diff
@@ -85,9 +85,14 @@ def main(input_uri, encoding, sample_rate, language_code='en-US'):
...
         time.sleep(1)
         operation = service.GetOperation(
-            operations_grpc_pb2.GetOperationRequest(name=name),
+            operations_pb2.GetOperationRequest(name=name),
             DEADLINE_SECS)
 
+        if operation.error.message:
+            print("")
+            print("Operation error:")
+            print(operation.error)
```
jerjou: What? Jon didn't wasn't like, "Please use single quotes for strings"?? WHO ARE YOU AND WHAT HAVE YOU DONE TO @jonparrott ?!? :)

I'd just combine this into one string, for brevity:

    print('\nOperation error:\n{}'.format(operation.error))

We should be paid proportional to like, the inverse of lines of code :) Except then we'd be incentivized to not write any code at all... hm...

speech/grpc/transcribe_async_test.py
```diff
@@ -0,0 +1,38 @@
...
+
+    # Run the transcribe_async sample on the audio file, verify correct results
+    speech_storage_uri = 'gs://' + os.environ.get('CLOUD_STORAGE_BUCKET')
+    speech_storage_uri += '/audio.raw'
+    main(speech_storage_uri, 'LINEAR16', 16000)
+    out, err = capsys.readouterr()
+    assert re.search(r'how old is the Brooklyn Bridge', out, re.DOTALL | re.I)
+
+    # Delete audio file from storage bucket
+    blob.delete()
```
jerjou: +1 just have the asset on the GCS bucket already :)

But for future reference, if you're going to do this, should probably put this in a context manager (so that eg if the assertion fails, the blob is still deleted), or in one of those fancy [pytest fixtures](http://doc.pytest.org/en/latest/fixture.html#fixture-finalization-executing-teardown-code) (same idea)

jerjou:
Minor comments. Mainly just Jon's thing :) Otherwise, looks good.

speech/grpc/transcribe_async.py
jonparrott: I blame my lack of sleep.

ryanmats:
@jonparrott @jerjou I pushed the changes based on your code review comments. Also updated the speech sample audio filename in scripts/prepare-testing-project.sh.

scripts/prepare-testing-project.sh
```diff
@@ -44,7 +44,7 @@ echo "Creating pubsub resources."
 gcloud alpha pubsub topics create gae-mvm-pubsub-topic
 
 echo "Creating speech resources."
-gsutil cp speech/api/resources/audio.flac gs://$GCLOUD_PROJECT/speech/
+gsutil cp speech/api-client/resources/audio.raw gs://$GCLOUD_PROJECT/speech/
```
jonparrott: Are there other tests that depend on audio.flac?

speech/grpc/transcribe_async_test.py
```diff
@@ -0,0 +1,27 @@
...
+import os
+import re
+
+from transcribe_async import main
+
+
+def test_main(resource, capsys):
+
+    # Run the transcribe_async sample on audio.raw, verify correct results
+    speech_storage_uri = 'gs://' + os.environ.get('CLOUD_STORAGE_BUCKET')
```
jonparrott: Use the `cloud_config` or `remote_resource` fixture.

scripts/prepare-testing-project.sh
ryanmats: Just searched through the repo and it doesn't look like there are. Also, the audio.flac file doesn't exist in that directory anymore.

jerjou: (doesn't have to be in this CL - just saying it'd probably be a good idea..)

ryanmats: changed in this PR

conftest.py
```diff
@@ -26,12 +26,15 @@ def __init__(self, **kwargs):
 @pytest.fixture(scope='session')
 def cloud_config():
     """Provides a configuration object as a proxy to environment variables."""
+    speech_storage_uri = 'gs://' + os.environ.get('CLOUD_STORAGE_BUCKET')
```
jonparrott: @ryanmats there's already the `storage_bucket` thing here, there isn't any need to add another one.

ryanmats: fixed

speech/grpc/transcribe_async_test.py
```diff
@@ -0,0 +1,26 @@
...
+
+import re
+
+from transcribe_async import main
+
+
+def test_main(resource, capsys, cloud_config):
+
+    # Run the transcribe sample on audio.raw, verify correct results
+    speech_storage_uri = 'gs://' + cloud_config.storage_bucket
```
jonparrott: Lol okay, I promise I'm not doing this just to annoy you:

Prefer:

```python
storage_uri = 'gs://{}/speech/audio.raw'.format(cloud_config.storage_bucket)
```

(Here and in the other test file).

ryanmats: fixed

PR #698, opened 2016-12-04
corrected runtime

ShuminKong:
Otherwise, the following error will be raised:
`google.appengine.tools.devappserver2.errors.InvalidAppConfigError: In env: flex, only the following runtimes are allowed: ('python-compat', 'java', 'java7', 'go', 'custom')`


jonparrott:
Thanks for the pull request, but this is as intended. `dev_appserver` is not used to run `env: flex` apps except for apps using the deprecated `python-compat` runtime.

You can just run the application directly when running locally.

PR #701, opened 2016-12-08
Allow ListShelves method to be called without API key

lookuptable:
This simplifies our quickstart such that reader does not need to go
through an extra step to create an API key before calling this API.


lookuptable:
@lizan @wlu2016 @kyessenov

Issue #707, opened 2016-12-09
 Permission denied running crud_object.py 


griffin2000:
## In which file did you encounter the issue?
https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/storage/api/crud_object.py

## Describe the issue

As written the crud_object.py file main function will fail with this error (on windows at least):
```
PermissionError: [Errno 13] Permission denied: 'C:\\Users\\owner\\AppData\\Local\\Temp\\tmp3preaezj'


Traceback (most recent call last):
  File "crud_object.py", line 150, in <module>
    main(args.bucket, args.filename, args.reader, args.owner)
  File "crud_object.py", line 47, in main
    if not filecmp.cmp(filename, tmpfile.name):
  File "E:\Python\Python35\lib\filecmp.py", line 62, in cmp
    outcome = _do_cmp(f1, f2)
  File "E:\Python\Python35\lib\filecmp.py", line 75, in _do_cmp
    with open(f1, 'rb') as fp1, open(f2, 'rb') as fp2:
```

The issue is that NamedTemporaryFile keeps the file handle open so that when filecmp attempts to open it, it will fail with a permission denied error. The correct way to do this is shown below:

```
    print('Fetching object..')
    with tempfile.NamedTemporaryFile(mode='w+b', delete=False) as tmpfile:
        get_object(bucket, filename, out_file=tmpfile)
        tmpfile.seek(0)
        tmpName = tmpfile.name
    if not filecmp.cmp(filename, tmpName):
        raise Exception('Downloaded file != uploaded object')
    print('Deleting object..')
    os.remove(tmpName)

```

jerjou:
Ended up removing those bits of code, since they were peripheral to the sample.

PR #710, opened 2016-12-09
Compare files manually.

jerjou:
Addresses #707

### DIFF
--- a/storage/api/crud_object.py
+++ b/storage/api/crud_object.py
@@ -24,7 +24,6 @@ For more information, see the README.md under /storage.
 """
 
 import argparse
-import filecmp
 import json
 import tempfile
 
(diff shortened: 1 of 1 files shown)


jonparrott:
Hmm, I liked the idea of just closing the handle to the named temporary file and then using filecmp.

jerjou:
Oh. It just seemed less clean. eg:
* You're now manually in charge of file cleanup, which feels like it should be the context manager's job
* The use of `filecmp` with a filename when you've already got a handle on the file always felt a bit hacky.
* The file comparison is now both inside and outside of the context manager, which breaks it up visually even though it's the same conceptual operation

Granted, this is a bit more verbose and low-level, but I feel like it flows better.

jonparrott:
Can we just remove the file comparison altogether from this sample?

jerjou:
Oh yeah - sure. Apparently this file isn't even referenced in the docs anymore.
Done.

storage/api/crud_object.py
```diff
@@ -24,14 +24,14 @@
...
 import argparse
-import filecmp
 import json
 import tempfile
 
 from googleapiclient import discovery
 from googleapiclient import http
 
 from oauth2client.client import GoogleCredentials
+from six.moves import zip_longest
```
jerjou: Ah. right.

PR #720, opened 2016-12-14
Replace 'vm: true' with 'env: flex' in yaml files.

janlugt:
vm: true is deprecated.

### DIFF
--- a/appengine/flexible/multiple_services/gateway-service/app.yaml
+++ b/appengine/flexible/multiple_services/gateway-service/app.yaml
@@ -1,6 +1,6 @@
 service: default
 runtime: python
-vm: true
+env: flex
 entrypoint: gunicorn -b :$PORT main:app
 
 runtime_config:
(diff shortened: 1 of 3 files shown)


janlugt:
I signed it!

jonparrott:
Thanks, @janlugt - I think the endpoints one was using vm: true for a reason.

@sepehre @kdeus is there any reason the endpoints sample won't work on `env: flex`?

janlugt:
I did the quickstart for endpoints this afternoon as part of a review, and gcloud did not accept the config until I changed vm:true to env:flex. I haven't tested the other two files, but I figure it would be the same. It's mentioned in the release notes here: https://cloud.google.com/appengine/docs/flexible/python/release-notes#november_15th_2016

kdeus:
Sounds fine.  Looks like we just hadn't updated this sample to env: flex yet.  Thanks!

sepehre:
Yes, flex everywhere! Onwards! :D

PR #732, opened 2016-12-20
Replace translate_text_with_model with translate_text

aman-ebay:
Fix model parameter.
Also delete mention of whitelist below replaced parameter.


jonparrott:
@aman-ebay can we get some more context on this change?

aman-ebay:
The code is correct.

aman-ebay:
Closed the request after clarification. No need for this change.

On Tue, Dec 20, 2016 at 1:01 PM, Jon Wayne Parrott <notifications@github.com

PR #746, opened 2017-01-05
More specific exception swallowing.

jerjou:


speech/grpc/transcribe_streaming.py
```diff
@@ -226,9 +226,13 @@ def main():
             listen_print_loop(recognize_stream)
 
             recognize_stream.cancel()
-        except grpc.RpcError:
-            # This happens because of the interrupt handler
-            pass
+        except grpc.RpcError as e:
+            code = e.code()
+            if code is code.CANCELLED:
```
jonparrott: why not:

```
if not code is code.CANCELLED:
   raise
```

jerjou: Because that would be way too clean, clearly.

PR #748, opened 2017-01-06
Updated memcache error checking to handle large object exception

ryanmats:
See bug #33780149 on buganizer.

For some reason memcache.add throws an exception for objects larger than 1 MB rather than simpling returning False. Thus, I changed the sample code to use try/except around memcache.add, while preserving the check to see if memcache.add returned True or False. The boolean returned does seem to be necessary since it is False when memcache already has an entry for a given key.

Also got rid of 'version: 1' from the app.yaml so that 'gcloud app deploy' works.

### DIFF
--- a/appengine/standard/memcache/guestbook/main.py
+++ b/appengine/standard/memcache/guestbook/main.py
@@ -86,14 +86,12 @@ class MainPage(webapp2.RequestHandler):
         if greetings is None:
             greetings = self.render_greetings(guestbook_name)
             try:
-                added = memcache.add('{}:greetings'.format(guestbook_name),
-                                     greetings, 10)
+                added = memcache.add(
+                    '{}:greetings'.format(guestbook_name), greetings, 10)
                 if not added:
                     logging.error('Memcache set failed.')
             except ValueError:
                 logging.error('Memcache set failed - data larger than 1MB')
-            except Exception as e:
-                logging.error('Memcache set failed - {}'.format(e))
         return greetings
     # [END check_memcache]
 
--- a/appengine/standard/memcache/guestbook/main.py
+++ b/appengine/standard/memcache/guestbook/main.py
@@ -85,8 +85,11 @@ class MainPage(webapp2.RequestHandler):
         greetings = memcache.get('{}:greetings'.format(guestbook_name))
         if greetings is None:
             greetings = self.render_greetings(guestbook_name)
-            if not memcache.add('{}:greetings'.format(guestbook_name),
-                                greetings, 10):
+            try:
+                if not memcache.add('{}:greetings'.format(guestbook_name),
+                                    greetings, 10):
+                    logging.error('Memcache set failed.')
+            except:
                 logging.error('Memcache set failed.')
         return greetings
     # [END check_memcache]
--- a/appengine/standard/memcache/guestbook/main.py
+++ b/appengine/standard/memcache/guestbook/main.py
@@ -86,11 +86,14 @@ class MainPage(webapp2.RequestHandler):
         if greetings is None:
             greetings = self.render_greetings(guestbook_name)
             try:
-                if not memcache.add('{}:greetings'.format(guestbook_name),
-                                    greetings, 10):
+                added = memcache.add('{}:greetings'.format(guestbook_name),
+                                     greetings, 10)
+                if not added:
                     logging.error('Memcache set failed.')
-            except:
-                logging.error('Memcache set failed.')
+            except ValueError:
+                logging.error('Memcache set failed - data larger than 1MB')
+            except Exception as e:
+                logging.error('Memcache set failed - {}'.format(e))
         return greetings
     # [END check_memcache]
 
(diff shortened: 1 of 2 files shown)


appengine/standard/memcache/guestbook/main.py
```diff
@@ -85,8 +85,11 @@ def get_greetings(self, guestbook_name):
         greetings = memcache.get('{}:greetings'.format(guestbook_name))
         if greetings is None:
             greetings = self.render_greetings(guestbook_name)
-            if not memcache.add('{}:greetings'.format(guestbook_name),
-                                greetings, 10):
+            try:
+                if not memcache.add('{}:greetings'.format(guestbook_name),
+                                    greetings, 10):
+                    logging.error('Memcache set failed.')
+            except:
```
waprin: Catch a more specific exception

appengine/standard/memcache/guestbook/main.py
```diff
@@ -85,8 +85,11 @@ def get_greetings(self, guestbook_name):
         greetings = memcache.get('{}:greetings'.format(guestbook_name))
         if greetings is None:
             greetings = self.render_greetings(guestbook_name)
-            if not memcache.add('{}:greetings'.format(guestbook_name),
-                                greetings, 10):
+            try:
+                if not memcache.add('{}:greetings'.format(guestbook_name),
+                                    greetings, 10):
+                    logging.error('Memcache set failed.')
```
waprin: since these are two separate cases, you want two separate error messages. Should be, `Memcache add failed, key already exists` and `Memcache add failed, data larger than 1MB`

appengine/standard/memcache/guestbook/main.py
```diff
@@ -85,8 +85,11 @@ def get_greetings(self, guestbook_name):
         greetings = memcache.get('{}:greetings'.format(guestbook_name))
         if greetings is None:
             greetings = self.render_greetings(guestbook_name)
-            if not memcache.add('{}:greetings'.format(guestbook_name),
-                                greetings, 10):
+            try:
+                if not memcache.add('{}:greetings'.format(guestbook_name),
```
waprin: Maybe break up this cramped up line with  `added = memcache.add()` then `if not added`

appengine/standard/memcache/guestbook/main.py
ryanmats: done

appengine/standard/memcache/guestbook/main.py
ryanmats: Added a more specific message for data larger than 1MB.

For the memcache.add False result, it might not necessarily be a key already exists error, so I'm leaving it as 'Memcache set failed.'

appengine/standard/memcache/guestbook/main.py
ryanmats: done

appengine/standard/memcache/guestbook/main.py
```diff
@@ -85,9 +85,15 @@ def get_greetings(self, guestbook_name):
...
         if greetings is None:
             greetings = self.render_greetings(guestbook_name)
-            if not memcache.add('{}:greetings'.format(guestbook_name),
-                                greetings, 10):
-                logging.error('Memcache set failed.')
+            try:
+                added = memcache.add('{}:greetings'.format(guestbook_name),
+                                     greetings, 10)
+                if not added:
+                    logging.error('Memcache set failed.')
```
waprin: just need `-value already exists` here right?

ryanmats: Do we know for sure if memcache.add returning False is necessarily a 'key already exists' type of error? The docs say memcache.add returns 'False on error'. (https://cloud.google.com/appengine/docs/python/refdocs/modules/google/appengine/api/memcache#Client.add). Wondering if the False could mean something else in certain circumstances.

waprin: Yeah I don't know, I guess we can leave it as is.

appengine/standard/memcache/guestbook/main.py
```diff
@@ -85,9 +85,15 @@ def get_greetings(self, guestbook_name):
         greetings = memcache.get('{}:greetings'.format(guestbook_name))
         if greetings is None:
             greetings = self.render_greetings(guestbook_name)
-            if not memcache.add('{}:greetings'.format(guestbook_name),
-                                greetings, 10):
-                logging.error('Memcache set failed.')
+            try:
+                added = memcache.add('{}:greetings'.format(guestbook_name),
```
jonparrott: please avoid hanging indents, newline after `add(`.

appengine/standard/memcache/guestbook/main.py
```diff
@@ -85,9 +85,15 @@ def get_greetings(self, guestbook_name):
...
+                                     greetings, 10)
+                if not added:
+                    logging.error('Memcache set failed.')
+            except ValueError:
+                logging.error('Memcache set failed - data larger than 1MB')
+            except Exception as e:
```
jonparrott: No need for this broad exception.

PR #750, opened 2017-01-09
Change service name from appspot.com to cloud.goog

qiwzhang:


qiwzhang:
@jonparrott could you review it?

PR #759, opened 2017-01-15
Updated README.md

dwardu89:
Updated the command gcloud beta app deploy to be gcloud app deploy, since the command is no longer in beta.

### DIFF
--- a/appengine/flexible/endpoints/README.md
+++ b/appengine/flexible/endpoints/README.md
@@ -41,7 +41,7 @@ Open the `openapi.yaml` file and in the `host` property, replace
 
 Then, deploy the sample using `gcloud`:
 ```bash
-$ gcloud beta app deploy
+$ gcloud app deploy
 ```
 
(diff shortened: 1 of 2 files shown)


dwardu89:
I signed it!

jonparrott:
I think this one is intentionally still beta because of endpoints.

@lookuptable @wlu2016 @qiwzhang?

qiwzhang:
The command is "gcloud app deploy" now.  Please refers to this

https://cloud.google.com/endpoints/docs/quickstart-app-engine#python

Thanks for updating the README.

jonparrott:
@qiwzhang it seems the quickstart no longer even uses this particular sample (instead uses /endpoints), should we just remove this one?

qiwzhang:
It is better to leave it here.  Maybe there are still some docs referring to here.

jonparrott:
I can search and verify, better to not have a duplicated sample. :)

qiwzhang:
OK,  thanks.  it can be removed if there are not doc referring to it.

PR #764, opened 2017-01-20
BigQuery - REST samples for adding labels.

tswast:
See docs: https://cloud.google.com/bigquery/docs/labeling-datasets

These samples uses raw REST calls because [adding labels is not yet supported in the cloud client libraries](https://github.com/GoogleCloudPlatform/google-cloud-python/issues/2931).


tswast:
I'm looking into the Travis failure.

```
{\n  "error" : "invalid_scope",\n  "error_description" : "Empty or missing scope not allowed."\n}
```

Strange that it worked on my machine. I guess maybe because Travis is using a service account and I was using `gcloud` credentials?

tswast:
I'll add a scope argument to the `google.auth.defaults()` call.

jonparrott:
> I'll add a scope argument to the google.auth.defaults() call.

yep, that should do it.

bigquery/rest/label_dataset.py
```diff
@@ -0,0 +1,79 @@
...
+        url_format.format(project_id=project_id, dataset_id=dataset_id),
+        params={'fields': 'labels'},
+        json={
+            'labels': {
+                label_key: label_value,
+            }
+        })
+
+    # Check the response for errors.
+    if response.status_code != 200:
```
jonparrott: use `response.raise_for_status()`.

bigquery/rest/label_dataset.py
```diff
@@ -0,0 +1,79 @@
...
+def label_dataset(dataset_id, label_key, label_value, project_id=None):
+    """Add or modify a label on a dataset."""
+    # Authenticate requests using Google Application Default credentials.
+    credentials, default_project_id = google.auth.default()
+
+    if project_id is None:
+        project_id = default_project_id
+
+    # Send a PATCH request to add or modify a label.
+    session = google.auth.transport.requests.AuthorizedSession(credentials)
```
jonparrott: put `session` in the auth section (two lines above)

bigquery/rest/label_table.py
```diff
@@ -0,0 +1,84 @@
+#!/usr/bin/env python
```
jonparrott: Would it make sense to combine these into one single file named `labels`?

bigquery/rest/label_dataset.py
tswast: Done.

bigquery/rest/label_dataset.py
tswast: Done.

bigquery/rest/label_table.py
tswast: Yes, it would. Combined.

PR #768, opened 2017-01-30
Removing redundant re-definition of 'Counter' class

Livshitz:


jonparrott:
I believe we did this on purpose.

I appreciate the contribution, however!

Livshitz:
OK. Can you explain the purpose behind it please? Thanks :)

jonparrott:
@Livshitz basically I think the reasoning was that we wanted to to be possible to deploy the worker from a completely separate codebase, where it wouldn't have access to the `application` module.

PR #770, opened 2017-01-31
Print transcripts so unicode is printed properly

jerjou:


speech/api-client/transcribe.py
```diff
@@ -75,7 +75,15 @@ def main(speech_file):
     # [END construct_request]
     # [START send_request]
     response = service_request.execute()
+
+    # First print the raw json response
     print(json.dumps(response))
```
jonparrott: might as well go ahead and add `indent=2` to this?

jerjou: Ah yes. I like.

puneith:
Sorry I am late. But why not print(json.dumps(item, encoding='utf8', ensure_ascii=False))?

jerjou:
Oh. Yup. That woulda worked too. I totally didn't think of the available parameters to json.dumps, clearly -_-;

jonparrott:
Send another PR?

PR #773, opened 2017-02-01
Just print json.

jerjou:


jerjou:
Ugh. python3 doesn't have the `encoding` argument for `json.dumps`. Doesn't seem worth the time to figure out how to this the Right Way - I rescind this PR.

puneith:
Bummer! You could do try/catch but leave it to you.

PR #774, opened 2017-02-01
Changes example code to use cloud client and adds usage examples.

gguuss:


jonparrott:
@gguuss the `api-client` folder is exclusively for samples using `google-api-python-client`.

PR #777, opened 2017-02-01
Add gcp repo tools as testing dep

waprin:


jonparrott:
So this will probably break the App Engine stuff because it expects gcprepotools to be installed before it actually installs the requirements.

PR #778, opened 2017-02-02
Removes examples on non-cloud library

gguuss:


gguuss:
This is for speech, creating the CL a little proactively because we now have the transcribe / transcribe_async examples in cloud-client.

jonparrott:
@gguuss typically we keep the api-client samples around for posterity. Do we absolutely not want users to have any samples for Speech using that client library? If so, then I'm fine with merging this.

gguuss:
@jonparrott A fair statement. If we want to keep them around, let's add usage examples to them.

jonparrott:
@gguuss it leave it up to you.

gguuss:
I'm going to abandon branch and will add usage examples in a separate PR.

PR #786, opened 2017-02-07
random generation of keyring / cryptokey names

ryanmats:


kms/api-client/snippets_test.py
```diff
@@ -22,10 +25,14 @@
 LOCATION = 'global'
 
 # Your Google Cloud Platform KeyRing name
-KEYRING = 'sample-keyring-43'
+KEYRING = ''
```
jonparrott: Here's a one-liner for you

```python
KEYRING = ''.join(random.choice(string.ascii_lowercase + string.digits) for _ in range(12))
```

ryanmats: done

jonparrott:
@dpebot please merge when travis passes

PR #797, opened 2017-02-10
Add get metric descriptor

waprin:


monitoring/api/v3/cloud-client/snippets.py
```diff
@@ -127,6 +127,13 @@ def get_monitored_resource_descriptor(type):
     # [END get_monitored_resources]
 
 
+def get_metric_descriptor(type):
```
jonparrott: nit: don't use `type`, it collides with built-in `type`.

waprin: done

PR #809, opened 2017-02-16
Add region tags for firebase auth to openapi.yaml

llatif:
In order to allow for includes


jonparrott:
FYI: you guys can use `indented_block="firebase:"` instead of region tags.

PR #810, opened 2017-02-16
Remove usage of GoogleCredentials

jonparrott:


jonparrott:
Travis is failing because of a weird bug with nox and linux. I'm gonna try to fix that in nox before I merge this.

PR #815, opened 2017-02-19
Update snippets.py

jasonmorton:
Seems like this should use the snapshot context


jasonmorton:
I signed it.

jonparrott:
Nice catch, thanks @jasonmorton!

Issue #811, opened 2017-02-17
ImportError: No module named longrunning


araa47:
## In which file did you encounter the issue?
While running trancribe_streaming.py for speech to text 

### Did you change the file? If so, how?
No

## Describe the issue

Akshays-iMac-2:grpc akshay$ python transcribe_streaming.py
Traceback (most recent call last):
  File "transcribe_streaming.py", line 29, in <module>
    from google.cloud.grpc.speech.v1beta1 import cloud_speech_pb2
  File "/usr/local/Cellar/python/2.7.13/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/google/cloud/grpc/speech/v1beta1/cloud_speech_pb2.py", line 17, in <module>
    from google.longrunning import operations_pb2 as google_dot_longrunning_dot_operations__pb2
ImportError: No module named longrunning

jonparrott:
@araa47 Are you using `virtualenv`? If so, can you tell me the output of `pip freeze`?

Neeraj-ky:
@jonparrott  I am facing same issue when trying to use RPC call for streaming request. 
I am using **virtualenv** and output for **pip freeze** is - 

$ pip freeze
Warning: cannot find svn location for distribute==0.6.24dev-r0
Cython==0.15.1
Django==1.4.22
Fabric==1.8.2
Flask==0.8
FormEncode==1.2.4
GeoIP-Python==1.2.4
GnuPGInterface==0.3.2
Jinja2==2.6
M2Crypto==0.21.1
Mako==0.7.0
MarkupSafe==0.15
MySQL-python==1.2.3
PIL==1.1.7
Paste==1.7.5.1
PyYAML==3.10
Pygments==1.5
SOAPpy==0.12.0
SpeechRecognition==3.6.0
Unidecode==0.04.9
Werkzeug==0.8.3
argparse==1.2.1
backports-abc==0.4
backports.ssl-match-hostname==3.5.0.1
boto==2.42.0
bpython==0.11
certifi==2016.08.31
chardet==2.0.1
colorama==0.2.4
## FIXME: could not find svn URL in dependency_links for this package:
distribute==0.6.24dev-r0
docutils==0.8.1
duplicity==0.6.18
enum34==1.1.6
euca2ools==2.0.2
filechunkio==1.5
fpconst==0.7.2
futures==3.0.5
google-auth==0.7.0
googleapis-common-protos==1.5.2
greenlet==0.4.7
[...]

PR #822, opened 2017-02-23
Update grpc python lib for speech.

jerjou:
Per @lukesneeringer's comment in #811, update grpc speech lib.


jonparrott:
@dpebot can you merge when travis passes?

PR #843, opened 2017-03-07
Updates translate version to match speech and language.

gguuss:


jonparrott:

PR #850, opened 2017-03-14
Fix ml_engine tests

elibixby:
* Rename tf_records -> examples everywhere, since TFRecords is technically a misnomer (or rather the name for the file container format for a number of tf.train.Example proto [or really any proto])

* Move both JSON and EXAMPLE tests to a new model, so that they share a common expected output

* Hopefully fix python 3 compat

### DIFF
--- a/ml_engine/online_prediction/predict.py
+++ b/ml_engine/online_prediction/predict.py
@@ -62,10 +62,10 @@ def predict_json(project, model, instances, version=None):
 
 
 # [START predict_tf_records]
-def predict_tf_records(project,
-                       model,
-                       example_bytes_list,
-                       version=None):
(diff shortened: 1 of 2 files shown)


elibixby:
Can't get the py35 tests to run locally @jonparrott . I get the following error:

https://gist.github.com/elibixby/ac829c83972c4fb9f5a3c8ff16d87e9f

jonparrott:
@elibixby travis seems fine with py35, might be your local install. Merging.

elibixby:
The failure was in the slow tests. I've kicked off https://jenkins.on2flixcloud.tv/view/python/job/python-doc-samples/731/ to verify they work in Jenkins

PR #877, opened 2017-03-24
Replaces test video for faces

gguuss:


jonparrott:

PR #878, opened 2017-03-24
Fix py2-3 compat issue in sending bytestring

elibixby:


jonparrott:

PR #896, opened 2017-04-11
Django flex sample: mysql -> postgres

ryanmats:
Corresponding tutorial changes coming very soon


appengine/flexible/django_cloudsql/mysite/settings.py
```diff
@@ -81,7 +81,7 @@
 # [START dbconfig]
 DATABASES = {
     'default': {
-        'ENGINE': 'django.db.backends.mysql',
```
jonparrott: I think it's probably useful to keep the mysql one around and add a comment that if you're using Cloud SQL for MySQL you should use that engine instead.

ryanmats: done

PR #900, opened 2017-04-14
Removes samples that don't use the cloud client library

gguuss:


jonparrott:
@dpebot please merge when travis passes.

PR #901, opened 2017-04-14
Translate unicode

gguuss:


translate/cloud-client/snippets.py
```diff
@@ -1,4 +1,6 @@
 #!/usr/bin/env python
+# -*- coding: iso-8859-15 -*-
```
jonparrott: You shouldn't need to specify an encoding in this file.

translate/cloud-client/snippets.py
```diff
@@ -23,7 +25,9 @@
 
 import argparse
 
+
 from google.cloud import translate
+from six import python_2_unicode_compatible
```
jonparrott: Oh no, not this. If `text` comes from the command line you should just be able to do `text.encode('utf-8')`. You' won't even have to reference the specific type (bytes/str).

jonparrott:

PR #906, opened 2017-04-19
name change for Google Translation API

merla18:
updated product name to Google Translation API in Readme


jonparrott:
@merla18 The readmes are generated, you'll need to update readme.rst.in instead.

PR #917, opened 2017-04-27
Changed the endpoint from v1beta1 to v1

rhaertel80:


jonparrott:
@elibixby can you review? I think we did this intentionally.

elibixby:
@rhaertel80 Do I need to redeploy the models as v1 models in order to send prediction requests to the v1 endpoint (I feel like I should know this)

puneith:
Why should the model need to be re-deployed? I am looking forward to @rhaertel80 answer on this.

rhaertel80:
Models deployed to v1beta1 (which has been officially deprecated and will be gone in < 2 months) are available on v1 under most circumstances, in particular if they are a SavedModel.

Given the deprecation status, we do need users to start migrating to v1 and this change is one step towards that.

elibixby:
@rhaertel80 I'm going to test this locally and then I'll merge.

PR #919, opened 2017-04-28
Add bucket-level IAM samples

jonparrott:


storage/cloud-client/iam.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+def view_bucket_iam_members(bucket_name):
+    storage_client = storage.Client()
+    bucket = storage_client.bucket(bucket_name)
+
+    policy = bucket.get_iam_policy()
+
+    for role in policy:
+        members = policy[role]
```
andrewsg: Why iterate over the keys and then do a dictionary lookup, instead of iterating over key, value using items()?

jonparrott: Because it's not a dictionary, see https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/core/google/cloud/iam.py#L64

andrewsg: Oh, ick.

If beginners like me are likely to read this code, maybe adding a comment about that here would prevent confusion. Up to you though, I don't have a lot of context yet.

storage/cloud-client/iam.py
```diff
@@ -0,0 +1,97 @@
...
+    bucket = storage_client.bucket(bucket_name)
+
+    policy = bucket.get_iam_policy()
+
+    policy[role].discard(member)
+
+    bucket.set_iam_policy(policy)
+
+    print('Removed {} with role {} from {}.'.format(
+        role, member, bucket_name))
```
ryanmats: should the order be 'member, role, bucket_name' instead ?

storage/cloud-client/iam.py
```diff
@@ -0,0 +1,97 @@
...
+    bucket = storage_client.bucket(bucket_name)
+
+    policy = bucket.get_iam_policy()
+
+    policy[role].add(member)
+
+    bucket.set_iam_policy(policy)
+
+    print('Added {} with role {} to {}.'.format(
+        role, member, bucket_name))
```
ryanmats: should the order be 'member, role, bucket_name' instead ?

jonparrott: Good catch.

jonparrott:
Failures unrelated (and reported to client library team). Merging.

PR #928, opened 2017-05-03
Update spanner sample to use bind parameters

jonparrott:
/cc @vkedia

### DIFF
--- a/spanner/cloud-client/snippets.py
+++ b/spanner/cloud-client/snippets.py
@@ -152,7 +152,8 @@ def add_index(instance_id, database_id):
     print('Added the AlbumsByAlbumTitle index.')
 
 
-def query_data_with_index(instance_id, database_id):
+def query_data_with_index(
+        instance_id, database_id, start_title='Aardvark', end_title='Goo'):
     """Queries sample data from the database using SQL and an index.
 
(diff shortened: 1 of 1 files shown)


spanner/cloud-client/snippets.py
```diff
@@ -413,7 +425,12 @@ def read_only_transaction(instance_id, database_id):
     subparsers.add_parser(
         'read_only_transaction', help=read_only_transaction.__doc__)
     subparsers.add_parser('add_index', help=add_index.__doc__)
-    subparsers.add_parser('query_data_with_index', help=insert_data.__doc__)
+    query_data_with_index_parser = subparsers.add_parser(
+        'query_data_with_index', help=query_data_with_index.__doc__)
+    query_data_with_index_parser.add_argument(
+        'start_title', default='Aardvark')
```
andrewsg: Why such an arbitrary default, and why specify a default here but also a keyword arg default in the function itself?

jonparrott: > Why such an arbitrary default

It was in the original sample decided on by the API team.

> and why specify a default here but also a keyword arg default in the function itself?

So that the default appears in the docs.

PR #930, opened 2017-05-04
Adds tutorials using Cloud Client

gguuss:
Existing tutorials use the Google API client library inline, this PR adds Cloud Client-based sources to replace them.


video/cloud-client/labels/labels.py
```diff
@@ -0,0 +1,80 @@
...
+        sys.stdout.flush()
+        time.sleep(20)
+
+    print('\nFinished processing.')
+    # [END check_operation]
+
+    # [START parse_response]
+    results = operation.result().annotation_results[0]
+
+    for i, label in enumerate(results.label_annotations):
```
jonparrott: Is the enumeration needed?

gguuss: It can be removed, at some point I may have used it to list the count the labels.

jonparrott: Your call.

PR #931, opened 2017-05-04
Create openapi-appengine.yaml

p-buse:


jonparrott:
@p-buse can I get some context?

p-buse:
@jonparrott Endpoints APIs that use App Engine as their backend need a different format for their _host_ field than other backends. This PR makes the code follow the quickstart guide for Endpoints on App Engine  more closely.

jonparrott:
Could this not be satisfied with a comment in the host field in the regular openapi.yaml file?

p-buse:
@jonparrott You're right. I'll do that instead.

PR #943, opened 2017-05-13
Add Bigtable Autoscaler sample

waprin:


waprin:
@rameshdharan thinking this sample is better here

bigtable/autoscaler/strategies_test.py
```diff
@@ -0,0 +1,30 @@
...
+
+"""Tests for autoscaler.py"""
+
+from strategies import DOWNSCALE_STRATEGIES
+from strategies import UPSCALE_STRATEGIES
+
+
+def test_downscale():
+    downscale_strategy = DOWNSCALE_STRATEGIES['incremental']
+    assert downscale_strategy(5) == 3
```
mbrukman: How do you run these tests without using the `unittest` module? Are they imported somewhere into a main file which runs all the tests?

waprin: Our main test runner is [pytest](https://docs.pytest.org/en/latest/) which scans for functions that start with `test_` and handles the Python asserts in a way similar to how unittest would the assert functions, so it's pretty neat.

If you do:

```
pip install pytest
pytest
```

it will run all the tests.

In this specific repo, since there's dozens of samples ,we use [nox](https://github.com/jonparrott/nox) which helps instrument different sessions etc. But really in here it will just call pytest.

bigtable/autoscaler/autoscaler.py
```diff
@@ -0,0 +1,134 @@
...
+import argparse
+import os
+import time
+
+from google.cloud import bigtable
+from google.cloud import monitoring
+
+import strategies
+
+CPU_METRIC = 'bigtable.googleapis.com/cluster/cpu_load'
```
jonparrott: while constants are good practice in production code, in sample code they add a layer of indirection that can be annoying when including a snippet in cloud site. Strongly prefer this to go near where it's used.

bigtable/autoscaler/autoscaler.py
```diff
@@ -0,0 +1,134 @@
...
+
+def get_cpu_load():
+    """Returns the most recent Bigtable CPU load measurement.
+
+    Returns:
+          float: The most recent Bigtable CPU usage metric
+    """
+    client = monitoring.Client()
+    query = client.query(CPU_METRIC, minutes=5)
+    return list(query)[0].points[0].value
```
jonparrott: For sample code, prefer pulling out discrete values and giving them names:

```python
results = client.query(CPU_METRIC, minutes=5)
results = list(query)
first_result = results[0]
cpu_load = first_result.points[0].value
```

Also, shouldn't that query have a limit of 1?

bigtable/autoscaler/autoscaler.py
```diff
@@ -0,0 +1,134 @@
...
+
+
+def scale_bigtable(bigtable_instance, up):
+    """Scales the number of Bigtable nodes up or down.
+
+    Args:
+           bigtable_instance (str): Cloud Bigtable instance id to scale
+           up (bool): If true, scale up, otherwise scale down
```
jonparrott: I'd recommend `scale_up` as a more more descriptive name. Then, you can probably drop the args section and just describe the behavior in the description.

bigtable/autoscaler/autoscaler.py
```diff
@@ -0,0 +1,134 @@
...
+             'scale operation',
+        default=60)
+    parser.add_argument(
+        '--long_sleep',
+        help='How long to sleep in seconds between checking metrics after a '
+             'scaling operation',
+        default=60 * 10)
+    args = parser.parse_args()
+
+    while True:
```
jonparrott: move the `while` to `main`.

bigtable/autoscaler/strategies.py
```diff
@@ -0,0 +1,50 @@
+# Copyright 2017 Google Inc.
```
jonparrott: Any reason to have this as a separate file? For the sake of samples, it seems it would be most useful to locate this near the code that uses these strategies.

bigtable/autoscaler/autoscaler.py
waprin: I had it in main, but I moved it out here to make it easier to test main. Otherwise testing `while True` loops is annoying. Would probably have to make a bunch of changes just to test the code, this seemed like the simpler option.

bigtable/autoscaler/strategies.py
waprin: The idea is we are going to add a few alternative strategies such as exponential, and have them all organized in here.

bigtable/autoscaler/autoscaler.py
waprin: `scale_up` seems misleading in the case of downscale.

bigtable/autoscaler/autoscaler.py
waprin: Done. API doesn't support limits on queries.

bigtable/autoscaler/autoscaler.py
waprin: done

bigtable/autoscaler/autoscaler.py
jonparrott: And `up` is less misleading?

bigtable/autoscaler/autoscaler.py
jonparrott: The tests shouldn't influence flow control. We can figure out how to break out of the loop. Likely by using mock to insert a keyboard interrupt.

bigtable/autoscaler/strategies.py
jonparrott: I see. If that's valuable and matches with how this will be presented in the docs, then that's fine.

bigtable/autoscaler/autoscaler.py
waprin: Sorry, misunderstood, thought you meant the function name. Done.

bigtable/autoscaler/strategies.py
waprin: Ya doc is still a WIP so once we have a draft of that I'll take another look at this and if it's pointlessly complicating things I'll move it back into `autoscaler.py`.

jonparrott:
@waprin why was this merged? There was at least [one outstanding comment](https://github.com/GoogleCloudPlatform/python-docs-samples/pull/943#discussion_r116534341) that I wanted to resolve before we merged.

I'm usually fine with compromises in the near term as long as we create bugs to follow up, but dismissing my review and merging does not seem like our usual mode of operation. What's the rush to merge this?

PR #969, opened 2017-05-31
Add samples for auth

jonparrott:


auth/api-client/snippets.py
```diff
@@ -0,0 +1,66 @@
...
+def implicit(project):
+    import googleapiclient.discovery
+
+    # If you don't specify credentials when constructing the client, the
+    # client library will look for credentials in the environment.
+    storage_client = googleapiclient.discovery.build('storage', 'v1')
+
+    # Make an authenticated API request
+    buckets = storage_client.buckets().list(project=project).execute()
+    print(buckets)
```
andrewsg: Normally I'd "from __future__ import print_function" in a file that uses print(), even if this technically works in 2.7 as-is. Do we have an explicit policy on this?

auth/api-client/snippets_test.py
```diff
@@ -0,0 +1,33 @@
...
+# limitations under the License.
+
+import os
+
+import mock
+
+import snippets
+
+
+def test_implicit():
```
andrewsg: I am confused about the structure of these tests; is there part of this (that does assertions etc) I'm missing?

auth/http-client/snippets.py
```diff
@@ -0,0 +1,79 @@
...
+
+"""Demonstrates how to authenticate to Google Cloud Platform APIs using the
+Requests HTTP library."""
+
+import argparse
+
+
+def implicit():
+    import google.auth
+    from google.auth.transport import requests
```
andrewsg: Why do we do these imports inside the function? And why do we have a vendored requests version?

auth/api-client/snippets.py
jonparrott: We don't, and generally because we use `.format` is isn't ever explicitly needed to use `print_function`.

auth/api-client/snippets_test.py
jonparrott: It's a system test, in this case an absence of an error is the assertion. We could technically capture the output and assert that it matches some idea of what we think the output should be, but it doesn't necessarily add any value.

auth/http-client/snippets.py
jonparrott: Because when these are included in the documentation it will only include the body of this function.

auth/api-client/snippets.py
andrewsg: Okay, my preference is to include the import because print with parens is confusing syntax in Py2, but I'll leave it to you.

auth/http-client/snippets.py
andrewsg: Okay, this strikes me as a bit confusing when you're reading the actual file, but not a big deal.

PR #974, opened 2017-06-02
Refactor HTML response code

dbarella:
Make the response writing more readable by pulling the HTML into a format string.


jonparrott:
One of our guiding principles for samples internally is to `Prefer less indirection`. This change introduces indirection. Is there any way you can improve the readability in a way that you like without introducing indirection?

dbarella:
Sure, check this version out!

PR #979, opened 2017-06-12
make_iap_request: Replace base URL with client ID

matthewg:
Client ID is the preferred thing to use for the aud claim.
  --msachs@, IAP TLM


matthewg:
I've now updated the test as well.  (I got the client ID by navigating to https://gcp-devrel-iap-reflect.appspot.com/ -- the client ID can be pulled from the resulting redirect URL.)  I don't have permission to merge, would you mind doing it?

PR #981, opened 2017-06-12
Add more auth samples

jonparrott:
Add samples for explicitly using compute and app engine credentials.

### DIFF
--- a/auth/cloud-client/snippets_test.py
+++ b/auth/cloud-client/snippets_test.py
@@ -14,6 +14,7 @@
 
 import os
 
+import google.auth
 import mock
 
 import snippets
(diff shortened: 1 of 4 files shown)


auth/cloud-client/snippets_test.py
```diff
@@ -31,3 +32,12 @@ def test_explicit():
 
     with mock.patch('io.open', open_mock):
         snippets.explicit()
+
+
+def test_explicit_compute_engine():
+    adc, project = google.auth.default()
+    credentials_patch = mock.patch(
```
andrewsg: Why assign this to a variable and then use it as a context manager instead of inlining it with the "with" as in the test above?

jonparrott: Because splitting context manager syntax over multiple lines is *extremely* awkward, so we've gotten into the habit of giving the patch manager its own variable.

PR #982, opened 2017-06-13
Add bigtable regions

waprin:


jonparrott:
Fix lint and you're good to go.

Feel free to FYI changes that only involve region tags in the future as long as they pass CI.

PR #984, opened 2017-06-14
[Storage] Fix for transfer storage sample

frankyn:
The sample positions of date parameters (year, month, day) aren't in the correct order that's expected.


frankyn:
Thanks Jon!

PR #987, opened 2017-06-15
Fixes broken example in docs

gguuss:


jonparrott:
@dpebot will you merge when Travis passes?

PR #988, opened 2017-06-15
Add query string authentication to Endpoints clients

p-buse:
Some of our users [have asked about a way to authenticate to the Endpoints proxy with a JWT in the query string](https://groups.google.com/forum/#!msg/google-cloud-endpoints/KquTCjR2TIc/YxTznNkKAAAJ) of the URL. This PR updates the Endpoints clients to expose that feature.

### DIFF
--- a/appengine/standard/ndb/overview/main.py
+++ b/appengine/standard/ndb/overview/main.py
@@ -22,6 +22,7 @@ For more information, see README.md
 
 # [START all]
 import cgi
+import textwrap
 import urllib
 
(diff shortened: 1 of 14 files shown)


jonparrott:
Two things:

1. I'd prefer a separate sample rather than complicating the existing samples.
2. Is this not dangerous?

p-buse:
It is dangerous. We're going to update the documentation to say that it's possible, but that header-based authentication is more secure.

PR #1000, opened 2017-06-25
sentiment.score -> sentiment.polarity ?

vochicong:
I modified this file a little so that the [tutorial](https://cloud.google.com/natural-language/docs/reference/libraries#client-libraries-usage-python) worked.


vochicong:
After `pip install --upgrade google-cloud-language` and `pip3 install --upgrade google-cloud-language` the problem was resolved.

PR #1002, opened 2017-06-26
Remove oauth2client usage in endpoints samples

jonparrott:


endpoints/getting-started/clients/google-id-token-client.py
```diff
@@ -19,25 +19,24 @@
...
-import oauth2client.tools
+import google_auth_oauthlib.flow
 import requests
 from six.moves import urllib
 
 
 def get_id_token(client_secrets_file, extra_args):
-    storage = oauth2client.file.Storage('credentials.dat')
-    credentials = storage.get()
+    """Obtains credentials form the user using OAuth 2.0 and then returns the
```
andrewsg: typo here (should be "from")

endpoints/getting-started/clients/google-id-token-client.py
```diff
@@ -19,25 +19,24 @@
 
 import argparse
 
-import oauth2client.client
-import oauth2client.file
-import oauth2client.tools
+import google_auth_oauthlib.flow
```
andrewsg: This seems to be accurate, but just out of curiosity, why is google.auth a thing and google_auth also a thing?

endpoints/getting-started/clients/google-id-token-client.py
jonparrott: Good catch. :)

endpoints/getting-started/clients/google-id-token-client.py
jonparrott: Two reasons:

1. We wanted to get google.auth to a v1.0.0 interface as quickly as possible to avoid thrashing. To do this, we [descoped several contentious features from the library](https://github.com/GoogleCloudPlatform/google-auth-library-python/issues/3). We split the "User Auth" featureset into a separate library, `google-auth-oauthlib`, in order to be able to iterate without worrying about breaking the `v1.0.0` interface of `google-auth`. Once that separate library has a solid interface, we can merge it into `google-auth` as a `1.x` feature.
2. We didn't want to make `google.auth` a namespace package because when we *do* promote the auxiliary packages we do not break existing code using the separate version. So the auxiliary packages like `google-auth-oauthlib` must use a new global top-level namespace, in this case `google_auth_oauthlib`.

viktorLisniak:
generated jwt worked when i added "scope":"https://www.googleapis.com/auth/admin.directory..." to payload.
maybe it should be there from the beginning?

PR #1013, opened 2017-07-05
Speech sample using microphone

jerjou:


speech/cloud-client/transcribe_streaming_mic_test.py
```diff
@@ -0,0 +1,69 @@
+# Copyright 2016, Google, Inc.
```
gguuss: Copyright date on new file should be 2017

PR #1015, opened 2017-07-10
Vision GAPIC client library

dizcology:
Updating the sample codes to use the GAPIC client library.

Merge only after the client library has been released.

### DIFF
--- a/vision/cloud-client/detect/detect.py
+++ b/vision/cloud-client/detect/detect.py
@@ -33,17 +33,17 @@ import io
 
 from google.cloud.vision import ImageAnnotatorClient
 from google.cloud.vision import types
-from google.cloud.vision import enums
 
 # Names of likelihood from google.cloud.vision.enums
 likelihood_name = ('UNKNOWN', 'VERY_UNLIKELY', 'UNLIKELY', 'POSSIBLE',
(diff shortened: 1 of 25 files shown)


vision/cloud-client/detect/detect.py
```diff
@@ -32,60 +32,77 @@
...
+
+# Names of likelihood from google.cloud.vision.enums
+likelihood_name = ('UNKNOWN', 'VERY_UNLIKELY', 'UNLIKELY', 'POSSIBLE',
+                   'LIKELY', 'VERY_LIKELY')
 
 
 def detect_faces(path):
     """Detects faces in an image."""
-    vision_client = vision.Client()
+    # Names of likelihood from google.cloud.vision.enums
```
jonparrott: Why have this global and local? pick one?

jonparrott:
This looks good, but you need to resolve the conflicts first.

vision/cloud-client/detect/detect.py
dizcology: updated.  keeping the local ones.

dizcology:
rebased and incremented client library version.

merge only after the client library is published and tested.

dizcology:
client library was published and tested against these updated samples.

PR #1018, opened 2017-07-12
Natural Language GAPIC client library

dizcology:
migrate samples to natural language gapic client library

Merge only after the client library has been released and tested.


language/cloud-client/v1beta2/snippets.py
```diff
@@ -175,17 +177,15 @@ def entity_sentiment_text(text):
...
-    language_client = language_service_client.LanguageServiceClient()
-    document = language_service_pb2.Document()
+    client = language_v1beta2.LanguageServiceClient()
 
-    document.gcs_content_uri = gcs_uri
-    document.type = enums.Document.Type.PLAIN_TEXT
+    document = types.Document(gcs_content_uri=gcs_uri, type='PLAIN_TEXT')
 
-    encoding = enums.EncodingType.UTF32
+    encoding = 'UTF32'
```
jonparrott: can you not use the enums here (should be `types.EncodingType.UTF32`?

jonparrott:
This LGTM, but you need to resolve the conflicts first.

language/cloud-client/v1beta2/snippets.py
dizcology: removing encoding, since it is no longer needed:  https://github.com/googleapis/googleapis/pull/408

dizcology: turns out we should still keep the encoding.

The client libraries are current designed to have all enums in the enums module, such as `google.cloud.language.enums`.

dizcology:
client library version 0.26.1 released, and tested against the updated samples.

PR #1019, opened 2017-07-13
Update validate_jwt to use new JWT header

matthewg:


iap/validate_jwt.py
```diff
@@ -27,18 +27,41 @@
...
 
+    One and only of of (app_engine_project_id, compute_engine_backend_service_id)
+    must be set.
+
     Args:
-      base_url: The URL from the incoming request, minus any path, query, etc.
-                For instance: "https://example.com:8443" or
-                "https://example.appspot.com" .
-      iap_jwt: The contents of the X-Goog-Authenticated-User-JWT header.
+      cloud_project_number: The project *number* for your Google Cloud project.
```
jonparrott: Oh no, project number? That's awful! Why is that needed instead of just the project ID?

iap/validate_jwt.py
```diff
@@ -27,18 +27,41 @@
...
+      compute_engine_backend_service_id: For Compute Engine and Container Engine
+          resources this must be set to the ID of the backend service used to
+          access the application. See
+          https://cloud.google.com/iap/docs/signed-headers-howto for details on
+          how to get this value.
 
     Returns:
       (user_id, user_email, error_str).
     """
+    if not (bool(app_engine_project_id) ^ bool(compute_engine_backend_service_id)):
```
jonparrott: This is a code smell, can you maybe split this into a couple of functions like `validate_iap_jwt_from_compute` that call a common helper with a `target_audience` or alternatively have one helper method that figures out the target audience and one to validate the jwt?

iap/validate_jwt.py
matthewg: The path through Google infrastructure that leads to calling the IAP auth component doesn't propagate the project ID. We _could_ look it up in the server, but then that'd be a latency and reliability hit on every single request. Yeah, agree that it's not the user experience we'd like to have.

iap/validate_jwt.py
matthewg: c4ea93c6a6bb96620281e67ccea559c8ee636698

iap/validate_jwt.py
```diff
@@ -27,18 +27,47 @@
...
 
     Returns:
       (user_id, user_email, error_str).
     """
+    expected_audience = '/projects/{}/apps/{}'.format(
+        cloud_project_number, cloud_project_id)
+    return _validate_iwp_jwt(iap_jwt, expected_audience)
+    
+
+def validate_iap_jwt_from_app_engine(iwp_jwt, cloud_project_number,
```
jonparrott: This says `app_engine`, did you mean `compute_engine`?

jonparrott:
@matthewg you'll need to placate travis: https://travis-ci.org/GoogleCloudPlatform/python-docs-samples/builds/253681888?utm_source=github_status&utm_medium=notification

matthewg:
Travis should be happy now.  I've updated the test, but you'll need to redeploy gcp-devrel-iap-reflect to pick up the new header name.

jonparrott:
Will do, thanks. :)

PR #1020, opened 2017-07-14
module -> service in app.yamls

### LINKED ISSUE #1016: Many .yaml files are using 'module', 'service' should now be used
https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/appengine/standard/taskqueue/counter/app.yaml#L4
https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/appengine/standard/taskqueue/counter/worker.yaml#L4

As we can read here: https://cloud.google.com/appengine/docs/standard/python/config/appref
Note: Modules are now named Services.

I suggest to use 'service' in all .yaml files to avoid confusion. Our docs are talking about Services, we should be coherent.

jonparrott:
Resolves #1016

### DIFF
--- a/appengine/standard/channel/app.yaml
+++ b/appengine/standard/channel/app.yaml
@@ -1,4 +1,4 @@
-module: tictactoe
+service: tictactoe
 runtime: python27
 version: 1
 api_version: 1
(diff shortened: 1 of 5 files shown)


jonparrott:

PR #1025, opened 2017-07-17
Fork cloud sql sample into a separate postgres sample

jonparrott:


appengine/flexible/cloudsql_postgresql/app.yaml
```diff
@@ -0,0 +1,21 @@
...
+    # when configuring your Cloud SQL instance.
+    SQLALCHEMY_DATABASE_URI: >-
+      postgresql+psycopg2://USER:PASSWORD@/DATABASE?host=/cloudsql/INSTANCE_CONNECTION_NAME
+#[END env]
+
+#[START cloudsql_settings]
+# Replace project and instance with the values obtained  when configuring your
+# Cloud SQL instance.
+beta_settings:
+    cloud_sql_instances: INSTANCE_CONNECTION_NAME
```
andrewsg: Why do we need both this beta_setting and the SQLALCHEMY_DATABASE_URI? Networking or something?

appengine/flexible/cloudsql_postgresql/main_test.py
```diff
@@ -0,0 +1,26 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import main
+
+
+def test_index():
+    main.db.create_all()
```
andrewsg: Do we have any way to safely tear these test tables down afterwards?

appengine/flexible/cloudsql_postgresql/app.yaml
jonparrott: the beta setting actually tells the cloud sql proxy running on your instance to create the socket for the cloud sql instance.

appengine/flexible/cloudsql_postgresql/main_test.py
jonparrott: We do, but it's not needed.

PR #1029, opened 2017-07-21
Properly forwarding the "region" parameter provided as an input argument in list_cluster.py

gioiab:
The `region` parameter is asked to the user but never used. This PR aims to properly forward the region argument to the other functions when this is provided and to set `global` as a default when it isn't provided. The `README` has been updated consequently to match with the required and optional arguments actually asked via `argparse`.


jonparrott:
Thank you, @gioiab. :)

PR #1034, opened 2017-07-27
Update settings.py

mzaforas:
ALLOWED_HOSTS is also check when DEBUG=True after Django 1.10.3, 1.9.11, and 1.8.16 to prevent a DNS rebinding attack. https://docs.djangoproject.com/en/1.11/ref/settings/#allowed-hosts


andrewsg:
Thanks for catching this, @mzaforas! I know it seems silly given the size of the commit, but could I ask you to sign the CLA as well? https://cla.developers.google.com/

PR #1041, opened 2017-07-31
Fix SQL query to return 10 rows & simplify

PicardParis:
- Original query returned a single nested row (instead of 10 rows as apparently expected in section print_results)
- Standard SQL specified directly in query
- Removed parameter max_results to avoid redundancy w/ query

### DIFF
--- a/bigquery/cloud-client/simple_app.py
+++ b/bigquery/cloud-client/simple_app.py
@@ -24,21 +24,20 @@ def query_shakespeare():
     client = bigquery.Client()
     # [END create_client]
     # [START run_query]
-    query_results = client.run_sync_query("""
-        SELECT
-            APPROX_TOP_COUNT(corpus, 10) as title,
-            COUNT(*) as unique_words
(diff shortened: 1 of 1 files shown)


PicardParis:
Added my corporate email

jonparrott:
@tswast @PicardParis This caused the [system tests to fail](https://travis-ci.org/GoogleCloudPlatform/python-docs-samples/builds/259529804). Can one of you send a PR to update the tests?

PicardParis:
Proposed change still gets a PASSED:
`simple_app_test.py::test_query_shakespeare PASSED`

Failures are related to 2 other tests.
I think I see why. Let me check...

PicardParis:
The 2 failures are in fact due to random queries in async_query_test.py and sync_query_test.py.

Proposed fixes in PRs:
https://github.com/GoogleCloudPlatform/python-docs-samples/pull/1043
https://github.com/GoogleCloudPlatform/python-docs-samples/pull/1044

jonparrott:
@PicardParis thanks!

PR #1046, opened 2017-08-01
enable_word_time_offsets=False to a few examples, and added enable_wor…

dlaqab:
Added enable_word_time_offsets=True to local file version of transcribe_async
Added enable_word_time_offsets=False to other recognize samples (except streaming)

### DIFF
--- a/speech/cloud-client/quickstart.py
+++ b/speech/cloud-client/quickstart.py
@@ -46,7 +46,8 @@ def run_quickstart():
     config = types.RecognitionConfig(
         encoding=enums.RecognitionConfig.AudioEncoding.LINEAR16,
         sample_rate_hertz=16000,
-        language_code='en-US')
+        language_code='en-US',
+        enable_word_time_offsets=False)
 
     # Detects speech in the audio file
     response = client.recognize(config, audio)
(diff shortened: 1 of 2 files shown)


dlaqab:
I am covered by employment agreement

jonparrott:
@dlaqab you need to register your github username at go/github, and make sure your @google.com address is associated with your github profile.

speech/cloud-client/quickstart.py
```diff
@@ -46,7 +46,8 @@ def run_quickstart():
     config = types.RecognitionConfig(
         encoding=enums.RecognitionConfig.AudioEncoding.LINEAR16,
         sample_rate_hertz=16000,
-        language_code='en-US')
+        language_code='en-US',
+        enable_word_time_offsets=False)
```
jonparrott: If this defaults to false, then we should just leave it out.

dizcology: Agreeing with @jonparrott on this:  This is definitely something we want to make clear in the API/client library reference, but adding this in the sample codes would make the API seem harder to use than it actually is (note that the other parameters `encoding`, `langauge_code`, and `sample_rate_hertz` are all required, but `enable_word_time_offsets` is optional and defaults to false).

What are some possible use case scenarios where not showing `enable_word_time_offsets=False` could cause friction for the users?

dlaqab:
@jonparrott thanks for the pointer - Done

remi:
The enable_word_time_offsets=False parameter in these samples is extraneous to the task these samples are intended to demonstrate. Closing.

remi:
Note: samples which demonstrate Enable Word Time Offsets were added to java-docs-samples in #1050

PR #1050, opened 2017-08-03
Add word time offset samples

jonparrott:


speech/cloud-client/transcribe_word_time_offsets.py
```diff
@@ -0,0 +1,111 @@
...
+            word = word_info.word
+            start_time = word_info.start_time
+            end_time = word_info.end_time
+            print('Word: {}, start_time: {}, end_time: {}'.format(
+                word,
+                start_time.seconds + start_time.nanos * 1e-9,
+                end_time.seconds + end_time.nanos * 1e-9))
+
+
+# [START def_transcribe_gcs]
```
remi: The `transcribe_file_with_word_time_offsets` function doesn't use a region tag but this does. Ignore comment if that's how Python often does samples!

If using region tags, the GCS tag is: `speech_async_recognize_gcs_words`

Again, if you normally use `def_...` region tags, then all good!

---

I'm guessing that we should embed these snippets using indented region for  `transcribe_file_with_word_time_offsets`  but region rag for `def_transcribe_gcs`?

speech/cloud-client/README.rst
```diff
@@ -140,6 +140,32 @@ To run this sample:
...
+
+    $ python transcribe_word_time_offsets.py
+
+    usage: transcribe_word_time_offsets.py [-h] path
+    
+    Google Cloud Speech API sample that demonstrates word time offsets.
+    
+    Example usage:
+        python transcribe_word_time_offsets.py resources/audio.raw
+        python transcribe_word_time_offsets.py         gs://cloud-samples-tests/speech/vr.flac
```
remi: Optional/Nit: the sample hardcodes LINEAR16 so Flac audio won't work.

(This has come up when people have tried to use the sample)

speech/cloud-client/transcribe_word_time_offsets.py
dizcology: If the ranges of indented blocks and region tags overlap then sometimes the code blocks would be incorrectly displayed in the include code widget.

jonparrott: `indented_block="def function_name"` is the preferred way for Python samples if the entire body of the function will be used.

jonparrott: I don't see how that would be an issue for these samples.

speech/cloud-client/README.rst
jonparrott: The tests use the flac audio and pass. No idea why that's the case, but this seems like a non-issue.

PR #1053, opened 2017-08-04
Fix region handling and allow to use an existing cluster.

gioiab:
This PR is related to [PR#1029](https://github.com/GoogleCloudPlatform/python-docs-samples/pull/1029): forwarding the region `global` to all the possible operations (list clusters, create clusters,...) doesn't work anymore. Since the `zone` parameter is required, the region is inferred from the `zone` and properly forwarded to all the dependent functions. This change also adds the chance to use directly an existing cluster when the `--create_new_cluster` flag is missing. Given the option of using an existing cluster, the script `create_cluster_and_submit_job.py` has been renamed to `submit_job_to_cluster.py`.

### DIFF
--- a/dataproc/README.md
+++ b/dataproc/README.md
@@ -35,14 +35,19 @@ To run list_clusters.py:
     python list_clusters.py <YOUR-PROJECT-ID> --region=us-central1
 
 
-To run create_cluster_and_submit_job, first create a GCS bucket, from the Cloud Console or with
+To run submit_job_to_cluster.py, first create a GCS bucket, from the Cloud Console or with
 gsutil:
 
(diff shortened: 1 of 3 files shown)


jonparrott:
@waprin can you take a look?

waprin:
Wow huge thanks for this nice PR! I want to take time to make sure it all works and we're running that tests on our e2e server so might take a bit to merge this but will aim for end of day.

waprin:
I am a bit confused as to the exact problem this is fixign as the previous sample seemed to work, but the change makes some sense to me and you seem like you've given it some thought, it still works for me so I'm just going to merge it.

PR #1062, opened 2017-08-08
show entity type name

dizcology:
This PR updates the analyze entities samples to show the names of entity types (as opposed to just the enums).

before:

```
name            : Golden Gate Bridge
type            : 2
```

after:

```
name            : Golden Gate Bridge
type            : LOCATION
```


jonparrott:
LGTM, but you need to placate Travis.

dizcology:
My bad - updated.

PR #1063, opened 2017-08-08
Pass multi_use=True to spanner read-only transaction

jonparrott:


jonparrott:
Merging as our system tests that exercise this sample now pass.

PR #1075, opened 2017-08-16
KMS: Make the encrypt/decrypt samples consistent with the other languages

russ-:
In particular, this change:
- changes to use a consistent "_id" suffix for resource name components
- uses "data" instead of "text", since encryption is not restricted to text
- substitutes "ciphertext" for "encrypted {text, data}"
- cases "cryptoKey" and "keyRing" consistently with the API

Tracking bug: http://b/64758639

### DIFF
--- a/kms/api-client/snippets.py
+++ b/kms/api-client/snippets.py
@@ -21,18 +21,18 @@ import googleapiclient.discovery
 
 
 # [START kms_create_keyring]
-def create_keyring(project_id, location, keyring):
+def create_key_ring(project_id, location_id, key_ring_id):
     """Creates a KeyRing in the given location (e.g. global)."""
 
     # Creates an API client for the KMS API.
     kms_client = googleapiclient.discovery.build('cloudkms', 'v1')
 
     # The resource name of the location associated with the KeyRing.
-    parent = 'projects/{}/locations/{}'.format(project_id, location)
+    parent = 'projects/{}/locations/{}'.format(project_id, location_id)
 
     # Create KeyRing
     request = kms_client.projects().locations().keyRings().create(
-        parent=parent, body={}, keyRingId=keyring)
+        parent=parent, body={}, keyRingId=key_ring_id)
     response = request.execute()
 
     print('Created KeyRing {}.'.format(response['name']))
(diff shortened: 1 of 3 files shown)


kms/api-client/snippets.py
```diff
@@ -61,70 +61,70 @@ def create_cryptokey(project_id, location, keyring, cryptokey):
...
-    # Read text from the input file.
+    # Read data from the input file.
     with io.open(plaintext_file_name, 'rb') as plaintext_file:
         plaintext = plaintext_file.read()
 
-    # Use the KMS API to encrypt the text.
-    cryptokeys = kms_client.projects().locations().keyRings().cryptoKeys()
-    request = cryptokeys.encrypt(
+    # Use the KMS API to encrypt the data.
+    cryptoKeys = kms_client.projects().locations().keyRings().cryptoKeys()
```
jonparrott: our Python style guide is lower snake case for variable names. I understand the desire to match the API, but we've stuck to using lower snake case consistency for variables regardless of the API.

kms/api-client/snippets.py
```diff
@@ -61,70 +61,70 @@ def create_cryptokey(project_id, location, keyring, cryptokey):
 
 
 # [START kms_encrypt]
-def encrypt(project_id, location, keyring, cryptokey, plaintext_file_name,
-            encrypted_file_name):
-    """Encrypts data from a plaintext_file_name using the provided CryptoKey
-    and saves it to an encrypted_file_name so it can only be recovered with a
-    call to decrypt."""
+def encrypt(project_id, location_id, key_ring_id, crypto_key_id,
```
jonparrott: Since you are changing variable names, can you update the other samples in this file to match?

russ-: Good catch, all fixed now.

russ-: Good catch. Fixed.

kms/api-client/snippets.py
russ-: Fixed. Sorry, I didn't intend to go against the style guide, just an oversight on my part.

PR #1079, opened 2017-08-17
Adding sample for making requests to IAP using a service account with user-managed keys

danieldeleo:


danieldeleo:
All commits have been rebased with Google author

PR #1082, opened 2017-08-17
Adding sample for making IAP request using service account with user-managed keys

danieldeleo:


jonparrott:
@danieldeleo thank you for doing this, but this sample already supports using a private key file: https://github.com/googlecloudplatform/python-docs-samples/tree/master/iap#using-a-downloaded-service-account-private-key

How does this change differ?

danieldeleo:
@jonparrott I completely overlooked that env variable! I'll close this request. Thanks!

PR #1085, opened 2017-08-18
Show the need for dealing with network connection issues.

Galabar001:
The current documentation for accessing the metadata server does
not show how to deal with network issues.  It is possible to see
a "connection reset by peer."  This change explicitly calls that
out.

### DIFF
--- a/compute/metadata/main.py
+++ b/compute/metadata/main.py
@@ -38,15 +38,15 @@ def wait_for_maintenance(callback):
 
     while True:
         try:
-          r = requests.get(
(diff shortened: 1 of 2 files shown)


Galabar001:
The CLA has already been signed (back in 2014).

jonparrott:
Thank you for this change, however, we generally do not exhaustively deal with error conditions unless the sample is directly intended to show how to deal with errors. In this case, it's best to allow requests to do its own retrying and surface the exception directly to the user.

PR #1086, opened 2017-08-21
BigQuery : max_results changes page size not full list size

tswast:
Also, running a query is not "preferred" if you do want all the data. If you do run a query, you end up reading data from the destination table anyway.

### DIFF
--- a/bigquery/cloud-client/snippets.py
+++ b/bigquery/cloud-client/snippets.py
@@ -125,9 +125,8 @@ def list_rows(dataset_name, table_name, project=None):
     # Reload the table so that the schema is available.
     table.reload()
 
-    # Load at most 25 results per page. You can change the max_results
-    # argument to load more rows from BigQuery at a time.
-    rows = list(itertools.islice(table.fetch_data(max_results=25), 25))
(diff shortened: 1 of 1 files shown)


jonparrott:
Can you just use `islice` without specifying `max_results`?

tswast:
Done.

bigquery/cloud-client/snippets.py
```diff
@@ -124,10 +125,8 @@ def list_rows(dataset_name, table_name, project=None):
     # Reload the table so that the schema is available.
     table.reload()
 
-    # Load at most 25 results. You can change the max_results argument to load
-    # more rows from BigQuery, but note that this can take some time. It's
-    # preferred to use a query.
-    rows = list(table.fetch_data(max_results=25))
+    # Load at most 25 results.
+    rows = list(itertools.islice(table.fetch_data(), 25))
```
jonparrott: since you're iterating through the rows below, you don't need to `list(...)`.

tswast: True. Done.

PR #1095, opened 2017-08-30
Language classify

dizcology:
Do not merge until the updated client library has been published and tested.


dizcology:
sample tests passed with the newest client library google-cloud-langage==0.29.0

PR #1098, opened 2017-08-31
Print all results, not all alternatives

jerjou:


speech/cloud-client/transcribe_streaming_mic.py
```diff
@@ -129,7 +129,9 @@ def listen_print_loop(responses):
         if not response.results:
             continue
 
-        # There could be multiple results in each response.
+        # The `results` list is consecutive. For streaming, we only care about
+        # the first result being considered, since once it's `is_final`, it
+        # moves on to considering the next utterance.
```
dizcology: Does this mean that there is always only one `result` in each `response` of `streaming_recognize`?

jerjou: a) What're you doing doing a code review on a holiday?? tsk tsk
b) Mm... if `interim_results` is `False`, then yes - there will only be one `result`. But if `interim_results` is `True`, then there can be 0 or one `is_final` result, followed by 0 to many interim (`is_final=False`) results. I believe different results correspond to consecutive parts of the audio stream. See [here](https://github.com/googleapis/googleapis/blob/master/google/cloud/speech/v1/cloud_speech.proto#L363)

PR #1101, opened 2017-09-07
BigQuery : add sample for writing query results to a destination table.

tswast:
See: https://cloud.google.com/bigquery/docs/writing-results

### DIFF
--- a/bigquery/cloud-client/query.py
+++ b/bigquery/cloud-client/query.py
@@ -63,6 +63,30 @@ def query_standard_sql(query):
         print(row)
 
 
+def query_destination_table(query, dest_dataset_id, dest_table_id):
+    client = bigquery.Client()
+    query_job = client.run_async_query(str(uuid.uuid4()), query)
+
+    # Allow for query results larger than the maximum response size.
+    query_job.allow_large_results = True
+
+    # When large results are allowed, a destination table must be set.
+    dest_dataset = client.dataset(dest_dataset_id)
+    dest_table = dest_dataset.table(dest_table_id)
+    query_job.destination = dest_table
+
+    # Allow the results table to be overwritten.
+    query_job.write_disposition = 'WRITE_TRUNCATE'
+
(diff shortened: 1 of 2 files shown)


bigquery/cloud-client/query.py
```diff
@@ -72,10 +96,19 @@ def query_standard_sql(query):
...
+        help=(
+            'Destination table to use for results. '
+            'Example: my_dataset.my_table'))
 
     args = parser.parse_args()
 
     if args.use_standard_sql:
         query_standard_sql(args.query)
+    elif args.destination_table:
+        dataset, table = args.destination_table.split('.')
```
davidcavazos: We could use a simple regular expression to validate that the destination table name is in the correct format, i.e. dataset_name.table_name

davidcavazos: However, this being a sample, it might be a better idea to keep unnecessary complexity down. So I'll leave that to you

davidcavazos:
Please do whatever you think best for the validation comment, otherwise LGTM

bigquery/cloud-client/query.py
tswast: I'd rather keep the complexity low.

bigquery/cloud-client/query.py
```diff
@@ -63,6 +63,30 @@ def query_standard_sql(query):
...
+
+    # When large results are allowed, a destination table must be set.
+    dest_dataset = client.dataset(dest_dataset_id)
+    dest_table = dest_dataset.table(dest_table_id)
+    query_job.destination = dest_table
+
+    # Allow the results table to be overwritten.
+    query_job.write_disposition = 'WRITE_TRUNCATE'
+
+    query_job.begin()
```
jonparrott: doesn't .result() call .begin()?

tswast: Oh, I didn't realize that. I can update the sample if so.

PR #1112, opened 2017-09-14
add another script to generate google id token.

qiwzhang:
Add another sample python script to generate google ID token.  There are many ways of doing it.  

The existing file google-id-token-client.py is using google_auth_oauthlib.flow package.
This file  google-id-token-client2.py is using google.auth package

### DIFF
--- a/endpoints/getting-started/clients/google-id-token-client2.py
+++ b/endpoints/getting-started/clients/google-id-token-client2.py
@@ -1,0 +1,54 @@
+import json
+import logging
+import requests
+import time
+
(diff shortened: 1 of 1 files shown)


qiwzhang:
@liminw I copied the original script here.

jonparrott:
Can I get some context for this?

ensonic:
Before we move on, lets discuss if this is a good practice and add a comment with the security implications.

jonparrott:
Closing due to inactivity, but feel free to open a new PR if this is still relevant.

PR #1124, opened 2017-09-19
Update readme

dizcology:
update readme to reflect the new method.

also putting back entity sentiment analysis to v1beta2 to avoid the false impression that it is removed from the API.


language/cloud-client/v1beta2/snippets.py
```diff
@@ -125,6 +126,66 @@ def entities_file(gcs_uri):
               entity.metadata.get('wikipedia_url', '-')))
 
 
+# [START def_entity_sentiment_text]
```
gguuss: FYI, the other blocks don't have includecode blocks.

PR #1134, opened 2017-09-22
Update README.md

VikramTiwari:
couldn't use appcfg.py and gcloud was suggested on the docs here:
https://cloud.google.com/appengine/docs/standard/python/getting-started/python-standard-env


jonparrott:
Thanks @VikramTiwari!

PR #1137, opened 2017-09-22
Update Task Queue samples for beta

andrewsg:
- Now that the discovery doc is published, remove discoveryServiceUrl from the client constructor.
- Move imports of the discovery command into the functions in preparation for adding region tags.
- Use flask's get_data() getter to process the payload bytes into unicode.
- Switch to using a print statement instead of logging.warn to avoid confusion over the term "warning".
- Update the JSON keys to reflect the newest version of the API.

### DIFF
--- a/appengine/flexible/tasks/create_app_engine_queue_task.py
+++ b/appengine/flexible/tasks/create_app_engine_queue_task.py
@@ -19,8 +19,6 @@ import base64
 import datetime
 import json
 
-from googleapiclient import discovery
-
 
 def seconds_from_now_to_rfc3339_datetime(seconds):
     """Return an RFC 3339 datetime string for a number of seconds from now."""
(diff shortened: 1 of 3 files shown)


appengine/flexible/tasks/create_app_engine_queue_task.py
```diff
@@ -31,16 +29,16 @@ def seconds_from_now_to_rfc3339_datetime(seconds):
 def create_task(project, queue, location, payload=None, in_seconds=None):
     """Create a task for a given queue with an arbitrary payload."""
 
+    from googleapiclient import discovery
```
jonparrott: Prefer `import googleapiclient.discovery` and `client = googleapiclient.discovery.build(...)` throughout.

andrewsg: Done

PR #1139, opened 2017-09-27
Small fixes to README.md

marciw:
1. gcloud alpha should be gcloud beta
2. --pubsub-topic should be --event-pubsub-topic 
3. Updated the link to the docs
4. Added info about the Publisher role for the Pub/Sub topic


gguuss:
@jonparrott FYI, a few fixes on the IoT docs.

PR #1142, opened 2017-09-27
Add support for overwrite attributes

BrandonY:
Adds support for displaying GCS overwrite-related attributes.

Also fixes a formatting bug for output.

### DIFF
--- a/storage/cloud-client/notification_polling.py
+++ b/storage/cloud-client/notification_polling.py
@@ -41,7 +41,7 @@ below:
    $ gcloud beta pubsub subscriptions create testsubscription --topic=testtopic
 
 6. Run this program:
-   $ python notification_polling my-project-id testsubscription
+   $ python notification_polling.py my-project-id testsubscription
 
 7. While the program is running, upload and delete some files in the testbucket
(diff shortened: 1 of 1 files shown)


storage/cloud-client/notification_polling.py
```diff
@@ -64,6 +64,8 @@ def summarize(message):
     bucket_id = attributes['bucketId']
     object_id = attributes['objectId']
     generation = attributes['objectGeneration']
+    overwroteGeneration = attributes['overwroteGeneration']
```
jonparrott: our style guide is `snake_case` for variable names, even if you're plucking them from the API repr.

BrandonY: Oh dangit, I do that every time. I just can't stop Javaing.

jonparrott: Hahaha. :)

jonparrott:
@BrandonY arg, this seems to fail with the system tests https://travis-ci.org/GoogleCloudPlatform/python-docs-samples/builds/280571893

Can you see about fixing it?

PR #1150, opened 2017-10-04
Change comment in ndb snippets to object property

### LINKED ISSUE #1149: ndb Key Snippet is Incorrect

[appengine/standard/ndb/entities/snippets.py](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/appengine/standard/ndb/entities/snippets.py#L121)

I have a fix which properly describes that keys are an instance property and not a class property. [You can see this change here.](https://github.com/GoogleCloudPlatform/python-docs-samples/compare/master...howinator:howinator/fix-entity-key-snippet)

## Describe the issue

I may have already properly described the issue, but the snipped in one of the ndb snippets indicates that the key property is a class attribute. It is actually an instance attribute. My branch above fixes this and I can submit a PR once I get the 👍 on this issue.

howinator:
Before this change, the documentation indicated that key property would
be on the Account class instead of the account object. This changes that
to properly reflect the key property is on the object.

Closes #1149


jonparrott:
Thank you, @howinator!

PR #1151, opened 2017-10-06
Dataproc GCS sample plus doc touchups

waprin:


dataproc/README.md
```diff
@@ -20,49 +20,78 @@ Go to the [Google Cloud Console](https://console.cloud.google.com).
...
+
 To install, run the following commands. If you want to use  [virtualenv](https://virtualenv.readthedocs.org/en/latest/)
 (recommended), run the commands within a virtualenv.
 
     * pip install -r requirements.txt
 
 Create local credentials by running the following command and following the oauth2 flow:
 
-    gcloud beta auth application-default login
+    gcloud auth application-default login
```
jonparrott: Our current auth guidance is to use service accounts.

dataproc/README.md
```diff
@@ -8,7 +8,7 @@ demonstrated here could also be accomplished using the Cloud Console or the gclo
 `list_clusters.py` is a simple command-line program to demonstrate connecting to the
```
jonparrott: Meta: This readme is enormous. Is there a documentation page that it could link to instead?

waprin: We do: https://cloud.google.com/dataproc/docs/tutorials/python-library-example . So yeah fair point, maybe I should cut this whole README, make sure all info's on that page?

jonparrott: That would be strongly preferable.

waprin: Ok, I pared down the README. I basically left all the information to actually run the samples, killed all the associated info and link to the tutorial at the top. I'll make a docs CL to add the note about reading from GCS with PySpark. LMK if that makes sense.

dataproc/README.md
waprin: Took this out, linked to our auth docs, noted we recommend service accounts.

jonparrott:
LGTM, but travis is failing.

PR #1159, opened 2017-10-11
Require latest Endpoints Framework version

inklesspen:
Given that users may go on to make real apps from this sample, we should keep to the latest known-good versions.


jonparrott:
Thanks, @inklesspen, keep in mind we have a weekly script that updates all dependencies automatically. :)

PR #1164, opened 2017-10-12
Changes to running the Python MQTT example

noerog:
* Remove required flag from --message_type flag
* Add 443 port as option for --mqtt_bridge_port flag


jonparrott:
@gguuss can you review?

noerog:
I removed the requirement for --message_type because it seemed unnecessary and also wasn't specified in any of the examples. Also, in the cloud shell, I had to use port 443.

PR #1170, opened 2017-10-20
Enable PyMySQL as an option for local development.

andrewsg:


andrewsg:
Let's hold off on merging until we also have satisfactory documentation changes in the pipe.

andrewsg:
I've updated this to remove the changes to the cloudsql sample in order to get the django sample changes out the door. I'll reconsider the cloudsql changes depending on how devappserver behaves in practice.

PR #1188, opened 2017-11-01
[endpoints] Update protoc commands to match documentation.

kryzthov:
@navinger 

 - Use long flags for better readability.
 - Change API descriptor file from `out.pb` to `api_descriptor.pb`.
 - Use same protoc command everywhere for consistency and simplicity.


navinger:
LGTM.

PR #1190, opened 2017-11-02
Adds JWT refresh to http example.

gguuss:


jonparrott:
LGTM, but must appease travis.

PR #1195, opened 2017-11-03
Updates pubsub client

gguuss:


jonparrott:
@gguuss seems you have a lint error

gguuss:
Fixed!

PR #1202, opened 2017-11-06
Language v1

dizcology:
copy classify_text samples and tests to v1


dizcology:
__DO NOT MERGE__ until the client library has been published and requirements.txt updated.

jonparrott:
Some lint errors on the build

dizcology:
tested with client library version 1.0.0

PR #1206, opened 2017-11-08
Fixed b/65407087

michaelawyu:
Import flask as a built-in third-party library


appengine/standard/flask/hello_world/app.yaml
```diff
@@ -5,3 +5,7 @@ threadsafe: true
 handlers:
 - url: /.*
   script: main.app
+
+libraries:
+- name: flask
+  version: latest
```
jonparrott: Can we pin a specific version here?

michaelawyu: Added :)

PR #1209, opened 2017-11-09
Ignore Flake8 import order error I202 to accommodate region tags

andrewsg:


andrewsg:
I decided to tell Flake to ignore this error because some of the detections were due to region tags inside import blocks, which is not an error given our use case. We could alternatively put # ignores inside each file where the region tags occur inside an import block; what do you think?

PR #1211, opened 2017-11-09
Vision beta

dizcology:
Update samples for some new features:
- safesearch
- using GPS geotagging data in images for label detection
- web entity search updates
- document text detection update

DO NOT MERGE until tested with released client library.


vision/cloud-client/web/web_entities.py
```diff
@@ -0,0 +1,81 @@
...
+import argparse
+import io
+
+from google.cloud import vision
+from google.cloud.vision import types
+# [END imports]
+
+
+# [START def_annotate_web_entities]
+def annotate_web_entities(path, include_geo_results=False):
```
jonparrott: Why wouldn't we just always set `include_geo_results` to true?

vision/cloud-client/web/web_entities.py
```diff
@@ -0,0 +1,81 @@
...
+    """Returns web annotations given the path to an image."""
+    client = vision.ImageAnnotatorClient()
+
+    if path.startswith('http') or path.startswith('gs:'):
```
jonparrott: I really don't like branches like this in samples, I prefer separate samples for local images and remote images.

dizcology: Good idea - I copied the pattern from `web_detect.py` without thinking much into it.  I will update both files when I get the chance to.

vision/cloud-client/web/web_entities.py
dizcology: The hope is to highlight how the results are different.  This being a new feature the user might want to compare the labels with this switched on/off to find out which suits their needs better.

jonparrott: I would suggest a separate sample altogether then.

dizcology: done.

jonparrott:
@dizcology LGTM, let me know when you're ready to merge.

remi:
Does this PR duplicate _all_ Vision samples into a new Beta directly? Are _all_ of the samples required to be duplicated or just a few which are different for Beta?

I advocate that the beta/ directory should be as small as possible and only include those exact snippets which we will display in our documentation to demonstrate _differences_ in the Beta API/library

Another option would be to add a `beta_snippets.py` or something similar with just the few snippets for Beta, rather than a whole new directory. This simplifies this quite a bit: no need to duplicate image assets, no need for a whole new README, etc

vision/cloud-client/detect/beta_snippets.py
```diff
@@ -0,0 +1,350 @@
...
+    response = client.web_detection(image=image)
+    web_detection = response.web_detection
+    # [END get_annotations]
+
+    return web_detection
+
+
+def report(annotations):
+    """Prints detected features in the provided web annotations."""
+    # [START print_annotations]
```
remi: Can we rename this to something that explicitly calls out this snippet demonstrates Web Detection Using Best Guess Labels? Perhaps something like `web_detection_using_best_guess_labels`?

vision/cloud-client/detect/beta_snippets.py
```diff
@@ -0,0 +1,350 @@
...
+
+    response = client.web_detection(image=image, image_context=image_context)
+
+    for entity in response.web_detection.web_entities:
+        print(u'\nDescription: {}'.format(entity.description))
+        print('Score: {}'.format(entity.score))
+# [END vision_web_entities_include_geo_results_file]
+
+
+# [START def_detect_document]
```
remi: Can this region tag be renamed to explicitly call out what the snippet demonstrates? Something like `confidence_score_on_text_detection`?

remi:
Note: for consistency with existing samples, all region tags should follow one of the following formats that we currently use for all Vision samples:

 - `vision_X_detection`
 - `vision_X_detection_gcs`

I think it would be very worthwhile to also consider using these, where it would add useful context:

 - `vision_X_detection_using_Y`
 - `vision_X_detection_using_Y_gcs`

remi:
Oh! Let's also add `beta` to ALL of the region tag IDs. This will make it very explicit and easier to track!

`vision_beta_XYZ`

vision/cloud-client/detect/beta_snippets.py
```diff
@@ -0,0 +1,350 @@
...
+    print('medical: {}'.format(likelihood_name[safe.medical]))
+    print('spoofed: {}'.format(likelihood_name[safe.spoof]))
+    print('violence: {}'.format(likelihood_name[safe.violence]))
+    print('racy: {}'.format(likelihood_name[safe.racy]))
+# [END def_detect_safe_search_uri]
+
+
+def annotate(path):
+    """Returns web annotations given the path to an image."""
+    # [START get_annotations]
```
remi: What Beta functionality does this snippet demonstrate?

vision/cloud-client/detect/beta_snippets.py
```diff
@@ -0,0 +1,350 @@
...
+
+# [START imports]
+import argparse
+import io
+
+from google.cloud import vision_v1p1beta1 as vision
+# [END imports]
+
+
+# [START vision_web_entities_uri]
```
remi: What Beta functionality does this snippet demonstrate?

vision/cloud-client/detect/beta_snippets.py
remi: Why are `annotate` and `report` separated?

dizcology:
@remi:
1. The whole `annotate` and `report` split is my fault.  Need to use a different snippet for this beta feature and I incorrectly used the code from the tutorial.  I'll fix this.

2. Not sure about `_gcs` in the region tags - in this case the uri could be either a `gcs` uri or an `http` uri.

3. I'll fix the region tags to use `vision_` prefixes in the beta file.  The v1 files will need to wait until after some of these features become GA.

4. The blocks of `vision_web_entities_uri/file` and `vision_web_entities_include_geo_results_uri/file` are there to highlight the subtle difference in the `web_detection` feature.  `web_detection` returns many things (entities, similar images, pages), and entities is the only one affected by the `include_geo_result` switch.  In the `vision_web_detection_uri/file` codes we would print all the interesting pieces from the API response, but that will make the subtle differences in entities even less visible.  Following your comments on the region tags, we could have `vision_web_detection_using_include_geo_result_uri/file` which prints only the entities (and none of the pages or similar images as in `vision_web_detection_uri/file`) with `include_geo_result` turned on.

PR #1216, opened 2017-11-13
update to use v1 client

dizcology:
Update the tutorial to use v1 client.  Sample tests passed.

### DIFF
--- a/language/classify_text/classify_text_tutorial.py
+++ b/language/classify_text/classify_text_tutorial.py
@@ -87,7 +87,7 @@ def index(path, index_file):
             print('Failed to process {}'.format(file_path))
 
     with io.open(index_file, 'w', encoding='utf-8') as f:
-        f.write(unicode(json.dumps(result)))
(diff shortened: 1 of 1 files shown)


language/classify_text/classify_text_tutorial.py
```diff
@@ -90,7 +87,7 @@ def index(path, index_file):
             print('Failed to process {}'.format(file_path))
 
     with io.open(index_file, 'w', encoding='utf-8') as f:
-        f.write(json.dumps(result))
+        f.write(unicode(json.dumps(result)))
```
jonparrott: unicode doesn't exist on python 3, what you really want here is:

```
json.dump(result, f)
```

dizcology: python 2.7's json.dump seems to use f.write and `json.dump(result, f)` still returns the following error:

```TypeError: write() argument 1 must be unicode, not str```

PR #1219, opened 2017-11-14
Convert tasks example to use logging for testability

andrewsg:


jonparrott:
We generally don't use logging in our samples, what about this makes the sample more testable?

andrewsg:
Previously we had print() which turned out to be hard to reliably mock in a way that works across Py2 and 3. Logging is easy to mock. Either way, the output goes to the same place; we need some sort of off-instance data storage to prove the task ran, and the logging console is the most convenient place without pulling in another dependency.

PR #1240, opened 2017-11-30
Update example_startup_script.sh

ahoying:
On latest Debian 9 image on GCP this startup script fails with this error:

Nov 30 17:03:33 localhost startup-script: INFO startup-script: E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?


jonparrott:
Thanks, @ahoying!

PR #1244, opened 2017-12-01
remove empty input_uri

dizcology:
The client library was updated to make the `input_uri` optional.


dizcology:
Tested with client library v1.0.0 with local file.

PR #1255, opened 2017-12-07
Tests for device over HTTP and adds example for getting configuration.

gguuss:


iot/api-client/http_example/cloudiot_http_example_test.py
```diff
@@ -0,0 +1,167 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import os
+import sys
+import time
+
+import cloudiot_http_example
+sys.path.append(os.path.join(os.path.dirname(__file__), '..', 'manager'))  # noqa
```
jonparrott: Can you add a comment on why you did this?

PR #1257, opened 2017-12-07
Use new Python image for Kokoro tests

jonparrott:


jonparrott:
I'm gonna go ahead and merge this as it's a tests-only change. I will follow up with any concerns post-hoc, @andrewsg.

PR #1258, opened 2017-12-07
[fix] Django 2.0 doesn't work with Python 2.7, rollback to 1.11.8

alvarowolfx:
Django 2.0 requires at least Python 3.4, but App Engine Standard uses Python 2.7. This error was introduced in PR #1245 by a bot that auto updates the packages. 

I made a rollback on the Django version to the latest compatible with Python 2.7, that is the version 1.11.8.


jonparrott:
I need to remember how to tell the bot to not update stuff.

PR #1266, opened 2017-12-11
Latest version of Werkzeug raises ImportError

awonak:
The latest version of Werkzeug (0.13) released on December 7th raises an ImportError when Flask is imported on app engine.

```ImportError: cannot import name SpooledTemporaryFile```

App engine standard does not support SpooledTemporaryFile, so we should explicitly instruct users to install Werkzeug==0.12.2 (last working version).

### DIFF
--- a/appengine/standard/flask/tutorial/requirements.txt
+++ b/appengine/standard/flask/tutorial/requirements.txt
@@ -1,1 +1,2 @@
+Werkzeug==0.12.2
 Flask==0.12.2
--- a/appengine/standard/flask/tutorial/requirements.txt
+++ b/appengine/standard/flask/tutorial/requirements.txt
@@ -1,2 +1,4 @@
-Werkzeug==0.12.2
 Flask==0.12.2
+# v0.13 uses tempfile.SpooledTemporaryFile which is not supported in the app
+# engine sandbox. See https://cloud.google.com/appengine/docs/standard/python/runtime#customized-libraries-in-python-version-27
+Werkzeug<=0.13

### CODE REVIEW

File: appengine/standard/flask/tutorial/requirements.txt [hunk 1]
```diff
@@ -1 +1,2 @@
+Werkzeug==0.12.2
```
jonparrott: To prevent our auto-updating script from messing with this, can you change it to a range? (& also add a note as to why we pin it)

```Werkzeug<=0.13```

jonparrott:
Do you mind filing a bug on Werkzeug for this?

jonparrott:
Thanks, @awonak!

awonak:
There is an [issue](https://github.com/pallets/werkzeug/issues/1210) over in Werkzeug. If they do revert back to `TemporaryFile` we should update the changes here to pin to that update.

davidism:
I am one of the maintainers of Werkzeug. Is there a reason you don't support `SpooledTemporaryFile`? I'm not familiar with GCP, so every time something breaks it's non-trivial for us to investigate and test this. Is there a better way we can be testing against GCP, similar to Travis-CI?

davidism:
I read https://cloud.google.com/appengine/docs/standard/python/runtime#customized-libraries-in-python-version-27. Is there a way for GCP to more comprehensively monkeypatch the `tempfile` module instead of restricting what can be imported?

jonparrott:
@davidism thanks for jumping in. :)

> Is there a reason SpooledTemporaryFile isn't supported?
> Is there a way for GCP to more comprehensively monkeypatch the tempfile module instead of restricting what can be imported?

App Engine Standard is heavily sandboxed and currently blocks all access to the filesystem. It's technically possible that we can emulate that with an in-memory file, but doing such changes is really difficult for us at the moment.

We're working on some under-the-covers changes to our runtime environment that will make it easier for us to get closer to a unforked experience and those will hopefully land early in the new year, so, if you want to say "no", that's perfectly acceptable.

> Is there a better way we can be testing against GCP, similar to Travis-CI?

The rest of GCP (App Engine flex, GCE, GKE) is fine it's just the sandboxed standard environment. I helped set up testing against GAE standard for urllib3 and I'm happy to do the same with Werkzeug if you want- but this is another case where it might be better to just wait for our changes that fix some of the deviation from the standard library (ideally, if it passes generic Python 2.7 it should work in GAE standard).

PR #1268, opened 2017-12-11
Adds MQTT device tests

gguuss:


gguuss:
@jonparrott can you merge?  I don't have perms >.<

PR #1279, opened 2017-12-15
Use exponential backoff in HTTP methods.

gguuss:


iot/api-client/http_example/cloudiot_http_example.py
```diff
@@ -27,11 +27,13 @@
 import json
 import time
 
+from google.api_core import retry
```
jonparrott: this can go in the same section as jwt and requests

iot/api-client/http_example/cloudiot_http_example.py
```diff
@@ -83,9 +88,16 @@ def publish_message(
     resp = requests.post(
             publish_url, data=json.dumps(body), headers=headers)
 
+    if (resp.status_code != 200):
```
jonparrott: If you wanted, you could use `resp.raise_for_status()`, but this is fine.

iot/api-client/http_example/cloudiot_http_example.py
gguuss: Done, ty!

iot/api-client/http_example/cloudiot_http_example.py
gguuss: ack.

PR #1292, opened 2017-12-22
Revert to django 1.11

michaelawyu:


jonparrott:
This is a no-op, so closing.

PR #1312, opened 2018-01-11
fix message.payload str casting.

nectarine:
Env: Python 3.6.4 (default, Jan  6 2018, 11:51:59)
[GCC 4.2.1 Compatible Apple LLVM 9.0.0 (clang-900.0.39.2)] on darwin

cloudiot_pubsub_example_mqtt_device.py.py in iot end_to_end_example contains wrong str casting.
json.loads takes byte parameter, so you don't have to cast for using it.
If you really want to convert byte to string, it should be 

```python3
payload = message.payload.decode()
```


nectarine:
I signed it!

nectarine:
:( commit log is messed up by cla, I'll re-open this.

PR #1322, opened 2018-01-17
Add comment about requirement order in endpoints sample.

inklesspen:
We've seen repeated issues where specifying packages such as `google-api-python-client` before the Frameworks packages causes an invalid version of `oauth2client` to be installed.

### DIFF
--- a/appengine/standard/endpoints-frameworks-v2/echo/requirements.txt
+++ b/appengine/standard/endpoints-frameworks-v2/echo/requirements.txt
@@ -1,2 +1,6 @@
+# If more requirements are added to an Endpoints project, the
+# google-endpoints and google-endpoints-api-management requirements
+# must remain at the top. See this url for details:
+# https://cloud.google.com/endpoints/docs/known-issues#error_message_importerror_cannot_import_name_locked_file
 google-endpoints==2.4.5
 google-endpoints-api-management==1.3.0


jonparrott:
Don't do this, just pin the specific oauth2client version.

Also - oauth2client is deprecated. Is there any chance we could chat about getting endpoints to upgrade to google-auth?

inklesspen:
Endpoints doesn't use oauth2client. Endpoints uses a fork (no, I don't know why) of google-apitools, which uses oauth2client. I am working on getting a replacement for google-apitools, but that is at least a few months out. In the meantime, people are having this problem, and with oauth2client being a dependency of a dependency, it is very hard to communicate the issue to users.

google-endpoints-api-management does [try to specify](https://github.com/cloudendpoints/endpoints-management-python/blob/ba475625680cc66dd44f2d95996773e3409ec8f6/setup.py#L37) the working versions of oauth2client, but that doesn't seem to stop people from installing nonworking versions.

jonparrott:
You should at least work with api_tools to support newer versions of oauth2client (it's not difficult and I'm happy to help) as there are important security-related changes.

inklesspen:
The capacity I have on that front is devoted to transitioning to a more maintained client library.

jonparrott:
We should chat in person soon - from what I understand, cloud sdk has moved to oauth2client > 4.0.0 and it uses apitools, so I don't see what's blocking endpoints.

PR #1325, opened 2018-01-22
Add a new sample service for Endpoints Frameworks for Python

inklesspen:
The IATA sample service provides examples of CRUD operations, managing
a list of airports and their corresponding IATA codes. This service
will be used in documentation for the Endpoints Frameworks.


appengine/standard/endpoints-frameworks-v2/iata/app.yaml
```diff
@@ -0,0 +1,34 @@
+runtime: python27
+threadsafe: true
+api_version: 1
+basic_scaling:
+  max_instances: 2
+
+#[START_EXCLUDE]
```
jonparrott: This shouldn't be necessary.

appengine/standard/endpoints-frameworks-v2/iata/appengine_config.py
```diff
@@ -0,0 +1,4 @@
+from google.appengine.ext import vendor
```
jonparrott: needs a license header.

appengine/standard/endpoints-frameworks-v2/iata/app.yaml
inklesspen: These are in the app.yaml in appengine/standard/endpoints-frameworks-v2/echo; do you know why they're there? I just mimicked that.

appengine/standard/endpoints-frameworks-v2/iata/appengine_config.py
inklesspen: This file is identical to (and was copied from) the equivalent file in appengine/standard/endpoints-frameworks-v2/echo.

jonparrott: Still needs a license header. :)

inklesspen:
Done.

PR #1340, opened 2018-02-02
Cut off streaming generator in transcribe_streaming_mic to 20 seconds

chemelnucfin:
Works with python3, python2 still has a bug (different from this)
See [google-cloud-python/4820](https://github.com/GoogleCloudPlatform/google-cloud-python/issues/4820)

### DIFF
--- a/speech/cloud-client/transcribe_streaming_mic.py
+++ b/speech/cloud-client/transcribe_streaming_mic.py
@@ -28,6 +28,8 @@ Example usage:
 # [START import_libraries]
 from __future__ import division
 
+from datetime import datetime
+from datetime import timedelta
 import re
 import sys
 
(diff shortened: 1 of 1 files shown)


jonparrott:
How/why does this fix the issue in question?

chemelnucfin:
It does not fix the Python2 issue.  Python3 times out by a little bit, I think it was 189.99 seconds, but it requires audio duration + 5 seconds, so something like 196 seconds or something.

The fix was to limit the number of seconds.  Plus, 196 seconds is too long for a doc sample anyway.

chemelnucfin:
I meant it as a python3 workaround to the issue, not a fix for the issue.

jonparrott:
Seems arbitrary. I'm not in favor of that sort of thing. Maybe @jerjou can provide some insight into this issue?

chemelnucfin:
https://github.com/GoogleCloudPlatform/google-cloud-python/issues/4820#issuecomment-362435841

This was the error for the issue.

jerjou:
I believe this PR patches the symptom rather than the problem. The root cause of the python3 stack is that the API [only supports 60 seconds](https://cloud.google.com/speech/quotas#content) of audio, and allocates 3*(audio length)+5 duration to transcribe it. Thus, it expects the client to set a timeout long enough that it has time to transcribe whatever length audio you've passed it. It looks like you just ran the sample for long enough that it hit the length of audio allowed for with the default timeout. The proper fix is likely to increase the default timeout to allow for the full 60 seconds. However, you'll still encounter this if you wait long enough (ie 60+ seconds), but that's a limitation of the API rather than the sample code.

PR #1358, opened 2018-02-09
WebSocket and Session Affinity beta samples

labadav:
Adding samples for WebSockets and Session Affinity beta for GAE flexible docs


labadav:
This will be re-submitted later. Need to re-write.

PR #1361, opened 2018-02-12
Speech model selection

dizcology:
This PR adds samples for Speech API's model selection feature.

Tested against cloud-google-speech==0.32.0.


dizcology:
Cleared to be merged.

PR #1366, opened 2018-02-16
Typo in method header.

scotmatson:
Switched letters to correct typo from craetes to creates.


scotmatson:
I signed it!

alixhami:
Hi @scotmatson - thanks for finding this! I deleted this file in PR #1455 because the samples were redundant with those in the google-cloud-python repo so this is no longer an issue.

PR #1368, opened 2018-02-20
Fixes python version flag & bad import in context manager

maciekrb:
Adds the `-p python3.5` flag to virtualenv creation step, as currently the flag is being added to an env var (a typo I believe) and the python binary ends up being `python 2.7`. 

Fixes a bad import in status context manager.

### DIFF
--- a/endpoints/bookstore-grpc/Dockerfile
+++ b/endpoints/bookstore-grpc/Dockerfile
@@ -5,7 +5,7 @@ FROM gcr.io/google_appengine/python
 
 # Create a virtualenv for dependencies. This isolates these packages from
 # system-level packages.
-RUN virtualenv -p python3.5 /env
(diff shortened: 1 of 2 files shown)


endpoints/bookstore-grpc/Dockerfile
```diff
@@ -5,11 +5,11 @@ FROM gcr.io/google_appengine/python
 
 # Create a virtualenv for dependencies. This isolates these packages from
 # system-level packages.
-RUN virtualenv /env
+RUN virtualenv -p python3.5 /env
```
jonparrott: use `python3.6` if you're gonna use python 3.

maciekrb: `python3.6` it is then.

PR #1372, opened 2018-02-26
Lesv patch 1

lesv:
b/64916781 remove log's samples


lesv:
@dpebot merge when green

PR #1373, opened 2018-02-26
Container Analysis Samples

DanSanche:
I the samples for the drydock (container analysis) API. This is my first python sample, so let me know if I'm doing anything weird/non-standard


container_registry/container_analysis/samples.py
```diff
@@ -0,0 +1,343 @@
...
+from google.cloud.devtools.containeranalysis_v1alpha1.proto \
+    import containeranalysis_pb2
+from google.cloud.devtools.containeranalysis_v1alpha1.proto \
+    import package_vulnerability_pb2
+from google.cloud.pubsub import SubscriberClient
+
+
+# [START create_note]
+def create_note(note_id, project_id):
+    """Creates and returns a new note
```
gguuss: We usually don't put docstrings this comprehensive into the samples as the snippets will typically be surfaced in the documentation where there is sufficient context to know what is going on.

container_registry/container_analysis/requirements.txt
```diff
@@ -0,0 +1,2 @@
+google-cloud-pubsub == 0.30.0
+./google_cloud_containeranalysis-0.1.0-py2.py3-none-any.whl
```
gguuss: Included wheel is less common, maybe add a comment:

    # Replace when client library is published.

container_registry/container_analysis/samples.py
```diff
@@ -0,0 +1,343 @@
...
+    In this case, we will simply print and count each message as it comes in
+    """
+    def __init__(self):
+        # initialize counter to 0 on initialization
+        self.msg_count = 0
+
+    def pubsub_callback(self, message):
+        # every time a pubsub message comes in, print it and count it
+        self.msg_count += 1
+        print("Message " + str(self.msg_count) + ": " + message.data)
```
gguuss: Prefer format to string concatenation:

    print('Message {}: {}'.format(self.msg_count, message.data)

container_registry/container_analysis/samples.py
```diff
@@ -0,0 +1,343 @@
...
+        google.api_core.exceptions.RetryError: If the request failed due
+            to a retryable error and retry attempts failed.
+        ValueError: If the parameters are invalid.
+    """
+    topic_id = "resource-notes-occurrences-v1alpha1"
+    client = SubscriberClient()
+    topic_name = client.topic_path(project_id, topic_id)
+    subscription_name = client.subscription_path(project_id, subscription_id)
+    success = True
+    try:
```
gguuss: If possible, avoid exceptions / and try/catch as they may cause flow control to be a little less clear

container_registry/container_analysis/samples_test.py
```diff
@@ -0,0 +1,178 @@
...
+    def test_update_note(self):
+        description = "updated"
+        self.note_obj.short_description = description
+        samples.update_note(self.note_obj, self.note_id, PROJECT_ID)
+
+        updated = samples.get_note(self.note_id, PROJECT_ID)
+        assert updated.short_description == description
+
+    def test_create_occurrence(self):
+        created = samples.create_occurrence(self.image_url,
```
gguuss: Prefer to avoid hanging indents (here and elsewhere):

    samples.create_occurrence(
            self.image_url,
            self.note_id,
            PROJECT_ID)

container_registry/container_analysis/samples.py
```diff
@@ -0,0 +1,343 @@
...
+        google.api_core.exceptions.RetryError: If the request failed due
+            to a retryable error and retry attempts failed.
+        ValueError: If the parameters are invalid.
+    """
+    client = container_analysis_client.ContainerAnalysisClient()
+    parent_name = client.note_path(project_id, parent_note_id)
+    project_name = client.project_path(project_id)
+    vul = package_vulnerability_pb2.VulnerabilityType.VulnerabilityDetails()
+
+    occurrence = containeranalysis_pb2.Occurrence(note_name=parent_name,
```
gguuss: Prefer to avoid hanging indents:

    occurence = containeranalysis_pb2.Occurence(
            note_name=parent_name,
            resource_url=image_url,
            vulnerability_details=vul)

container_registry/container_analysis/samples.py
```diff
@@ -0,0 +1,343 @@
...
+        google.api_core.exceptions.RetryError: If the request failed due
+            to a retryable error and retry attempts failed.
+        ValueError: If the parameters are invalid.
+    """
+    filterStr = "kind=\"DISCOVERY\" AND resourceUrl=\"" + image_url + "\""
+    client = container_analysis_client.ContainerAnalysisClient()
+    project_name = client.project_path(project_id)
+    response = client.list_occurrences(project_name, filter_=filterStr)
+    for occ in response:
+        print occ
```
gguuss: **print** is a function, `print(occ)`

container_registry/container_analysis/samples.py
```diff
@@ -0,0 +1,343 @@
...
+        google.api_core.exceptions.RetryError: If the request failed due
+            to a retryable error and retry attempts failed.
+        ValueError: If the parameters are invalid.
+    """
+    client = container_analysis_client.ContainerAnalysisClient()
+    note_name = client.note_path(project_id, note_id)
+
+    response = client.list_note_occurrences(note_name)
+    count = 0
+    for o in response:
```
gguuss: Does `return len(response)` do the same?

container_registry/container_analysis/samples.py
```diff
@@ -0,0 +1,343 @@
...
+    :param project_id: the project's unique identifier
+    :return: number of occurrences found
+    :raises
+        google.api_core.exceptions.GoogleAPICallError: If the request
+            failed for any reason.
+        google.api_core.exceptions.RetryError: If the request failed due
+            to a retryable error and retry attempts failed.
+        ValueError: If the parameters are invalid.
+    """
+    filterStr = "resourceUrl=\"" + image_url + "\""
```
gguuss: 1. Prefer format over string concatenation
2. snake_case and not camelCase
3. Single-quotes over double-quotes

e.g. 

    filter_str = 'resourceUrl="{}".format(image_url)'

container_registry/container_analysis/samples.py
```diff
@@ -0,0 +1,343 @@
...
+    client = container_analysis_client.ContainerAnalysisClient()
+    project_name = client.project_path(project_id)
+
+    response = client.list_occurrences(project_name, filter_=filterStr)
```
gguuss: Does `return len(response)` work here too?

container_registry/container_analysis/samples_test.py
```diff
@@ -0,0 +1,178 @@
...
+from google.cloud.pubsub import SubscriberClient
+
+import samples
+
+PROJECT_ID = environ['GCLOUD_PROJECT']
+SLEEP_TIME = 1
+TRY_LIMIT = 20
+
+
+class TestContainerAnalysisSamples:
```
gguuss: Prefer just having test methods, [see this example](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/vision/cloud-client/detect/detect_test.py) for a pattern you can use.

container_registry/container_analysis/samples_test.py
```diff
@@ -0,0 +1,178 @@
...
+        try:
+            samples.get_occurrence(created.name)
+        except NotFound:
+            pass
+        else:
+            # didn't raise exception we expected
+            assert False
+
+    def test_update_occurrence(self):
+        new_type = "newType"
```
gguuss: Prefer single-quote over double-quote

container_registry/container_analysis/samples.py
DanSanche: I considered it, but I thought it would be useful to demonstrate how to access the individual responses so the sample is easier to build real uses cases off of, and to keep it more consistent with other languages. Do you think it's better to keep it simple?

gguuss: If you want to leave in the loop, maybe add a comment 

    # you can process results here

container_registry/container_analysis/samples.py
DanSanche: Here I'm just catching and responding to an exception thrown by the pubsub library, so it might be hard to avoid. I thought about doing get_subscription before create_subscription to explicitly check if it already exists first, but it looks like get_subscription throws an exception if the subscription doesn't exist too, so i'd have to try-catch either way. Is there a batter way to handle this I'm not thinking of?

container_registry/container_analysis/samples.py
DanSanche: Ok, should I just remove the param/return/raises to make it simpler? Or should I move the docstrings outside the region tags?

container_registry/container_analysis/samples_test.py
DanSanche: Does this pattern work well with setup/teardown functions? I'm having trouble finding examples

gguuss: [Here is an example](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/iot/api-client/http_example/cloudiot_http_example_test.py#L45-L54) where the fixture yields the dependency before the test and destroys it after.

container_registry/container_analysis/samples.py
gguuss: > Ok, should I just remove the param/return/raises to make it simpler?
Yup, I'd remove them.

PR #1383, opened 2018-03-05
CDN samples dir with snippet for signing urls

djmailhot:
Add a new directory for CDN.

Include a snippet for signing urls, as specified by the Cloud Storage documentation at https://cloud.google.com/storage/docs/access-control/create-signed-urls-program


cdn/snippets.py
```diff
@@ -0,0 +1,77 @@
+"""This application demonstrates how to perform operations on data (content)
+when using the Google Cloud CDN (Content Delivery Network).
+
+For more information, see the README.md under /cdn and the documentation
+at https://cloud.google.com/cdn/docs.
+"""
+
+import argparse
+import base64
+<<<<<<< HEAD
```
ahmetb: looks like merge conflict indicator is a leftover here

cdn/snippets.py
```diff
@@ -0,0 +1,77 @@
...
+    sign_url_parser.add_argument('key_name',
+            help="Signed URL key name to use for the 'KeyName=' query parameter.")
+    sign_url_parser.add_argument('key',
+            help="The 16-byte unencoded secret key value to use for signing.")
+    sign_url_parser.add_argument('expiration_timestamp',
+            help="Expiration time expessed as seconds since the epoch.")
+
+    args = parser.parse_args()
+
+    args.url = "http://35.186.234.33/index.html"
```
ahmetb: delete these 3 to not to overwrite the args?

cdn/snippets.py
```diff
@@ -0,0 +1,77 @@
...
+            url=stripped_url,
+            separator='&' if query_params else '?',
+            expires=expiration_timestamp,
+            key_name=key_name)
+
+    signature = base64.urlsafe_b64encode(
+            hmac.new(key, url_to_sign, hashlib.sha1).digest())
+
+    return '{url}&Signature={signature}'.format(
+            url=url_to_sign, signature=signature)
```
ahmetb: can you put a region tag from the beginning of `import`s to the end of `def sign_url` (so that we don't include what's below in the documentation)?

cdn/snippets.py
djmailhot: Fixed

cdn/snippets.py
djmailhot: Fixed

cdn/snippets.py
djmailhot: Fixed

ahmetb:
Algorithm LGTM. Needs a Python approver.

cdn/snippets.py
```diff
@@ -0,0 +1,95 @@
...
+For more information, see the README.md under /cdn and the documentation
+at https://cloud.google.com/cdn/docs.
+"""
+
+import argparse
+import base64
+import datetime
+import hashlib
+import hmac
+import urlparse
```
jonparrott: you'll need to use `six` in order to get python 2/3 compatibility.

cdn/snippets.py
```diff
@@ -0,0 +1,95 @@
...
+
+    """
+    stripped_url = url.strip()
+    parsed_url = urlparse.urlsplit(stripped_url)
+    query_params = urlparse.parse_qs(parsed_url.query, keep_blank_values=True)
+    epoch = datetime.datetime.utcfromtimestamp(0)
+    expiration_timestamp = int((expiration_datetime - epoch).total_seconds())
+
+    url_to_sign = '{url}{separator}Expires={expires}&KeyName={key_name}'.format(
+            url=stripped_url,
```
jonparrott: This is indented way too far. Indents in python are 4 spaces with a single indent level for continuations.

cdn/snippets.py
```diff
@@ -0,0 +1,95 @@
...
+            'encoded_base64_key',
+            help='The base64 encoded secret key value to use for signing.')
+    sign_url_parser.add_argument(
+            'expiration_timestamp',
+            help='Expiration time expessed as seconds since the epoch.')
+
+    args = parser.parse_args()
+
+    if args.command == 'sign-url':
+        expiration_datetime = datetime.datetime.utcfromtimestamp(
```
jonparrott: don't put any logic here, move this into the sample.

ahmetb: I think this part is correct (and is compatible with other language samples). We want `sign_url` to accept a language-native date type (in this case `datetime`).

So here, we're converting Unix epoch CLI arg to `datetime` to pass to `sign_url`. Most people would be integrating `sign_url` to their programs, and not this part. Hence, we should probably keep it as is.

cdn/snippets.py
```diff
@@ -0,0 +1,99 @@
...
+    from urlparse import urlsplit, parse_qs
+
+# [BEGIN sign_url]
+def sign_url(url, key_name, key, expiration_time):
+    """Gets the Signed URL string for the specified URL and configuration.
+
+    Args:
+        url: URL to sign as a string.
+        key_name: name of the signing key as a string.
+        key: signing key as a urlsafe base64 encoded string.
```
ahmetb: `base64url-encoded` to be precise

cdn/snippets.py
```diff
@@ -0,0 +1,99 @@
...
+        expiration_time: expiration time as a UTC datetime object.
+
+    Returns:
+        Returns the Signed URL appended with the query parameters based on the
+        specified configuration.  Roughly of the form:
```
ahmetb: > Roughly of the form:  {url}{separator}Expires={expiration}&KeyName={key_name}&Signature={signature}

seems unnecessary.

ahmetb:
CLA bot is glitchy nowadays. This is likely ready to merge.

PR #1387, opened 2018-03-08
Change to endpoints managed rollout

simonz130:
We introduce a new feature where we can specify in YAML that service configuration for Endpoints is going to be managed. This means that the service will pull latest deployed service configuration automatically, rather than requiring customer to specify specific version of service configuration.

New flag:
**--rollout_strategy=managed**
Previously the only way to specify a version was using the following flag:
**--version=SERVICE_CONFIG_ID**

### DIFF
--- a/endpoints/getting-started-grpc/README.md
+++ b/endpoints/getting-started-grpc/README.md
@@ -49,13 +49,10 @@ Cloud account and [SDK](https://cloud.google.com/sdk/) configured.
 
     ```bash
     gcloud endpoints services deploy api_descriptor.pb api_config.yaml
-    # The Config ID should be printed out, looks like: 2017-02-01r0, remember this
 
     # Set your project ID as a variable to make commands easier:
     GCLOUD_PROJECT=<Your Project ID>
 
-    # Print out your Config ID again, in case you missed it:
-    gcloud endpoints configs list --service hellogrpc.endpoints.${GCLOUD_PROJECT}.cloud.goog
     ```
 
 1. Also get an API key from the Console's API Manager for use in the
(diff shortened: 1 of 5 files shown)


endpoints/getting-started-grpc/deployment.yaml
```diff
@@ -44,7 +44,7 @@ spec:
           "--http2_port=9000",
           "--backend=grpc://127.0.0.1:50051",
           "--service=SERVICE_NAME",
-          "--version=SERVICE_CONFIG_ID",
+          "--rollout_strategy==managed",
```
wmwong: Change "==" to "=".

endpoints/getting-started-grpc/README.md
```diff
@@ -95,7 +95,6 @@ Cloud account and [SDK](https://cloud.google.com/sdk/) configured.
     ```bash
```
wmwong: Up above after "gcloud endpoints services deploy ...":
- Remove comment about noting the config ID.
- Remove command about getting the config ID.

endpoints/getting-started-grpc/deployment.yaml
simonz130: thanks!

endpoints/getting-started-grpc/README.md
```diff
@@ -49,13 +49,10 @@ Cloud account and [SDK](https://cloud.google.com/sdk/) configured.
 
     ```bash
     gcloud endpoints services deploy api_descriptor.pb api_config.yaml
-    # The Config ID should be printed out, looks like: 2017-02-01r0, remember this
 
     # Set your project ID as a variable to make commands easier:
     GCLOUD_PROJECT=<Your Project ID>
```
wmwong: This blank line should probably also be removed.

simonz130: I think it's better to have the blank line - for better clarity of the output.

Issue #1352, opened 2018-02-07
Sample code is out of sync


talonx:
## In which file did you encounter the issue?
https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/monitoring/api/v3/api-client/custom_metric.py#L176

### Did you change the file? If so, how?
No

## Describe the issue
The function get_client() does not exist in list_resources

engelke:
PR #1396 addressed this issue.

PR #1392, opened 2018-03-10
Add get_client function

campionfellin:
This pertains to issue https://github.com/GoogleCloudPlatform/python-docs-samples/issues/1352

It looks like `list_resources.py` did not have a `get_client` function, so I added it.

### DIFF
--- a/monitoring/api/v3/api-client/list_resources.py
+++ b/monitoring/api/v3/api-client/list_resources.py
@@ -98,7 +98,7 @@ def list_timeseries(client, project_resource, metric):
 def get_client():
     client = googleapiclient.discovery.build('monitoring', 'v3')
     return client
-    
+
 
 def main(project_id):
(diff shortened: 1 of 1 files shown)


jonparrott:
Thanks, @campionfellin, but we actually construct the client [here](https://github.com/campionfellin/python-docs-samples/blob/984fc18275955e74410f74624a2e6c5fed551f73/monitoring/api/v3/api-client/list_resources.py#L104) in these samples, so this function isn't necessary.

campionfellin:
With regards to #1352 how would you suggest we fix https://github.com/GoogleCloudPlatform/python-docs-samples/blob/b56ac42df20ec374821791c86481f22444d1b63d/monitoring/api/v3/api-client/custom_metric.py#L176 ?

Would it be best to just import googleapiclient.discovery and instantiate an instance there?

jonparrott:
Yep!

PR #1398, opened 2018-03-13
Update inspect_content; pin other samples at v2beta1

andrewsg:


andrewsg:
Thanks Jon, I'll back out my second commit and put it in a separate PR to make the reviews more manageable.

PR #1402, opened 2018-03-14
Spanner Batch Query Sample

kurtisvg:


spanner/cloud-client/batch_sample.py
```diff
@@ -0,0 +1,99 @@
...
+# limitations under the License.
+
+"""This application demonstrates how to do batch operations using Cloud
+Spanner.
+
+For more information, see the README.rst under /spanner.
+"""
+
+import argparse
+from multiprocessing.pool import ThreadPool
```
jonparrott: I'd suggest you use `concurrent.futures` which is a preferred over directly using `multiprocessing.pool`.

spanner/cloud-client/batch_sample.py
```diff
@@ -0,0 +1,99 @@
...
+import time
+
+from google.cloud import spanner
+
+
+# [START spanner_batch_client]
+def run_batch_query(instance_id, database_id):
+    """Runs an example batch query."""
+
+    # CREATE TABLE Singers (
```
jonparrott: Why is this here?

kurtisvg: It's the definition of the table the sample expects to work on.

spanner/cloud-client/batch_sample.py
kurtisvg: Fixed in 172ee2c8c4a64f32f0713292153969e009854543.

kurtisvg:
Updated for 1.2.0. I think the only reason it is failing travis is because the version isn't on PyPI yet.

spanner/cloud-client/batch_sample.py
jonparrott: Can you leave a comment that explains that?

jonparrott:
@kurtisvg it should be now.

spanner/cloud-client/batch_sample.py
```diff
@@ -0,0 +1,92 @@
...
+        columns=('SingerId', 'FirstName', 'LastName',),
+        keyset=spanner.KeySet(all_=True)
+    )
+
+    # Create a pool of workers for the tasks
+    start = time.time()
+    with concurrent.futures.ThreadPoolExecutor() as executor:
+        futures = [executor.submit(process, snapshot, p) for p in partitions]
+
+        for future in futures:
```
jonparrott: Use `as_completed`, it's a wonderful thing.

spanner/cloud-client/requirements.txt
```diff
@@ -1 +1,2 @@
-google-cloud-spanner==1.1.0
+google-cloud-spanner==1.2.0
+futures==3.1.0
```
jonparrott: DPEbot will bump this back up.

Just do

```
futures; python_version < 3
```

spanner/cloud-client/batch_sample.py
```diff
@@ -0,0 +1,92 @@
...
+def process(snapshot, partition):
+    """Processes the requests of a query in an separate process."""
+    print('Started processing partition.')
+    try:
+        row_ct = 0
+        for row in snapshot.process_read_batch(partition):
+            print(u'SingerId: {}, AlbumId: {}, AlbumTitle: {}'.format(*row))
+            row_ct += 1
+        return time.time(), row_ct
+    except Exception as e:
```
jonparrott: I would just let the exception bubble up. The executor will print the whole stack trace, this will just print the error message.

kurtisvg: Done.

PR #1415, opened 2018-03-19
Update DLP samples for release

andrewsg:


dlp/metadata.py
```diff
@@ -21,11 +21,12 @@
 
 
 # [START list_info_types]
```
ace-n: This should be `dlp_list_info_types`, unless Python doesn't use region tags. (The lack of region tags may confuse @jmdobry's sample-tracking tools,  but I'm not sure.)

dlp/quickstart.py
```diff
@@ -51,22 +54,30 @@ def quickstart():
...
+    parent = dlp.project_path(project)
+
     # Call the API.
-    response = dlp.inspect_content(inspect_config, items)
+    response = dlp.inspect_content(parent, inspect_config, item)
 
     # Print out the results.
-    if response.results[0].findings:
-        for finding in response.results[0].findings:
+    if response.result.findings:
```
ace-n: Nit: is this necessary?

dlp/inspect_content_test.py
```diff
@@ -247,6 +249,7 @@ def test_inspect_gcs_file_no_results(
     assert 'No findings' in out
 
 
+@pytest.mark.skip(reason='nondeterministically failing')
```
ace-n: Add a 500ms timeout between Pub/Sub completion and job retrieval?

PR #1419, opened 2018-03-21
[Storage] Define additional properties in create bucket snippet

frankyn:
Users are confused by the `Client.create_bucket()` method as it doesn't support additional properties that would normally be used. This PR is to update the method to use `Bucket.create()` instead.


engelke:
Please address the reviewer's requested changes and update this PR.

engelke:
Closed with no activity. Reopen if needed.

PR #1420, opened 2018-03-21
[DO NOT MERGE] Vision API OCR PDF/TIFF sample

dizcology:
Add sample for Vision API document text detection support for PDF/TIFF file types.


vision/cloud-client/detect/docpdf.py
```diff
@@ -0,0 +1,89 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+"""OCR with PDF/TIFF as source files on GCS
+
+Example:
+    python docpdf.py --gcs-source-uri gs://python-docs-samples-tests/HodgeConj.pdf \
```
nnegrey: Personal wish: file name change. detect_beta or detect_pdf

vision/cloud-client/detect/docpdf.py
```diff
@@ -0,0 +1,89 @@
...
+
+    feature = vision.types.Feature(
+        type=vision.enums.Feature.Type.DOCUMENT_TEXT_DETECTION)
+
+    gcs_source = vision.types.GcsSource(uri=gcs_source_uri)
+    input_config = vision.types.InputConfig(
+        gcs_source=gcs_source, mime_type=mime_type)
+
+    gcs_destination = vision.types.GcsDestination(uri=gcs_destination_uri)
+    output_config = vision.types.OutputConfig(gcs_destination=gcs_destination, batch_size=batch_size)
```
nnegrey: Random thoughts: Python library is starting to look like Java.

vision/cloud-client/detect/docpdf.py
```diff
@@ -0,0 +1,89 @@
...
+
+    async_request = vision.types.AsyncAnnotateFileRequest(
+        features=[feature], input_config=input_config, output_config=output_config)
+
+    operation = client.async_batch_annotate_files(
+        requests=[async_request])
+
+    print('Waiting for the operation to finish.')
+    result = operation.result(90)
```
nnegrey: In a different sample, you had `operation.result(timeout=90)` is that clarification possible here?

vision/cloud-client/detect/docpdf.py
```diff
@@ -0,0 +1,89 @@
...
+
+    # Retrieve the first output file from GCS
```
nnegrey: You mentioned that the result object might be changing so this may be out of date.

Should we clarify that this is almost a 2nd step. 
"Once the request has completed and the output has been written to GCS. Retrieve ..."

vision/cloud-client/detect/docpdf.py
```diff
@@ -0,0 +1,89 @@
...
+    storage_client = storage.Client()
+
+    match = re.match(r'gs://([^/]+)/(.+)', gcs_destination_uri)
+    bucket_name = match.group(1)
+    object_name = match.group(2) + 'output-1-to-2.json'
```
nnegrey: Add a comment how you got to 'output-1-to-2.json'

I think you passed 'output' in the call. Since only '-1-to-2.json' gets attached by the api.

"Example gcs_destination_uri = 'gs://BUCKET_NAME/OCR/output' results in:
output-1-to-2.json for pages 1 and 2
output-3-to-4.json for pages 3 and 4, etc.
"

vision/cloud-client/detect/docpdf.py
remi: +1 let's make this more conventional with the other CLIs, I like detect_pdf

vision/cloud-client/detect/docpdf.py
remi: I +1 this thought. I'm not a Pythonista, but they've been Python GAPIC samples have starting to look kinda gnarly to me 🤷🏼

vision/cloud-client/detect/docpdf.py
```diff
@@ -0,0 +1,89 @@
...
+    operation = client.async_batch_annotate_files(
+        requests=[async_request])
+
+    print('Waiting for the operation to finish.')
+    result = operation.result(90)
+
+    # Retrieve the first output file from GCS
+    storage_client = storage.Client()
+
+    match = re.match(r'gs://([^/]+)/(.+)', gcs_destination_uri)
```
remi: Let's try to keep Regex out of all samples :)

We're in charge of what get passed into the CLI, so how about we pass Bucket and Object separately into the CLI?

vision/cloud-client/detect/docpdf.py
dizcology: changing to detect_pdf.py

vision/cloud-client/detect/docpdf.py
dizcology: best of the two worlds.

vision/cloud-client/detect/docpdf.py
dizcology: added.

vision/cloud-client/detect/docpdf.py
dizcology: changing this to a list_blobs, and then download the first.

vision/cloud-client/detect/docpdf.py
dizcology: for the CLI, sure - but the way the client is used is by passing in gcs_uris.

dizcology: The `urlsplit` methods included in the client library won't really help - for one thing it might be only in the python client library, and it doesn't automatically parse the prefix correctly:

```
In [1]: from google.cloud import storage

In [2]: storage.blob.urlsplit('gs://BUCKET/PREFIX/')
Out[2]: SplitResult(scheme='gs', netloc='BUCKET', path='/PREFIX/', query='', fragment='')
```

vision/cloud-client/detect/detect_pdf.py
```diff
@@ -0,0 +1,127 @@
...
+    # The actual response for the first page of the input file.
+    first_page_response = response.responses[0]
+
+    # Print the full text from the first page.
+    print(u'Full text:\n{}'.format(
+        first_page_response.full_text_annotation.text))
+
+    # The response additionally includes individual detected symbol's
+    # confidence and bounding box.
+    for page in first_page_response.full_text_annotation.pages:
```
nnegrey: What if we just made a comment that this information exists and didn't print it all out. Seems like a lot of output, when we could just print the full text and its confidence score. Thoughts?

nnegrey: or at least not every symbol needs to get printed.

dizcology: I copied that part from the vision detect document text sample.  the argument for showing this might be more for those who read code on cloud.google.com (and not so much those who actually come and run the samples), in that this way we expose more interesting structure of the response message and the reader can extract parts of this code for their use without having to run it nor having go through API reference docs.

vision/cloud-client/detect/detect_pdf.py
```diff
@@ -90,31 +90,14 @@ def async_detect_document(gcs_source_uri, gcs_destination_uri):
 
     # The actual response for the first page of the input file.
     first_page_response = response.responses[0]
+    annotation = first_page_response.full_text_annotation
 
-    # Print the full text from the first page.
+    # Here we print the full text from the first page.
+    # The response contains more information:
+    # annotation/pages/blocks/paragraphs/words/symbols
+    # including conficence score and bounding boxes
```
nnegrey: confidence

PR #1421, opened 2018-03-21
Add text-to-speech beta samples

nnegrey:


texttospeech/cloud-client/synthesize_text.py
```diff
@@ -0,0 +1,107 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Google Cloud Text-To-Speech API sample application .
+
+Example usage:
+    python synthesize_text.py --text "hello" --output hello.mp3
```
remi: All of the snippets currently write to `output.mp3` and sample functions don't accept a parameter for output file

`--output` is not an argument

Either add support for `--output` or remove from the example usage :)

nnegrey: Fixed.

texttospeech/cloud-client/list_voices.py
```diff
@@ -0,0 +1,55 @@
...
+
+        # Display the supported language codes for this voice. Example: "en-US"
+        for language_code in voice.language_codes:
+            print('Supported language: {}'.format(language_code))
+
+        # Names of SSML voice genders from google.cloud.texttospeech.enums
+        ssml_voice_genders = ['SSML_VOICE_GENDER_UNSPECIFIED', 'MALE',
+                              'FEMALE', 'NEUTRAL']
+
+        # Display the supported SSML - gender for this voice. Example: FEMALE
```
remi: How about:

```py
# SSML Voice Gender values from google.cloud.texttospeech.enums
ssml_voice_genders = ['SSML_VOICE_GENDER_UNSPECIFIED', 'MALE',
                      'FEMALE', 'NEUTRAL']

# Display the SSML Voice Gender
print('SSML Voice Gender: {}'.format(ssml_voice_genders[voice.ssml_gender]))
```

(And apply same to Java)

Simply making it as clear as possible that the values used map to the SSML spec, see Voice element here: https://www.w3.org/TR/speech-synthesis11/#g13

remi: #nit

It's good as is! The only thing I really would like to change is this:

    the supported SSML - gender for this voice

Rather than "SSML - gender" which separates the 2, say "SSML gender" to be clear that the `gender` is SSML defined

nnegrey: Updated

texttospeech/cloud-client/resources/hello.ssml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0"?>
```
remi: Simplify this :)

For out samples, let's just use simple SSML:

```xml
<speak>Hello there.</speak>
```

Per request from product :)

texttospeech/cloud-client/quickstart.py
```diff
@@ -0,0 +1,61 @@
...
+    # Instantiates a client
+    client = texttospeech.TextToSpeechClient()
+
+    # Set the text input to be synthesized
+    synthesis_input = texttospeech.types.SynthesisInput(text="Hello, World!")
+
+    # Build the voice request, select the language code ("en-US") and the ssml
+    # voice gender ("neutral")
+    voice = texttospeech.types.VoiceSelectionParams(language_code='en-US',
+                                                    ssml_gender='NEUTRAL')
```
dizcology: ssml_gender should have its own enums, please use enums to specify this field.

texttospeech/cloud-client/synthesize_file_test.py
```diff
@@ -0,0 +1,31 @@
...
+
+TEXT_FILE = 'resources/hello.txt'
+SSML_FILE = 'resources/hello.ssml'
+
+
+def test_synthesize_text_file(capsys):
+    synthesize_file.synthesize_text_file(text_file=TEXT_FILE)
+    out, err = capsys.readouterr()
+
+    assert 'Audio content written to file' in out
```
dizcology: if possible assert that the output file has size greater than zero.  (doing this in only one of the tests should be enough)

texttospeech/cloud-client/quickstart.py
nnegrey: done

texttospeech/cloud-client/synthesize_file_test.py
nnegrey: done

dizcology: looks good.  just need a new line between import os and import synthesize_file/text.

nnegrey: Question: is there a flake8 config file I should grab, that would catch that?

andrewsg: Use nox (`pip install nox-automation`; then `nox -l` with grep to find the lint session for your particular subproject, then `nox -s "lint(your-subproject-here)") to use the repo's standard flake8 config.

PR #1446, opened 2018-04-10
add Speech API auto punctuation sample

dizcology:


telpirion:
The auto punctuation sample helps greatly. Thank you!

speech/cloud-client/beta_snippets.py
```diff
@@ -99,6 +100,32 @@ def transcribe_file_with_metadata(path):
 # [END speech_transcribe_file_with_metadata]
 
 
+# [START speech_transcribe_file_with_auto_punctuation]
+def transcribe_file_with_auto_punctuation(path):
+    """Transcribe the given audio file using an enhanced model."""
```
nnegrey: the "enhanced model" bit, this is where I'm confused between the one type of enhanced models and the model selected enhanced models. (Not sure if this point really matters though)

dizcology: I failed to update the docstring - fixing it now.

jgeorge-burwood:
Is this functionality available in python client library 0.33.0?  Trying to test  but its coming back with:  Protocol message RecognitionConfig has no "enable_automatic_punctuation" field.

jgeorge-burwood:
solved -- I needed the types and enums from v1p1beta1 as well.  doh!

dalmam11:
jgeorge-burwood I have the same issue where did you found v1p1beta1?

dalmam11:
cannot import name 'speech_v1p1beta1'

jgeorge-burwood:
dalmam11 make sure you're using at least 0.33.0 of the client library.   
pip install --upgrade google-cloud-speech

Then just add the imports, you'll probably want something like:
...
from google.cloud import speech_v1p1beta1 as speech
from google.cloud.speech_v1p1beta1 import enums
...

PR #1449, opened 2018-04-13
Fixing Endpoints GRPC samples

simonz130:
Endpoints GRPC samples were broken and outdated.

This change is the first to fix that - currently getting-started sample only:

1. Update dependencies
2. Update generated grpc code
3. Update copyrights to 2018


endpoints/getting-started-grpc/api_config.yaml
```diff
@@ -1,4 +1,4 @@
-# Copyright 2017 Google Inc.
+# Copyright 2018 Google Inc.
```
kurtisvg: I think we are supposed to leave these at the original dates since they aren't new files.

Everything else LGTM.

PR #1459, opened 2018-04-18
BigQuery: Removes unused API samples

alixhami:
Removes API samples, which are no longer used on devsite.


bigquery/api/resources/data.csv
```diff
@@ -1 +0,0 @@
-Gandalf,2000,140.0,1
```
tswast: You might want to double-check that none of the cloud-client sample tests use these test data files.

alixhami: The snippets tests are passing and I didn't see any usage of data files, which prompted me to also delete the unused cloud-client data files.

PR #1468, opened 2018-05-01
Composer: initial workflow and REST samples

tswast:
These samples are to be included in the Cloud Composer docs at https://cloud.google.com/composer/docs/

Note that these samples have already been reviewed via Gerrit.


tswast:
Question: Are we still using Travis? I see the Kokoro configs and thought we had migrated.

michaelawyu:
Hi Tim! I think Travis is still used.

tswast:
A composer environment is currently being created for the test project, and I've updated the Travis environment secrets file to reflect it. 🤞

PR #1481, opened 2018-05-15
Adding ability to make reqeusts of all method types: GET, POST, PUT etc

danieldeleo:


iap/make_iap_request.py
```diff
@@ -30,12 +30,16 @@
...
-def make_iap_request(url, client_id):
+def make_iap_request(url, client_id, method='GET', **kwargs):
     """Makes a request to an application protected by Identity-Aware Proxy.
 
     Args:
       url: The Identity-Aware Proxy-protected URL to fetch.
       client_id: The client ID used by Identity-Aware Proxy.
+      method: The request method to use
+              ('GET', 'OPTIONS', 'HEAD', 'POST', 'PUT', 'PATCH', 'DELETE')
+      **kwargs: Any of the parameters defined for the Request class:
```
andrewsg: In this case, I think it's the `request function` rather than the request class.

iap/make_iap_request.py
```diff
@@ -30,12 +30,16 @@
...
+                https://github.com/requests/requests/blob/9616f82905f86173477fdb3bf24f2f7228eed3a0/requests/api.py#L16
```
andrewsg: Let's link to the default branch, without a line specified (https://github.com/requests/requests/blob/master/requests/api.py). Our requests version is updated by an automated system, so unless we are very general in our link it will be out of date almost immediately.

danieldeleo: Done

PR #1491, opened 2018-05-16
PubSub: adds region tags and updates existing to standard

alixhami:


pubsub/cloud-client/quickstart.py
```diff
@@ -16,7 +16,7 @@
 
 
 def run_quickstart():
-    # [START pubsub_quickstart]
+    # [START pubsub_quickstart_create_topic]
```
tswast: I assume you've locked the docs via the branch parameter already?

PR #1493, opened 2018-05-17
Remove  mysqlclient==1.3.12 as it is not mandatory

ltupin:
And it could make troubles. For example on my side I got a `OSError: mysql_config not found` error.


ltupin:
I signed it!

andrewsg:
Hmm... I agree with the sentiment, but right now the docs have instructions on how to switch from postgres to mysql, and if we remove this, those instructions will no longer be accurate. I think we will have to decline this PR until we can resolve that in a satisfactory manner. Thank you for bringing the problem to my attention, though, I appreciate your interest!

PR #1494, opened 2018-05-18
BigQuery: Moves BigQuery tutorial for Dataproc to python-docs-samples

alixhami:
Adds BigQuery portion of [Dataproc tutorial](https://cloud.google.com/dataproc/docs/tutorials/bigquery-sparkml) to this repo so it can be regularly tested (previously hard coded in the docs).

### DIFF
--- a/bigquery/cloud-client/natality_tutorial.py
+++ b/bigquery/cloud-client/natality_tutorial.py
@@ -21,11 +21,9 @@ def run_natality_tutorial():
 
     In the code below, the following actions are taken:
     * A new dataset is created "natality_regression."
-    * A new table "regression_input" is created to hold the inputs for our
-        linear regression.
     * A query is run against the public dataset,
         bigquery-public-data.samples.natality, selecting only the data of
-        interest to the regression, the output of which is stored in the
+        interest to the regression, the output of which is stored in a new
         "regression_input" table.
     * The output table is moved over the wire to the user's default project via
         the built-in BigQuery Connector for Spark that bridges BigQuery and
--- a/bigquery/cloud-client/natality_tutorial.py
+++ b/bigquery/cloud-client/natality_tutorial.py
@@ -53,9 +53,6 @@ def run_natality_tutorial():
     # Set the destination table to the table reference created above.
     job_config.destination = table_ref
 
-    # BigQuery can auto-detect the schema based on the source table.
-    job_config.autodetect = True
-
     # Set up a query in Standard SQL, which is the default for the BigQuery
     # Python client library.
     # The query selects the fields of interest.
--- a/bigquery/cloud-client/natality_tutorial.py
+++ b/bigquery/cloud-client/natality_tutorial.py
@@ -1,0 +1,95 @@
+#!/usr/bin/env python
+
+# Copyright 2018 Google Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


bigquery/cloud-client/natality_tutorial.py
```diff
@@ -0,0 +1,95 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+def run_natality_tutorial():
+    # [START bigquery_query_natality_tutorial]
```
tswast: Question: do we want to own this as a BigQuery sample or should it go under Dataproc?

I'm leaning towards BigQuery, too, since there doesn't appear to be anything Dataproc-specific here.

bigquery/cloud-client/natality_tutorial.py
```diff
@@ -0,0 +1,95 @@
...
+    """Create a Google BigQuery linear regression input table.
+
+    In the code below, the following actions are taken:
+    * A new dataset is created "natality_regression."
+    * A new table "regression_input" is created to hold the inputs for our
```
tswast: We don't have to actually create the table, right? The query job should be able to do that for us, so long as the [`configuration.query.createDisposition` BigQuery job property](https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs) is correct.

bigquery/cloud-client/natality_tutorial.py
```diff
@@ -0,0 +1,95 @@
...
+    # Prepare a reference to the new dataset.
+    dataset_ref = client.dataset(dataset_id)
+    dataset = bigquery.Dataset(dataset_ref)
+
+    # Create the new BigQuery dataset.
+    dataset = client.create_dataset(dataset)
+
+    # In the new BigQuery dataset, create a new table.
+    table_ref = dataset.table('regression_input')
+    # The table needs a schema before it can be created and accept data.
```
tswast: I actually think the query job will run just fine without a schema, since BigQuery knows all they correct types when it runs the query.

bigquery/cloud-client/natality_tutorial.py
```diff
@@ -0,0 +1,95 @@
...
+    table = client.create_table(table)
+
+    # Set up a query in Standard SQL, which is the default for the BigQuery
+    # Python client library.
+    # The query selects the fields of interest.
+    query = """
+        SELECT weight_pounds, mother_age, father_age, gestation_weeks,
+        weight_gain_pounds, apgar_5min
+        FROM `bigquery-public-data.samples.natality`
+        WHERE weight_pounds is not null
```
tswast: Let's use consistent casing for SQL keywords `AND` `IS` `NOT` `NULL`.

bigquery/cloud-client/natality_tutorial.py
alixhami: I put it as BigQuery and didn't include Dataproc in the name because there's nothing Dataproc-related in this sample (except the docstring, which we could move into the docs instead of the sample) and we could potentially leverage this sample for other purposes.

bigquery/cloud-client/natality_tutorial.py
```diff
@@ -38,52 +38,43 @@ def run_natality_tutorial():
...
-    table = client.create_table(table)
+
+    # Configure the query job.
+    job_config = bigquery.QueryJobConfig()
+
+    # Set the destination table to the table reference created above.
+    job_config.destination = table_ref
+
+    # BigQuery can auto-detect the schema based on the source table.
+    job_config.autodetect = True
```
tswast: QueryJobConfig doesn't have an `autodetect` parameter. This line is not needed.

Sorry if my previous comment was confusing. I mean that when BigQuery runs a query, it knows the types of the various columns in the query results based on the query text, so no explicit schema is needed.

alixhami: Oops removed the nonsense attribute

PR #1495, opened 2018-05-18
Update monitoring samples to use version 0.29.0 of the client library.

SurferJeffAtGoogle:


monitoring/api/v3/cloud-client/snippets.py
```diff
@@ -13,126 +13,215 @@
...
+def project_id():
+    """Retreives the project id from the environment variable.
+
+    Raises:
+        MissingProjectIdError -- When not set.
+
+    Returns:
+        str -- the project name
+    """
+    project_id = os.environ['GCLOUD_PROJECT']
```
tswast: Could we check the `GOOGLE_CLOUD_PROJECT` env var first and use `GCLOUD_PROJECT` as a fallback?

SurferJeffAtGoogle: I haven't seen that in any of the python samples I've come across.  Is that a new standard?

PR #1500, opened 2018-05-23
Update Endpoints gRPC bookstore example

kryzthov:
- Now works with Python2 and Python3
 - Updated ProtoBuf/gRPC generated code

@simonz130 Could you take a look?


endpoints/bookstore-grpc/bookstore_pb2.py
```diff
@@ -23,7 +23,6 @@
...
   ,
   dependencies=[google_dot_protobuf_dot_empty__pb2.DESCRIPTOR,])
-_sym_db.RegisterFileDescriptor(DESCRIPTOR)
```
simonz130: Do we need to checkin this file?

endpoints/bookstore-grpc/bookstore_pb2_grpc.py
```diff
@@ -0,0 +1,170 @@
+# Generated by the gRPC Python protocol compiler plugin. DO NOT EDIT!
```
simonz130: do we need to checkin this file?
I think it's better to update instructions that the user will generate the file rather than checkin autogenerated code.

endpoints/bookstore-grpc/README.md
```diff
@@ -55,8 +55,8 @@ gRPC libraries and tools are installed, run:
         --include_imports \
         --include_source_info \
         --proto_path=. \
-        --python_out=generated_pb2 \
-        --grpc_python_out=generated_pb2 \
+        --python_out=. \
```
simonz130: will this put the files in the same directory?

endpoints/bookstore-grpc/bookstore_pb2_grpc.py
kryzthov: This file was checked-in before.
It is checked-in in all gRPC examples, as far as I can tell.

endpoints/bookstore-grpc/README.md
kryzthov: Yes, this wasn't previously needed because the sample relied on an beta version of gRPC that combined the protobuf generated source and the grpc generated source.

With this update, the protoc command generates 2 files, one of which depends on the other, with a specific module path (hence the location of these file at the root).

endpoints/bookstore-grpc/bookstore_pb2_grpc.py
simonz130: I want to change it so autogenerated files will not be checked in - everywhere. However I'm fine to do it in stages - first unblock this sample as-is and then update all the places to not include autogenerated files (separate checkin).

PR #1503, opened 2018-05-26
modify middleware settings and url paths for django 2.0

hirokts:
Please check the django 2.0 middleware settings  of the official document.

(pip requirements.txt:    Django==2.0.3)

https://docs.djangoproject.com/en/2.0/topics/http/middleware/


container_engine/django_tutorial/Dockerfile
```diff
@@ -21,11 +21,11 @@ FROM gcr.io/google_appengine/python
 
 # Create a virtualenv for the application dependencies.
 # # If you want to use Python 3, add the -p python3.4 flag.
-RUN virtualenv /env
+RUN virtualenv -p python3.4 /env
```
sketchytechky: I bumped this to 3.6 and confirmed it works 👌

sketchytechky: I bumped this to 3.6 and confirmed it works 👌

hirokts: Thanks. I changed the version.

container_engine/django_tutorial/Dockerfile
```diff
@@ -21,11 +21,11 @@ FROM gcr.io/google_appengine/python
 
 # Create a virtualenv for the application dependencies.
 # # If you want to use Python 3, add the -p python3.4 flag.
```
andrewsg: We should change this comment, since python36 is preselected.

hirokts: Thanks, I changed it.

hirokts:
I confirmed it's ok to do so.
The reason of this change is "middleware settings needs to be changed due to differences in version of django", I prefer declarative version specification. Although, assuming automatically updating the version of django (in pip requirements text) or python (in docker image),  it's better to use "-p python3" or "python3 -m venv".

andrewsg:
We do often update versions of dependencies such as Django as part of our maintenance (via bots). Given that, should we change it to -p python3 or do you think it should stay as it is?

morenoh149:
I fixed these issues myself. I also had to make the changes described here https://gist.github.com/morenoh149/39be1ace1f99b9571fd2b6b96098075d I could submit these as a separate PR but may make logical sense to include in this PR

hirokts:
Thanks.
Thinking about the situation, I changed python version specification.

I think it's better to change url path settings for django 2.
https://docs.djangoproject.com/en/2.0/ref/urls/
But I cannot confirm that immediately.

morenoh149:
lgtm

hirokts:
@andrewsg ,  could you check this PR?　Editing my comment may have been wrong if you did not notice the PR changing.

andrewsg:
lgtm. Thanks!

PR #1506, opened 2018-05-29
Update cloudiot_pubsub_example_server.py

subhamoydey:
Doing an learning using Google codelab

### DIFF
--- a/iot/api-client/end_to_end_example/cloudiot_pubsub_example_server.py
+++ b/iot/api-client/end_to_end_example/cloudiot_pubsub_example_server.py
@@ -56,10 +56,9 @@ SERVICE_NAME = 'cloudiot'
 
 class Server(object):
     """Represents the state of the server."""
-
-    def __init__(self, service_account_json):
-        credentials = ServiceAccountCredentials.from_json_keyfile_name(
-            service_account_json, API_SCOPES)
+def __init__(self, service_account_json):        
+        from google.auth import compute_engine
+        credentials = compute_engine.Credentials()
         if not credentials:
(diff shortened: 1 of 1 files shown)


iot/api-client/end_to_end_example/cloudiot_pubsub_example_server.py
```diff
@@ -56,10 +56,9 @@
 
 class Server(object):
     """Represents the state of the server."""
-
-    def __init__(self, service_account_json):
-        credentials = ServiceAccountCredentials.from_json_keyfile_name(
-            service_account_json, API_SCOPES)
+def __init__(self, service_account_json):
```
gguuss: I'm not sure this is the intended indentation as the __init__ block should be inside the Server class.

iot/api-client/end_to_end_example/cloudiot_pubsub_example_server.py
```diff
@@ -56,10 +56,9 @@
 
 class Server(object):
     """Represents the state of the server."""
-
-    def __init__(self, service_account_json):
-        credentials = ServiceAccountCredentials.from_json_keyfile_name(
-            service_account_json, API_SCOPES)
+def __init__(self, service_account_json):        
+        from google.auth import compute_engine
```
gguuss: This looks like it's the content from the codelab that I authored for Google I/O this year -- this sample (as opposed to the codelab) is more flexible without this import and credential initialization.

gguuss:
Closing - for now, let's use explicit credential loading instead of DAC. If you feel strongly about this, please reopen the PR and explain.

PR #1510, opened 2018-05-30
Storage: add KMS samples

broady:


broady:
Still TODO:

* Set up `CLOUD_KMS_KEY` environment variable for CI.
* Add test for bucket default KMS key function.

@andrewsg @theacodes can you PTAL whenever you get a chance? wanted to get this up quick for review.

broady:
Should I make the `CLOUD_KMS_KEY` env var optional? Probably yes?

That'd be something like moving it into the test body and catching `KeyError` for py36 and checking if empty for py27?

theacodes:
Nah you don't need to make it optional.

andrewsg:
It looks like our Travis build is failing because of the CLOUD_KMS_KEY thing. Let's either revisit the question of whether to make it optional, or else add it to the env vars on our test environment.

frankyn:
Can we add it to env vars?

frankyn:
Is there a way I can help here?

frankyn:
Added CLA:yes, I'm okay with my contributions being added to repo.

andrewsg:
Thanks!

frankyn:
Two items pending and will send a follow-up PR for:
* Add test for bucket default KMS key function.
* Add region tags.

PR #1512, opened 2018-06-05
Fixed Initialize Firebase section and add snippet

mdejong3:
RE: https://cloud.google.com/appengine/docs/standard/python/authenticating-users-firebase-appengine#adding_the_firebase_authentication_user_interface
This fails because the projectId and messagingSenderId are missing. I have also added snippet anchors to include snippet in the doc.

### DIFF
--- a/appengine/standard/firebase/firenotes/frontend/main.js
+++ b/appengine/standard/firebase/firenotes/frontend/main.js
@@ -19,7 +19,7 @@ $(function(){
   // backend's app.yaml file.
   var backendHostUrl = '<your-backend-url>';
 
-  // [START config]
+  // [START gae_firenotes_config]
   // Obtain the following from the "Add Firebase to your web app" dialogue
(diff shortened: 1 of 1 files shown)


mdejong3:
Registered id for cla

appengine/standard/firebase/firenotes/frontend/main.js
```diff
@@ -19,15 +19,19 @@ $(function(){
   // backend's app.yaml file.
   var backendHostUrl = '<your-backend-url>';
 
+  // [START config]
```
andrewsg: My understanding is that now that we have the sample tracker, we need to use tag titles that are non-ambiguous even across files. Perhaps something like [START gae_firenotes_config]?

mdejong3: Done!

andrewsg:
Great, thanks!

PR #1517, opened 2018-06-11
[Storage] Update kms samples

frankyn:
Adding:
* region tags
* missing test


frankyn:
@andrewsg PTAL, thank you

PR #1518, opened 2018-06-11
Fix the vision geo test.

SurferJeffAtGoogle:
Sometimes, Vision sees Zepra.  Othertimes, it sees Electra Tower.


dizcology:
Looks fine.  Let's plan to update these sample tests to make them more robust and less dependent on the results.

PR #1520, opened 2018-06-13
Composer: Add schedule to simple workflow.

tswast:
Without a schedule this DAG gets scheduled way too often when I upload it to a Cloud Composer environment.


tswast:
Merged as TBR since this is blocking an internal CL to set up integration tests for these workflows.

PR #1522, opened 2018-06-13
Move ocr pdf/tiff samples to GA

nnegrey:


vision/cloud-client/detect/README.rst
```diff
@@ -94,12 +94,13 @@ To run this sample:
     python detect.py web-uri http://wheresgus.com/dog.JPG
     python detect.py web-geo ./resources/city.jpg
     python detect.py faces-uri gs://your-bucket/file.jpg
+    python detect_pdf.py ocr-uri gs://python-docs-samples-tests/HodgeConj.pdf     gs://BUCKET_NAME/PREFIX/
```
andrewsg: This example has a bunch of extra whitespace, is that intentional?

vision/cloud-client/detect/detect.py
```diff
@@ -636,6 +641,77 @@ def detect_document_uri(uri):
...
 
 
+# [START vision_async_detect_document_ocr]
+def async_detect_document(gcs_source_uri, gcs_destination_uri):
+    """OCR with PDF/TIFF as source files on GCS"""
+    # Supported mime_types are: 'application/pdf' and 'image/tiff'
+    mime_type = 'application/pdf'
+
+    # How many pages should be grouped into each json output file.
+    # With a file of 5 pages
```
andrewsg: I don't understand this part, "with a file of 5 pages", is it cut off?

nnegrey: The API lets you decide how many pages you want each output file to be. 

So its reads a PDF and it writes a JSON file. Then you can decide each json file gets 1-### of pages of the PDF.

vision/cloud-client/detect/README.rst
nnegrey: Ah, well the line was too long in detect.py, so I added spaces to make it easier to read, but I'll remove the spaces and regen the README to make it cleaner.

vision/cloud-client/detect/detect.py
andrewsg: That makes sense, but what does the sentence fragment "With a file of 5 pages" mean specifically?

nnegrey: Oh... Yea, I think that is referring to the sample PDF we provide, I'll pull that.

andrewsg:
Thanks!

PR #1524, opened 2018-06-13
Add DLP code samples for custom info types

mwdaub:
The DLP team is in the process of expanding code samples to include features that we have launched since the initial samples were written.

In this pull request, I updated inspect_content.py to include dictionary and regex based custom info types, and added some tests in inspect_content_test.java. Also updated deid.py and deid_test.py to fix issues with the tests since launching the V2 API.

### DIFF
--- a/dlp/inspect_content.py
+++ b/dlp/inspect_content.py
@@ -66,7 +66,7 @@ def inspect_string(project, content_string, info_types,
     regexes = [{
         'info_type': {'name': 'CUSTOM_REGEX_{}'.format(i)},
         'regex': {'pattern': custom_regex}
-    } for i in enumerate(custom_regexes)]
+    } for i, custom_regex in enumerate(custom_regexes)]
     custom_info_types = dictionaries + regexes
 
     # Construct the configuration dictionary. Keys which are None may
@@ -154,7 +154,7 @@ def inspect_file(project, filename, info_types, min_likelihood=None,
     regexes = [{
         'info_type': {'name': 'CUSTOM_REGEX_{}'.format(i)},
         'regex': {'pattern': custom_regex}
-    } for i in enumerate(custom_regexes)]
+    } for i, custom_regex in enumerate(custom_regexes)]
     custom_info_types = dictionaries + regexes
 
     # Construct the configuration dictionary. Keys which are None may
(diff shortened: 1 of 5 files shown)


mwdaub:
Note that the Travis CI build failure had nothing to do with my changes, and seems to be some other problem with installing python3.6 - not sure how to restart it.

dlp/inspect_content.py
```diff
@@ -50,10 +51,29 @@ def inspect_string(project, content_string, info_types,
...
+    # Prepare custom_info_types by parsing the dictionary word lists and
+    # regex patterns.
+    if custom_dictionaries is None:
+        custom_dictionaries = []
+    dictionaries = [{
+        'info_type': {'name': 'CUSTOM_DICTIONARY_{}'.format(i)},
+        'dictionary': {
+            'word_list': {'words': custom_dictionaries[i].split(',')}
+        }
+    } for i in range(len(custom_dictionaries))]
```
andrewsg: Rather than using `for i in range(len())` can we iterate directly over the values and use the `enumerate()` builtin to get the numbers? (Same throughout)

dlp/inspect_content.py
```diff
@@ -648,6 +769,16 @@ def callback(message):
              'include "FIRST_NAME", "LAST_NAME", "EMAIL_ADDRESS". '
              'If unspecified, the three above examples will be used.',
         default=['FIRST_NAME', 'LAST_NAME', 'EMAIL_ADDRESS'])
+    parser_gcs.add_argument(
+        '--custom_dictionaries', action='append',
+        help='Strings representing comma-delimited lists of dictionary words'
```
andrewsg: Is this intended to be a single string of comma-delimited words, or multiple strings of one word each, or multiple comma-delimited strings?

mwdaub:
Working on this today.

dlp/inspect_content.py
mwdaub: Done.

dlp/inspect_content.py
mwdaub: It is intended to be multiple strings, each of which is a string of comma-delimited words; I have updated the help message to make this more clear. This is how the other code samples are written.

andrewsg:
Tests appear to be red currently due to this:

```
./inspect_content.py:68:30: F821 undefined name 'custom_regex'
        'regex': {'pattern': custom_regex}

```
Let me know when you're ready for another look.

mwdaub:
Tests all pass. PTAL

andrewsg:
Great, thanks!

PR #1533, opened 2018-06-20
Make Django 2.0 docker image build

morenoh149:
Django 2.0 requires python3, update dockerfile to use python3.

related https://github.com/GoogleCloudPlatform/python-docs-samples/pull/1503


morenoh149:
#1503 now contains these changes

PR #1535, opened 2018-06-21
BigQuery: Adds authorized view tutorial

alixhami:
Python version of [authorized view tutorial](https://cloud.google.com/bigquery/docs/share-access-views)

I'm open to ideas on phrasing region tags. My intent was to have sections to match up with the tutorial, but also have a region tag that encompassed the whole thing so I could show it at the end of the tutorial.

### DIFF
--- a/bigquery/cloud-client/authorized_view_tutorial.py
+++ b/bigquery/cloud-client/authorized_view_tutorial.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 
-# Copyright 2018 Google Inc. All Rights Reserved.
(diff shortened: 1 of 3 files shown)


bigquery/cloud-client/authorized_view_tutorial.py
```diff
@@ -0,0 +1,108 @@
+#!/usr/bin/env python
+
+# Copyright 2018 Google Inc. All Rights Reserved.
```
tswast: Nit. Per go/releasing/preparing Google LLC, drop the "all rights reserved."

PR #1541, opened 2018-06-25
update Tasks Sample for App Engine

averikitsch:


appengine/flexible/tasks/app.yaml
```diff
@@ -1,3 +1,17 @@
+# Copyright 2018 Google Inc. All Rights Reserved.
```
andrewsg: We don't need the copyright header for .yaml files, I believe.

andrewsg:
Thanks! Go ahead and merge yourself when you're ready (optionally addressing the comment above)

PR #1565, opened 2018-07-15
Create quickstart python

sangramql:
Create quickstart for bigtable in python


sduskis:
This should have had a readme, and just a single print statement.

PR #1569, opened 2018-07-16
Downgrades Django to 1.11

### LINKED ISSUE #1451: Django 2.0 doesn't support Python 2.7 mismatch with official Cloud Platform tutorial
https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/appengine/flexible/django_cloudsql/requirements.txt

No.

## Describe the issue
In the following [official GCP tutorial], it states "This tutorial requires Python 2.7 or 3.4." and then it makes you use [this GitHub repo]. In the "requirements.txt" file, Django version is: "Django==2.0.3" . However [Django 2.0 doesn't support Python 2.7] ("Django 1.11.x series is the last to support Python 2.7")

[official GCP tutorial]: https://cloud.google.com/python/django/flexible-environment 
[this GitHub repo]: https://github.com/GoogleCloudPlatform/python-docs-samples/tree/master/appengine/flexible/django_cloudsql
[Django 2.0 doesn't support Python 2.7]: https://docs.djangoproject.com/en/2.0/releases/2.0/#python-compatibility

gakeseaton:
Downgrades Django to 1.11 to close #1451 

Removes unnecessary mysqlclient (tutorial uses postgresql)

### DIFF
--- a/appengine/flexible/django_cloudsql/requirements.txt
+++ b/appengine/flexible/django_cloudsql/requirements.txt
@@ -1,4 +1,4 @@
-Django==1.11.13
+Django<=1.11.13
 wheel==0.31.0
 gunicorn==19.7.1
 psycopg2==2.7.4
--- a/appengine/flexible/django_cloudsql/requirements.txt
+++ b/appengine/flexible/django_cloudsql/requirements.txt
@@ -1,5 +1,4 @@
-Django==2.0.3
-mysqlclient==1.3.12
+Django==1.11.13
 wheel==0.31.0
 gunicorn==19.7.1
 psycopg2==2.7.4


appengine/flexible/django_cloudsql/requirements.txt
```diff
@@ -1,5 +1,4 @@
-Django==2.0.3
-mysqlclient==1.3.12
+Django==1.11.13
```
kurtisvg: It may be better to specify the version based on the pip version with [requirement specifiers](http://pip.readthedocs.io/en/stable/reference/pip_install/#requirement-specifiers)

tswast: This LGTM, but dpebot will probably swing through and update it again if we don't also change https://github.com/GoogleCloudPlatform/repository-gardener

tswast:
I talked about this PR with Kurtis. I think the tutorial should be updated to indicate that it runs with the latest version of Django.

If we use an old version of Django in the tutorial there is a risk that Django developers will assume the latest version isn't supported.

PR #1572, opened 2018-07-17
[DO_NOT_MERGE] Add samples for object localization and handwritten ocr

nnegrey:


vision/cloud-client/detect/beta_snippets.py
```diff
@@ -0,0 +1,210 @@
...
+
+    with open(path, 'rb') as image_file:
+        content = image_file.read()
+    image = vision.types.Image(content=content)
+
+    objects = client.object_localization(
+        image=image).localized_object_annotations
+
+    print('Number of objects found: {}'.format(len(objects)))
+    for object in objects:
```
dizcology: because `object` is a Pyhton built-in global variable, perhaps use `object_` instead.  the plural `objects` is fine though.

PR #1574, opened 2018-07-17
Cleans up corrupting Git breadcrumbs

kmonsoor:
Cleans up corrupting Git breadcrumbs from `requirements.txt`, caused by previous commit: https://github.com/GoogleCloudPlatform/python-docs-samples/commit/0f26000946455cba56ee0b9e07ed88a2d2858116


engelke:
Thanks for contributing this change, but it turns out to not be needed. The requirements versions are regularly updated by a bot and tested before merging, so this change is redundant.

PR #1575, opened 2018-07-17
automl beta

sirtorry:
add automl samples and tests


translate/automl/automl_translation_model.py
```diff
@@ -0,0 +1,298 @@
...
+    """Get operation status.
+    Args:
+        operation_full_id: Full name of a operation.
+            For example, the name of your operation is
+            projects/<projectId>/locations/us-central1/operations/<operationId>
+    """
+    client = automl.AutoMlClient()
+
+    # Get the latest state of a long-running operation.
+    response = client.transport._operations_client.get_operation(
```
crwilcox: Can you explain why you need to be able to get the operation status? If you think there is a legitimate use case for this we may want to consider adding a method on the client object for this, since you are relying on an internal object here.

sirtorry: these training jobs can be long running, up to a day.

crwilcox: Alright. I am going to create an issue on the library for us to consider this as a change. Thanks!

language/automl/automl_natural_language_dataset.py
```diff
@@ -0,0 +1,300 @@
+#!/usr/bin/env python
+
+# Copyright 2018 Google Inc. All Rights Reserved.
```
happyhuman: I was told we are supposed to start using LLC instead of Inc from now on.

language/automl/automl_natural_language_dataset.py
```diff
@@ -0,0 +1,300 @@
...
+# limitations under the License.
+
+"""This application demonstrates how to perform basic operations on Dataset
+with the Google AutoML Natural Language API.
+
+For more information, see the tutorial page at
+https://cloud.google.com/natural-language/automl/docs/
+"""
+
+# [START automl_natural_language_import]
```
happyhuman: Do we need region tags for imports too? I did not know.

language/automl/automl_natural_language_dataset.py
```diff
@@ -0,0 +1,300 @@
...
+import argparse
+import os
+
+from google.cloud import automl_v1beta1 as automl
+
+# [END automl_natural_language_import]
+
+
+# [START automl_natural_language_create_dataset]
```
happyhuman: Tim Swast advised me to put the region tags inside the function body. He told me this is the new standard.

language/automl/automl_natural_language_dataset.py
```diff
@@ -0,0 +1,300 @@
...
+
+For more information, see the tutorial page at
+https://cloud.google.com/natural-language/automl/docs/
+"""
+
+# [START automl_natural_language_import]
+import argparse
+import os
+
+from google.cloud import automl_v1beta1 as automl
```
happyhuman: I believe for samples this import should be inside the function body. That way then the code snippet is selected using the region tags, the imports will also be included in the selected snippet.

language/automl/automl_natural_language_dataset.py
```diff
@@ -0,0 +1,300 @@
...
+    Args:
+        project_id: Id of the project.
+        compute_region: Region name.
+        dataset_name: Name of the dataset.
+        multilabel: Type of the classification problem.
+            False - MULTICLASS, True - MULTILABEL.
+            Default is False.
+    """
+    client = automl.AutoMlClient()
+
```
happyhuman: You need these comment lines here (or something like this):

    # TODO(developer): Uncomment and set to a path to your audio file.
    # project_id = 'PROJECT_ID_HERE'
    # compute_region = 'COMPUTE_REGION_HERE'
    # dataset_name = 'DATASET_NAME_HERE'

The reason is that once the function body is selected to be shown in some web page using the region tags, these comments tell the user what these parameters should be.

language/automl/automl_natural_language_dataset.py
```diff
@@ -0,0 +1,300 @@
...
+import os
+
+from google.cloud import automl_v1beta1 as automl
+
+# [END automl_natural_language_import]
+
+
+# [START automl_natural_language_create_dataset]
+def create_dataset(project_id, compute_region, dataset_name, multilabel=False):
+    """Create a dataset.
```
happyhuman: Not sure if you need to describe the parameters in this comment. Just a one line comment to tell what the sample is should be good enough.

language/automl/automl_natural_language_dataset.py
```diff
@@ -0,0 +1,300 @@
...
+    """
+    client = automl.AutoMlClient()
+
+    # A resource that represents Google Cloud Platform location.
+    project_location = client.location_path(project_id, compute_region)
+
+    # Classification type is assigned based on multilabel value.
+    classification_type = "MULTICLASS"
+    if multilabel:
+        classification_type = "MULTILABEL"
```
happyhuman: Nothing wrong with the lines above, a more compact way is of course:
classification_type = "MULTILABEL" if multilabel else "MULTICLASS"

language/automl/automl_natural_language_dataset.py
```diff
@@ -0,0 +1,300 @@
...
+    print("Dataset example count: {}".format(dataset.example_count))
+    print("Dataset create time:")
+    print("\tseconds: {}".format(dataset.create_time.seconds))
+    print("\tnanos: {}".format(dataset.create_time.nanos))
+
+
+# [END automl_natural_language_create_dataset]
+
+
+# [START automl_natural_language_list_datasets]
```
happyhuman: Same as above comments. Region tag should be inside the function body.

language/automl/automl_natural_language_dataset.py
```diff
@@ -0,0 +1,300 @@
...
+        print("Dataset display name: {}".format(dataset.display_name))
+        print("Text classification dataset metadata:")
+        print("\t{}".format(dataset.text_classification_dataset_metadata))
+        print("Dataset example count: {}".format(dataset.example_count))
+        print("Dataset create time:")
+        print("\tseconds: {}".format(dataset.create_time.seconds))
+        print("\tnanos: {}".format(dataset.create_time.nanos))
+
+
+# [END automl_natural_language_list_datasets]
```
happyhuman: This too should be inside the function

language/automl/automl_natural_language_dataset.py
```diff
@@ -0,0 +1,300 @@
...
+    )
+    delete_dataset_parser.add_argument("dataset_id")
+
+    project_id = os.environ["PROJECT_ID"]
+    compute_region = os.environ["REGION_NAME"]
+
+    args = parser.parse_args()
+
+    if args.command == "create_dataset":
+        multilabel = True if args.multilabel == "True" else False
```
happyhuman: This line can be simplified to:
multilabel = (args.multilabel == "True")

happyhuman: This line can be simplified to:
multilabel = (args.multilabel == "True")

translate/automl/automl_translation_model.py
crwilcox: I am not 100% sure but I think response.operation.refresh() will give the current status.

sirtorry: what would response be in response.operation.refresh()?

language/automl/automl_natural_language_dataset.py
sirtorry: done

language/automl/automl_natural_language_dataset.py
sirtorry: doesn't affect snippet quality or functionality. won't fix.

sirtorry: doesn't affect snippet quality or functionality. won't fix.

language/automl/automl_natural_language_dataset.py
sirtorry: removed

language/automl/automl_natural_language_dataset.py
sirtorry: done

language/automl/automl_natural_language_dataset.py
sirtorry: done

language/automl/automl_natural_language_dataset.py
sirtorry: done

language/automl/automl_natural_language_dataset.py
sirtorry: done

language/automl/automl_natural_language_dataset.py
sirtorry: doesn't really affect snippet quality or functionality. won't fix.

language/automl/automl_natural_language_dataset.py
sirtorry: done

language/automl/automl_natural_language_dataset.py
sirtorry: done

PR #1578, opened 2018-07-17
Updates BQ Tutorial chart to count of births by gestation weeks

alixhami:


alixhami:
Test failure is unrelated. Looks like the authorized view tutorial had failed previously and didn't clean up the dataset. @tswast how can we clean up the datasets and/or fix the tests to make sure this can be avoided?

tswast:
Yeah, it looks like the test could fail before we add the datasets to `to_delete()`.

https://github.com/GoogleCloudPlatform/python-docs-samples/blob/7169f29ae41c694b4ec04e6c18960925bbd91740/bigquery/cloud-client/authorized_view_tutorial_test.py#L40

The quick fix would be to add both to `to_delete()` _before_ running `run_authorized_view_tutorial()`. Might want to pass in the dataset IDs as parameters to `run_authorized_view_tutorial()` to prevent conflicts on concurrent test runs too.

PR #1580, opened 2018-07-18
Product search

dizcology:
Add product search samples.


vision/cloud-client/product_search/import_product_sets.py
```diff
@@ -0,0 +1,96 @@
+#!/usr/bin/env python
+
+# Copyright 2018 Google Inc. All Rights Reserved.
```
crwilcox: The generated libraries  use 'llc'
```
# Copyright 2018 Google LLC
```

dizcology: Some other python samples use Inc.  Keeping these samples consistent for now.

PR #1585, opened 2018-07-19
Added samples for Storage, OCR, and Slack tutorial

michaelawyu:


functions/gcs/main_test.py
```diff
@@ -0,0 +1,39 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import main
+
+
+class Context(object):
```
andrewsg: Can we import unittest.mock's Mock object and use it here instead of creating this class? (I've updated the other Functions tests to do that as well)

functions/gcs/requirements.txt
```diff
@@ -0,0 +1 @@
+google-cloud-logging==1.6.0
```
andrewsg: Where is this used?

functions/ocr/app/main.py
```diff
@@ -0,0 +1,160 @@
...
+from google.cloud import storage
+from google.cloud import translate
+from google.cloud import vision
+
+vision_client = vision.ImageAnnotatorClient()
+translate_client = translate.Client()
+publisher = pubsub_v1.PublisherClient()
+storage_client = storage.Client()
+
+project_id = os.environ['GCLOUD_PROJECT']
```
andrewsg: I think we're discouraged from using this env variable; is this the variable present in production on GCF or is it for local use?

functions/slack/main.py
```diff
@@ -0,0 +1,100 @@
...
+import json
+
+import apiclient
+from flask import jsonify
+
+with open('config.json', 'r') as f:
+    data = f.read()
+config = json.loads(data)
+
+kgsearch = apiclient.discovery.build('kgsearch', 'v1',
```
andrewsg: We have to use apiclient here? What's this for? I'm not familiar with "kgsearch"

functions/slack/main.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+# [START functions_slack_request]
+def make_search_request(query):
+    try:
+        req = kgsearch.entities().search(query=query, limit=1)
+        res = req.execute()
+        print(res)
+        return format_slack_message(query, res)
+    except apiclient.errors.Error as err:
```
andrewsg: Why are we casting this to a new error?

functions/slack/main.py
```diff
@@ -0,0 +1,100 @@
...
+
+# [START functions_slack_search]
+def kg_search(request):
+    if request.method != 'POST':
+        return 'Only POST requests are accepted', 400
+    try:
+        verify_web_hook(request.form)
+        kg_search_response = make_search_request(request.form['text'])
+        return jsonify(kg_search_response)
+    except ValueError as err:
```
andrewsg: Is it necessary to catch this? What happens if we just let it bubble up instead?

functions/slack/main.py
```diff
@@ -0,0 +1,100 @@
...
+        return format_slack_message(query, res)
+    except apiclient.errors.Error as err:
+        raise ValueError(str(err))
+# [END functions_slack_request]
+
+
+# [START functions_slack_search]
+def kg_search(request):
+    if request.method != 'POST':
+        return 'Only POST requests are accepted', 400
```
andrewsg: This should be the 405 error code, if it's intended to be an HTTP code.

functions/slack/main.py
```diff
@@ -0,0 +1,100 @@
...
+        res = req.execute()
+        print(res)
+        return format_slack_message(query, res)
+    except apiclient.errors.Error as err:
+        raise ValueError(str(err))
+# [END functions_slack_request]
+
+
+# [START functions_slack_search]
+def kg_search(request):
```
andrewsg: Where is this function used, is it an HTTP handler for a Function? In that case Flask has a good way of returning a 405 error, and we don't have to catch errors and print them -- just let them bubble up.

functions/slack/main.py
michaelawyu: It's the Knowledge Graph API, which is used in the Slack tutorial to return search results. The API is not a part of GCP products and doesn't have veneer client library :(

functions/gcs/requirements.txt
michaelawyu: My bad. The sample was created when GCF Python cannot write stdout to Stackdriver Logging; I forgot to remove the reference after the support came online.

michaelawyu:
Note: One day we will be able to use F-string in this repo...

andrewsg:
I don't suppose there's an easy version bump or other change to flake8 that will allow us to use f-strings easily?

functions/ocr/app/main.py
michaelawyu: I talked to Ace about the same question when I wrote the sample. Basically it is used because the Cloud Pub/Sub client library for Python explicitly requires the `project_id` parameter for the `topic_path` method (the Node.js client library, however, doesn't). Cloud Functions runtime does support it, though it is only documented in the [Firebase documentation](https://firebase.google.com/docs/functions/config-env).

michaelawyu:
I assume we can add `session.interpreter = 'python3.6'` to `nox.py`, though it will run python3 flake8 on all samples.

andrewsg:
Okay, since you've already changed the f-strings we can revisit the lint issue later.

functions/helloworld/sample_pubsub_test_system.py
```diff
@@ -58,7 +58,7 @@ def test_print_name(publisher_client):
         start_time
     ], stdout=subprocess.PIPE)
     logs = str(log_process.communicate()[0])
-    print logs
-    print start_time
+    print(logs)
```
andrewsg: Are these print statements necessary to leave in, or were they only added for development purposes?

michaelawyu: I will remove them 👌

functions/ocr/app/main.py
multiple1902: GCF support for env variables was documented here (added really recently) https://cloud.google.com/functions/docs/env-var

In short, env["GCP_PROJECT"] is preferred than env["GCLOUD_PROJECT"]

michaelawyu: @multiple1902 Thanks! I'll update it to use `GCP_PROJECT` instead.

multiple1902: Sent in PR https://github.com/GoogleCloudPlatform/python-docs-samples/pull/1594

PR #1586, opened 2018-07-19
Diarization Output Modified

happyhuman:
Printing the last paragraph only.


speech/cloud-client/beta_snippets.py
```diff
@@ -156,21 +153,18 @@ def transcribe_file_with_diarization(speech_file):
...
-        print('-' * 20)
-        print('First alternative of result {}: {}'
-              .format(i, alternative.transcript))
-        print('Speaker Tag for the first word: {}'
-              .format(alternative.words[0].speaker_tag))
+    result = response.results[-1]
+    words_info = result.alternatives[0].words
+    pieces = ['%s (%s)' % (word_info.word, word_info.speaker_tag)
+              for word_info in words_info]
+    print(' '.join(pieces))
```
dizcology: I feel this might not illustrate the "typical" use case, where the developer might more likely want to group and join the words according to their speaker_tag.

happyhuman: Interesting point. Tough to say what the right use case is. But I see it just as a sample. to show them the API, and not the use case. Do you think we can keep it as is, or should we change it?

dizcology: I agree - in that sense perhaps let's just iterate through words_info and print everything without the nice formatting of `'%s (%s)'`.  the expected output of the test really confused me.

speech/cloud-client/beta_snippets.py
```diff
@@ -156,21 +153,18 @@ def transcribe_file_with_diarization(speech_file):
...
     response = client.recognize(config, audio)
 
-    for i, result in enumerate(response.results):
-        alternative = result.alternatives[0]
-        print('-' * 20)
-        print('First alternative of result {}: {}'
-              .format(i, alternative.transcript))
-        print('Speaker Tag for the first word: {}'
-              .format(alternative.words[0].speaker_tag))
+    result = response.results[-1]
```
jerjou: A comment here explaining why you're only taking the last result (instead of all of them) would probably be helpful.

speech/cloud-client/beta_snippets.py
```diff
@@ -46,7 +46,6 @@ def transcribe_file_with_enhanced_model(speech_file):
     audio = speech.types.RecognitionAudio(content=content)
     config = speech.types.RecognitionConfig(
         encoding=speech.enums.RecognitionConfig.AudioEncoding.LINEAR16,
-        sample_rate_hertz=8000,
```
jerjou: Any particular reason for omitting this? For WAV files, the API can infer this, but in the general case it's probably a good idea to include the sample rate

speech/cloud-client/beta_snippets.py
```diff
@@ -156,21 +153,18 @@ def transcribe_file_with_diarization(speech_file):
...
-    for i, result in enumerate(response.results):
-        alternative = result.alternatives[0]
-        print('-' * 20)
-        print('First alternative of result {}: {}'
-              .format(i, alternative.transcript))
-        print('Speaker Tag for the first word: {}'
-              .format(alternative.words[0].speaker_tag))
+    result = response.results[-1]
+    words_info = result.alternatives[0].words
+    pieces = ['%s (%s)' % (word_info.word, word_info.speaker_tag)
```
jerjou: Isn't `'{} ({})'.format(word_info.word, word_info.speaker_tag)` the preferred way to do this these days? Or is Thea no longer benevolent overlord?

speech/cloud-client/beta_snippets.py
happyhuman: I keep going back and forth about it. Sometimes it is useful, and other times it is causing an error when the input file has a different sample rate.

speech/cloud-client/beta_snippets.py
happyhuman: Good point. I will modify it.

speech/cloud-client/beta_snippets.py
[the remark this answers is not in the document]
happyhuman: We just had a discussion about it with Roy and Jerjou. If the input file has a different sample rate, it will cause an error. It is simpler just to omit it and the API figures it out on its own.

speech/cloud-client/beta_snippets_test.py
```diff
@@ -51,10 +51,10 @@ def test_transcribe_file_with_auto_punctuation(capsys):
 
 def test_transcribe_diarization(capsys):
     transcribe_file_with_diarization(
-        os.path.join(RESOURCES, 'Google_Gnome.wav'))
+        os.path.join(RESOURCES, 'commercial_mono.wav'))
```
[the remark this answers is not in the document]
happyhuman: Even for the unit tests?

speech/cloud-client/beta_snippets.py
happyhuman: I made some changes to make the output to look like this:

Speaker #1: I'm here
Speaker #2: hi I'd like to buy a Chrome Cast and I was wondering whether you could help me

happyhuman:
@dizcology , @puneith, I made some changes based on the comments made by all the reviewers? Can you please take a final look and approve the PR if it looks good?

speech/cloud-client/beta_snippets.py
```diff
@@ -156,33 +152,42 @@ def transcribe_file_with_diarization(speech_file):
...
 
-    for i, result in enumerate(response.results):
-        alternative = result.alternatives[0]
-        print('-' * 20)
-        print('First alternative of result {}: {}'
-              .format(i, alternative.transcript))
-        print('Speaker Tag for the first word: {}'
-              .format(alternative.words[0].speaker_tag))
+    # response.results contains partial results with the last item
+    # containing the entire result:
```
jerjou: Mm... not quite. The transcript within each result is separate and sequential per result. However, the words list within an alternative (for whatever reason) includes all the words from all the results thus far. Thus, to get all the words with speaker tags, you only have to take the words list from the last result.

speech/cloud-client/beta_snippets.py
```diff
@@ -156,33 +152,42 @@ def transcribe_file_with_diarization(speech_file):
...
+
+    words_info = result.alternatives[0].words
+
+    # Separating the words by who said what:
+    speakers_words = []
+    for word_info in words_info:
+        if speakers_words and speakers_words[-1][0] == word_info.speaker_tag:
+            speakers_words[-1][1].append(word_info.word)
+        else:
+            speakers_words.append((word_info.speaker_tag, [word_info.word, ]))
```
jerjou: This is a bit hard to read. An intermediate variable and a `namedtuple` would go a long way to making this clearer:

```py
Speaker = collections.namedtuple('Speaker', ['tag', 'words'])
speaker_words = [Speaker(tag=0, words=[])]
for word_info in words_info:
  current_speaker = speaker_words[-1]
  if current_speaker.tag == word_info.speaker_tag:
    current_speaker.words.append(word_info.word)
  else:
    speaker_words.append(Speaker(tag=word_info.speaker_tag, words=[word_info.word]))
```

speech/cloud-client/beta_snippets.py
```diff
@@ -279,28 +283,20 @@ def transcribe_file_with_word_level_confidence(speech_file):
...
-        transcribe_file_with_auto_punctuation(args.path)
+        transcribe_file_with_auto_punctuation()
     elif args.command == 'diarization':
-        transcribe_file_with_diarization(args.path)
+        transcribe_file_with_diarization()
     elif args.command == 'multi-channel':
-        transcribe_file_with_multichannel(args.path)
+        transcribe_file_with_multichannel()
     elif args.command == 'multi-language':
-        transcribe_file_with_multilanguage(args.path, args.first, args.second)
```
jerjou: For future reference, argparse's [sub-commands feature](https://docs.python.org/2/library/argparse.html#sub-commands) would be helpful to avoid having args that only matter for one command or another.

speech/cloud-client/beta_snippets.py
```diff
@@ -153,30 +152,42 @@ def transcribe_file_with_diarization(speech_file):
...
+
     words_info = result.alternatives[0].words
-    pieces = ['%s (%s)' % (word_info.word, word_info.speaker_tag)
-              for word_info in words_info]
-    print(' '.join(pieces))
+
+    # Separating the words by who said what:
+    speakers_words = []
+    for word_info in words_info:
+        if speakers_words and speakers_words[-1][0] == word_info.speaker_tag:
```
puneith: If I understand correctly what this piece of for loop is doing (creating list of words for tag) isn't is better to use hashmap. Does this loop work what it's supposed to do?

puneith: Jerjou is already reviewing this code so I should stop reviewing. You are already is great hands :)

speech/cloud-client/beta_snippets.py
happyhuman: I see. Thanks for the clarification. Let me update the comment.

speech/cloud-client/beta_snippets.py
jerjou: Also, maybe `speaker_sequence` or something, to make it clear it's not just a speaker->words_they_spoke mapping, and is actually the conversation / words spoken in sequence.

happyhuman: Interesting idea. However, I think the code readability is mainly the result of the intermediate`Speaker(tag=0, words=[])` as the first element in the list, allowing us to define `current_speaker` every time.
The downside of it is that we are introducing a new object that is not returned by the API and we will have to handle it in the next step separately (either by removing it, or by skipping it), which reduces the code cleanness in another way.
So, while I am not against the suggested solution, I am also not sure if it is really helping with the code readability that much.
What do you think?

speech/cloud-client/beta_snippets.py
happyhuman: Sounds good.

speech/cloud-client/beta_snippets.py
happyhuman: What about this:

```
    speakers = []
    words = []
    for word_info in words_info:
        if speakers and speakers[-1] == word_info.speaker_tag:
            words[-1].append(word_info.word)
        else:
            speakers.append(word_info.speaker_tag)
            words.append([word_info.word, ])
```
I think this is more readable that the current code, without introducing the intermediate variable.

speech/cloud-client/beta_snippets.py
happyhuman: Sounds good.

speech/cloud-client/beta_snippets.py
```diff
@@ -156,21 +153,18 @@ def transcribe_file_with_diarization(speech_file):
 
     config = speech.types.RecognitionConfig(
         encoding=speech.enums.RecognitionConfig.AudioEncoding.LINEAR16,
-        sample_rate_hertz=16000,
         language_code='en-US',
         enable_speaker_diarization=True,
         diarization_speaker_count=2)
 
     print('Waiting for operation to complete...')
```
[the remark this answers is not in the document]
tswast: FWIW nearly all our other Python samples do `print()`. It's true that it's not always the recommended practice in production, but it's easy to understand. With logging there's always the risk that the developer has some weird config where the logs end up where maybe they don't expect.

speech/cloud-client/beta_snippets.py
```diff
@@ -156,33 +152,46 @@ def transcribe_file_with_diarization(speech_file):
...
 
-    for i, result in enumerate(response.results):
-        alternative = result.alternatives[0]
-        print('-' * 20)
-        print('First alternative of result {}: {}'
-              .format(i, alternative.transcript))
-        print('Speaker Tag for the first word: {}'
-              .format(alternative.words[0].speaker_tag))
+    # The transcript within each result is separate and sequential per result.
+    # However, the words list within an alternative (for whatever reason)
```
tswast: I understand that this comment is to explain why the `[-1]`, but I'm having trouble understanding what this is referring to. (Probably because I don't know what an "alternative" means in this context.)

I think you could probably get by with fewer words within the sample with just
```
# To get all words with speaker tags, you only have to take the words list from the last result.
```
and leave the extra explanation about alternatives for the actual cloud.google.com docs.

speech/cloud-client/beta_snippets.py
```diff
@@ -156,33 +152,46 @@ def transcribe_file_with_diarization(speech_file):
...
+    # the last result:
+    result = response.results[-1]
+
+    words_info = result.alternatives[0].words
+
+    # Separating the words by who said what:
+    speakers = []
+    words = []
+    for word_info in words_info:
+        if (not speakers) or speakers[-1] != word_info.speaker_tag:
```
tswast: I think this implementation means that if speakers appear in certain orders you could end up with duplicates. For example.

`word_info` speakers: `A, B, B, A` would result in a `speakers` lists: `A, B, A`.

I think this sample would be clearer with a single dictionary of speakers to words rather than two lists.

speech/cloud-client/beta_snippets.py
tswast: I'm not really understanding your comment, @jerjou but this sounds like something that needs to be documented on the cloud.google.com docs with a briefer explanation in the sample itself.

speech/cloud-client/beta_snippets.py
happyhuman: We had a long discussion about the code readability of this part of the sample. The API returns each word with an associated speak_tag. So, if I am speaker #1 and I say "I am ok", the words 'I' and 'am' and 'okay' will all be tagged with speaker_tag: 1.

Since the point of the sample is mostly just to showcase the API call and how to unbox the response object, I am beginning to think that this particular implementation is adding more confusion than clarity. I am leaning back toward a simpler implementation that just iterates over words and prints each word along with the associated speaker_tag. What do you think?

speech/cloud-client/beta_snippets.py
happyhuman: I initially had a shorter comment. But Jerjou suggested to explain it more, so I added the longer comment. I am okay either way (just want to wrap up this PR).

speech/cloud-client/beta_snippets.py
tswast: Yes, printing each word with a speaker tag would be much clearer.

You could do something like
```
current_speaker = None
for word_info in words_info:
    if current_speaker is None or current_speaker != word_info.speaker_tag:
        current_speaker = word_info.speaker_tag
        print()
        print(current_speaker)
    print(word_info.word)
```
if you wanted to keep them grouped.

jerjou:
Shoot. I totally dropped off this PR, didn't I. Sorry - bad email management -_-; Glad y'all just went on without me.

PR #1594, opened 2018-07-22
Functions: Use the documented way to fetch current GCP project ID.

weisi:
The document at https://cloud.google.com/functions/docs/env-var suggests using `env['GCP_PROJECT']` for current project ID, as `env['GCLOUD_PROJECT']` is deprecated/legacy.


ace-n:
Thanks for the PR - merging. 😄

PR #1601, opened 2018-07-25
Use example node pool names for affinity argument

tswast:
I believe this will make it clearer what the actual value should be for the affinity label.

@noremac201 please review.


Noremac201:
LGTM 👍

PR #1609, opened 2018-08-02
Add IAM doc snippets

awkoren:


awkoren:
Addressed all feedback

PR #1610, opened 2018-08-02
Kokoro: fix + add GCF unit tests

ace-n:


functions/helloworld/sample_http_test.py
```diff
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 # [START functions_http_unit_test]
-from unittest.mock import Mock
```
andrewsg: Why change this?

.kokoro/presubmit_tests_functions.cfg
```diff
@@ -0,0 +1,15 @@
...
+
+# Tell the trampoline which build file to use.
+env_vars: {
+    key: "TRAMPOLINE_BUILD_FILE"
+    value: "github/python-docs-samples/.kokoro/system_tests.sh"
+}
+
+env_vars: {
+    key: "NOX_SESSION"
+    value: "functions and py27 and not venv"
```
andrewsg: I'm not familiar with this file, what does it do? Esp. this part

ace-n: This file tells Kokoro what tests should be run. The `NOX_SESSION` variable is in _Pytest-style_ [key format](http://nox.readthedocs.io/en/latest/usage.html#specifying-one-or-more-sessions).

PR #1616, opened 2018-08-07
Cloud Bigtable Quickstarts

billyjacobson:
Quickstarts for Python and Happybase


bigtable/quickstart/main.py
```diff
@@ -0,0 +1,56 @@
...
+
+    column_family_id = 'cf1'
+    column_id = 'c1'.encode('utf-8')
+    value = row.cells[column_family_id][column_id][0].value.decode('utf-8')
+
+    print('Row key: {}\nData: {}'.format(key, value))
+
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(
```
tswast: In my opinion, the argparse stuff might be a distraction from the key pieces of the quickstart, but we haven't done UX studies on that as far as I know, so I'll allow it. :-)

bigtable/quickstart/main.py
```diff
@@ -0,0 +1,56 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# [START bigtable_quickstart]
+import argparse
+
+from google.cloud import bigtable
+
+
+def main(project_id, instance_id, table_id):
```
tswast: Even though you don't use them, I think it'd be helpful to show default values so that people can see what a project ID, instance ID, table ID look like.

```
def main(project_id="your-project-id", instance_id="your_instance", table_id="your_table"):
```

bigtable/quickstart_happybase/main.py
```diff
@@ -0,0 +1,61 @@
...
+# limitations under the License.
+# [START bigtable_quickstart_happybase]
+import argparse
+import json
+
+from google.cloud import bigtable
+from google.cloud import happybase
+
+
+def main(project_id, instance_id, table_id):
```
tswast: Ditto on showing default values here so it's clear right away the types and what the values should look like.

bigtable/quickstart/main.py
billyjacobson: That is a good idea, but how would you suggest pulling in the information? Env variables? Also I wasn't sure if there is an easy way to customize the documentation since so much of it seems to autogenerated and especially using this argument parser as part of that generation.

bigtable/quickstart/main.py
billyjacobson: Done in 56393fe

bigtable/quickstart_happybase/main.py
billyjacobson: Done in 56393fe

bigtable/quickstart/main.py
jabubake: Take a look at https://github.com/GoogleCloudPlatform/golang-samples/blob/master/bigtable/quickstart/main.go
There's a way to do it with region tags so that you can see only the default arguments in the docs, yet offer command line arg support if you were to run it from the command line.

bigtable/quickstart/main.py
```diff
@@ -19,7 +19,8 @@
 from google.cloud import bigtable
 
 
-def main(project_id, instance_id, table_id):
+def main(project_id="project-id", instance_id="instance-id",
+    table_id="my-table"):
```
tswast: This probably fails flake8. Needs `table_id` needs to be aligned with `project_id` in the line above.

```
def main(project_id="project-id", instance_id="instance-id",
         table_id="my-table"):
```

bigtable/quickstart/main.py
tswast: Yeah, it's okay as-is. I've seen other quickstarts hard-code a project ID but that comes with its own problems.

bigtable/quickstart/main.py
billyjacobson: Fixed before you commented this 😛 IntelliJ autoformat typically is better about this stuff 🤷‍♂️

bigtable/quickstart/main.py
tswast: Ah, yeah that Go example looks like a good compromise.

PR #1624, opened 2018-08-10
Updated google-cloud-pubsub to version 0.35

anguillanneuf:


pubsub/cloud-client/publisher_test.py
```diff
@@ -43,6 +45,19 @@ def topic(client):
     yield topic_path
 
 
+def _make_sleep_patch():
```
tswast: You should use the `monkeypatch` fixture built in to pytest. The reason is that that fixture will automatically clean up any patched functions.

tswast: Then again, I see you are using this in a context manager, which has the same effect. Nevermind.

anguillanneuf: Ohh that's good to know!

PR #1628, opened 2018-08-10
Add samples for Redis Labs redis for standard py37

andrewsg:


appengine/standard_python37/redis/main.py
```diff
@@ -0,0 +1,43 @@
...
+from flask import Flask
+import redis
+
+
+app = Flask(__name__)
+
+
+# [START gae_py37_redis_client]
+redis_host = os.environ.get('REDIS_HOST', 'localhost')
+redis_port = int(os.environ.get('REDIS_PORT', 6379))
```
ace-n: Nit: will this fail [gracefully] if `REDIS_PORT` is set to a non-integer?

appengine/standard_python37/redis/main_test.py
```diff
@@ -0,0 +1,35 @@
...
+        main.redis_client.set('counter', 0)
+    except Exception:
+        pytest.skip('Redis is unavailable.')
+
+    main.app.testing = True
+    client = main.app.test_client()
+
+    r = client.get('/')
+    assert r.status_code == 200
+    assert '1' in r.data.decode('utf-8')
```
ace-n: Nit: what is the format of `r.data`? Is there any chance of spurious `1`s or `2`s occurring in it?

appengine/standard_python37/redis/main.py
andrewsg: It'll fail with an exception that shows pretty clearly what the problem is, which is the best we can ask for given there is no way to operate without that number.

appengine/standard_python37/redis/main_test.py
andrewsg: No, this test is copied from Flex and we didn't have any issues. The format is an integer but it'll be starting from zero every time, and the feature that determines that is not a component under test.

PR #1635, opened 2018-08-15
Vision region tag update

alixhami:

### DIFF
--- a/vision/cloud-client/crop_hints/crop_hints.py
+++ b/vision/cloud-client/crop_hints/crop_hints.py
@@ -20,19 +20,19 @@ Examples:
     python crop_hints.py resources/cropme.jpg draw
     python crop_hints.py resources/cropme.jpg crop
 """
-# [START full_tutorial]
-# [START imports]
+# [START vision_crop_hints_tutorial]
+# [START vision_crop_hints_tutorial_imports]
 import argparse
 import io
 
 from google.cloud import vision
 from google.cloud.vision import types
 from PIL import Image, ImageDraw
-# [END imports]
+# [END vision_crop_hints_tutorial_imports]
 
 
 def get_crop_hint(path):
-    # [START get_crop_hint]
+    # [START vision_crop_hints_tutorial_get_crop_hints]
     """Detect crop hints on a single image and return the first result."""
     client = vision.ImageAnnotatorClient()
 
@@ -49,14 +49,14 @@ def get_crop_hint(path):
 
     # Get bounds for the first crop hint using an aspect ratio of 1.77.
     vertices = hints[0].bounding_poly.vertices
-    # [END get_crop_hint]
+    # [END vision_crop_hints_tutorial_get_crop_hints]
 
     return vertices
 
 
 def draw_hint(image_file):
     """Draw a border around the image using the hints in the vector list."""
-    # [START draw_hint]
+    # [START vision_crop_hints_tutorial_draw_crop_hints]
     vects = get_crop_hint(image_file)
 
     im = Image.open(image_file)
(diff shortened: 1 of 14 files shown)


vision/cloud-client/crop_hints/crop_hints.py
```diff
@@ -49,14 +49,14 @@ def get_crop_hint(path):
...
-    # [END get_crop_hint]
+    # [END vision_crop_hints_tutorial_get_crop_hints]
 
     return vertices
 
 
 def draw_hint(image_file):
     """Draw a border around the image using the hints in the vector list."""
-    # [START draw_hint]
+    # [START vision_crop_hints_tutorial_draw_crop_hints]
```
dizcology: does the region tag need to be consistent with the function name?  here the tag says `...draw_crop_hints` while the function's name is `draw_hint`.

vision/cloud-client/detect/beta_snippets.py
```diff
@@ -59,7 +59,7 @@ def localize_objects(path):
 # [END vision_localize_objects]
 
 
-# [START vision_localize_objects_uri]
+# [START vision_localize_objects_gcs]
```
dizcology: if i remember correctly the API accepts both GCS URI and HTTP URI, in this case the `gcs` suffix might not be appropriate.

vision/cloud-client/detect/beta_snippets.py
```diff
@@ -130,7 +130,7 @@ def detect_handwritten_ocr(path):
 # [END vision_handwritten_ocr]
 
 
-# [START vision_handwritten_ocr_uri]
+# [START vision_handwritten_ocr_gcs]
```
dizcology: on the other hand here i think the uri must be a GCS uri.

vision/cloud-client/detect/detect_pdf.py
```diff
@@ -31,7 +31,7 @@
 from google.protobuf import json_format
 
 
-# [START vision_async_detect_document_ocr]
+# [START vision_text_detection_pdf_gcs]
```
dizcology: the segment `pdf` might not be appropriate, since the API also accepts TIFF files.

vision/cloud-client/face_detection/faces.py
```diff
@@ -18,14 +18,14 @@
...
+# [START vision_face_detection_tutorial_imports]
 from google.cloud import vision
-# [END import_client_library]
 from google.cloud.vision import types
 from PIL import Image, ImageDraw
+# [END vision_face_detection_tutorial_imports]
 
 
-# [START def_detect_face]
+# [START vision_face_detection_tutorial_send_request]
```
dizcology: `send_request` is a bit vague/generic.

vision/cloud-client/product_search/product_in_product_set_management.py
```diff
@@ -21,14 +21,17 @@
 https://cloud.google.com/vision/product-search/docs/
 """
 
-# [START product_search_import]
 import argparse
 
+# [START vision_product_search_add_product_to_product_set]
+# [START vision_product_search_remove_product_from_product_set]
```
dizcology: these two tags only enclose the import line.  does not look right.

vision/cloud-client/product_search/product_management.py
```diff
@@ -21,14 +21,23 @@
...
 """
 
-# [START product_search_import]
 import argparse
 
+# [START vision_product_search_create_product]
+# [START vision_product_search_delete_product]
+# [START vision_product_search_list_products]
+# [START vision_product_search_get_product]
+# [START vision_product_search_update_product_labels]
```
dizcology: these too.  they are also duplicate tags - missing `imports` suffix?

vision/cloud-client/crop_hints/crop_hints.py
alixhami: The region tag does not need to have the same name as the function

vision/cloud-client/detect/detect_pdf.py
alixhami: The API might accept TIFF, but this sample demonstrates PDF

vision/cloud-client/face_detection/faces.py
alixhami: It's within the context of being the face detection tutorial, so I think it's an ok suffix. This was done to standardize across languages. `def_detect_face` is too vague because there is also a face detection sample for the how to guides.

vision/cloud-client/product_search/product_in_product_set_management.py
alixhami: These tags serve the purpose of including the import with the samples.

vision/cloud-client/product_search/product_management.py
alixhami: It increases maintenance time to create a special import tag for every single sample, so I am just using the same region tags to extend the sample to include the import.

vision/cloud-client/detect/beta_snippets.py
alixhami: _gcs and _uri were previously inconsistently used across languages and samples. I standardized to choose _gcs when it was a GCS URI. I will keep in mind to change if it is an HTTP URI.

PR #1636, opened 2018-08-15
 Added timeout in error handling

anguillanneuf:
Highlights the timeout feature - if not specified, Pub/Sub will try to publish or subscribe indefinitely.

### DIFF
--- a/pubsub/cloud-client/publisher.py
+++ b/pubsub/cloud-client/publisher.py
@@ -131,7 +131,7 @@ def publish_messages_with_error_handler(project, topic_name):
     topic_path = publisher.topic_path(project, topic_name)
 
     def callback(message_future):
-        # Attempts to publish for 10 seconds, then times out.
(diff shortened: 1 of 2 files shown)


pubsub/cloud-client/publisher.py
```diff
@@ -131,7 +131,9 @@ def publish_messages_with_error_handler(project, topic_name):
     topic_path = publisher.topic_path(project, topic_name)
 
     def callback(message_future):
-        if message_future.exception():
+        # Attempts to publish for 10 seconds, then times out.
+        # Defaults to None.
```
tswast: What's the behavior when timeout is None?

This comment should be more specific. Something like: When timeout is unspecified or None, the exception method waits indefinitely.

anguillanneuf: That's much better. Let me rewrite it.

PR #1639, opened 2018-08-16
Video Intelligence region tag update

alixhami:

### DIFF
--- a/video/cloud-client/labels/labels.py
+++ b/video/cloud-client/labels/labels.py
@@ -27,29 +27,29 @@ Usage Example:
 
 """
 
-# [START full_tutorial]
-# [START imports]
+# [START video_label_tutorial]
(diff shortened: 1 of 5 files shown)


video/cloud-client/labels/labels.py
```diff
@@ -27,29 +27,29 @@
 
 """
 
-# [START full_tutorial]
-# [START imports]
+# [START video_label_tutorial]
```
dizcology: should this use plural `labels` to be more consistent with the directory and file name?

alixhami: I can make this update if you prefer, but the other context to note is that page in the docs is called label-tutorial and the title of the tutorial is Label Detection Tutorial.
https://cloud.google.com/video-intelligence/docs/label-tutorial

PR #1643, opened 2018-08-20
Language region tag update

alixhami:

### DIFF
--- a/language/cloud-client/v1/snippets.py
+++ b/language/cloud-client/v1/snippets.py
@@ -39,12 +39,12 @@ def sentiment_text(text):
         text = text.decode('utf-8')
 
     # Instantiates a plain text document.
-    # [START migration_document_text]
-    # [START migration_analyze_sentiment]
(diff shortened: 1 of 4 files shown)


language/cloud-client/v1/snippets.py
```diff
@@ -39,45 +39,45 @@ def sentiment_text(text):
...
     print('Score: {}'.format(sentiment.score))
     print('Magnitude: {}'.format(sentiment.magnitude))
-    # [END migration_analyze_sentiment]
-# [END def_sentiment_text]
+    # [END language_python_migration_sentiment_text]
+# [END language_sentiment_text]
 
 
-# [START def_sentiment_file]
+# [START language_sentiment_file_gcs]
```
dizcology: should this be just the gcs suffix instead of file_gcs?

alixhami: To be consistent with the other samples, there is a `_gcs` suffix when the file is from GCS, and no suffix when it is a local file.

PR #1644, opened 2018-08-20
Speech region tag update

alixhami:


speech/cloud-client/transcribe_word_time_offsets.py
```diff
@@ -60,7 +60,7 @@ def transcribe_file_with_word_time_offsets(speech_file):
                 end_time.seconds + end_time.nanos * 1e-9))
 
 
-# [START def_transcribe_gcs]
+# [START speech_transcribe_async_time_offsets_gcs]
```
dizcology: i would suggest keeping this as `word_time_offset`.  there could be other types of time offset.

PR #1652, opened 2018-08-24
Added samples for handling CORS requests.

michaelawyu:


michaelawyu:
@ace-n Completed requested changes. PTAL.

functions/http/main.py
```diff
@@ -118,3 +118,56 @@ def get_signed_url(request):
...
+        headers = {
+            'Access-Control-Allow-Origin': '*',
+            'Access-Control-Allow-Methods': 'GET',
+            'Access-Control-Allow-Headers': 'Content-Type',
+            'Access-Control-Max-Age': '3600'
+        }
+
+        return ('', 204, headers)
+
+    # Set CORS headers for the main request
```
andrewsg: "The main request" is ambiguous here. Maybe we should just say "Set CORS headers" or perhaps "Set CORS headers for non-OPTIONS requests" if that is necessary? Also: same comment on similar line in next function.

functions/http/main.py
```diff
@@ -118,3 +118,56 @@ def get_signed_url(request):
 
     return url
 # [END functions_http_signed_url]
+
+
+# [START functions_http_cors]
+def cors_enabled_function(request):
+    # Set CORS headers for the preflight request
+    # e.g. allows GETs from any origin with the Content-Type header
+    # and cache preflight response for an 3600s
```
andrewsg: I don't understand this comment block. Can you try and explain it to me in more detail, and then we can use that explanation to clarify it?

functions/http/main.py
```diff
@@ -118,3 +118,56 @@ def get_signed_url(request):
...
+
+    return ('Hello World!', 200, headers)
+# [END functions_http_cors]
+
+
+# [START functions_http_cors_auth]
+def cors_enabled_function_auth(request):
+    # Set CORS headers for preflight requests
+    # e.g. allows GETS from origin https://mydomain.com with Authorization
+    # header
```
andrewsg: Likewise I'm not sure I'm clear on this comment meaning yet.

michaelawyu:
Hi @andrewsg ,

Basically under certain circumstances browsers need to send one `OPTIONS` request first (preflight request) before accessing a resource cross-region ("main" request). They require different sets of headers so I put these comments in the sample to separate them in a more clear way. Should I change the comment to something else?

andrewsg:
Okay. Here are some suggestions:
- Change `GETS` into `GET requests`
- Add a link to https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request to explain preflight to people unfamiliar with the terminiology
- Reconsider use of "e.g."

michaelawyu:
@ace-n @andrewsg Updated as requested; PTAL :)

ace-n:
@andrewsg I'll LGTM this once you do.

PR #1653, opened 2018-08-24
Added sample for Using Environment Variables.

michaelawyu:


functions/env_vars/main.py
```diff
@@ -0,0 +1,21 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+
+
+# [START functions_env_vars]
+def env_vars(request):
+    return os.environ['FOO']
```
andrewsg: This will throw an exception if FOO isn't set. If that's the desired outcome, that's fine; otherwise perhaps use os.getenv with a default value of something like "environment variable not set"

michaelawyu: Hi @andrewsg ! The documentation has one step specified before using this snippet, instructing developers to set up the `FOO` env var; though I agree it would be better to add a default value. Will add a new commit.

PR #1666, opened 2018-08-28
Updated appengine/flexible/cloudsql_postgresql samples.

kurtisvg:
Manually updated requirements for appengine/flexible/cloudsql_postgresql. The test seemed to fail the first time because it ran out of space.


appengine/flexible/cloudsql_postgresql/requirements.txt
```diff
@@ -1,4 +1,4 @@
-Flask==0.12.2
+Flask==1.0.2
```
frankyn: That's a major version update. This should be okay, but could you verify associated documentation?

PR #1669, opened 2018-08-29
Update AutoML region tags to use standard product prefixes

alixhami:
automl_natural_language => automl_language
automl_translation => automl_translate

This update is to be consistent with the Natural Language and Translation region tags.


sirtorry:
I actually think we should change everything else to natural_language and translation

alixhami:
Discussed in person with Torry. We will consider making the prefix update in the future if it affects user searching.

PR #1672, opened 2018-08-29
Added samples for using Cloud SQL with App Engine Python 3.7 Standard

michaelawyu:
Pending approval of the new doc page.


michaelawyu:
The linter only supports Python 2.7 :(

appengine/standard_python37/cloudsql/main_mysql.py
```diff
@@ -0,0 +1,57 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# [START gae_python37_cloudsql_mysql]
+import os
+
+from flask import Flask
+import mysql.connector
```
andrewsg: I think we might be intentionally standardizing on pymysql as it has the largest usage share and is easiest for our users to install. Is there a particular reason to go with mysql connector?

appengine/standard_python37/cloudsql/main_mysql.py
```diff
@@ -0,0 +1,57 @@
...
+        # Cloud SQL instance
+        host = '127.0.0.1'
+
+    cnx = mysql.connector.connect(user=db_user, password=db_password,
+                                  host=host, database=db_name)
+    cursor = cnx.cursor()
+    cursor.execute('SELECT NOW() as now;')
+    result = cursor.fetchall()
+    current_time = result[0][0]
+    cursor.close()
```
andrewsg: The adaptor doesn't support anything like "with" statements (context managers) here, do they?

appengine/standard_python37/cloudsql/main_mysql_pooling.py
```diff
@@ -0,0 +1,68 @@
...
+
+@app.route('/')
+def main():
+    cnx = cnxpool.get_connection()
+    cursor = cnx.cursor()
+    cursor.execute('SELECT NOW() as now;')
+    result = cursor.fetchall()
+    current_time = result[0][0]
+    cursor.close()
+    # If the connection comes from a pool, close() will send the connection
```
andrewsg: Good comment here!

michaelawyu: Thanks :)

appengine/standard_python37/cloudsql/main_mysql.py
michaelawyu: I checked the reference and it does not seem to have the `__enter__` and `__exit__` method.

andrewsg:
CLA bot is due to lack of github tag in some commits for Chen, should be resolved going forward

PR #1673, opened 2018-08-29
Added sample for synchronous pull

anguillanneuf:

### DIFF
--- a/pubsub/cloud-client/subscriber.py
+++ b/pubsub/cloud-client/subscriber.py
@@ -214,11 +214,13 @@ def receive_messages_synchronously(project, subscription_name):
     subscription_path = subscriber.subscription_path(
         project, subscription_name)
 
-    # Builds a pull request with a specified number of messages to return.
+    # Builds a pull request with a specific number of messages to return.
+    # `return_immediately` is set to False so that the system waits (for a
+    # bounded amount of time) until at lease one message is available.
     response = subscriber.pull(
         subscription_path,
         max_messages=3,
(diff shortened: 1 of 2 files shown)


pubsub/cloud-client/subscriber.py
```diff
@@ -207,6 +207,31 @@ def callback(message):
     # [END pubsub_subscriber_flow_settings]
 
 
+def receive_messages_synchronously(project, subscription_name):
+    """Pulling messages synchronously."""
+    # [START pubsub_subscriber_sync_pull]
+    subscriber = pubsub_v1.SubscriberClient()
+    subscription_path = subscriber.subscription_path(
+        project, subscription_name)
```
tswast: Is there a public subscription name we can use instead of using this as a parameter? I recall I saw [one for taxis in the Dataflow templates quickstart](https://cloud-dot-devsite.googleplex.com/dataflow/docs/quickstarts/quickstart-templates).

If not, we should show a comment like `# subscription_name = '/does/a/sub/name/look/like/a/path?'`

anguillanneuf: The `subscription_path` method will build `/projects/{project}/subscription/{subscription_name}`. 

Should we leave `subscription_name` as it is because it's just a string that's the subscription name?

tswast: It's not clear that it's a string without some kind of comment or example. (Some APIs like Bigtable use bytes for names, so it's not obvious from context)

PR #1674, opened 2018-08-30
Upgrading the metrics query to the latest version

billyjacobson:


bigtable/metricscaler/metricscaler.py
```diff
@@ -16,10 +16,14 @@
 programmatically scale a Google Cloud Bigtable cluster."""
 
 import argparse
+import os
 import time
 
 from google.cloud import bigtable
-from google.cloud import monitoring
+from google.cloud import monitoring_v3
+from google.cloud.monitoring_v3.query import Query
```
tswast: Import modules not classes. `from google.cloud.monitoring_v3 import query`

bigtable/metricscaler/metricscaler.py
```diff
@@ -16,10 +16,14 @@
...
+
+PROJECT = os.environ['GCLOUD_PROJECT']
```
tswast: Usually we use `GOOGLE_CLOUD_PROJECT` but I forget if we set that for the Python tests.

PR #1682, opened 2018-08-31
Memorystore GCE : Replace  GCS_APP_LOCATION with GCS_BUCKET_NAME

jabubake:
Note: GCS_APP_LOCATION prone to user error , simplifying to GCS_BUCKET_NAME param


jabubake:
@kurtisvg FYI

PR #1684, opened 2018-09-05
Add GCF imagemagick samples

ace-n:
~~**TODO** - add Kokoro test config. (@andrewsg this is good to review though.)~~

**EDIT:** Kokoro configs need to be added for GCF separately. We should do this in a separate PR.

### DIFF
--- a/functions/imagemagick/main.py
+++ b/functions/imagemagick/main.py
@@ -31,25 +31,25 @@ def blur_offensive_images(data, context):
     blob = data
 
     # Exit if this is a deletion or a deploy event.
-    if blob.resource_state == 'not_exists':
-        print('This is a deletion event.')
-        return
-    elif 'name' not in blob:
+    if 'name' not in blob:
         print('This is a deploy event.')
+        return
+    elif blob['resource_state'] == 'not_exists':
+        print('This is a deletion event.')
         return
 
     file_name = blob['name']
-    file = storage_client.bucket(blob.bucket).file(file_name)
-    file_path = 'gs://%s/%s' % (blob.bucket, file_name)
+    file = storage_client.bucket(blob['bucket']).file(file_name)
+    file_path = 'gs://%s/%s' % (blob['bucket'], file_name)
 
-    print('Analyzing %s' % file_name)
+    print('Analyzing %s.' % file_name)
 
     result = vision_client.safe_search_detection(file_path)
     detected = result.safe_search_annotation
 
     if detected.adult == 'VERY_LIKELY' or detected.violence == 'VERY_LIKELY':
         print('The image %s was detected as inappropriate.' % file_name)
-        return blur_image(file)
+        return __blur_image(file)
     else:
         print('The image %s was detected as OK.' % file_name)
 # [END functions_imagemagick_analyze]
@@ -57,8 +57,8 @@ def blur_offensive_images(data, context):
 
 # [START functions_imagemagick_blur]
 # Blurs the given file using ImageMagick.
-def blur_image(file):
-    file_name = file["name"]
+def __blur_image(file):
+    file_name = file['name']
     temp_local_filename = '/tmp/%s' % os.path.basename(file_name)
 
     # Download file from bucket.
--- a/functions/imagemagick/main.py
+++ b/functions/imagemagick/main.py
@@ -15,6 +15,7 @@
 
 # [START functions_imagemagick_setup]
 import os
+import tempfile
 
 from google.cloud import storage, vision
 from wand.image import Image
@@ -51,7 +52,7 @@ def blur_offensive_images(data, context):
 # Blurs the given file using ImageMagick.
 def __blur_image(blob):
     file_name = blob.name
-    temp_local_filename = '/tmp/%s' % os.path.basename(file_name)
+    _, temp_local_filename = tempfile.mkstemp()
 
     # Download file from bucket.
     blob.download_to_filename(temp_local_filename)
--- a/functions/imagemagick/main.py
+++ b/functions/imagemagick/main.py
@@ -28,28 +28,29 @@ vision_client = vision.ImageAnnotatorClient()
 # [START functions_imagemagick_analyze]
 # Blurs uploaded images that are flagged as Adult or Violence.
 def blur_offensive_images(data, context):
-    blob = data
+    file = data
 
     # Exit if this is a deletion or a deploy event.
-    if 'name' not in blob:
+    if 'name' not in file:
         print('This is a deploy event.')
         return
-    elif blob['resource_state'] == 'not_exists':
+    elif file.get('resource_state', None) == 'not_exists':
         print('This is a deletion event.')
         return
 
-    file_name = blob['name']
-    file = storage_client.bucket(blob['bucket']).file(file_name)
-    file_path = 'gs://%s/%s' % (blob['bucket'], file_name)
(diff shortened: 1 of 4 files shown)


functions/imagemagick/main.py
```diff
@@ -0,0 +1,86 @@
...
+    if 'name' not in file:
+        print('This is a deploy event.')
+        return
+    elif file.get('resource_state', None) == 'not_exists':
+        print('This is a deletion event.')
+        return
+
+    file_name = file['name']
+    blob = storage_client.bucket(file['bucket']).get_blob(file_name)
+    blob_uri = 'gs://%s/%s' % (file['bucket'], file_name)
```
andrewsg: Is there an attribute on the blob object itself that contains the full URI? If not, we can construct it like this.

functions/imagemagick/main.py
```diff
@@ -0,0 +1,86 @@
...
+        return __blur_image(blob)
+    else:
+        print('The image %s was detected as OK.' % file_name)
+# [END functions_imagemagick_analyze]
+
+
+# [START functions_imagemagick_blur]
+# Blurs the given file using ImageMagick.
+def __blur_image(blob):
+    print(blob)
```
andrewsg: What does this look like when it reaches the logs, stringified? Should we be preface it so it's clear to the user what it means? Or was this a debugging print left in by accident?

functions/imagemagick/main.py
```diff
@@ -0,0 +1,86 @@
...
+
+    file_name = blob.name
+    temp_local_filename = '/tmp/%s' % os.path.basename(file_name)
+
+    # Download file from bucket.
+    blob.download_to_filename(temp_local_filename)
+    print('Image %s was downloaded to %s.' % (file_name, temp_local_filename))
+
+    # Blur the image using ImageMagick.
+    subprocess.check_call([
```
andrewsg: Is there any in-Python way to do this, for instance by using Pillow or an ImageMagick binding for Python, that doesn't involve a subprocess call?

functions/imagemagick/main_test.py
```diff
@@ -0,0 +1,134 @@
...
+# limitations under the License.
+
+import uuid
+
+from mock import MagicMock, patch
+
+import main
+
+
+class DictObject(dict):
```
andrewsg: Depending on what you need this for, you might be able to use Python 3's "UserDict".

ace-n: UserDict only works in Python 3 - I'm assuming we don't care about supporting 2.7?

functions/imagemagick/main.py
ace-n: Yes - Python has [Wand](http://docs.wand-py.org/en/0.4.4/).

For consistency, we should also do this in Node.js - which uses `[gm](https://www.npmjs.com/package/gm)`

functions/imagemagick/main.py
ace-n: The latter - I removed it.

functions/imagemagick/main.py
ace-n: No - the blob object's attributes have the form `b/[BUCKET]/o/[OBJECT]`, whereas the Vision API requires `gs://[BUCKET]/[OBJECT]`.

functions/imagemagick/main.py
```diff
@@ -0,0 +1,86 @@
...
+
+storage_client = storage.Client()
+vision_client = vision.ImageAnnotatorClient()
+# [END functions_imagemagick_setup]
+
+
+# [START functions_imagemagick_analyze]
+# Blurs uploaded images that are flagged as Adult or Violence.
+def blur_offensive_images(data, context):
+    file = data
```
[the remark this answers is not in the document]
ace-n: Changed to `file_data`

functions/imagemagick/main.py
```diff
@@ -0,0 +1,86 @@
...
+
+    # Exit if this is a deletion or a deploy event.
```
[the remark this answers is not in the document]
ace-n: Given that we can configure triggers to fire only on certain events (e.g. [file writes](https://cloud.google.com/functions/docs/calling/storage#object_finalize)), probably not.

We can delete this code from the Node sample as well - but we'll want to explicitly specify which event to use in the docs.

functions/imagemagick/main_test.py
ace-n: Done.

functions/imagemagick/main.py
```diff
@@ -0,0 +1,73 @@
...
+    else:
+        print('The image %s was detected as OK.' % file_name)
+# [END functions_imagemagick_analyze]
+
+
+# [START functions_imagemagick_blur]
+# Blurs the given file using ImageMagick.
+def __blur_image(blob):
+    file_name = blob.name
+    temp_local_filename = '/tmp/%s' % os.path.basename(file_name)
```
andrewsg: Rather than hardcoding /tmp, can we use `tempfile.mkdtemp` to get a random, guaranteed-temporary directory? (note, we should clean this up afterwards as well)

functions/imagemagick/main.py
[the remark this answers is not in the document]
andrewsg: What do you want to do in this situation then, leave it or change it somehow? We should be careful to follow the most straightforward, best-documented practice we can in either this or the Node case.

functions/imagemagick/main.py
andrewsg: Great, `wand` is a good solution here.

functions/imagemagick/main.py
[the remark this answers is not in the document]
andrewsg: Oh, I see it's been changed in the latest revision. Never mind. Thanks!

functions/imagemagick/main.py
andrewsg: (One reason is to avoid filename conflicts or race conditions.)

ace-n: Done (with `mkstemp` instead.)

ace-n:
Best-practices question: this function triggers _twice_ - once when the original file is uploaded, and a second time when the file is blurred. (The Node sample also does this.)

Should we address this in the sample and/or docs?

andrewsg:
That's also the workaround I was thinking of. An alternative would be to
prefix the names of blur uploads with a constant like "blurred-" and have
the function look for that prefix and abort, with a comment explaining the
issue.

I would prefer some workaround or another be implemented because it seems
like not fixing this would open the function up to an infinite loop if the
content identifier malfunctions upstream, and that customers modifying this
function to meet their needs could easily also run into a similar problem.
The workaround can be commented to alert customers to the potential issue
as well.

ace-n:
In the _documented_ use case, it seems like it wouldn't be an issue. (Eventually, the blurring will obscure the image so much that the Vision API doesn't detect anything.)

I agree we should document how to handle _user-provided_ cases though. Would commenting on this in the sample be enough?

andrewsg:
I would personally prefer if we actually architected the sample to make
certain it only runs once. This will make it clearer to the user how to do
it than any comment, and also I'm iffy about relying on the vision
detection algorithm to work and prevent an infininte loop 100% of the time.

Andrew

ace-n:
(Marking as `DO NOT MERGE` until @andrewsg's feedback is addressed.)

PR #1688, opened 2018-09-06
Test Composer samples on Python 3

tswast:
Note: I had to skip the use_local_deps workflow sample because of a
change in relative versus absolute imports between Python 2 and Python
3. I have not yet checked the workflow within Airflow to see if it runs
under Python 3.

@TrevorEdwards please review

### DIFF
--- a/.kokoro/presubmit_tests_composer.cfg
+++ b/.kokoro/presubmit_tests_composer.cfg
@@ -11,5 +11,13 @@ env_vars: {
 
 env_vars: {
     key: "NOX_SESSION"
-    value: "composer and py27 and not appengine"
+    value: "composer and py36 and not appengine"
+}
+
+# Explicitly allow GPL dependencies to let the apache-airflow package
+# installation continue. See:
+# https://github.com/apache/incubator-airflow/pull/3660
+env_vars: {
+    key: "AIRFLOW_GPL_UNIDECODE"
+    value: "yes"
 }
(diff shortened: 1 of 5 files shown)


.kokoro/presubmit_tests_composer.cfg
```diff
@@ -11,5 +11,13 @@ env_vars: {
...
     key: "NOX_SESSION"
-    value: "composer and py27 and not appengine"
+    value: "composer and py36 and not appengine"
+}
+
+# Explicitly allow GPL dependencies to let the apache-airflow package
+# installation continue. See:
+# https://github.com/apache/incubator-airflow/pull/3660
+env_vars: {
+    key: "AIRFLOW_GPL_UNIDECODE"
```
TrevorEdwards: Would SLUGIFY_USES_TEXT_UNIDECODE work?

composer/workflows/use_local_deps_test.py
```diff
@@ -12,7 +12,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import sys
 
+import pytest
+
+
+@pytest.mark.skipif(
```
tswast: Note to self: I can test this on Python 3 by updating the sys.path (PYTHONPATH).

See: composer/workflows/bq_copy_across_locations_test.py added in https://github.com/GoogleCloudPlatform/python-docs-samples/pull/1678

.kokoro/presubmit_tests_composer.cfg
tswast: Done in https://github.com/GoogleCloudPlatform/python-docs-samples/pull/1688/commits/495e7653c5a139a3f37f33d97843d4de08ae61ed

Added environment variable to Travis, too. Looks like this repo hasn't 100% migrated to Kokoro yet.

PR #1692, opened 2018-09-07
Add IAM custom roles and access snippets

awkoren:


iam/api-client/access.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+# [START iam_get_policy]
+def get_policy(project_id):
+    """Gets IAM policy for a project."""
+
+    # pylint: disable=no-member
+    policy = service.projects().getIamPolicy(
+        resource=project_id, body={}).execute()
+    print(policy)
```
tswast: Please introspect in the objects so that we see the relevant fields in the actual code. e.g.

```
print(policy['some_relevant_field_name'])
```

iam/api-client/access.py
```diff
@@ -0,0 +1,131 @@
...
+    return policy
+# [END iam_get_policy]
+
+
+# [START iam_modify_policy_add_member]
+def modify_policy_add_member(policy, role, member):
+    """Adds a new member to a role binding."""
+    binding = next(b for b in policy['bindings'] if b['role'] == role)
+    binding['members'].append(member)
+    print(binding)
```
tswast: Is this missing an API request to actually update the policy?

iam/api-client/custom_roles.py
```diff
@@ -0,0 +1,240 @@
...
+    # pylint: disable=no-member
+    role = service.roles().get(name=name).execute()
+    print(role['name'])
+    for permission in role['includedPermissions']:
+        print(permission)
+# [END iam_get_role]
+
+
+# [START iam_create_role]
+def create_role(name, project, title, description, permissions, stage):
```
tswast: We'd like to avoid general samples if at all possible. Could you remove the arguments and use specific values?

Otherwise, provide example values as comments if not possible to hardcode values for these.

iam/api-client/access.py
```diff
@@ -0,0 +1,131 @@
+# !/usr/bin/env python
+#
+# Copyright 2018 Google Inc. All Rights Reserved.
```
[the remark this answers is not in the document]
awkoren: Fixed.

iam/api-client/access.py
```diff
@@ -0,0 +1,131 @@
...
+For more information, see the documentation at
+https://cloud.google.com/iam/docs/granting-changing-revoking-access.
+"""
+
+import argparse
+import os
+
+from google.oauth2 import service_account
+import googleapiclient.discovery
+
```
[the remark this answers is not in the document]
awkoren: Fixed.

iam/api-client/access.py
```diff
@@ -0,0 +1,131 @@
...
+credentials = service_account.Credentials.from_service_account_file(
+    filename=os.environ['GOOGLE_APPLICATION_CREDENTIALS'],
```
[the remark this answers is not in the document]
awkoren: This is the documented approach: https://developers.google.com/api-client-library/python/auth/service-accounts
google.auth.default() does not seem to work.

iam/api-client/access.py
awkoren: print(policy) dumps the policy object as JSON. This is good since users need to understand and manipulate this object to edit IAM policy. This is explained in the companion article: https://cloud.google.com/iam/docs/granting-changing-revoking-access#overview_of_cloud_iam_policy

iam/api-client/access.py
awkoren: The set_policy() method on like 72 demonstrates that.

iam/api-client/access.py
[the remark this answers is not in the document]
tswast: Odd. It should do the same thing if the environment variable is set.

Note that its signature is slightly different: `google.auth.default()` returns a tuple with the credentials and a default project ID.

iam/api-client/access.py
tswast: I find that a bit strange to separate them. Most how-to guides that I've seen try to keep each code snippet self-contained, but you're right that this matches the docs.

tswast:
@awkoren Travis is showing some lint errors. Please fix.

```
nox > flake8 --show-source --builtin gettext --max-complexity 20 --import-order-style google --exclude .nox,.cache,env,lib,generated_pb2,*_pb2.py,*_pb2_grpc.py --ignore=E121,E123,E126,E226,E24,E704,W503,W504,I100,I201,I202 --application-import-names grantable_roles,grantable_roles_test,quickstart,quickstart_test,service_account_keys,service_accounts,service_accounts_test,access,access_test,custom_roles,custom_roles_test .
./custom_roles.py:4:2: W291 trailing whitespace
# 
 ^
./custom_roles_test.py:2:2: W291 trailing whitespace
# 
 ^
```

PR #1696, opened 2018-09-11
Update django app.yaml to fix threadsafe to a valid value (true vs yes)

Feni:


engelke:
Can not be merged without a signed CLA.

PR #1698, opened 2018-09-11
Update Cloud Tasks Push Queue Sample

averikitsch:


appengine/flexible/tasks/requirements.txt
```diff
@@ -1,3 +1,3 @@
-Flask==1.0.2
-gunicorn==19.9.0
-google-cloud-tasks==0.2.0
+Flask==0.12.2
```
andrewsg: Why are we downgrading dependencies here?

averikitsch: Merging mistake. Fixed now.

PR #1700, opened 2018-09-11
Remove appengine BigQuery sample

tswast:
google-api-python-client no longer uses oauth2client by default. Since
this sample is only linked to and not included in any tutorials,
removing it for now rather than updating it to use google-auth.


kurtisvg:
Do we need an additional CL to remove the links to it?

PR #1703, opened 2018-09-12
GCF SQL: Use proper environment variables

ace-n:
(Missed a spot in #1655)

Replace `SUPERVISOR_HOSTNAME` with something more stable

@jasonpolites please review (and tell me if there's a better way)

### DIFF
--- a/functions/sql/mysql_sample.py
+++ b/functions/sql/mysql_sample.py
@@ -16,8 +16,7 @@
 from os import getenv
 
 import pymysql
-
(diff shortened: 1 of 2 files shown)


ace-n:
Spoke with @jasonpolites and @stew-r internally - we should ditch the env. var approach (except perhaps for Node, since `NODE_ENV` is idiomatic) and use a `try/catch` here.

ace-n:
**Edit**

PR #1705, opened 2018-09-13
Add Composer Environment Migration script

tswast:
Add a script for creating a clone of an existing Cloud Composer
Environment using the latest Composer image version.

CC @janczak10


tswast:
I've confirmed that I can create a new environment from an old environment with this script.

PR #1708, opened 2018-09-16
Update dataproc/submit_job_to_cluster.py

jameswinegar:
switch region to new 'global' region and remove unnecessary function.


jameswinegar:
I signed it!

PR #1710, opened 2018-09-17
Fix DPEBot auto-update failure

ace-n:


testing/test-env.tmpl.sh
```diff
@@ -1,5 +1,6 @@
 # Environment variables for system tests.
 export GCLOUD_PROJECT=your-project-id
+export GCP_PROJECT=$GCLOUD_PROJECT
```
kurtisvg: Rather than adding an additional environment variable as an alias, can we update the sample to use the correct variable?

ace-n: `GCLOUD_PROJECT` is marked as "legacy" in the [GCF docs](https://cloud.google.com/functions/docs/env-var#reserved_keys_key_validation).

PR #1713, opened 2018-09-21
Updates dependencies

gguuss:


gguuss:
@kurtisvg PTAL, I saw we had issues automatically updating the dependencies here.

PR #1715, opened 2018-09-21
Remove unused and outdated Natural Language samples

alixhami:
Removes unused and outdated tutorials and samples. Specifically:
+ v1beta2/ folder because no samples are in beta
+ Tutorials using googleapiclient.discovery


beccasaurus:
My recommendation would be to do away with all stale code using googleapiclient.discovery. It's confusing to users.

Also, out standard practice is:

> if it's not in the docs (either cloud.google.com product pages or cloud.google.com/community) then it's safe to delete

If we want to keep code, it needs to be published somewhere

LGTM

alixhami:
cc @beccasaurus

PR #1716, opened 2018-09-21
Remove unused and outdated Vision tutorial

alixhami:
Unused and outdated (using googleapiclient.discovery)


dizcology:
Should be okay since these samples seem to all have their equivalent using google-cloud-vision client.

PR #1722, opened 2018-09-25
add test modules to requirements.txt for iot examples

hongalex:
Noticed requirements.txt for python virtualenv was missing the test modules (pytest/mock are explicitly required, and flaky is specified in pytest.ini in python-docs-samples root directory)

### DIFF
--- a/iot/api-client/http_example/requirements.txt
+++ b/iot/api-client/http_example/requirements.txt
@@ -5,3 +5,6 @@ google-auth==1.5.1
 google-cloud-pubsub==0.37.2
 pyjwt==1.6.4
 requests==2.19.1
+flaky
+mock
+pytest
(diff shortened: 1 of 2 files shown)


iot/api-client/http_example/requirements.txt
```diff
@@ -5,3 +5,6 @@ google-auth==1.5.1
 google-cloud-pubsub==0.37.2
 pyjwt==1.6.4
 requests==2.19.1
+flaky
```
gguuss: Lock the modules to a version number in case the testing syntax changes, also note that the testing infrastructure will install these modules so we generally don't have all users install these modules when running the samples. I recommend leaving as-is (without test dependencies) for consistency with other samples. If you feel strongly otherwise, happy to discuss.

hongalex: Sounds good to me. I figured that was the case, but I also saw a few other projects in here that had the test modules included [like here](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/bigquery/bqml/requirements.txt) so wanted to double check

PR #1723, opened 2018-09-25
KMS changes

DanSanche:
implements changes suggested in b/113025362 and b/113025156

### DIFF
--- a/kms/api-client/asymmetric.py
+++ b/kms/api-client/asymmetric.py
@@ -63,7 +63,7 @@ def decryptRSA(ciphertext, client, key_path):
 
 
 # [START kms_encrypt_rsa]
-def encryptRSA(message, client, key_path):
+def encryptRSA(plaintext, client, key_path):
     """
     Encrypt message locally using an 'RSA_DECRYPT_OAEP_2048_SHA256' public
     key retrieved from Cloud KMS
@@ -72,7 +72,7 @@ def encryptRSA(message, client, key_path):
     pad = padding.OAEP(mgf=padding.MGF1(algorithm=hashes.SHA256()),
                        algorithm=hashes.SHA256(),
                        label=None)
-    ciphertext = public_key.encrypt(message, pad)
+    ciphertext = public_key.encrypt(plaintext, pad)
     return base64.b64encode(ciphertext)
 # [END kms_encrypt_rsa]
 
--- a/kms/api-client/asymmetric.py
+++ b/kms/api-client/asymmetric.py
@@ -48,7 +48,7 @@ def decryptRSA(ciphertext, client, key_path):
     Decrypt a given ciphertext using an 'RSA_DECRYPT_OAEP_2048_SHA256' private
     key stored on Cloud KMS
     """
-    request_body = {'ciphertext': base64.b64encode(ciphertext).decode()}
+    request_body = {'ciphertext': ciphertext.decode()}
     request = client.projects() \
                     .locations() \
                     .keyRings() \
@@ -72,7 +72,8 @@ def encryptRSA(message, client, key_path):
     pad = padding.OAEP(mgf=padding.MGF1(algorithm=hashes.SHA256()),
                        algorithm=hashes.SHA256(),
                        label=None)
-    return public_key.encrypt(message, pad)
+    ciphertext = public_key.encrypt(message, pad)
+    return base64.b64encode(ciphertext)
 # [END kms_encrypt_rsa]
 
 
--- a/kms/api-client/asymmetric.py
+++ b/kms/api-client/asymmetric.py
@@ -82,7 +82,7 @@ def signAsymmetric(message, client, key_path):
     """
     # Note: some key algorithms will require a different hash function
     # For example, EC_SIGN_P384_SHA384 requires SHA384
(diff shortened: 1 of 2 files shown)


kms/api-client/asymmetric_test.py
```diff
@@ -99,44 +99,55 @@ def test_get_public_key(self):
...
-                                       self.client,
-                                       self.rsaDecrypt)
+        ciphertext_bytes = sample.encryptRSA(self.message_bytes,
+                                             self.client,
+                                             self.rsaDecrypt)
+        ciphertext = ciphertext_bytes.decode('utf-8')
         # ciphertext should be 344 characters with base64 and RSA 2048
         assert len(ciphertext) == 344, \
             'ciphertext should be 344 chars; got {}'.format(len(ciphertext))
         assert ciphertext[-2:] == '==', 'cipher text should end with =='
```
russ-: This is an odd detail to assert, and I'm surprised the test isn't flaky because of it. The API doesn't guarantee anything about the content of the ciphertext. I'd remove this assertion.

kms/api-client/asymmetric_test.py
```diff
@@ -99,44 +99,55 @@ def test_get_public_key(self):
...
         assert len(sig) == 344, \
             'sig should be 344 chars; got {}'.format(len(sig))
         assert sig[-2:] == '==', 'sig should end with =='
         success = sample.verifySignatureRSA(sig,
-                                            self.message,
+                                            self.message_bytes,
                                             self.client,
                                             self.rsaSign)
         assert success is True, 'RSA verification failed'
+        changed_bytes = (self.message+".").encode('utf-8')
```
russ-: Nit: alternatively, `changed+bytes = message_bytes + b'.'`.

kms/api-client/asymmetric.py
```diff
@@ -69,9 +72,8 @@ def encryptRSA(message, client, key_path):
     pad = padding.OAEP(mgf=padding.MGF1(algorithm=hashes.SHA256()),
                        algorithm=hashes.SHA256(),
                        label=None)
-    ciphertext = public_key.encrypt(message.encode('ascii'), pad)
-    ciphertext = base64.b64encode(ciphertext).decode('utf-8')
-    return ciphertext
+    ciphertext = public_key.encrypt(plaintext, pad)
+    return base64.b64encode(ciphertext)
```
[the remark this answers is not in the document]
DanSanche: make sense, I moved the encoding into the decrypt request body

kms/api-client/asymmetric_test.py
DanSanche: removed

kms/api-client/asymmetric_test.py
DanSanche: That does look nicer. Changed

kms/api-client/asymmetric.py
```diff
@@ -102,11 +103,11 @@ def signAsymmetric(message, client, key_path):
...
     """
     Verify the validity of an 'RSA_SIGN_PSS_2048_SHA256' signature for the
-    specified plaintext message
+    specified message
     """
     public_key = getAsymmetricPublicKey(client, key_path)
 
-    digest_bytes = hashlib.sha256(message.encode('ascii')).digest()
+    digest_bytes = hashlib.sha256(message).digest()
     sig_bytes = base64.b64decode(signature)
```
russ-: The signature passed to verifySignature{RSA,EC} should be raw binary. signAsymmetric should b64decode the signature received in the response.

kms/api-client/asymmetric_test.py
```diff
@@ -99,44 +100,54 @@ def test_get_public_key(self):
...
         # ciphertext should be 344 characters with base64 and RSA 2048
         assert len(ciphertext) == 344, \
             'ciphertext should be 344 chars; got {}'.format(len(ciphertext))
-        assert ciphertext[-2:] == '==', 'cipher text should end with =='
-        plaintext = sample.decryptRSA(ciphertext, self.client, self.rsaDecrypt)
+        plaintext_bytes = sample.decryptRSA(ciphertext_bytes,
+                                            self.client,
+                                            self.rsaDecrypt)
+        assert plaintext_bytes == self.message_bytes
+        plaintext = plaintext_bytes.decode('utf-8')
```
russ-: Is there much value in checking that plaintext == self.message if you've already checked that plaintext_bytes == self.message_bytes?

DanSanche: Probably not, but I'd rather err on the side of too many tests, especially when dealing with all these encoding changes

kms/api-client/asymmetric.py
DanSanche: fixed

kms/api-client/asymmetric_test.py
engelke: Most of the time you specify utf-8 for string.encode() and bytes.decode(), except on line 51 where you just assume the default (which is utf-8). Both ways work fine, but this is a minor inconsistency.

DanSanche: good point, we should probably keep that consistent. Fixed

PR #1728, opened 2018-10-03
Fix for error when missing MySQLdb

iotmani:
While looking into the cause of an exception:

    ModuleNotFoundError: No module named 'MySQLdb'

I noticed the pymysql initialisation was missing, yet it was [done for the python2 tutorial settings.py counterpart](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/appengine/standard/django/mysite/settings.py#L95-L103)..

Note that this was noticed even when running the app in ["production" on GAE](https://console.cloud.google.com/errors/CIv9orry98OZRw?time=PT1H&project=expanded-talon-217815) (link for my own error in case it's accessible by Google employees). 
So worth looking into if this is the case here:

    ...This is a convenience feature for developers who cannot install
    # MySQLdb locally; when running in production on Google App Engine Standard
    # Environment, MySQLdb will be used.

### DIFF
--- a/appengine/standard_python37/django/mysite/settings.py
+++ b/appengine/standard_python37/django/mysite/settings.py
@@ -95,7 +95,7 @@ if os.getenv('GAE_APPLICATION', None):
             'HOST': '/cloudsql/[YOUR-CONNECTION-NAME]',
             'USER': '[YOUR-USERNAME]',
             'PASSWORD': '[YOUR-PASSWORD]',
-            'NAME': 'polls',
+            'NAME': '[YOUR-DATABASE]',
         }
     }
 else:
--- a/appengine/standard_python37/django/mysite/settings.py
+++ b/appengine/standard_python37/django/mysite/settings.py
@@ -95,6 +95,7 @@ if os.getenv('GAE_APPLICATION', None):
             'HOST': '/cloudsql/[YOUR-CONNECTION-NAME]',
             'USER': '[YOUR-USERNAME]',
             'PASSWORD': '[YOUR-PASSWORD]',
+            'NAME': 'polls',
         }
     }
 else:
--- a/appengine/standard_python37/django/mysite/settings.py
+++ b/appengine/standard_python37/django/mysite/settings.py
@@ -79,15 +79,11 @@ WSGI_APPLICATION = 'mysite.wsgi.application'
 # Database
 # https://docs.djangoproject.com/en/2.1/ref/settings/#databases
 
-# Check to see if MySQLdb is available; if not, have pymysql masquerade as
-# MySQLdb. This is a convenience feature for developers who cannot install
-# MySQLdb locally; when running in production on Google App Engine Standard
-# Environment, MySQLdb will be used.
-try:
-    import MySQLdb  # noqa: F401
-except ImportError:
-    import pymysql
-    pymysql.install_as_MySQLdb()
+# Install PyMySQL as mysqlclient/MySQLdb to use Django's mysqlclient adapter
+# See https://docs.djangoproject.com/en/2.1/ref/databases/#mysql-db-api-drivers for
+# more information
+import pymysql
+pymysql.install_as_MySQLdb()
 
 # [START db_setup]
 if os.getenv('GAE_APPLICATION', None):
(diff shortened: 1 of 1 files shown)


iotmani:
@theacodes can you please review :)? Ref https://github.com/GoogleCloudPlatform/python-docs-samples/pull/1170/commits/90bfd6a0c63f93d775a32562932041eb18b21f8d.

theacodes:
@andrewsg should review, not me.

andrewsg:
Thank you! This is indeed an issue but actually I think we can just use pymysql across the board here, since it's in the requirements file (mysqldb is not) and pymysql is fully compatible with the new Standard runtime (whereas it's not with the Python 2 one).

@michaelawyu Can you please switch over to using pymysql here entirely, and also look into why our tests didn't catch the mismatch?

michaelawyu:
@iotmani I am afraid there's some confusion here. [MySQLdb does not support Python 3](https://pypi.org/project/MySQL-python/); Django 2 however, requires Python 3.

iotmani:
I see... I think we might have a bug in Django 2 in that case (that we'll probably want to handle here nonetheless)...

The [sample code uses engine `'django.db.backends.mysql'`](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/appengine/standard_python37/django/mysite/settings.py#L88).
Looking at the Django 2 code, [it does import MySQLdb](https://github.com/django/django/blob/stable/2.0.x/django/db/backends/mysql/base.py#L14-L20).


Stacktrace I got while trying to deploy the sample app in the first comment:

    django.core.exceptions.ImproperlyConfigured: Error loading MySQLdb module.
    at <module> (/env/lib/python3.7/site-packages/django/db/backends/mysql/base.py:20)
    at import_module (/env/lib/python3.7/importlib/__init__.py:127)
    at load_backend (/env/lib/python3.7/site-packages/django/db/utils.py:110)
    at __getitem__ (/env/lib/python3.7/site-packages/django/db/utils.py:202)
    at __getattr__ (/env/lib/python3.7/site-packages/django/db/__init__.py:33)
    at contribute_to_class (/env/lib/python3.7/site-packages/django/db/models/options.py:203)
    at add_to_class (/env/lib/python3.7/site-packages/django/db/models/base.py:305)
    at __new__ (/env/lib/python3.7/site-packages/django/db/models/base.py:101)
    at <module> (/srv/polls/models.py:7)
    at import_module (/env/lib/python3.7/importlib/__init__.py:127)
    at import_models (/env/lib/python3.7/site-packages/django/apps/config.py:198)
    at populate (/env/lib/python3.7/site-packages/django/apps/registry.py:112)
    at setup (/env/lib/python3.7/site-packages/django/__init__.py:24)
    at <module> (/srv/main.py:1)
    at import_app (/env/lib/python3.7/site-packages/gunicorn/util.py:350)
    at load_wsgiapp (/env/lib/python3.7/site-packages/gunicorn/app/wsgiapp.py:41)
    at load (/env/lib/python3.7/site-packages/gunicorn/app/wsgiapp.py:52)
    at wsgi (/env/lib/python3.7/site-packages/gunicorn/app/base.py:67)
    at load_wsgi (/env/lib/python3.7/site-packages/gunicorn/workers/base.py:138)
    at init_process (/env/lib/python3.7/site-packages/gunicorn/workers/base.py:129)
    at init_process (/env/lib/python3.7/site-packages/gunicorn/workers/gthread.py:104)
    at spawn_worker (/env/lib/python3.7/site-packages/gunicorn/arbiter.py:583)

michaelawyu:
Hi @iotmani,

I looked further into this issue; Django 2 requires `mysqlclient`, which is a fork of `MySQLdbv1` (that actually supports Python 3), so technically speaking Django's error message is not wrong. At this moment, the best solution might be to ask `PyMySQL` to install as `MySQLdb`, as suggested in your patch, though the comment will need to be updated as App Engine Python 3 runtime no longer uses `MySQLdb` or `mysqlclient`. I just take the liberty to edit your PR; if you are OK with it I will approve and merge it.

Thanks again for reporting and fixing the problem :)

iotmani:
Perfect, thanks @michaelawyu!

PR #1729, opened 2018-10-03
Database name missing for production configuration

iotmani:
Name added as per the instructions in:
https://cloud.google.com/python/django/appengine#configure_the_database_settings

Note that this is a discrepancy compared to [python2 example](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/appengine/standard/django/mysite/settings.py#L113):

    'NAME': 'polls',


iotmani:
Replaced by https://github.com/GoogleCloudPlatform/python-docs-samples/pull/1747

PR #1730, opened 2018-10-03
Fix serving static files on production

iotmani:
Django collectstatic adds files to ``static/`` folder rather than ``static_files/``.


andrewsg:
@michaelawyu Same here, please check on this and if it's correct, look into why our tests or other process didn't catch it. Thanks @iotmani!

PR #1738, opened 2018-10-05
Access Display Names of enum fields via enum object

beccasaurus:
To get the display name of enum fields of protobuf objects, we had been copy/pasting the enum field names from the .proto into our code and using the int index of the enum field to get the display name

Updated various samples to get enum field display names the newer, improved way:

```py
from google.cloud.mylibrary import enums

enum_object = enums.EnumName(response.enum_field)

print('The display name of enum_field is: {}'.format(enum_object.name))
```

Related: https://github.com/googleapis/google-cloud-python/issues/6169

Language library needed to be updated to make use of this feature. Worked out-of-the-box for Video Intelligence and Text-to-Speech


video/cloud-client/analyze/analyze_test.py
```diff
@@ -22,6 +22,9 @@
 
 
 BUCKET = os.environ['CLOUD_STORAGE_BUCKET']
+
+
+# Videos used for these tests can be found in the gs://demomaker/ public bucket
```
alixhami: I'm going to send a PR to update these tests to explicitly use the demomaker bucket to save future contributor friction, so this comment will no longer be necessary. Please remove so we can avoid merge conflicts.

PR #1739, opened 2018-10-06
Add Firestore reactive sample

ace-n:
**Lint** and **tests** are passing locally.


functions/firebase/main.py
```diff
@@ -69,3 +73,25 @@ def hello_auth(data, context):
...
+
+# [START functions_firebase_reactive]
+client = firestore.Client()
+
+
+# Converts strings added to /messages/{pushId}/original to uppercase
+def make_upper_case(data, context):
+    path_parts = context.resource.split('/documents/')[1].split('/')
+    collection_path = path_parts[0]
+    document_path = '/'.join(path_parts[1:])
```
ace-n: I don't suppose there are helpers for this?

andrewsg: os.path handles paths intelligently, is that something we can leverage? If that's not a good fit, can you explain what you're intending to do? I can't seem to process this code in my head right now.

andrewsg: Honestly I am not sure. There may also be useful url/uri manipulation features elsewhere in the standard library if that is better.

ace-n: Everything I've found works with the _right_ side of a path, not the _left_. Only functions that do the latter would be convenient here.

Also, Firestore document names cannot contain the `/` character (according to the Cloud Console) - so we _shouldn't_ have to worry about character escaping.

ace-n:
Related sub-builds are passing - merging.

PR #1740, opened 2018-10-09
Add analytics sample

ace-n:
~~**Do not review** until I have confirmed that this works and the `WIP` in the title has been removed.~~

- Add Firebase analytics sample
- Switch to static values in tests


ace-n:
Kokoro failures are unrelated - merging.

PR #1742, opened 2018-10-09
Added Spanner DML/PDML samples. 

kurtisvg:


spanner/cloud-client/snippets.py
```diff
@@ -748,6 +748,245 @@ def query_nested_struct_field(instance_id, database_id):
 # [END spanner_field_access_on_nested_struct_parameters]
 
 
+# [START spanner_dml_standard_insert]
+def insert_data_with_dml(instance_id, database_id):
```
tswast: Ideally we'd show example values for instance_id and database_id

PR #1745, opened 2018-10-10
Remove unused code

krissdap:
There is an extra line of
<code>
    parser = argparse.ArgumentParser()
</code>

which is unused. 
May be it can be removed.


krissdap:
I signed it!

krissdap:
Thank you @alixhami .

PR #1747, opened 2018-10-10
Database name missing for on production

iotmani:
Name added as per the instructions in:
https://cloud.google.com/python/django/appengine#configure_the_database_settings

    5. Replace [YOUR-DATABASE] with the name you choose during the "Initialize your Cloud SQL instance" step.

Note the [python2 example](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/appengine/standard/django/mysite/settings.py#L113) the name was put in directly, so worth aligning them:

    'NAME': 'polls',


iotmani:
@michaelawyu if you can please take a look at this too, it's minor but it's referenced by the official documentation.

Thanks in advance.

iotmani:
Change done as part of https://github.com/GoogleCloudPlatform/python-docs-samples/pull/1728.
This has now been implemented.

PR #1751, opened 2018-10-11
Add missing requirements.txt for a quickstart

engelke:


engelke:
Now passes presubmit tests.

PR #1753, opened 2018-10-12
Pub/Sub: moved import statements inside region tags

anguillanneuf:
- moved imports inside region tags
- updated var names 
- added more comments for the quickstart code samples


pubsub/cloud-client/subscriber.py
```diff
@@ -90,32 +105,41 @@ def create_push_subscription(project,
...
     its topic, are not modifiable.
     """
     # [START pubsub_update_push_configuration]
-    # endpoint = "https://my-test-project.appspot.com/push"
+    from google.cloud import pubsub_v1
+
+    # TODO project_id = "Your Google Cloud Project ID"
+    # TODO topic_name = "Your Pub/Sub topic name"
+    # TODO subscription_name = "Your Pub/Sub subscription name"
+    # TODO endpoint = (for example, "https://my-test-project.appspot.com/push")
```
tswast: Why "for example" here and not elsewhere? Seems obvious to me that these are all examples, so I'd drop it.

PR #1754, opened 2018-10-12
bigtable code had print statements instead of print function calls

engelke:
Broken under Python 3


engelke:
Failed presubmit. New PR will fix.

PR #1760, opened 2018-10-15
Fix block variable used instead of page variable in doctext.py for vision tutorial

rcourivaud:


rcourivaud:
I signed it!

engelke:
Thanks for your contribution. Unfortunately, it caused a test to fail because the page object does not have that property.

PR #1770, opened 2018-10-16
Added Bu Sun's updates, fixed some lint errors

engelke:


noxfile.py
```diff
@@ -147,13 +149,17 @@ def _setup_appengine_sdk(session):
     '--show-source', '--builtin', 'gettext', '--max-complexity', '20',
     '--import-order-style', 'google',
     '--exclude', '.nox,.cache,env,lib,generated_pb2,*_pb2.py,*_pb2_grpc.py',
-    '--ignore=E121,E123,E126,E226,E24,E704,W503,W504,I202',
+    '--ignore=E121,E123,E126,E226,E24,E704,W503,W504,I100,I201,I202',
```
tmatsuo, line 152: This ignore list is applied to all the samples, but I don't think it's right.
We'll remove them in https://github.com/GoogleCloudPlatform/python-docs-samples/pull/3727 
@engelke FYI

PR #1773, opened 2018-10-19
Add beta snippets for object tracking / text detection

nnegrey:
Here's the beta samples @anguillanneuf and I paired on.


nnegrey:
We tried to create tests that would be less flaky

PR #1776, opened 2018-10-19
Time series tests use random unique name to avoid limits

engelke:


monitoring/api/v3/cloud-client/snippets.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2017 Google Inc.
```
sirtorry: keep this copyright line

PR #1783, opened 2018-10-22
Add CHC API alpha datasets and FHIR samples and tests.

noerog:


michaelawyu:
Will merge after the test passes.

PR #1785, opened 2018-10-22
Updates needed to use in the docs

navinger:
I don't know why we have two directories -- echo and quickstart -- with almost the same exact sample. The  Endpoints Frameworks Python tutorial (https://cloud.google.com/endpoints/docs/frameworks/python/get-started-frameworks-python#code) is currently pointing to the sample in the echo directory. I think I should change the docs to use the one in the quickstart directory.  

When I do that, I'll need the region tag so that I can display the section of app.yaml that people need to modify in a docwidget. I also changed the domain name to appspot.com since the cloud.goog doesn't work on App Engine.


navinger:
@DanSanche PTAL

navinger:
Looks like my fork wasn't up to date when I made this PR. I'm closing this PR and will clean things up on my end.

PR #1791, opened 2018-10-25
more info on alternatives

dizcology:
Improvement proposed by the product team to give more information about how alternatives work based on beta users' feedback.


video/cloud-client/analyze/beta_snippets.py
```diff
@@ -55,21 +55,26 @@ def speech_transcription(input_uri):
...
-        word = word_info.word
-        start_time = word_info.start_time
-        end_time = word_info.end_time
-        print('\t{}s - {}s: {}'.format(
-            start_time.seconds + start_time.nanos * 1e-9,
-            end_time.seconds + end_time.nanos * 1e-9,
-            word))
+    for speech_transcription in annotation_results.speech_transcriptions:
+
+        # The number of alternatives for each transcription is limited by
```
nnegrey: Would it also be good to state what an alternative is?
My understanding from the name is that there are several different guesses at how to transcribe this video here are our different results of what we think this transcript is during this time frame?

dizcology: Added a few words.

PR #1810, opened 2018-11-01
Update main.py

josteinl:
Fix crash if no translation is needed. Change so that detect_text(), save_result() and translate_text() all use 'lang' in message.
Add check for 'und' (undefined) language. Then don't try to translate.


josteinl:
I signed it!

engelke:
Thanks, but the change to 'lang' for the property caused test failures because it is not in the event object. Take a look and see if you want to update this PR.

PR #1811, opened 2018-11-01
Samples for pandas-gbq migration guide

alixhami:


bigquery/library-migration/requirements.txt
```diff
@@ -0,0 +1,4 @@
+# [START bigquery_migration_pandas_gbq_requirements]
```
tswast: Note: DPEBot will likely strip out these region tags next time there's a client library version bump. You might want to have a separate directory for the pandas-gbq sample so that you can include the whole file without getting the datalab stuff.

alixhami: Because of this, I made the directory specific to pandas-gbq and put all the samples in one file. It'll be easier to maintain with all of them in one file, plus they both need the same fixture for temporary datasets.

PR #1829, opened 2018-11-09
[Asset] Add quick start code for ExportAssets API

peter-zheng-g:


peter-zheng-g:
I don't have permission to merge. Please help merge if you have the permission and feel good to do so. I do have another change for the other API, but I'd like to put it into next pull request (not this one). Thanks a lot for your help!

peter-zheng-g:
Thanks very much!

PR #1834, opened 2018-11-13
[Asset] Add requirements.txt

peter-zheng-g:


asset/cloud-client/requirements.txt
```diff
@@ -0,0 +1,2 @@
+google-cloud-storage==1.38.0
```
gguuss: I'm not sure these are valid module versions, I'm using:

```
google-cloud-storage==1.13.0
google-cloud-asset==0.1.1
```

gguuss: There are more things I missed in my initial review.  

You may want to add some entropy to the bucket name in case it doesn't get deleted.  The test seems to be failing the second time it's run for me. An example of adding some randomness to your tests [can be found here](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/iot/api-client/manager/manager_test.py#L27) where the time is appended to the filename.

I'm seeing the error:

```
TypeError: ('gs://', 'bucket-for-assets2', '/assets-dump.txt') has type tuple, but expected one of: bytes, unicode
``` 

in your test when running on my machine. Are you testing with both Python3 and Python2?

gguuss: The error you're seeing appears to be coming from how you're doing string concatenation:

Replace:
```
-    dump_file_path = "gs://", asset_bucket, "/assets-dump.txt"
```
in your test with
```
+    dump_file_path = 'gs://{}/assets-dump.txt'.format(asset_bucket)
```

You're missing parenthesis on result in `quickstart_exportassets.py`, it should be:

```
print(response.result())
```

In your test file, you're testing for your personal project ID.  This should instead be using the string from earlier:

```
-    assert "uri: \"gs://cai-prober-prod-for-assets/phython-test.txt\"" in out
+    assert dump_file_path in out
```

Anyways, sorry I missed all these before, you caught me sleeping.  Let's get the fixes in tomorrow am.

asset/cloud-client/quickstart_exportassets_test.py
```diff
@@ -43,8 +44,8 @@ def asset_bucket(storage_client):
 
 
 def test_export_assets(asset_bucket, capsys):
-    dump_file_path = "gs://", asset_bucket, "/assets-dump.txt"
+  dump_file_path = 'gs://{}/assets-dump.txt'.format(asset_bucket)
```
gguuss: Insert two spaces here, looks like your indention was off.  Please also add the `result()` addition to `quickstart_exportassets.py` so I can comment on it as well.

peter-zheng-g:
Thanks!

gguuss:
Please update the branch.

asset/cloud-client/quickstart_exportassets.py
```diff
@@ -31,7 +31,7 @@ def export_assets(project_id, dump_file_path):
     output_config = asset_service_pb2.OutputConfig()
     output_config.gcs_destination.uri = dump_file_path
     response = client.export_assets(parent, output_config)
-    print(response.result)
+    print(response.result())
```
gguuss: On line 26-27, please replace the double-quotes with single-quotes. LINT is not picking it up because the lines are commented out.

gguuss: On line 45, please avoid using the hanging indent. I recommend as:

```
parser.add_argument(
        'dump_file_path',
        help='The file ExportAssets will dump assets to, '
        'e.g.: gs://<bucket-name>/asset_dump_file')
```

asset/cloud-client/quickstart_exportassets_test.py
```diff
@@ -43,8 +44,8 @@ def asset_bucket(storage_client):
 
 
 def test_export_assets(asset_bucket, capsys):
-    dump_file_path = "gs://", asset_bucket, "/assets-dump.txt"
```
gguuss: It seems strange that you're passing on line 40 and silencing the exception -- are you regularly seeing exceptions when cleaning up?

asset/cloud-client/quickstart_exportassets_test.py
```diff
@@ -36,15 +37,16 @@ def asset_bucket(storage_client):
 
     try:
         storage_client.delete_bucket(BUCKET)
-    except Exception:
-        pass
+    except Exception as e:
+        print('Failed to delete bucket{}'.format(BUCKET))
```
gguuss: Space between bucket and {}:

```
print('Failed to delete bucket: {}'.format(BUCKET))
```

PR #1840, opened 2018-11-14
Remove extra lines between region tag and method name

noerog:


gguuss:
FYI, lint doesn't like the newline format you're using:

```
./dicomweb.py:49:1: E302 expected 2 blank lines, found 1
def dicomweb_store_instance(
```

noerog:
Whoops, thanks. Fixed.

PR #1841, opened 2018-11-15
[DO NOT MERGE] Pub/Sub: end-to-end standard usage

anguillanneuf:
Performant code meant to demonstrate how to achieve high-throughput publishing/subscribing.


pubsub/cloud-client/quickstart.py
```diff
@@ -19,76 +19,110 @@
 
 
 def end_to_end(project_id, topic_name, subscription_name, num_messages):
-    # [START pubsub_end_to_end]
+    # [START pubsub_quickstart_end2end_basic]
```
kir-titievsky: Could you please keep a version of the old sample with the same tag (pubsub_end_to_end) alongside this one, to minimize docs breakages.  Maybe rename the function in the old sample?

pubsub/cloud-client/quickstart.py
```diff
@@ -19,76 +19,110 @@
...
     import time
 
+    from google.api_core.exceptions import NotFound
     from google.cloud import pubsub_v1
 
     # TODO project_id = "Your Google Cloud Project ID"
```
kir-titievsky: Isn't there a way to grab this from the environment, in case it's configured. 
So something like
```
# replace with you own way of grabbing project id
project_id = google_magic_client_library_that_understands_env_variables.get_active_project()
```

pubsub/cloud-client/quickstart.py
```diff
@@ -19,76 +19,110 @@
...
     # TODO topic_name = "Your Pub/Sub topic name"
+    # TODO subscription_name = "Your Pub/Sub subscription name"
     # TODO num_messages = number of messages to test end-to-end
```
kir-titievsky: Here, I think it's fine to (1) make this a constant (NUM_MESSAGES). and (2) give it an actual value rather than leaving it as TODO for the user.

pubsub/cloud-client/quickstart.py
```diff
@@ -19,76 +19,110 @@
...
 
-    # Instantiates a publisher and subscriber client
+    # Instantiate a publisher and a subscriber client
     publisher = pubsub_v1.PublisherClient()
     subscriber = pubsub_v1.SubscriberClient()
 
     # The `topic_path` method creates a fully qualified identifier
```
kir-titievsky: I don't think these comments are helpful -- they use a lot of space. A curious user could add print topic_path to see what that looks like.

pubsub/cloud-client/quickstart.py
```diff
@@ -19,76 +19,110 @@
...
 
     # The `subscription_path` method creates a fully qualified identifier
     # in the form `projects/{project_id}/subscriptions/{subscription_name}`
     subscription_path = subscriber.subscription_path(
         project_id, subscription_name)
 
-    # Create the topic.
-    topic = publisher.create_topic(topic_path)
-    print('\nTopic created: {}'.format(topic.name))
+    # Create a topic.
```
kir-titievsky: This create & delete subscription functionality should live in their own functions. There is a lot of corner condition checking that distracts from the flow.  Whereas if you read ```create_topic_safely()``` or ```delete_topic_safely()``` the intent is very clear.

pubsub/cloud-client/quickstart.py
```diff
@@ -19,76 +19,110 @@
...
+        subscriber.delete_subscription(subscription_path)
+    except NotFound:
+        pass
+    finally:
+        subscription = subscriber.create_subscription(
+            subscription_path, topic_path)
+        print('Subscription created: \"{}\"'.format(subscription.name))
+
+    # `data` is roughly 10 Kb.
+    data = 'x' * 9600
```
kir-titievsky: A small thing, but I wonder if bytearray https://docs.python.org/2/library/functions.html#bytearray would give you a very precise object size.

pubsub/cloud-client/quickstart.py
```diff
@@ -19,76 +19,110 @@
...
+    pubsub_time = 0.0
 
     # Publish messages.
-    for n in range(num_messages):
-        data = u'Message number {}'.format(n)
-        # Data must be a bytestring
-        data = data.encode('utf-8')
-        # When you publish a message, the client returns a future.
+    for i in range(num_messages):
+        # When we publish a message, the client returns a future.
```
kir-titievsky: I don't think this comment adds anything I can't infer from the code. You named the return time very informatively -- "future" -- so I can see what's happening.  Maybe instead of calling it a "future" call it a "publish_future"

pubsub/cloud-client/quickstart.py
```diff
@@ -19,76 +19,110 @@
...
         future = publisher.publish(topic_path, data=data)
-        print('Published {} of message ID {}.'.format(data, future.result()))
 
-    publish_time = time.time() - publish_begin
+        tracker.update({i: {'index': i,
+                            'pubtime': time.time(),
+                            'subtime': None}})
+
+        # `future.result()` blocks and returns a unique message ID per message.
+        tracker[future.result()] = tracker.pop(i)
```
kir-titievsky: This seems wrong.  You should not have to block on this.  One unique identifier per message you have is the future itself.  So you can have a dictionary of tracker[publisher_future].  When that future returns, you could set it's message id in a callback or put it in a separate tracker keyed on the message id.

pubsub/cloud-client/quickstart.py
```diff
@@ -19,76 +19,110 @@
...
 
-    messages = set()
+    print('\nPublished all messages.')
 
     def callback(message):
-        print('Received message: {}'.format(message))
         # Unacknowledged messages will be sent again.
```
kir-titievsky: I think this is something that is clear from the ack() call -- no need to comment it.   I would change the function name here, though, to make it more descriptive: "process_message_callback_function"

kir-titievsky: ... particularly because this is between the publish and subscribe parts. I don't know if I'm still reading about publish or not.

pubsub/cloud-client/quickstart.py
```diff
@@ -19,76 +19,110 @@
...
-            subscribe_time = time.time() - subscribe_begin
-            print("\nReceived all messages.")
-            print("Publish time lapsed: {:.2f}s.".format(publish_time))
-            print("Subscribe time lapsed: {:.2f}s.".format(subscribe_time))
+        # Deplete messages in `tracker` every time we have complete info
+        # of a message.
+        for msg_id in list(tracker):
+            pubtime = tracker[msg_id]['pubtime']
+            subtime = tracker[msg_id]['subtime']
+            if subtime is not None:
```
kir-titievsky: This seems unnecessarily wordy.  I would say:

for message_record in list(tracker):
if message_record['subtime'] is not None: delivery_time += message_record['subtime'] - message_record['pubtime']
del message_record # this one I'm not completely sure about.  It might just delete the reference.

pubsub/cloud-client/quickstart.py
```diff
@@ -19,76 +19,110 @@
...
+                pubsub_time += subtime - pubtime
+                del tracker[msg_id]
+
+        # Exit if `tracker` is empty i.e. all the messages' publish-subscribe
+        # time have been accounted for.
+        if len(tracker) == 0:
+            print('\nTotal publish to subscribe time for {} messages: \
```
kir-titievsky: I think it would be much more useful to print statistics here.  The total time is mostly determined by the number of messages and batching settings, not system performance.  So I'd recommend making pubsub_time a list of latencies.  And then printing stats of that: average, best time, worst time, and if number of messages > 100, say. 
 10th, 25th ptile, 75th ptile, 90th ptile (if number of record > 100

pubsub/cloud-client/quickstart.py
```diff
@@ -19,76 +19,110 @@
...
+
+
+def end_to_end_standard(project_id, topic_name, subscription_name,
+                        num_messages):
+    # [START pubsub_quickstart_end2end_standard]
+    pass
+    # [END pubsub_quickstart_end2end_standard]
 
 
 if __name__ == '__main__':
```
kir-titievsky: This should have a comment with infocation instructions for the script.

pubsub/cloud-client/quickstart.py
anguillanneuf: When `pubtime` is captured, I feel it's captured just in time after `publish` is called. `tracker[future.result()] = tracker.pop(i)` switches in a new key for the info we already captured. Why do we need a separate tracker? 

```Python
    # Publish messages.
    for i in range(10):
        future = publisher.publish(topic_path, data=data)
        tracker.update({future: {'pubtime': time.time(),
                                 'subtime': None}})
        # Update the old key, which is the publish future, with the unique
        # identifier per message returned by the `result()` method.
        tracker[future.result()] = tracker.pop(future)
```

pubsub/cloud-client/quickstart.py
anguillanneuf: I shall use a simple example without size info for the basic example, and save this for the standard example where batch setting is incorporated. But FWIW: 
```python
>>> import sys
>>> data = 'x' * 9600
>>> data = data.encode('utf-8')
>>> sys.getsizeof(data) # 9.633Kb
9633
```

pubsub/cloud-client/quickstart.py
anguillanneuf: Yes there is! Maybe we can address this in a different PR because currently we use the same TODO format for all our code samples. What do you think?

```Python
project_id = os.environ['GCLOUD_PROJECT']
```

pubsub/cloud-client/quickstart.py
anguillanneuf: The old region tag hasn't been used in the docs yet. So it is still safe to decide what name we want. I will revert to the old one `pubsub_end_to_end`.

pubsub/cloud-client/quickstart.py
kir-titievsky: sg.  not critical.

pubsub/cloud-client/quickstart.py
kir-titievsky: sg

pubsub/cloud-client/quickstart.py
kir-titievsky: I think what I was thinking here is that the call to future.result() blocks on the return of the publish and therefore all messages get published sequentially. This is neither good practice nor fast.  

The solution to this would be to do it in a non-blocking way.  Meaning,  I think, all updates to the tracker would have to be done in a callback for the future. 

Does that make any sense?  It could well be me misunderstanding your intent here.

anguillanneuf: It makes sense now. Thank you!

This is the standard output after moving the code to resolve the future and to update the tracker into a callback: 

```
Published all messages.

Listening for messages...
7
5
1
6
3
8
9
0
2
4

Delivery Statistics
Average time: 2.336982s
Best time: 1.828190s
```

PR #1861, opened 2018-11-21
update samples for product search GA

dizcology:
Update Python samples for Vision API's Product Search GA.  Tested with the updated client library google-cloud-vision==0.35.0 with `from google.cloud.vision_v1 import ProductSearchClient` added to `google/cloud/vision.py`.


engelke:
Consider changing the timeout to 120 from 90. That seems to result in fewer failures.

PR #1867, opened 2018-11-22
[Asset] Add quickstart code for BatchGetAssetsHistory API.

peter-zheng-g:


asset/cloud-client/quickstart_batchgetassetshistory_test.py
```diff
@@ -0,0 +1,55 @@
+#!/usr/bin/env python
+
+# Copyright 2018 Google LLC. All Rights Reserved.
```
gguuss: Remove "All Rights Reserved."

asset/cloud-client/quickstart_batchgetassetshistory_test.py
```diff
@@ -0,0 +1,55 @@
...
+
+def test_batch_get_assets_history(asset_bucket, capsys):
+    bucket_asset_name = '//storage.googleapis.com/{}'.format(BUCKET)
+    asset_names = [bucket_asset_name, ]
+    quickstart_batchgetassetshistory.batch_get_assets_history(
+        PROJECT, asset_names)
+    out, _ = capsys.readouterr()
+
+    if not out:
+        assert bucket_asset_name in out
```
gguuss: Why is the `if not out` in this check?

asset/cloud-client/quickstart_batchgetassetshistory.py
```diff
@@ -0,0 +1,58 @@
...
+    # ["//storage.googleapis.com/[BUCKET_NAME]",]'
+
+    client = asset_v1beta1.AssetServiceClient()
+    parent = client.project_path(project_id)
+    content_type = enums.ContentType.RESOURCE
+    read_time_window = assets_pb2.TimeWindow()
+    read_time_window.start_time.GetCurrentTime()
+    response = client.batch_get_assets_history(
+        parent, content_type, read_time_window, asset_names)
+    print(response)
```
gguuss: It may be better explicitly print the response members using `print('Member: {}'.format(reponse.member))`

asset/cloud-client/quickstart_batchgetassetshistory_test.py
peter-zheng-g: In rare case, out can be empty when the bucket resource change has not been updated into our backend yet. It still means the API call was successful though, otherwise error response will be returned.

asset/cloud-client/quickstart_batchgetassetshistory.py
peter-zheng-g: Committed with the suggested print. Thanks!

gguuss:
I think you need a different member of the response object for tests to pass.

asset/cloud-client/quickstart_batchgetassetshistory.py
peter-zheng-g: > I actually didn't mean member, I meant the relevant member that is included in the response and that we want to show here.
I see... Updated the code to print response.assets. Thanks.

peter-zheng-g:
> I think you need a different member of the response object for tests to pass.
Thanks. Sorry didn't get what you meant at the beginning. Now the code has been updated and the test are passing.

asset/cloud-client/quickstart_batchgetassetshistory.py
```diff
@@ -0,0 +1,58 @@
...
+    read_time_window = assets_pb2.TimeWindow()
+    read_time_window.start_time.GetCurrentTime()
+    response = client.batch_get_assets_history(
+        parent, content_type, read_time_window, asset_names)
+    print('assets: {}'.format(response.assets))
+    # [END asset_quickstart_batch_get_assets_history]
+
+
+if __name__ == '__main__':
+
```
gguuss: I'm not sure you want / need this newline here

peter-zheng-g:
I've resolved the nit. Thanks!

peter-zheng-g:
Thanks!

PR #1884, opened 2018-12-03
Updates library versions and tests

gguuss:


iot/api-client/mqtt_example/cloudiot_mqtt_example.py
```diff
@@ -206,7 +213,7 @@ def parse_command_line_args():
     parser.add_argument(
             '--mqtt_bridge_port',
             choices=(8883, 443),
-            default=8883,
+            default=443,
```
hongalex: shouldn't 8883 be the default (and 443 be backup for firewall issues/testing)?

gguuss: It'd appeared that the tests were failing because of the default port, updating test / fixing.

gguuss: I'm seeing the issue again now.

PR #1898, opened 2018-12-05
speech: add missing import for code sample

anguillanneuf:


speech/cloud-client/transcribe_enhanced_model.py
```diff
@@ -28,7 +28,10 @@
 def transcribe_file_with_enhanced_model(path):
     """Transcribe the given audio file using an enhanced model."""
     # [START speech_transcribe_enhanced_model]
+    import io
```
nnegrey: This import exists at the top of the file.

nnegrey: @beccasaurus Should we also move standard imports inside the functions for imports or just leave them outside.

beccasaurus: Ones like IO for reading files are 👍  – folks will find it helpful so things don't blow up

nnegrey: Sounds good. @anguillanneuf
I'd say we should just make sure we remove the import above then and replace it inside the function where needed. :)

PR #1903, opened 2018-12-05
KMS: Updated for new client library

DanSanche:
Updated all samples to use the new libraries

tracking bug: b/120216809


kms/api-client/asymmetric.py
```diff
@@ -13,71 +13,100 @@
...
+    # Creates an API client for the KMS API.
+    client = kms_v1.KeyManagementServiceClient()
+
+    # The resource name of the KeyRing associated with the CryptoKey.
+    parent = client.key_ring_path(project_id, location_id, key_ring_id)
+
+    # Create the CryptoKey object template
+    purpose = enums.CryptoKey.CryptoKeyPurpose.ASYMMETRIC_DECRYPT
+    algorithm = enums.CryptoKeyVersion.CryptoKeyVersionAlgorithm.\
+        RSA_DECRYPT_OAEP_2048_SHA256,
```
engelke: Remove trailing comma. It turns algorithm into a one-element tuple which makes the create_crypto_key call fail.

kms/api-client/asymmetric.py
```diff
@@ -13,71 +13,100 @@
...
-                    .locations() \
-                    .keyRings() \
-                    .cryptoKeys() \
-                    .cryptoKeyVersions() \
-                    .getPublicKey(name=key_path)
-    response = request.execute()
-    key_txt = response['pem'].encode('ascii')
+
+    client = kms_v1.KeyManagementServiceClient()
+    response = client.get_public_key(key_name)
```
engelke: This function requires the name of a CryptoKeyVersion, not just a CryptoKey, so it fails with names like the example key_name.

kms/api-client/asymmetric.py
```diff
@@ -13,71 +13,100 @@
...
+
+    client = kms_v1.KeyManagementServiceClient()
+    response = client.asymmetric_decrypt(key_name, ciphertext)
+    return response.plaintext
 # [END kms_decrypt_rsa]
 
 
 # [START kms_encrypt_rsa]
-def encryptRSA(plaintext, client, key_path):
+def encrypt_rsa(plaintext, key_name):
```
engelke: Needs a CryptoKeyVersion name, not just a CryptoKey name, or else get_public_key will fail.

kms/api-client/asymmetric.py
```diff
@@ -13,71 +13,100 @@
...
-def decryptRSA(ciphertext, client, key_path):
+def decrypt_rsa(ciphertext, key_name):
     """
     Decrypt the input ciphertext (bytes) using an
     'RSA_DECRYPT_OAEP_2048_SHA256' private key stored on Cloud KMS
 
-    Requires:
-      base64
+    Example key_name:
+      "projects/PROJECT_ID/locations/global/keyRings/RING_ID/cryptoKeys/KEY_ID"
```
engelke: Needs to be a key version name, not just a key name.

kms/api-client/asymmetric.py
```diff
@@ -86,46 +115,51 @@ def encryptRSA(plaintext, client, key_path):
 
 
 # [START kms_sign_asymmetric]
-def signAsymmetric(message, client, key_path):
+def sign_asymmetric(message, key_name):
```
engelke: The only key creator is for RSA encryption keys, not for RSA signing nor EC signing. So how will a user exercise this (and following) snippets?

kms/api-client/asymmetric_test.py
```diff
@@ -16,135 +16,116 @@
...
+             .format(parent, keyring_id, ecSignId)
 
     message = 'test message 123'
     message_bytes = message.encode('utf-8')
 
-    client = discovery.build('cloudkms', 'v1')
-
     def test_get_public_key(self):
-        rsa_key = sample.getAsymmetricPublicKey(self.client, self.rsaDecrypt)
+        rsa_key = sample.get_asymmetric_public_key(self.rsaDecrypt)
```
engelke: Should also test getting an RSA signing key for completeness.

kms/api-client/asymmetric_test.py
```diff
@@ -16,135 +16,116 @@
...
         plaintext = plaintext_bytes.decode('utf-8')
         assert plaintext == self.message
 
     def test_rsa_sign_verify(self):
-        sig = sample.signAsymmetric(self.message_bytes,
-                                    self.client,
-                                    self.rsaSign)
+        sig = sample.sign_asymmetric(self.message_bytes,
+                                     self.rsaSign)
         # ciphertext should be 344 characters with base64 and RSA 2048
```
engelke: Fix comment to reference signature, not ciphertext

kms/api-client/requirements.txt
```diff
@@ -1,4 +1,2 @@
-google-api-python-client==1.7.4
-google-auth==1.6.1
-google-auth-httplib2==0.0.3
-cryptography==2.4.2
+google-cloud-kms==0.2
+cryptography==2.3.1
```
engelke: Is this lower version number deliberate?

kms/api-client/asymmetric_test.py
```diff
@@ -16,135 +16,116 @@
 from os import environ
 from time import sleep
 
+import asymmetric as sample
```
engelke: The other tests in this folder all simply import the module under test. Is there a reason to rename it as sample for this one? If not, please just leave the name as asymmetric.

kms/api-client/asymmetric.py
DanSanche: good catch

kms/api-client/asymmetric_test.py
```diff
@@ -16,135 +16,116 @@
...
         assert plaintext == self.message
 
     def test_rsa_sign_verify(self):
-        sig = sample.signAsymmetric(self.message_bytes,
-                                    self.client,
-                                    self.rsaSign)
+        sig = sample.sign_asymmetric(self.message_bytes,
+                                     self.rsaSign)
         # ciphertext should be 344 characters with base64 and RSA 2048
         assert len(sig) == 256, \
```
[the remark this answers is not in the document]
DanSanche: you're right, 256 bytes. I believe that comment was from when we were returning encoded strings instead of raw bytes

kms/api-client/quickstart.py
```diff
@@ -13,32 +13,37 @@
...
 def run_quickstart():
     # [START kms_quickstart]
     # Imports the Google APIs client library
-    import googleapiclient.discovery
+    from google.cloud import kms_v1
 
     # Your Google Cloud Platform project ID
     project_id = 'YOUR_PROJECT_ID'
+    # [END kms_quickstart]
+    project_id = os.environ['GCLOUD_PROJECT']
```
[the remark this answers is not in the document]
DanSanche: The idea is that the dummy variable will show up in the docs, but it will inject the ENV VAR when running tests. It's a little strange, but I wanted to make sure the project_id variable is consistent with the other languages in the docs, while still keeping the function testable

kms/api-client/quickstart.py
```diff
@@ -13,32 +13,37 @@
...
+    # [START kms_quickstart]
```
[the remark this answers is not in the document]
DanSanche: See above. The region tag ends and then restarts so that line 27 is skipped when pulled into the docs

kms/api-client/requirements.txt
DanSanche: nope, fixed

kms/api-client/asymmetric_test.py
DanSanche: I can't remember what the logic was behind this, but it probably made more sense before a refactor. Fixed

DanSanche:
@engelke I addressed your comments, let me know if there's anything else

PR #1910, opened 2018-12-06
Decode base64.b64decode output as string for Python 3 compatibility.

### LINKED ISSUE #1895: Composer copy_environment.py script fails under Python3
https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/composer/tools/copy_environment.py

No

## Describe the issue

When copying an environment, the script fails with the following stack trace:
```
Composer operation successful.
Starting database transfer.
Obtaining service account `REDACTED` credentials to access SQL database.
File "copy_environment.py", line 732, in <module>
"diskSizeGb": args.override_disk_size_gb,
File "copy_environment.py", line 711, in clone_environment
copy_database(project, existing_env, new_env, running_as_service_account)
File "copy_environment.py", line 598, in copy_database
iam_client, project, gke_service_account_name
File "copy_environment.py", line 298, in create_service_account_key
base64.b64decode(service_account_key.get("privateKeyData", ""))
File "/usr/lib/python3.5/ast.py", line 84, in literal_eval
return _convert(node_or_string)
File "/usr/lib/python3.5/ast.py", line 83, in _convert
raise ValueError('malformed node or string: ' + repr(node))
ValueError: malformed node or string: b'{\n "type": "service_account",\n "project_id": "...", ....}'
```

Root cause is that in Python 3, base64.b64decode returns a byte string which gets passed to ast.literal_eval(), which expects a string.

wwlian:
Closes #1895


jmcarp:
It looks like we're decoding json here, so do we need to use `ast.literal_eval`? I switched to `json.loads` in #1919, and it's working fine for me.

engelke:
Kokoro tests are not running automatically, and are not passing, for reasons unrelated to this PR.

PR #1916, opened 2018-12-07
Updates Cloud IoT core betas

gguuss:


gguuss:
Not intentional, I'll make sure tests still pass after upgrading and will merge.

PR #1919, opened 2018-12-08
Miscellaneous updates to copy composer environment script.

jmcarp:
* Fail fast if required executables missing
* Pass project to storage client
* Set bucket permissions without subprocess
* Handle fuse umount on mac os


jmcarp:
cc @engelke

PR #1925, opened 2018-12-12
Fix update test to use new endpoint

engelke:


pubsub/cloud-client/subscriber_test.py
```diff
@@ -155,13 +155,10 @@ def _():
 
 
 def test_update(subscriber_client, subscription, capsys):
-    ACK_DEADLINE_SECONDS = 100
-
-    subscriber.update_subscription(PROJECT, SUBSCRIPTION, ACK_DEADLINE_SECONDS)
+    subscriber.update_subscription(PROJECT, SUBSCRIPTION, ENDPOINT)
```
anguillanneuf: Instead of updating subscription with itself, let's use a new endpoint, like say

`new_endpoint=ew_endpoint = 'https://{}.appspot.com/push2'.format(PROJECT)`

And then call `subscriber.update_subscription(PROJECT, SUBSCRIPTION, new_endpoint)`.

engelke: Check it now.

PR #1929, opened 2018-12-12
Sample code for Cloud Scheduler instance scheduling tutorial

djmailhot:
https://cloud.google.com/scheduler/docs/start-and-stop-compute-engine-instances-on-a-schedule


gguuss:
@djmailhot Update your branch and the tests should re-run, I don't think we're still running the Kokoro tests as they were flaky.

djmailhot:
This code is so old it no longer works with the intended tutorial.  I will close this PR and submit a new PR when I've updated the code.

PR #1931, opened 2018-12-13
Add DLP code samples for redacting text with the "replace with info type" transformation

jmorrise:
he DLP team is in the process of expanding code samples to include features that we have launched since the initial samples were written.

This pull request adds examples of using the "replace with info type" transformation.


dlp/deid.py
```diff
@@ -557,7 +627,10 @@ def write_data(data):
 
     args = parser.parse_args()
 
-    if args.content == 'deid_mask':
+    if args.content == 'deid_replace_with_info_type':
+        deidentify_replace_with_info_type(args.project, args.item,
```
gguuss: This is valid formatting but I prefer to avoid hanging indent:

```python

deidentify_replace_with_info_type(
    args.project, args.item, args.info_types)

```

leahecole:
Closing due to inactivity

PR #1933, opened 2018-12-14
Per internal documentation complaint, fix the naming.

realjordanna:
The documentation for DLP uses 'dlp' as the instance name.  As this is also the name of the python package, it could be confusing for people new to the API object model so switch to dlp_client.


theacodes:
Some tests seem to be failing/hanging. This needs to be investigated, but they certainly weren't caused by this change. Merging.

PR #1940, opened 2018-12-18
Speech continuous

nnegrey:


speech/cloud-client/transcribe_streaming_indefinite.py
```diff
@@ -0,0 +1,268 @@
...
+
+"""Google Cloud Speech API sample application using the streaming API.
+
+NOTE: This module requires the additional dependency `pyaudio`. To install
+using pip:
+
+    pip install pyaudio
+
+Example usage:
+    python transcribe_streaming_indefinite.py
```
dizcology: `indefinite` or `infinite`?

speech/cloud-client/transcribe_streaming_indefinite.py
```diff
@@ -0,0 +1,268 @@
...
+from six.moves import queue
+
+# Audio recording parameters
+STREAMING_LIMIT = 55000
+SAMPLE_RATE = 16000
+CHUNK_SIZE = int(SAMPLE_RATE / 10)  # 100ms
+
+
+def get_current_time():
+    return int(round(time.time() * 1000))
```
dizcology: probably don't need both `int` and `round`

speech/cloud-client/transcribe_streaming_indefinite.py
```diff
@@ -0,0 +1,268 @@
...
+
+def duration_to_secs(duration):
+    return duration.seconds + (duration.nanos / float(1e9))
+
+
+class ResumableMicrophoneStream:
+    """Opens a recording stream as a generator yielding the audio chunks."""
+    def __init__(self):
+        self._rate = SAMPLE_RATE
+        self._chunk_size = CHUNK_SIZE
```
dizcology: nit: these should be passed into `__init__` as arguments, possible defaulting to those global variables.

speech/cloud-client/transcribe_streaming_indefinite.py
```diff
@@ -0,0 +1,268 @@
...
+        self.start_time = get_current_time()
+
+        # 2 bytes in 16 bit samples
+        self._bytes_per_sample = 2 * self._num_channels
+        self._bytes_per_second = self._rate * self._bytes_per_sample
+
+        self._bytes_per_chunk = (self._chunk_size * self._bytes_per_sample)
+        self._chunks_per_second = (
+                self._bytes_per_second // self._bytes_per_chunk)
+        self._untranscribed = collections.deque(
```
dizcology: it is not very easy to trace through the code to understand how this second buffer `self._untranscribed` is needed.

speech/cloud-client/transcribe_streaming_indefinite.py
nnegrey: Updated

speech/cloud-client/transcribe_streaming_indefinite.py
nnegrey: Right, removed. Before it was holding over audio that might have failed, because it was waiting for the Error of streaming for too long, but updated to restart before that error making it's use redundant.

PR #1949, opened 2019-01-01
 Removed method that causes exception

albertpadin:
Removed unnecessary method (?) that causes exception when following the endpoints tutorial here: https://cloud.google.com/endpoints/docs/frameworks/python/get-started-frameworks-python#generating_the_openapi_document

If I don't remove this method, he error that I get when following the tutorial is: `endpoints.api_exceptions.ApiConfigurationError: audiences must be a dict when third-party issuers (auth0, firebase, etc) are in use.`


gguuss:
Looks like we took so long to approve / resolve that master has drifted from your source branch (sorry about that!). Can you resolve the conflicts and we'll take another look.

andrewferlitsch:
from gguuss: I'd say go ahead and close <== Doing cloud samples issue rotation. It looks like the (non-googler) never responded to your comments in July/Aug. Can we close it. Your call?

PR #1953, opened 2019-01-02
Add PROJECT_ID

engelke:


trace/main.py
```diff
@@ -21,7 +22,9 @@
 from opencensus.trace.exporters import stackdriver_exporter
 import opencensus.trace.tracer
 
-exporter = stackdriver_exporter.StackdriverExporter()
+exporter = stackdriver_exporter.StackdriverExporter(
+    project_id=os.environ['GCLOUD_PROJECT'])
```
SurferJeffAtGoogle: When I search for code in this repo that inspects the environment variable GCLOUD_PROJECT, I find tests primarily, not samples:
https://github.com/GoogleCloudPlatform/python-docs-samples/search?p=4&q=GCLOUD_PROJECT&unscoped_q=GCLOUD_PROJECT

This feels more like a quickstart where we should hard-code a place-holder string.  Is there more context that I don't see that supports inspecting the environment variable here?

PR #1957, opened 2019-01-05
Set END for detach_listener region tag

jlara310:


jlara310:
@kurtisvg can you review?

PR #1967, opened 2019-01-10
iot: update misnamed function in gateway_demo

hongalex:
Mike found out that `gateway_demo.py` was still using the old function name `listen_for_config_messages` instead of `listen_for_config_and_error_messages` which prevented the demo from running


gguuss:
I'll see if I can address this in [my PR to fix flaky tests](https://github.com/GoogleCloudPlatform/python-docs-samples/pull/1960).

PR #1969, opened 2019-01-13
Create firebase.py

meenakshi91:


nnegrey:
Hi, not sure what this is for, but looks like it might be an accidental push of a private token. 

Going to close. 

Thanks!

PR #1976, opened 2019-01-24
storage: bucket policy only samples

frankyn:
Hi,

Samples for the GCS Bucket Policy Only feature. The client library is not yet released to the Python package manager and requires a requirements.txt update before merge.


storage/cloud-client/requirements.txt
```diff
@@ -1,2 +1,3 @@
 google-cloud-pubsub==0.38.0
-google-cloud-storage==1.13.2
+#google-cloud-storage==1.13.2
+../google-cloud-python/storage
```
engelke: Need to have a valid requirements.txt and then pass Kokoro presubmit tests.

frankyn: Sg, should the review wait til then? 

Clarifying context: The client is not yet released. I ran tests locally and they passed.

PR #1988, opened 2019-01-31
Endpoints Auth Cleanup

DanSanche:
Cleaning up endpoints auth between services code as part of multi-language update

related to b/109951097 and cl/228405078

### DIFF
--- a/endpoints/getting-started/clients/google-jwt-client.py
+++ b/endpoints/getting-started/clients/google-jwt-client.py
@@ -22,64 +22,55 @@ import time
 
 import google.auth.crypt
 import google.auth.jwt
+
 import requests
-from six.moves import urllib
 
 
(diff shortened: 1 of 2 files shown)


endpoints/getting-started/clients/google-jwt-client.py
```diff
@@ -22,65 +22,56 @@
...
+                 expiry_length=3600):
 
-    signer = google.auth.crypt.RSASigner.from_service_account_file(
-        service_account_file)
+    """Generates a signed JSON Web Token using a Google API Service Account."""
 
     now = int(time.time())
-    expires = now + 3600  # One hour in seconds
 
+    # build pyload
```
engelke: Nit: payload

DanSanche: good catch

PR #1989, opened 2019-01-31
Update to use new subscribe() syntax

engelke:


dlp/risk.py
```diff
@@ -49,8 +49,34 @@ def numerical_risk_analysis(project, table_project_id, dataset_id, table_id,
...
+                    if prev_value != value:
+                        print('Value at {}% quantile: {}'.format(
+                              percent, value))
+                        prev_value = value
+                subscription.set_result(None)
+            else:
+                # This is not the message we're looking for.
+                message.drop()
+        except Exception as e:
+            # Because this is executing in a thread, an exception won't be
```
andrewsg: Is this still running in a thread even with the new system, or is it running in the main thread? It's possible we no longer need this section.

Repeated below as well

PR #1994, opened 2019-02-04
BigQuery Storage API sample for reading pandas dataframe

tswast:
How to get a pandas DataFrame, fast!

The first two examples use the existing BigQuery client. These examples
create a thread pool and read in parallel. The final example shows using
just the new BigQuery Storage client, but only shows how to read with a
single thread.


bigquery_storage/to_dataframe/main.py
```diff
@@ -0,0 +1,156 @@
...
+    # [END bigquerystorage_pandas_read_table]
+
+
+def query_to_dataframe(bqclient, bqstorageclient, dataset_id):
+    from google.cloud import bigquery
+
+    # [START bigquerystorage_pandas_read_query_results]
+    import uuid
+
+    # Due to a known issue in the BigQuery Storage API (TODO: link to
```
shollyman: Should we consider simply running a large query that emits enough data to avoid the inline?  Pros:  better demonstrates the perf of the new API, and avoids us having to revisit the sample.  Cons:  test time overhead and potential pitfalls for people kicking tires with small results.   Part of this is dependent on how the team will be maintaining their KI list.

tswast: > Cons: test time overhead.

Long test time is my main reason for avoiding queries that return big results. I guess it's not so bad since this repo can test the different directories independently.

> Cons: potential pitfalls for people kicking tires with small results.

The current failure case is rather bad: it returns a successful response, but gives you an empty result set. That's a pretty big pitfall, because catching it requires noticing that you didn't get the data you thought you were going to get.

shollyman: Works for me.  Let's confirm how the KIs will be maintained so you can link them.

bigquery_storage/to_dataframe/main.py
```diff
@@ -0,0 +1,154 @@
...
+    reader = bqstorageclient.read_rows(position)
+
+    # Parse all Avro blocks and create a dataframe. This call requires a
+    # session, because the session contains the schema for the row blocks.
+    dataframe = reader.to_dataframe(session)
+    print(dataframe.head())
+    # [END bigquerystorage_pandas_read_session]
+
+
+if __name__ == "__main__":
```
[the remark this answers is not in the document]
tswast: This is going into a tutorial on how to use the BigQuery Storage API to get your pandas dataframes faster. All the snippets in this file need to make some sense together. Honestly though, I don't intend to have people run this file directly. I can remove the unnecessary argparse logic.

bigquery_storage/to_dataframe/main.py
```diff
@@ -0,0 +1,154 @@
...
+    bqclient = bigquery.Client(credentials=credentials, project=project_id)
+    bqstorageclient = bigquery_storage_v1beta1.BigQueryStorageClient(
+        credentials=credentials
+    )
+    # [END bigquerystorage_pandas_create_client]
+    return bqclient, bqstorageclient, project_id
+
+
+def table_to_dataframe(bqclient, bqstorageclient):
+    from google.cloud import bigquery
```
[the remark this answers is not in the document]
tswast: It felt weird to repeat imports multiple times in the same tutorial, but you're right that it's better to show them.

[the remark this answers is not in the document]
alixhami: Oh it's all for one tutorial? Then I'd just have the imports once in the client initialization snippet. I'd also add "tutorial" somewhere in the region tag name so it's clear they aren't modular samples.

And possibly if you want, you could have a region tag that wraps all the snippets so the user can see what all the code looks like together. That would make this file look kinda weird though because the region tag for the whole tutorial would just have to wrap each of the individual step region tags. Up to you!

bigquery_storage/to_dataframe/main.py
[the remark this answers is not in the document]
alixhami: Yeah I think it's best to remove since it isn't intended to be used, doesn't show up on the docs, and isn't tested.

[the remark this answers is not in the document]
alixhami: nvm, looks like it's tested. But you could remove that test if you removed the CLI :D

bigquery_storage/to_dataframe/main_test.py
```diff
@@ -0,0 +1,38 @@
...
+def temporary_dataset():
+    client = bigquery.Client()
+    dataset_id = "bqstorage_to_dataset_{}".format(uuid.uuid4().hex)
+    dataset_ref = client.dataset(dataset_id)
+    client.create_dataset(dataset_ref)
+    yield dataset_id
+    client.delete_dataset(dataset_id, delete_contents=True)
+
+
+def test_main(capsys, temporary_dataset):
```
alixhami: Are main and test_main necessary now that the CLI is gone?

tswast: You're right. No reason to have anything but `main_test.py`. This inspired me to add another region tag for the temporary dataset creation, too, since I'll need to talk about that in the tutorial as well.

bigquery_storage/to_dataframe/main_test.py
```diff
@@ -0,0 +1,178 @@
...
+def test_query_to_dataframe(capsys, clients, temporary_dataset):
+    bqclient, bqstorageclient, _ = clients
+    dataset_id = temporary_dataset
+
+    # [START bigquerystorage_pandas_tutorial_read_query_results]
+    import uuid
+
+    from google.cloud import bigquery
+
+    # Due to a known issue in the BigQuery Storage API (TODO: link to
```
alixhami: Is this a todo for you?

bigquery_storage/to_dataframe/main_test.py
```diff
@@ -0,0 +1,178 @@
...
+    # [END bigquerystorage_pandas_tutorial_create_dataset]
+    yield dataset_id
+    bqclient.delete_dataset(dataset_ref, delete_contents=True)
+
+
+def test_table_to_dataframe(capsys, clients):
+    bqclient, bqstorageclient, _ = clients
+
+    # [START bigquerystorage_pandas_tutorial_read_table]
+    from google.cloud import bigquery
```
alixhami: I thought since it's a tutorial you were gonna just have these imports in the client initialization snippet?

bigquery_storage/to_dataframe/main_test.py
```diff
@@ -0,0 +1,178 @@
...
+@pytest.fixture
+def temporary_dataset(clients):
+    bqclient, _, _ = clients
+    dataset_id = "bqstorage_to_dataset_{}".format(uuid.uuid4().hex)
+
+    # [START bigquerystorage_pandas_tutorial_create_dataset]
+    from google.cloud import bigquery
+
+    # TODO: Set the dataset_id to the dataset used to store temporary results.
+    # dataset_id = "query_results_dataset"
```
alixhami: Nit (feel free to ignore): I think it would be clearer if you removed the TODO did something like this:

```python
dataset_id = "query_results_dataset"
# [END bigquerystorage_pandas_tutorial_create_dataset]
# Create a unique dataset ID for testing
dataset_id = "bqstorage_to_dataset_{}".format(uuid.uuid4().hex)
# [START bigquerystorage_pandas_tutorial_create_dataset]
```
(This would apply to all variable replacement TODOs in the tutorial)

bigquery_storage/to_dataframe/main_test.py
```diff
@@ -0,0 +1,178 @@
...
+    table.dataset_id = "new_york_trees"
+    table.table_id = "tree_species"
+
+    # Select columns to read with read options. If no read options are
+    # specified, the whole table is read.
+    read_options = bigquery_storage_v1beta1.types.TableReadOptions()
+    read_options.selected_fields.append("species_common_name")
+    read_options.selected_fields.append("fall_color")
+
+    parent = "projects/{}".format(project_id)
```
alixhami: This might be confusing to the user if they don't know it's their project ID, not the "bigquery-public-data" project above. What about using `bqstorageclient.project` or `bqclient.project` instead?

bigquery_storage/to_dataframe/main_test.py
```diff
@@ -0,0 +1,178 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import uuid
+
+import pytest
+
+
+@pytest.fixture
+def clients():
```
alixhami: Nit - it's a little weird that the third thing returned from something called "clients" is project_id. See note below; it may work out to not return project_id here and just pull it from the client.

bigquery_storage/to_dataframe/main_test.py
tswast: Yeah. I won’t submit until we have that filed publicly somewhere.

alixhami: I think it would probably be better to have it be vague (no TODO or bug link) than have a random TODO in here. I think it looks a bit unprofessional to have our TODOs show up in samples.

bigquery_storage/to_dataframe/main_test.py
tswast: Done.

bigquery_storage/to_dataframe/main_test.py
tswast: Done.

bigquery_storage/to_dataframe/main_test.py
tswast: Renamed the variable to `your_project_id`

PR #1999, opened 2019-02-06
Fix a typo in output message. Also, remove duplicate parser assignment.

dawoonchung:
A very minor typo error in draw_hint function, but since the tutorial is one of the starting points for new users, I thought it's worth correcting to avoid confusion.


dawoonchung:
I signed it!

PR #2004, opened 2019-02-07
Cloud Pub/Sub Quickstart V2

anguillanneuf:
Code samples to go along with Cloud Pub/Sub Quickstart V2. 

@kir-titievsky for comments/suggestions.


pubsub/cloud-client/quickstart/pub.py
```diff
@@ -0,0 +1,71 @@
...
+    data = data.encode('utf-8')
+
+    # When you publish a message, the client returns a future.
+    api_future = client.publish(topic_path, data=data)
+    api_future.add_done_callback(get_callback(api_future, data))
+
+    # Keeps the main thread from exiting to handle message processing
+    # in the background.
+    while True:
+        time.sleep(60)
```
kir-titievsky: I think this should be something like while api_future.running(): sleep (0.1) -- this will prevent the code from hanging any longer than necessary.

pubsub/cloud-client/quickstart/pub.py
```diff
@@ -0,0 +1,71 @@
...
+# [START pubsub_quickstart_pub_deps]
+from google.cloud import pubsub_v1
+# [END pubsub_quickstart_pub_deps]
+
+
+def get_callback(api_future, data):
+    """Wraps message data in the context of the callback function."""
+    def callback(api_future):
+        if api_future.exception():
+            print("There was a problem with message {}".format(data))
```
kir-titievsky: This should re-raise the exception. Otherwise we get into the "Marshall problem."

pubsub/cloud-client/quickstart/sub.py
```diff
@@ -0,0 +1,62 @@
...
+    # Initializes the Subscriber client
+    client = pubsub_v1.SubscriberClient()
+    # [END pubsub_quickstart_sub_client]
+    # Creates a fully qualified identifier in the form of
+    # `projects/{project_id}/subscriptions/{subscription_name}`
+    subscription_path = client.subscription_path(
+        project_id, subscription_name)
+
+    def callback(message):
+        print('Received message {} of message ID {}'.format(
```
kir-titievsky: Nit: do logging (print) after acking.

pubsub/cloud-client/quickstart/pub.py
anguillanneuf: This is much better. I'm removing `while true`.

pubsub/cloud-client/quickstart/pub.py
anguillanneuf: Done.

pubsub/cloud-client/quickstart/sub.py
anguillanneuf: Done.

pubsub/cloud-client/quickstart/pub.py
```diff
@@ -0,0 +1,74 @@
...
+    # [START pubsub_quickstart_pub_client]
+    # Initializes the Publisher client
+    client = pubsub_v1.PublisherClient()
+    # [END pubsub_quickstart_pub_client]
+    # Creates a fully qualified identifier in the form of
+    # `projects/{project_id}/topics/{topic_name}`
+    topic_path = client.topic_path(project_id, topic_name)
+
+    # Data sent to Cloud Pub/Sub must be a bytestring
+    data = "Hello, World!"
```
tswast: Optional: Since you only use ASCII, you could write this as `data = b"Hello, World!"` and skip the encode step.

Maybe you want to include some unicode like an emoji to show why encoding is needed?

pubsub/cloud-client/quickstart/pub.py
```diff
@@ -0,0 +1,74 @@
...
+            print("A problem occurred when publishing {}: {}\n".format(
+                data, api_future.exception()))
+            raise
+    return callback
+
+
+def pub(project_id, topic_name):
+    """Publishes a message to a Pub/Sub topic."""
+    # [START pubsub_quickstart_pub_client]
+    # Initializes the Publisher client
```
tswast: Nit: When explaining what the code does (as opposed to "why"), write instructions like you are talking to the developer and telling them the code to write.

Prefer: Initialize a Publisher client

Likewise for the other comments in these samples.

pubsub/cloud-client/quickstart/pub.py
```diff
@@ -0,0 +1,74 @@
...
+
+    # Data sent to Cloud Pub/Sub must be a bytestring
+    data = "Hello, World!"
+    data = data.encode('utf-8')
+
+    # When you publish a message, the client returns a future.
+    api_future = client.publish(topic_path, data=data)
+    api_future.add_done_callback(get_callback(api_future, data))
+
+    # Keeps the main thread from exiting to handle message processing
```
tswast: Prefer:

```
# Keep the main thread from exiting until background message processing is complete.
```

pubsub/cloud-client/quickstart/pub_test.py
```diff
@@ -0,0 +1,52 @@
...
+@pytest.fixture(scope='module')
+def publisher_client():
+    yield pubsub_v1.PublisherClient()
+
+
+@pytest.fixture(scope='module')
+def topic(publisher_client):
+    topic_path = publisher_client.topic_path(PROJECT, TOPIC)
+
+    try:
```
tswast: Why not continue if the topic already exists and delete after the yield? This test will leave a resource on the project after the test runs otherwise.

pubsub/cloud-client/quickstart/sub.py
```diff
@@ -0,0 +1,64 @@
...
+            message, message.message_id))
+        # Acknowledges the message. Unack'ed messages will be redelivered.
+        message.ack()
+        print('Acknolwedged message of message ID {}\n'.format(
```
tswast: Typo:

s/Acknolwedged/Acknowledged

pubsub/cloud-client/quickstart/sub.py
```diff
@@ -0,0 +1,64 @@
...
+            message.message_id))
+
+    client.subscribe(subscription_path, callback=callback)
+    print('Listening for messages on {}..\n'.format(subscription_path))
+
+    # The subscriber is non-blocking. We keep the main thread from exiting
```
tswast: Non-blocking doesn't convey to me that it actually starts up a background thread to wait for messages.

Also, nit: In general, don't use "We" in our technical writing. Tell the developer what to do actively.

Prefer:

Keep the main thread from exiting so that the subscriber can process messages in the background.

("asynchronously in the background" seems a little redundant to me.)

pubsub/cloud-client/quickstart/sub.py
```diff
@@ -0,0 +1,64 @@
...
+    # so it can process messages asynchronously in the background.
+    while True:
```
tswast: Never exit? What if there's an unrecoverable error in the subscriber, can we detect it?

pubsub/cloud-client/quickstart/sub_test.py
```diff
@@ -0,0 +1,98 @@
...
+@pytest.fixture(scope='module')
+def publisher_client():
+    yield pubsub_v1.PublisherClient()
+
+
+@pytest.fixture(scope='module')
+def topic(publisher_client):
+    topic_path = publisher_client.topic_path(PROJECT, TOPIC)
+
+    try:
```
tswast: Ditto. Seems to me that we should delete after `yield`. Plus choose a random topic name in case multiple tests are running in parallel (such as by CI against multiple versions of Python).

pubsub/cloud-client/quickstart/sub_test.py
```diff
@@ -0,0 +1,98 @@
...
+
+    yield SUBSCRIPTION
+
+
+def _publish_messages(publisher_client, topic):
+    data = u'Hello, World!'.encode('utf-8')
+    publisher_client.publish(topic, data=data)
+
+
+def _make_sleep_patch():
```
tswast: Prefer using the `monkeypatch` fixture for pytest. It can clean up after your tests so that the sleep function is restored after the test is done.

pubsub/cloud-client/quickstart/pub.py
anguillanneuf: Done.

pubsub/cloud-client/quickstart/pub_test.py
anguillanneuf: I added a cleanup section in test. Is it what you mean by "delete after the yield"?

pubsub/cloud-client/quickstart/sub.py
anguillanneuf: Done.

pubsub/cloud-client/quickstart/sub.py
anguillanneuf: Error messages will be printed as long as the connection stays open.

pubsub/cloud-client/quickstart/sub_test.py
anguillanneuf: Same as above. I added some cleanup code in test. So far I haven't run into problems with CI against multiple version of Python.

pubsub/cloud-client/quickstart/sub_test.py
anguillanneuf: I need some help with this. In the meanwhile, since other tests are also written using `_make_sleep_patch` in the same repo, shall we start another PR just for this purpose?

pubsub/cloud-client/quickstart/pub_test.py
tswast: No. If the test fails, then the cleanup won't run.

See

https://github.com/googleapis/google-cloud-python/blob/0ba121b5646ce2bc1537ac8be99ad87461a15a06/bigquery/docs/snippets.py#L86-L96

as an example. The fixture yields an object, then the cleanup code runs after the tests finish.

pubsub/cloud-client/quickstart/sub_test.py
tswast: This is the pattern I'm describing: https://stackoverflow.com/a/29110609/101923,

Now that I have a closer look, `mock.patch` does the same thing, but this still seems much more complicated than it needs to be.

pubsub/cloud-client/quickstart/pub_test.py
anguillanneuf: That makes sense. This is cool. 
Done.

pubsub/cloud-client/quickstart/sub_test.py
anguillanneuf: Updated. I used a pytest fixture `to_delete` to collect resources to be deleted this time.

pubsub/cloud-client/quickstart/sub_test.py
anguillanneuf: I tried `monkeypatch` but it created an infinite loop. 

```python
def mock_sleep(sleep):
    time.sleep(10)
    raise RuntimeError('sigil')

monkeypatch.setattr(time, 'sleep', mock_sleep)
```

`while True: time.sleep(60)` would call `time.sleep(10)`, which calls `time.sleep(10)` and never reaches `raise RuntimeError('sigil')`.

PR #2010, opened 2019-02-11
Update datasets.py from v1alpha to v1alpha2

noerog:
Only backwards-incompatible change is deidentify operation. Tests are unaffected.


noerog:
Tests actually are affected because of the change from whiteList to keepList. Made a separate PR in https://github.com/GoogleCloudPlatform/python-docs-samples/pull/2011.

PR #2019, opened 2019-02-21
Fix Firestore order_by descending syntax

willbattel:
The docs currently suggest that the proper way to get documents in descending order is with `firestore.Query.DESCENDING`.

However, this throws an error for me: `Module 'firebase_admin.firestore' has no 'Query' member`.

I've resolved this by instead using `db.Query.DESCENDING`, where `db = firestore.Client()`.

I don't know if this is actually a problem, but this is what I needed to do.


willbattel:
I signed it!

willbattel:
I was mistaken. For some reason pylint is complaining about `firestore.Query.DESCENDING` but it is actually the correct syntax.

PR #2020, opened 2019-02-21
storage: Update signed url sample

frankyn:
Short PR to modify a value used to be consistent with gsutil. 

gcs -> storage


frankyn:
Thank you @engelke

PR #2022, opened 2019-02-25
storage: S3 SDK sample to list buckets

frankyn:
Hi,

The PR adds a new sample to list GCS buckets using the S3 SDK in Python.

- [x] Sample and system test
- [x] Set up integration tests for HMAC key and secret.
- [ ] Code review


storage/s3-sdk/README.rst
```diff
@@ -0,0 +1,73 @@
...
+
+
+Install Dependencies
+++++++++++++++++++++
+
+#. Clone python-docs-samples and change directory to the sample directory you want to use.
+
+    .. code-block:: bash
+
+        $ git clone https://github.com/GoogleCloudPlatform/python-docs-samples.git
```
engelke: Suggest you add the cd command after the clone.

engelke: Also, should this tell the user to set the authentication environment variables?

frankyn: I modified the example to pass in credentials using CLI parameter.

I looked at other samples and I didn't see `cd ./../` when generating README's with templates. For consistency, I'm going to keep as is as it impacts other samples.

frankyn:
Thanks @engelke, I updated the PR with credentials necessary to test this sample. 

PTAL.

PR #2027, opened 2019-03-02
Add Profiler code samples

jqll:
Add code samples for https://cloud.google.com/profiler.

### DIFF
--- a/profiler/quickstart/main.py
+++ b/profiler/quickstart/main.py
@@ -18,19 +18,46 @@
 # [START profiler_python_quickstart]
 import googlecloudprofiler
 
-if __name__ == '__main__':
-    # Profiler initialization, best done as early as possible.
+
+def main():
+    # Profiler initialization. It starts a daemon thread which continuously
+    # collects and uploads profiles. Best done as early as possible.
     try:
         googlecloudprofiler.start(
-            service='service-name',
-            service_version='version',
+            service='hello-profiler',
+            service_version='1.0.1',
             verbose=3,
             # project_id must be set if not running on GCP.
             # project_id='my-project-id',
         )
     except (ValueError, NotImplementedError) as e:
         print(e)  # Handles errors here
-
 # [END profiler_python_quickstart]
+    busyloop()
+
+
+# A loop function which spends 30% CPU time on loop3() and 70% CPU time
+# on loop7().
+def busyloop():
     while True:
+        loop3()
+        loop7()
+
+
+def loop3():
+    for _ in range(3):
+        loop()
+
+
+def loop7():
+    for _ in range(7):
+        loop()
+
+
+def loop():
+    for _ in range(10000):
         pass
+
+
+if __name__ == '__main__':
+    main()
--- a/profiler/quickstart/main.py
+++ b/profiler/quickstart/main.py
@@ -26,12 +26,14 @@ def main():
         googlecloudprofiler.start(
             service='hello-profiler',
             service_version='1.0.1',
+            # verbose is the logging level. 0-error, 1-warning, 2-info,
+            # 3-debug. It defaults to 0 (error) if not set.
             verbose=3,
             # project_id must be set if not running on GCP.
             # project_id='my-project-id',
         )
-    except (ValueError, NotImplementedError) as e:
-        print(e)  # Handles errors here
+    except (ValueError, NotImplementedError) as exc:
+        print(exc)  # Handles errors here
 # [END profiler_python_quickstart]
     busyloop()
 
--- a/profiler/quickstart/main.py
+++ b/profiler/quickstart/main.py
@@ -1,0 +1,36 @@
+#!/usr/bin/env python
+
+# Copyright 2019 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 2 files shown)


jqll:
@tswast sounds like I can't add reviewers. Could you take a look at this PR? We'd like to add some code samples for https://cloud.google.com/profiler. Also, could we request write access to this repo?

@aalexand @kalyanac could you also take a look? This change adds a quickstart sample that works for GCE, GKE and local environment. App Engine samples are to be added.

profiler/quickstart/main.py
```diff
@@ -0,0 +1,36 @@
...
+            verbose=3,
+            # project_id must be set if not running on GCP.
+            # project_id='my-project-id',
+        )
+    except (ValueError, NotImplementedError) as e:
+        print(e)  # Handles errors here
+
+# [END profiler_python_quickstart]
+    while True:
+        pass
```
aalexand: The flame graph is probably going to look quite boring. Would it make sense to make it more interesting by at least having a named function?

profiler/quickstart/main.py
```diff
@@ -0,0 +1,36 @@
...
+"""An example of starting https://cloud.google.com/profiler."""
+
+# [START profiler_python_quickstart]
+import googlecloudprofiler
+
+if __name__ == '__main__':
+    # Profiler initialization, best done as early as possible.
+    try:
+        googlecloudprofiler.start(
+            service='service-name',
```
aalexand: In https://github.com/GoogleCloudPlatform/golang-samples/blob/master/profiler/profiler_quickstart/main.go we specify more realistic service name (and no service version) rather than placeholder string. Should we try to be consistent?

profiler/quickstart/main.py
```diff
@@ -0,0 +1,36 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""An example of starting https://cloud.google.com/profiler."""
+
+# [START profiler_python_quickstart]
+import googlecloudprofiler
+
+if __name__ == '__main__':
+    # Profiler initialization, best done as early as possible.
```
aalexand: Explain why "as early is possible" is important?

profiler/quickstart/main.py
jqll: Updated to use a more realistic servce name. This will be used as the code snippet for using Python agent. I slightly prefer keeping the service version, like this: https://github.com/GoogleCloudPlatform/golang-samples/blob/master/profiler/snippets/snippets.go#L28?

profiler/quickstart/main.py
jqll: Added some functions so that the profile shows CPU time percentage as expected.

profiler/quickstart/main.py
```diff
@@ -0,0 +1,63 @@
...
+    try:
+        googlecloudprofiler.start(
+            service='hello-profiler',
+            service_version='1.0.1',
+            verbose=3,
+            # project_id must be set if not running on GCP.
+            # project_id='my-project-id',
+        )
+    except (ValueError, NotImplementedError) as e:
+        print(e)  # Handles errors here
```
tswast: Can you provide some explanation for when ValueError or NotImplementedError might occur?

profiler/quickstart/main.py
```diff
@@ -0,0 +1,63 @@
...
+    # collects and uploads profiles. Best done as early as possible.
+    try:
+        googlecloudprofiler.start(
+            service='hello-profiler',
+            service_version='1.0.1',
+            verbose=3,
+            # project_id must be set if not running on GCP.
+            # project_id='my-project-id',
+        )
+    except (ValueError, NotImplementedError) as e:
```
tswast: Nit: `exc` not `e`

profiler/quickstart/main.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+def main():
+    # Profiler initialization. It starts a daemon thread which continuously
+    # collects and uploads profiles. Best done as early as possible.
+    try:
+        googlecloudprofiler.start(
+            service='hello-profiler',
+            service_version='1.0.1',
+            verbose=3,
```
tswast: 3 is a magic number. Does the client library provide any constants which may be used instead? Otherwise create a constant for the sample.

jqll: Done.

profiler/quickstart/main.py
jqll: Done.

profiler/quickstart/main.py
jqll: It's a little long and it will be presented in the doc surrounding this code snippet. I'm a little hesitate to make this code snippet too long.

profiler/quickstart/main.py
jqll: I just added a comment. The client code doesn't provide constants for logging levels. This sample will serve as a code snippet of calling the start function. I'm a little hesitate to create a constant, which may confuse the reader.

tswast: Is there a FR to add such constants to the client? I think such constants would be useful.

jqll: Created a FR. Though I'm not sure if we definitely want to do that change, the number is consistent with the profiling agents in other languages.

PR #2028, opened 2019-03-05
Add Profiler code samples for App Engine

jqll:


jqll:
@tswast  @aalexand @kalyanac, could you take a look? It's a copy-paste of the appengine hello-world samples, with profiler initialization code added in all main.py. Tested that the samples can be deployed to appengine as is, and the profiles are collected as expected.

profiler/appengine/flexible/main.py
```diff
@@ -0,0 +1,45 @@
...
+
+# Profiler initialization. It starts a daemon thread which continuously
+# collects and uploads profiles. Best done as early as possible.
+try:
+    # service and service_version can be automatically inferred when
+    # running on App Engine. project_id must be set if not running
+    # on GCP.
+    googlecloudprofiler.start(verbose=3)
+except (ValueError, NotImplementedError) as exc:
+    print(exc)  # Handles errors here
```
kalyanac: nit: handle not handles

profiler/appengine/standard_python37/main.py
```diff
@@ -0,0 +1,48 @@
...
+
+# Profiler initialization. It starts a daemon thread which continuously
+# collects and uploads profiles. Best done as early as possible.
+try:
+    # service and service_version can be automatically inferred when
+    # running on App Engine. project_id must be set if not running
+    # on GCP.
+    googlecloudprofiler.start(verbose=3)
+except (ValueError, NotImplementedError) as exc:
+    print(exc)  # Handles errors here
```
kalyanac: nit: handle not handles

profiler/appengine/flexible/app.yaml
```diff
@@ -0,0 +1,17 @@
+runtime: python
+env: flex
+entrypoint: gunicorn -b :$PORT main:app
+
+runtime_config:
+  python_version: 3
```
kalyanac: It would be helpful to add a comment here that using python_version 3, Python 3.7.x is used which supports both Wall & CPU profiling. 
https://cloud.google.com/appengine/docs/flexible/python/runtime

kalyanac:
@jqll do these simple apps generate any interesting profiles? Would it help to add a specific duration of CPU workload?

profiler/appengine/flexible/main.py
jqll: Done.

profiler/appengine/standard_python37/main.py
jqll: Done.

profiler/appengine/flexible/app.yaml
jqll: Done.

jqll:
> @jqll do these simple apps generate any interesting profiles? Would it help to add a specific duration of CPU workload?

The server keeps checking if there are requests, so we get some interesting profiles even if nobody sends requests. See the screenshot.
screenshot from 2019-03-05 11-08-42

jqll:
> This is a bit of a rubberstamp approval - the code makes sense but it's important to test the samples to make sure they actually work, I might easily miss something.

Tested that both samples can be deployed and we get profiles as expected.

kalyanac:
> The server keeps checking if there are requests, so we get some interesting profiles even if nobody sends requests. See the screenshot.

thanks for the clarification. LGTM.

PR #2034, opened 2019-03-07
Added two samples for "OCR with PDF/TIFF as source files"

happyhuman:
The two new samples are for online document annotation.


happyhuman:
@nnegrey , @dizcology can you please review this PR?

vision/cloud-client/detect/detect.py
```diff
@@ -661,6 +661,95 @@ def detect_document_uri(uri):
 # [END vision_fulltext_detection_gcs]
 
 
+# [START vision_fulltext_detection_pdf]
```
nnegrey: # [START vision_fulltext_detection_pdf_beta]

nnegrey:
Should move these to the `beta_snippets.py` file

vision/cloud-client/detect/detect.py
```diff
@@ -661,6 +661,95 @@ def detect_document_uri(uri):
...
+def detect_pdf_document(path):
+    """Detects document features in a PDF/TIFF/GIF file."""
+    from google.cloud import vision_v1p4beta1 as vision
+    client = vision.ImageAnnotatorClient()
+
+    with io.open(path, 'rb') as pdf_file:
+        content = pdf_file.read()
+
+    # Can be other like 'image/tiff' or 'image/gif'
```
nnegrey: `# Other supported mime_types: image/tiff' or 'image/gif'`

vision/cloud-client/detect/detect.py
```diff
@@ -661,6 +661,95 @@ def detect_document_uri(uri):
...
+    mime_type = 'application/pdf'
+    file_bytes_input_config = vision.types.InputConfig(
```
nnegrey: Let's just name it `input_config` to match the method's parameter.

vision/cloud-client/detect/detect.py
```diff
@@ -661,6 +661,95 @@ def detect_document_uri(uri):
...
+        content=content, mime_type=mime_type)
+
+    feature = vision.types.Feature(
+        type=vision.enums.Feature.Type.DOCUMENT_TEXT_DETECTION)
+    image_context = vision.types.ImageContext()
```
nnegrey: If this is empty, is it still required?

vision/cloud-client/detect/detect.py
```diff
@@ -661,6 +661,95 @@ def detect_document_uri(uri):
...
+    # Annotate the first two pages and the last one (max 5 pages)
+    # First page starts at 1, and not 0. Last page is -1.
+    pages = [1, 2, -1]
```
nnegrey: Is that a max of 5 pages per PDF document?
I don't understand the API's pages, syntax.

vision/cloud-client/detect/detect.py
```diff
@@ -661,6 +661,95 @@ def detect_document_uri(uri):
...
+
+    online_one_request = vision.types.AnnotateFileRequest(
+        input_config=file_bytes_input_config,
+        features=[feature],
+        image_context=image_context,
+        pages=pages)
+
+    response = client.batch_annotate_files(requests=[online_one_request])
```
nnegrey: Does this feature not have a handwritten helper? 
Most `vision` features usually include one.

vision/cloud-client/detect/detect.py
happyhuman: Yeah, took me a while to understand it too. 5 Pages per PDF; and you pass nonzero indices as a list.

vision/cloud-client/detect/detect.py
happyhuman: Not sure. The alpha sample did not have it. Is it a plus or a must to have in this sample?

vision/cloud-client/detect/detect.py
happyhuman: Good point. It is not needed.

vision/cloud-client/detect/detect.py
happyhuman: Done

vision/cloud-client/detect/detect.py
happyhuman: Changed.

vision/cloud-client/detect/detect.py
happyhuman: Done

vision/cloud-client/detect/beta_snippets.py
```diff
@@ -174,6 +174,99 @@ def detect_handwritten_ocr_uri(uri):
...
+def detect_document_features(path):
+    """Detects document features in a PDF/TIFF/GIF file.
+
+    Args:
+    path: The path to the local file.
+    """
+    from google.cloud import vision_v1p4beta1 as vision
+    client = vision.ImageAnnotatorClient()
+
+    with io.open(path, 'rb') as pdf_file:
```
dizcology: if you want to use `io` here then it should be imported inside the function definition so the code snippet on the documentation page does not miss that import.

vision/cloud-client/detect/beta_snippets.py
```diff
@@ -174,6 +174,99 @@ def detect_handwritten_ocr_uri(uri):
...
+    input_config = vision.types.InputConfig(
+        content=content, mime_type=mime_type)
+
+    feature = vision.types.Feature(
+        type=vision.enums.Feature.Type.DOCUMENT_TEXT_DETECTION)
+    # Annotate the first two pages and the last one (max 5 pages)
+    # First page starts at 1, and not 0. Last page is -1.
+    pages = [1, 2, -1]
+
+    online_one_request = vision.types.AnnotateFileRequest(
```
dizcology: why is this called `online_one_request`?

vision/cloud-client/detect/beta_snippets.py
```diff
@@ -174,6 +174,99 @@ def detect_handwritten_ocr_uri(uri):
 # [END vision_handwritten_ocr_gcs_beta]
 
 
+# [START vision_fulltext_detection_pdf_beta]
+def detect_document_features(path):
+    """Detects document features in a PDF/TIFF/GIF file.
```
dizcology: i was under the impression that **document features** includes not only paragraphs/words/symbols but also stuff like tables or images.

vision/cloud-client/detect/beta_snippets_test.py
```diff
@@ -46,9 +47,25 @@ def test_handwritten_ocr(capsys):
...
+    uri = GCS_ROOT + 'handwritten.jpg'
 
     beta_snippets.detect_handwritten_ocr_uri(uri)
 
     out, _ = capsys.readouterr()
     assert 'Cloud Vision API' in out
+
+
+def test_detect_pdf_document(capsys):
+    file_name = os.path.join(RESOURCES, 'kafka.pdf')
```
dizcology: test also TIFF and GIF.

vision/cloud-client/detect/beta_snippets.py
happyhuman: Turns out the builtin function works just fine. I removed io

vision/cloud-client/detect/beta_snippets.py
happyhuman: Just copied the code from alpha. You think request is a better name?

vision/cloud-client/detect/beta_snippets_test.py
happyhuman: That means we are testing the actual API which is beyond the scope of this test.

vision/cloud-client/detect/beta_snippets.py
happyhuman: Done

vision/cloud-client/detect/beta_snippets.py
happyhuman: Possibly. I did not have a good PDF to test that out.

vision/cloud-client/detect/beta_snippets.py
```diff
@@ -174,6 +174,99 @@ def detect_handwritten_ocr_uri(uri):
...
+    from google.cloud import vision_v1p4beta1 as vision
+    client = vision.ImageAnnotatorClient()
+
+    with open(path, 'rb') as pdf_file:
+        content = pdf_file.read()
+
+    # Other supported mime_types: image/tiff' or 'image/gif'
+    mime_type = 'application/pdf'
+    input_config = vision.types.InputConfig(
+        content=content, mime_type=mime_type)
```
nnegrey: nit: (up to you if you want to change) we could even remove the `mime_type` variable
```
input_config = vision.types.InputConfig(
        content=content,
        mime_type='application/pdf')  # Other mime_types: image/tiff, image/gif
```

vision/cloud-client/detect/beta_snippets.py
```diff
@@ -174,6 +174,99 @@ def detect_handwritten_ocr_uri(uri):
...
+
+    feature = vision.types.Feature(
+        type=vision.enums.Feature.Type.DOCUMENT_TEXT_DETECTION)
+    # Annotate the first two pages and the last one (max 5 pages)
+    # First page starts at 1, and not 0. Last page is -1.
+    pages = [1, 2, -1]
```
nnegrey: nit: Since I was confused the first time could add like:
```
pages = [1, 2, -1]  # Syntax for a 3 page pdf file
```

vision/cloud-client/detect/beta_snippets.py
```diff
@@ -196,15 +289,27 @@ def detect_handwritten_ocr_uri(uri):
...
             detect_handwritten_ocr_uri(args.uri)
+        elif 'doc-features' in args.command:
+            detect_handwritten_ocr_uri(args.uri)
     else:
         if 'object-localization' in args.command:
             localize_objects(args.path)
         elif 'handwritten-ocr' in args.command:
             detect_handwritten_ocr(args.path)
+        elif 'doc-features' in args.command:
+            detect_handwritten_ocr(args.uri)
```
nnegrey: args.path

vision/cloud-client/detect/beta_snippets.py
happyhuman: To clarify: your PDF can have many pages, but you get to choose up to 5 pages of it to be processed.

vision/cloud-client/detect/beta_snippets.py
happyhuman: Yeah, but it may be a good idea to keep it for clarity anyway.

vision/cloud-client/detect/beta_snippets_test.py
nnegrey: I thought we usually try to also test/validate the API. 

We get back the result and verify it's what we expected the API to return.

vision/cloud-client/detect/beta_snippets.py
happyhuman: oops. Correct. Let me fix that.

vision/cloud-client/detect/beta_snippets.py
nnegrey: Oh... interesting. That itself might be worth commenting in the docs.

beccasaurus: I appreciate having this as part of the example, with the comment!

As a user, I may not have stumbled upon this otherwise :)

vision/cloud-client/detect/beta_snippets_test.py
beccasaurus: > I thought we usually try to also test/validate the API.
> We get back the result and verify it's what we expected the API to return.

+1 yep

As for whether to test multiple file types, I leave that up to your discretion. I consider that "above and beyond." I wouldn't complain if we had that test. But I also don't think it's necessary.

Precedents:
 - We do not test all of the various types of audio formats which Speech accepts, e.g. mp3/wav/raw/...
 - We do not test all of the various image formats which Vision accepts, e.g. png/jpg/...

nnegrey:
Oh should probably re-run the README generator

vision/cloud-client/detect/beta_snippets_test.py
```diff
@@ -46,9 +47,25 @@ def test_handwritten_ocr(capsys):
...
+    beta_snippets.detect_document_features(file_name)
+    out, _ = capsys.readouterr()
+    assert 'Symbol' in out
+    assert 'Word text' in out
```
nnegrey: Should add a check that we got some `word` back out from the pdf, something like:
weather, textile, business or some other word in the API that we expect. 

Make sure the API works. Like we have in the above tests looking for "Dog" or verifying the handwritten text.

beccasaurus:
@happyhuman is there going to be a follow-up PR which demonstrates [`async_batch_annotate_images`][]?

@d-torres noticed that these 2 samples demonstrate calling the new synchronous version of annotating files/documents: [`batch_annotate_files`][].

 - `vision_fulltext_detection_pdf_beta`
 - `vision_fulltext_detection_pdf_gcs_beta`

----

The `V1` Cloud Vision API currently provides:
 - [`rpc BatchAnnotateImages `](https://github.com/googleapis/googleapis/blob/master/google/cloud/vision/v1/image_annotator.proto#L44) which is a synchronous method for analyzing _images_ using Cloud Vision, e.g. PNG/JPG
 - [`AsyncBatchAnnotateFiles `](https://github.com/googleapis/googleapis/blob/master/google/cloud/vision/v1/image_annotator.proto#L58) which is an asynchronous (Long-Running Operation) method for analyzing _files/documents_ using Cloud Vision, e.g. PDF/TIFF 

The `v1p4beta` API provides these methods:
 - Images
   - [`BatchAnnotateImages`][`batch_annotate_images`] ~ _already exists_
   - **New!** ✅ [`AsyncBatchAnnotateImages`][`async_batch_annotate_images`] ~ long-running requests for big batches of images! 🎉
 - Files/Documents
   - **New!** ✅  [`BatchAnnotateFiles`][`batch_annotate_files`] ~ synchronous requests for small batches of docs/files! 🎉
   - [`AsyncBatchAnnotateFiles`][`async_batch_annotate_files`] ~ _already exists_

---

This PR demonstrates how to use the new synchronous API for files/docs, but it doesn't demonstrate the new long-running operation for large batches of images.

[`batch_annotate_images`]: https://github.com/googleapis/googleapis/blob/master/google/cloud/vision/v1p4beta1/image_annotator.proto#L44
[`batch_annotate_files`]: https://github.com/googleapis/googleapis/blob/master/google/cloud/vision/v1p4beta1/image_annotator.proto#L58
[`async_batch_annotate_files`]: https://github.com/googleapis/googleapis/blob/master/google/cloud/vision/v1p4beta1/image_annotator.proto#L89 
[`async_batch_annotate_images`]: https://github.com/googleapis/googleapis/blob/master/google/cloud/vision/v1p4beta1/image_annotator.proto#L75

PR #2036, opened 2019-03-07
Improvements to facilitate the documentation instructions

dwightjl:
- Add hostname and account arguments to the argument parser.
- Remove unnecessary argument defaults.
- Modify main to accommodate the new arguments.

These changes allow users to run this example outside of a Compute Engine environment if they want.


compute/oslogin/service_account_ssh.py
```diff
@@ -146,16 +146,23 @@ def main(
...
+    parser.add_argument(
+        '--account',
+        help='The service account email.')
+    parser.add_argument(
+        '--hostname',
+        help='The external IP address or hostname for the target instance.')
     args = parser.parse_args()
 
-    main(args.cmd, args.project, args.instance, args.zone)
+    main(args.cmd, args.project, args.instance, args.zone,
```
engelke: Should this be instance=args.instance, zone=args.zone here?

compute/oslogin/service_account_ssh.py
```diff
@@ -97,17 +97,17 @@ def run_ssh(cmd, private_key_file, username, hostname):
 
 
 # [START main]
-def main(
-        cmd, project, instance, zone, oslogin=None, account=None, hostname=None
-        ):
+def main(cmd, project, instance=None, zone=None,
+         oslogin=None, account=None, hostname=None):
```
engelke: oslogin is not provided by the call from the body below.

dwightjl: It shouldn't be necessary unless your application is initializing the API manually. service_account_ssh_test.py does this because it cannot manage SSH keys on the behalf of another service account, so it has to obtain the API key for the service account and initialize the API manually. If you don't provide an oslogin object, main() will generate one for you using whatever application default credentials it can obtain from the instance or environment variables.

compute/oslogin/service_account_ssh.py
dwightjl: Good catch this could fail in some situations. Thanks!

PR #2041, opened 2019-03-14
storage: Add Boto3 List Objects in GCS

frankyn:
Hi @engelke,

Sample to list objects in GCS using S3 SDK.


storage/s3-sdk/list_gcs_objects.py
```diff
@@ -0,0 +1,60 @@
...
+                          endpoint_url="https://storage.googleapis.com",
+                          aws_access_key_id=google_access_key_id,
+                          aws_secret_access_key=google_access_key_secret)
+
+    # Call GCS to list objects in bucket_name
+    response = client.list_objects(Bucket=bucket_name)
+
+    # Print object names
+    print("Objects:")
+    for bucket in response["Contents"]:
```
engelke: Since this is listing objects in a bucket, it would be clearer for this to say `for object in ...` instead of `for bucket in ...`. (Or another word for object, just not bucket.)

engelke: Or "blobs". That's a word I like.

storage/signed_urls/generate_signed_urls.py
```diff
@@ -78,8 +78,10 @@ def generate_signed_url(service_account_file, bucket_name, object_name,
     ordered_headers = collections.OrderedDict(sorted(headers.items()))
     for k, v in ordered_headers.items():
         lower_k = str(k).lower()
```
engelke: lower_k and strip_v? name and value would be better variable name.

storage/signed_urls/generate_signed_urls.py
```diff
@@ -78,8 +78,10 @@ def generate_signed_url(service_account_file, bucket_name, object_name,
...
-        strip_v = str(v).lower()
+        strip_v = v
         canonical_headers += '{}:{}\n'.format(lower_k, strip_v)
+    canonical_headers = "host:storage.googleapis.com\nx-goog-encryption-algorithm:AES56\n" # "x-goog-encryption-key:AAryxNglNkXQY0Wa+h9+7BLSFMhCzPo22MtXUWjOBbI=\nx-goog-encryption-key-sha256:QlCdVONb17U1aCTAjrFvMbnxW/Oul8VAvnG1875WJ3k=\n"
```
engelke: Why is this comment here?

engelke: Shouldn't this be += instead of =? Otherwise earlier values are lost.

storage/signed_urls/generate_signed_urls.py
```diff
@@ -88,6 +90,7 @@ def generate_signed_url(service_account_file, bucket_name, object_name,
         lower_k = str(k).lower()
```
engelke: Similar issue with names. Also, these aren't signed_headers, they're names of the headers that need to be signed, so a different name would be more clear. And no need to use items() since you just want the keys, no values.

storage/signed_urls/generate_signed_urls.py
```diff
@@ -88,6 +90,7 @@ def generate_signed_url(service_account_file, bucket_name, object_name,
...
         signed_headers += '{};'.format(lower_k)
     signed_headers = signed_headers[:-1]  # remove trailing ';'
+    signed_headers = 'host;x-goog-encryption-algorithm' #;x-goog-encryption-key;x-goog-encryption-key-sha256'
```
engelke: Again, why the comment? If not used, shouldn't they be removed?

storage/signed_urls/generate_signed_urls.py
```diff
@@ -105,7 +108,7 @@ def generate_signed_url(service_account_file, bucket_name, object_name,
     for k, v in ordered_query_parameters.items():
```
engelke: Variable names. Something like safe_name, safe_value would be clearer than encoded_k and encoded_v

storage/signed_urls/generate_signed_urls.py
```diff
@@ -159,9 +164,21 @@ def generate_signed_url(service_account_file, bucket_name, object_name,
     parser.add_argument('expiration', help='Expiration Time.')
 
     args = parser.parse_args()
+    headers = {'X-Goog-Encryption-Key': 'AAryxNglNkXQY0Wa+h9+7BLSFMhCzPo22MtXUWjOBbI=',
```
engelke: We don't usually include secrets in samples, instead using a placeholder or getting them from command line arguments or environment variables.

PR #2042, opened 2019-03-14
BigQuery sample writes results to Google Sheets.

kylelandonwright:
This sample writes BigQuery results to Google Sheets.


tswast:
Thanks for the contribution @kylelandonwright , but since we don't have a tutorial for this in the BigQuery docs, we can't accept this contribution. I recommend you submit this as a Cloud Community Tutorial: https://cloud.google.com/community/tutorials/write Tag me on the PR you send to https://github.com/GoogleCloudPlatform/community and I'll help review it.

Oh, and by the way, you can use user credentials for both BigQuery and Sheets. I'd recommend getting a credentials object once and passing it to the constructor for both the BigQuery client and the Sheets API client.

You may also be interested in https://pydata-google-auth.readthedocs.io/en/latest/, which makes the user-based authentication much easier.

PR #2045, opened 2019-03-14
Added the sample for async image batch annotation

happyhuman:
I added the sample for async image batch annotation. I also renamed the function names and region tags for the previous two samples in my last PR.

### DIFF
--- a/vision/cloud-client/detect/beta_snippets.py
+++ b/vision/cloud-client/detect/beta_snippets.py
@@ -177,7 +177,7 @@ def detect_handwritten_ocr_uri(uri):
 # [END vision_handwritten_ocr_gcs_beta]
 
 
-# [START detect_batch_annotate_files_beta]
+# [START vision_batch_annotate_files_beta]
 def detect_batch_annotate_files(path):
     """Detects document features in a PDF/TIFF/GIF file.
 
@@ -225,10 +225,10 @@ def detect_batch_annotate_files(path):
                         for symbol in word.symbols:
                             print('\t\t\tSymbol: {} (confidence: {})'.format(
                                 symbol.text, symbol.confidence))
-# [END detect_batch_annotate_files_beta]
+# [END vision_batch_annotate_files_beta]
 
 
-# [START detect_batch_annotate_files_gcs_beta]
+# [START vision_batch_annotate_files_gcs_beta]
 def detect_batch_annotate_files_uri(gcs_uri):
     """Detects document features in a PDF/TIFF/GIF file.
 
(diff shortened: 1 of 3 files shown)


happyhuman:
@nnegrey and @beccasaurus can you please review this PR?

vision/cloud-client/detect/beta_snippets.py
```diff
@@ -176,8 +177,8 @@ def detect_handwritten_ocr_uri(uri):
 # [END vision_handwritten_ocr_gcs_beta]
 
 
-# [START vision_fulltext_detection_pdf_beta]
-def detect_document_features(path):
+# [START detect_batch_annotate_files_beta]
```
nnegrey: Keep the `vision_` at the start of the region tags.

vision/cloud-client/detect/beta_snippets.py
```diff
@@ -272,7 +273,75 @@ def detect_document_features_uri(gcs_uri):
                         for symbol in word.symbols:
                             print('\t\t\tSymbol: {} (confidence: {})'.format(
                                 symbol.text, symbol.confidence))
-# [END vision_fulltext_detection_pdf_gcs_beta]
+# [END detect_batch_annotate_files_gcs_beta]
+
+
+# [START detect_async_batch_annotate_images_beta]
+def async_batch_annotate_images_uri(image_uri, gcs_uri):
```
nnegrey: `input_image_uri` and `output_uri` might help people distinguish, since both are gcs links, according to the description below.

vision/cloud-client/detect/beta_snippets.py
nnegrey: And for the ones below.

vision/cloud-client/detect/beta_snippets.py
```diff
@@ -272,7 +273,75 @@ def detect_document_features_uri(gcs_uri):
                         for symbol in word.symbols:
                             print('\t\t\tSymbol: {} (confidence: {})'.format(
                                 symbol.text, symbol.confidence))
-# [END vision_fulltext_detection_pdf_gcs_beta]
+# [END detect_batch_annotate_files_gcs_beta]
+
+
+# [START detect_async_batch_annotate_images_beta]
```
nnegrey: nit: `vision_detect_async_batch_annotate_images_gcs_beta`
I feel like we usually put `gcs` in the region tag when the parameters are gcs links, but not sure if this still applies to async.

vision/cloud-client/detect/beta_snippets.py
happyhuman: Not sure what you mean. Can you elaborate a bit more? You mean something like beta_detect_batch_annotate_files?

vision/cloud-client/detect/beta_snippets.py
happyhuman: How strongly do you feel about this? The other methods I see just have 'uri' for the output.

vision/cloud-client/detect/beta_snippets.py
nnegrey:
```
# [vision_detect_batch_annotate_files_beta]
```
We keep the product name at the start of every region tag, so that sample tracker can pick it up.

vision/cloud-client/detect/beta_snippets.py
nnegrey: The other ones use `uri` as input

vision/cloud-client/detect/beta_snippets.py
happyhuman: oops. I read it VERSION the first time. Good catch.  Fixed...

vision/cloud-client/detect/beta_snippets.py
happyhuman: Done

PR #2046, opened 2019-03-14
[Cloud Tasks] Add Http Push Queue Sample

averikitsch:


averikitsch:
Test projects will have to be granted permission or wait until whitelisting is off.

appengine/flexible/tasks/app.flexible.yaml
```diff
@@ -0,0 +1,20 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+runtime: python
+env: flex
+entrypoint: gunicorn -b :$PORT main:app
```
andrewsg: Can we change this to "gunicorn -b :$PORT --threads=4 main:app" to ensure the user does not run into concurrency issues while dealing with asynchronously triggered tasks?

PR #2064, opened 2019-03-22
Set a timeout when waiting for query job to finish.

lbristol88:
Part of the BigQuery App Engine redesign.


appengine/standard_python37/bigquery/templates/query_result.html
```diff
@@ -1,21 +1,31 @@
...
-            <td>{{ result[0] }}</td>
-            <td>{{ result[1] }}</td>
-        </tr>
-        {% endfor %}
-    </table>
-</body>
-</html>
\ No newline at end of file
+<!DOCTYPE html>
+{#
```
engelke: Is it intended to make this text not appear in the page? If so, the syntax for HTML comments is different:

appengine/standard_python37/bigquery/templates/timeout.html
```diff
@@ -0,0 +1,20 @@
+<!DOCTYPE html>
+{#
```
engelke: HTML comment syntax

tswast: This is Jinja2 template syntax so that the license header doesn't appear in the rendered result.

engelke: Got it. Thanks.

PR #2069, opened 2019-03-25
Created new route for results and added redirect for main. 

lbristol88:
Part of the BigQuery App Engine redesign.


appengine/standard_python37/bigquery/main_test.py
```diff
@@ -47,3 +47,9 @@ def fake_query(query):
     r = flask_client.get('/')
     assert r.status_code == 200
     assert 'Query Timeout' in r.data.decode('utf-8')
+
+
+def test_results(flask_client):
+    r = flask_client.get('/')
```
tswast: Does this test pass? I think by default the test client doesn't follow redirects. You'll need to pass in `follow_redirects=True`. See: http://flask.pocoo.org/docs/1.0/testing/#logging-in-and-out

tswast: Also, this test is now identical to `test_main`. You should update `test_main` to check that the redirect goes where expected.

tswast: `test_main_timeout` will also need `follow_redirects=True`. You should rename that test to `test_results_timeout`.

lbristol88: Renamed the functions as directed and made changes to tests in order for them to pass. Thx!

appengine/standard_python37/bigquery/main.py
```diff
@@ -35,20 +36,40 @@ def main():
...
+
+
+@app.route("/results")
+def results():
+    project_id = flask.request.args.get("project_id")
+    job_id = flask.request.args.get("job_id")
+    location = flask.request.args.get("location")
+
+    bigquery_client = bigquery.Client(project=project_id, location=location)
+    query_job = bigquery_client.get_job(job_id)
```
tswast: `get_job` takes a `project` and `location` argument. (See [reference](https://googleapis.github.io/google-cloud-python/latest/bigquery/generated/google.cloud.bigquery.client.Client.html#google.cloud.bigquery.client.Client.get_job)). You don't need to create a client in the line above. Instead, pass the `project` and `location` here to `get_job` on the existing (global) `bigquery_client`.

lbristol88: Removed line 58 and changed to:
`query_job = bigquery_client.get_job(job_id, project=project_id, location=location)` as suggested!

tswast:
@dpebot merge on green

PR #2077, opened 2019-03-29
Demo webserver for managed instances tutorial

djmailhot:
A simple demo web server used in a new interactive tutorial:  [Highly Available Deployments with Autohealing](https://cloud-dot-devsite.googleplex.com/compute/docs/tutorials/highly-available-deployments-with-autohealing)


djmailhot:
The linter fails on a few lines over the line limit that are part of a multi-line HTML template string:

'''IN-LINE
HTML
TEMPLATE
WITH-A-REALLY-REALLY-REALLY-LONG-LINE
AND-SOME-IF-STATEMENTS
'''

Let me know if this is, in fact, an ugly way to define an index.html page for a python HTTP server, and I'll find a way to refactor.

djmailhot:
I've updated the server to use jinja2 templates for the HTML files.  Much less ugly now.

compute/managed-instances/demo/main.py
```diff
@@ -0,0 +1,91 @@
+#!/usr/bin/python3
```
andrewsg: We should add the copyright boilerplate here; copy from any another file.

compute/managed-instances/demo/main.py
```diff
@@ -0,0 +1,91 @@
...
+from http.server import BaseHTTPRequestHandler, HTTPServer
```
andrewsg: It's very unusual for us to use http.server here; almost universally we'd use Flask for a project like this in sample code and then gunicorn to serve (see the newer App Engine samples, App Engine Flex, etc.). Is there a specific reason we should use http.server? It seems like socketserver, ThreadingMixIn, etc. are boilerplate that distract from the main point.

compute/managed-instances/demo/main.py
```diff
@@ -0,0 +1,91 @@
...
+
+class request_handler(BaseHTTPRequestHandler):
+    def get_zone(self):
+        r = get('http://metadata.google.internal/'
+                'computeMetadata/v1/instance/zone',
+                headers={'Metadata-Flavor': 'Google'})
+        if r.status_code == 200:
+            return sub(r'.+zones/(.+)', r'\1', r.text)
+        else:
+            return ''
```
andrewsg: Are there any failure cases here that we should be throwing errors on instead of returning the empty string? This seems like unexpected behavior.

compute/managed-instances/demo/main.py
```diff
@@ -0,0 +1,91 @@
...
+HEALTHY = True
+
+env = Environment(
+    loader=FileSystemLoader(HTML_TEMPLATES_PATH),
+    autoescape=select_autoescape('html')
+)
+
+
+class request_handler(BaseHTTPRequestHandler):
+    def get_zone(self):
```
andrewsg: These helper functions don't actually use "self" passed in, so perhaps they should be top-level standalone functions instead of instance methods on this class. I was confused when I read them because I expected that since they are instance methods of a request handler, they would handle requests and return HTTP responses.

compute/managed-instances/demo/main.py
```diff
@@ -0,0 +1,91 @@
...
+from jinja2 import Environment, FileSystemLoader, select_autoescape
+from os import path
+from requests import get
+from re import sub
+from socket import gethostname
+from socketserver import ThreadingMixIn
+
+PORT_NUMBER = 80
+HTML_TEMPLATES_PATH = path.dirname(path.realpath(__file__)) + '/templates'
+HEALTHY = True
```
andrewsg: Does this sample ever have multiple instances behind a load balancer, or is it intended to run on one machine at a time only?

compute/managed-instances/demo/main.py
```diff
@@ -0,0 +1,91 @@
...
+                self.end_headers()
+            html = env.get_template('health.jinja2').render(healthy=HEALTHY)
+            self.wfile.write(html.encode('UTF-8'))
+
+        else:
+            HOSTNAME = gethostname()
+            ZONE = self.get_zone()
+            TEMPLATE = self.get_template()
+
+            self.send_response(200 if HEALTHY else 500)
```
andrewsg: In this case, is it intended behavior that we may be returning a 500 response but with the exact same template and information as if it were 200?

compute/managed-instances/demo/main.py
djmailhot: In the future, I intend to run this sample with a load balancer.

The global 'healthy' is meant to simply hold state for each individual server.  It is expected that a user might be redirected to different servers by the load balancer.  Please let me know if there is a more idiomatic way of hosting a simple boolean state variable on a flask server (all I can find are examples using databases which seems too overcomplicated for this task).

compute/managed-instances/demo/main.py
djmailhot: Okay, re-written using Flask.

Which invalidates a few other comments here, which I have marked.

compute/managed-instances/demo/main.py
djmailhot: Done

compute/managed-instances/demo/main.py
djmailhot: Refactored with update to Flask

compute/managed-instances/demo/main.py
djmailhot: I believe this was a coding mistake.  This represented the root path '/'.  I've fixed this in the Flask refactoring.

compute/managed-instances/demo/main.py
djmailhot: Unsuccessfully getting the zone is not a critical failure.  The zone (and the template) are fetched simply for informational purposes on the UI.  The demo still works perfectly fine if both fetches fail.  If I were to put something in the failure case, it would be an error message along the lines of 'zone not found'/'template not found'.

compute/managed-instances/demo/app.py
```diff
@@ -0,0 +1,91 @@
...
+    else:
+        return ''
+
+
+def _get_template():
+    r = get('http://metadata.google.internal/'
+            'computeMetadata/v1/instance/attributes/instance-template',
+            headers={'Metadata-Flavor': 'Google'})
+    if r.status_code == 200:
+        return sub(r'.+instanceTemplates/(.+)', r'\1', r.text)
```
andrewsg: What's the purpose of this part? I don't think I fully understand what it's doing. It might benefit from a couple of comments, especially to describe the functionality of the function itself.

djmailhot: Added comments for the entire file.  Let me know if they don't sufficiently describe functionality.

andrewsg:
Looks like there's a linting error with the new comments... sorry it's so nitpicky. If that's resolved I can merge right away. Thanks for your work on this.

djmailhot:
Comments fixed, all checks passed!

andrewsg:
Thanks, I'll merge now.

PR #2078, opened 2019-03-29
Print actual number of messages pulled

olivercoad:
The samples for pulling from Cloud Pub/Sub now prints the actual number of messages received instead of `max_messages` for the request.

### DIFF
--- a/pubsub/cloud-client/subscriber.py
+++ b/pubsub/cloud-client/subscriber.py
@@ -284,7 +284,8 @@ def synchronous_pull(project_id, subscription_name):
     # Acknowledges the received messages so they will not be sent again.
     subscriber.acknowledge(subscription_path, ack_ids)
 
-    print("Received and acknowledged {} messages. Done.".format(len(response.received_messages)))
+    print('Received and acknowledged {} messages. Done.'.format(
+        len(response.received_messages)))
     # [END pubsub_subscriber_sync_pull]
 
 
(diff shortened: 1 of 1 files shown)


olivercoad:
I signed it!

pubsub/cloud-client/subscriber.py
```diff
@@ -284,7 +284,7 @@ def synchronous_pull(project_id, subscription_name):
     # Acknowledges the received messages so they will not be sent again.
     subscriber.acknowledge(subscription_path, ack_ids)
 
-    print("Received and acknowledged {} messages. Done.".format(NUM_MESSAGES))
+    print("Received and acknowledged {} messages. Done.".format(len(response.received_messages)))
```
gguuss: Two small things:

1. Line is over 80 column limit, causing lint to fail.
2. Although the previous string was using a doublequote, as per our style guide, prefer single-quote strings.

pubsub/cloud-client/subscriber.py
```diff
@@ -357,7 +357,7 @@ def worker(msg):
         if processes:
             time.sleep(SLEEP_TIME)
 
-    print("Received and acknowledged {} messages. Done.".format(NUM_MESSAGES))
+    print("Received and acknowledged {} messages. Done.".format(len(response.received_messages)))
```
gguuss: Same feedback as above, prefer single-quotes and add a break before len, e.g.:

```
    print('Received and acknowledged {} messages. Done.'.format(
        len(response.received_messages)))
```

olivercoad:
No problem.

PR #2079, opened 2019-04-02
Add OpenCensus metrics quickstart

c24t:
Add quickstart docs for exporting metrics to stackdriver via the opencensus client.

See previous PRs for the same feature in [go](https://github.com/GoogleCloudPlatform/golang-samples/pull/646), [java](https://github.com/GoogleCloudPlatform/java-docs-samples/pull/1241), and [node](https://github.com/GoogleCloudPlatform/nodejs-docs-samples/pull/1173).


c24t:
cc @mayurkale22

opencensus/metrics_quickstart.py
```diff
@@ -0,0 +1,74 @@
...
+    # Create the Stackdriver stats exporter and start exporting metrics in the
+    # background, once every 60 seconds by default.
+    exporter, transport = stats_exporter.new_stats_exporter()
+
+    # Record 100 fake latency values between 0 and 5 seconds.
+    for ii in range(100):
+        ms = random() * 5 * 1000
+        print("Latency {}: {}".format(ii, ms))
+
+        mmap = stats.stats.stats_recorder.new_measurement_map()
```
andrewsg: This is already imported as from `opensensus.stats import stats`, so that would make this `opencensus.stats.stats.stats.stats_recorder`. Is that really accurate? It's not a typo of too many stats?

opencensus/metrics_quickstart.py
```diff
@@ -0,0 +1,74 @@
...
+    # Register the view. Measurements are only aggregated and exported if
+    # they're associated with a registered view.
+    stats.stats.view_manager.register_view(LATENCY_VIEW)
+
+    # Create the Stackdriver stats exporter and start exporting metrics in the
+    # background, once every 60 seconds by default.
+    exporter, transport = stats_exporter.new_stats_exporter()
+
+    # Record 100 fake latency values between 0 and 5 seconds.
+    for ii in range(100):
```
andrewsg: Why `ii` here instead of `i`? Actually, can we use a more descriptive variable?

c24t: Just to avoid single-letter names. I don't know what would be more descriptive here, possibly `_` to indicate that we're not using the var for anything.

c24t:
@andrewsg let me know if this is what you're looking for in the README and tests.

andrewsg:
Looks great, thanks!

PR #2076, opened 2019-03-28
[CloudAsset] Update client library version to use v1 endpoint.

peter-zheng-g:

PR #2081, opened 2019-04-02
Revert "[CloudAsset] Update client library version to use v1 endpoint."

andrewsg:
Reverts GoogleCloudPlatform/python-docs-samples#2076 due to test failures


busunkim96:
@andrewsg Cloud Asset was re-generated recently, so I think this is a real error. It doesn't have system tests. https://github.com/googleapis/google-cloud-python/tree/master/asset 

CC @peter-zheng-g

peter-zheng-g:
It should be fixed with https://github.com/GoogleCloudPlatform/python-docs-samples/pull/2082. Please take a look. Thanks!

PR #2082, opened 2019-04-02
[CloudAsset] Test: Delay seconds between bucket creation and API call.

peter-zheng-g:
There's some latency between bucket creation and when it's reflected in
the backend. Need to delay in order to make the API call successfully.


busunkim96:
LGTM. 

@andrewsg Could you also take a look? (I don't have write permissions to this repository).

PR #2090, opened 2019-04-04
Tables Notebooks

sirtorry:


beccasaurus:
This is non typical Python docs-samples code, i.e. it's not standard client library samples

These are Notebooks

We don't currently have a clear way of testing these (& automl is already a challenge to test)

I'm merging these as-is. Ping me / these folks is you find this PR and have questions about these.

Happy to revisit these after GCP Next and look into testing / etc, which we're very interested in doing.

Contacts: @beccasaurus @sirtorry @merla18 @andrewferlitsch

PR #2095, opened 2019-04-05
Add Cloud SCC tool app packages

Tabrarian:
Adding app packages previously available to beta customers via .zip download. Docs will be published on https://cloud.google.com/security-command-center/docs/ as of Wednesday, April 10.


Tabrarian:
Closing as I now understand we need to go through a different process for this.

PR #2097, opened 2019-04-05
Cloud Pub/Sub: authenticated push in GAE Standard for Python 3

anguillanneuf:
TL;DR: Authenticated Push is a new Cloud Pub/Sub beta feature. After it has been set up [as such](https://cloud.google.com/sdk/gcloud/reference/beta/pubsub/subscriptions/create), Cloud Pub/Sub generates a JWT and attaches it to the "Authorization" header in the push request. Users can verify the token using the Google Auth Python library to confirm that the push requests indeed come from Cloud Pub/Sub. 

Things to check extra carefully in this PR: 
- `main.py`: `receive_messages_handler`
- `main_test.py`: `test_push_endpoint`

I create a mock token using the private key in `data/privatekey.pem`. I monkeypatched the function to verify oauth2 tokens so that it doesn't get checked against the real Google public certs but what's in `data/public_cert.pem`. These `pem` files are the same as those from the tests in [`google-auth-library-python`](https://github.com/googleapis/google-auth-library-python/tree/81477855ae3de643b8138d646c55b198c4fecedd/tests/data).


appengine/standard_python37/pubsub/README.md
```diff
@@ -0,0 +1,78 @@
...
+
+This demonstrates how to send and receive messages using [Google Cloud Pub/Sub](https://cloud.google.com/pubsub) on [Google App Engine Standard Environment](https://cloud.google.com/appengine/docs/standard/).
+
+## Setup
+
+Before you can run or deploy the sample, you will need to do the following:
+
+1. Enable the Cloud Pub/Sub API in the [Google Developers Console](https://console.developers.google.com/project/_/apiui/apiview/pubsub/overview).
+
+2. Create a topic and subscription. The pusn auth service account must have Service Account Token Creator Role assigned, which can be done in the Cloud Console [IAM & admin](https://console.cloud.google.com/iam-admin/iam) UI.
```
engelke: Typo - 'pusn' instead of 'push'

appengine/standard_python37/pubsub/README.md
```diff
@@ -0,0 +1,78 @@
...
+    $ export PUBSUB_VERIFICATION_TOKEN=[your-verification-token]
+    $ export PUBSUB_TOPIC=[your-topic]
+    $ python main.py
+
+### Simulating push notifications
+
+The application can send messages locally, but it is not able to receive push messages locally. You can, however, simulate a push message by making an HTTP request to the local push notification endpoint. There is an included ``sample_message.json``. You can use
+``curl`` or [httpie](https://github.com/jkbrzt/httpie) to POST this:
+
+    $ curl -i --data @sample_message.json ":8080/_ah/push-handlers/receive_messages?token=[your-token]"
```
engelke: My curl installation doesn't support leaving the hostname off as shown here. I needed "localhost:8080", not just ":8080".

appengine/standard_python37/pubsub/templates/index.html
```diff
@@ -0,0 +1,53 @@
+{#
+# Copyright 2015 Google Inc. All Rights Reserved.
```
engelke: Consider updating the license text (2019, Google LLC).

appengine/standard_python37/pubsub/main.py
```diff
@@ -0,0 +1,108 @@
...
+        return 'Invalid request', 400
+
+    # Verify that the push request originates from Cloud Pub/Sub.
+    try:
+        # Get the Cloud Pub/Sub-generated JWT in the "Authorization" header.
+        bearer_token = request.headers.get('Authorization')
+        token = bearer_token.split(' ')[1]
+        TOKENS.append(token)
+
+        header = jwt.decode_header(token)
```
engelke: Is this verifying the audience ('aud') field? Does it need to? This makes sure it is from pub/sub, but it could possibly be relayed from a push notification to a different app. Or do all push notifications use the same aud value?

anguillanneuf: The 'aud' field is in the JWT claim. If `--push-auth-token-audience` isn't set when creating the push subscription, then the `aud` field will be set to the push endpoint. `jwt.decode_header` decodes the first section of the JWT; it's a convenient function to showcase the `google.auth.jwt` module but there's not much else to it. If `aud` is set, all push notifications use the same value. 

The checking of the JWT signature happens in `id_token.verify_oauth2_token()`. 

I don't have code that specifically checks the `aud` field, should I add? 

```
# JWT header
{"alg":"RS256","kid":"7d680d8c70d44e947133cbd499ebc1a61c3d5abc","typ":"JWT"}

# JWT claim
{
   "aud":"https://example.com",
   "azp":"113774264463038321964",
   "email":"gae-gcp@appspot.gserviceaccount.com",
   "sub":"113774264463038321964",
   "email_verified":true,
   "exp":1550185935,
   "iat":1550182335,
   "iss":"https://accounts.google.com"
  }
```

engelke: No, I don't think so. This all looks very good to me. Go ahead and merge when you're ready, or let me know if you want me to do it.

appengine/standard_python37/pubsub/README.md
```diff
@@ -0,0 +1,77 @@
...
+            --topic [your-topic-name] \
+            --push-endpoint \
+                https://[your-app-id].appspot.com/_ah/push-handlers/receive_messages/token=[your-token] \
+            --ack-deadline 30
+
+3. Update the environment variables in ``app.yaml``.
+
+## Running locally
+
+Refer to the [top-level README](../README.md) for instructions on running and deploying.
```
plamut: This relative link does not work, at least not when previewing the file during the PR. The same is probably the case for several other (sub)pages, thus an absolute URL should probably be used?

appengine/standard_python37/pubsub/README.md
```diff
@@ -0,0 +1,77 @@
...
+
+    OK
+
+After the request completes, you can refresh ``localhost:8080`` and see the message in the list of received messages.
+
+## Running on App Engine
+
+Deploy using `gcloud`:
+
+    gcloud app deploy app.yaml
```
plamut: (minor)
Just for consistency, this command could also be prefixed with `$`, as are all the others in this document.

appengine/standard_python37/pubsub/main_test.py
```diff
@@ -37,20 +80,35 @@ def test_post_index(client):
...
-        })
-    )
+    r = None
+    with _verify_oauth2_token_patch():
+        r = client.post(
+            url,
+            data=json.dumps({
+                "message": {
+                    "data": base64.b64encode(
+                        u'Test message'.encode('utf-8')
```
plamut: (minor)
This could simply be written as `b'Test message'`

appengine/standard_python37/pubsub/main.py
```diff
@@ -57,13 +61,31 @@ def index():
...
+
+        # Verify and decode the token. Underneath it checks the signature
+        # with the signed section using Google's public certs at
+        # https://www.googleapis.com/oauth2/v1/certs
+        req = requests.Request()
+        claim = id_token.verify_oauth2_token(token, req)
+        CLAIMS.append(claim)
+    except Exception as e:
+        CLAIMS.append(e)
+        return 'Unable to verify: ' + e, 400
```
plamut: This line will result in a `TypeError`. In order to avoid it, the exception object must first be converted to a string before concatenation.

appengine/standard_python37/pubsub/main_test.py
```diff
@@ -37,20 +80,35 @@ def test_post_index(client):
...
+                    ).decode('utf-8')
+                }
+            }),
+            header=json.dumps({
+                "Content-Type": "application/json",
+                "Authorization": "Bearer " + token_factory()
```
plamut: If I read this correctly, `token_factory()` returns the result of [google.auth.jwt.encode()](https://google-auth.readthedocs.io/en/latest/reference/google.auth.jwt.html#google.auth.jwt.encode), which returns `bytes`. 

Concatenating a byte string and a unicode string will fail under Python 3, thus if this is a concern, the returned token must be explicitly decoded before concatenation.

appengine/standard_python37/pubsub/main.py
```diff
@@ -40,20 +40,21 @@
...
 
     data = request.form.get('payload', 'Example payload').encode('utf-8')
 
     publisher = pubsub_v1.PublisherClient()
     topic_path = publisher.topic_path(app.config['GCLOUD_PROJECT'],
                                       app.config['PUBSUB_TOPIC'])
-    publisher.publish(topic_path, data)
-
+    future = publisher.publish(topic_path, data)
+    future.result()
```
plamut: Sanity check - are we fine with letting any exception here to simply bubble up, or do we want to explicitly handle it?

appengine/standard_python37/pubsub/main_test.py
```diff
@@ -80,36 +82,27 @@ def test_post_index(client):
...
-                    "data": base64.b64encode(
-                        u'Test message'.encode('utf-8')
-                    ).decode('utf-8')
-                }
-            }),
-            header=json.dumps({
-                "Content-Type": "application/json",
-                "Authorization": "Bearer " + token_factory()
-            })
+    # "".join(chr(x) for x in fake_token)
```
plamut: Seems like a leftover?

appengine/standard_python37/pubsub/data/privatekey.pem
```diff
@@ -0,0 +1,27 @@
+-----BEGIN RSA PRIVATE KEY-----
```
lesv: Are you sure you want to check these in, rather than explain how to generate them in the README?

appengine/standard_python37/pubsub/data/public_cert.pem
```diff
@@ -0,0 +1,19 @@
+-----BEGIN CERTIFICATE-----
```
lesv: ditto

appengine/standard_python37/pubsub/main.py
lesv: @anguillanneuf I'm just getting started reading this, so apologies if you are doing this, but the rule of thumb on these things is to verify the signature before touching any of the fields, then look at the expires and then the audience. (in that order).   Then everything else.

appengine/standard_python37/pubsub/main_test.py
anguillanneuf: Thank you Peter. I fixed it by encoding the result of google.auth.jwt.encode() now.

appengine/standard_python37/pubsub/main_test.py
anguillanneuf: Deleted. Good catch!

appengine/standard_python37/pubsub/.gcloudignore
```diff
@@ -0,0 +1,19 @@
+# This file specifies files that are *not* uploaded to Google Cloud Platform
```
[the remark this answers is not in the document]
anguillanneuf: This file has been removed.

appengine/standard_python37/pubsub/README.md
```diff
@@ -0,0 +1,78 @@
+# Python 3 Google Cloud Pub/Sub sample for Google App Engine Standard Environment
```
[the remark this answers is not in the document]
anguillanneuf: I added a paragraph in ## Running on App Engine that explains the purposes of these pem files.

appengine/standard_python37/pubsub/data/privatekey.pem
anguillanneuf: Yes, I do want to check them in for testing purposes. :) Users shouldn't need to generate these when using authenticated push.

appengine/standard_python37/pubsub/README.md
anguillanneuf: I removed this line because it's not really needed.

appengine/standard_python37/pubsub/README.md
anguillanneuf: Done.

appengine/standard_python37/pubsub/main.py
anguillanneuf: @lesv 1. verifying signature and 2. checking expiration time are both done by the Google Auth Python library. I added 3. checking the audience field in the claim. Thanks for catching this!

appengine/standard_python37/pubsub/main.py
anguillanneuf: I didn't do publish with error handling here. The idea would be to add `future.add_done_callback(callback)`. I think it's okay to let the exception bubble up here.

appengine/standard_python37/pubsub/main_test.py
anguillanneuf: I think it's good to show it this way so it applies to a more general case.

appengine/standard_python37/pubsub/README.md
```diff
@@ -0,0 +1,79 @@
...
+    Server: Werkzeug/0.15.2 Python/3.7.3
+    Date: Sat, 06 Apr 2019 04:56:12 GMT
+
+    Invalid token: 'NoneType' object has no attribute 'split'
+
+The simulated push request fails because it does not have a Cloud Pub/Sub-generated JWT in the "Authorization" header.
+
+## Running on App Engine
+
+Note: Not all the files in the current directory are needed to run your code on App Engine. Specifically, `main_test.py` and the `data` directory, which contains a mocked private key file and a mocked public certs file, are for testing purposes only. You may feel free to leave them out or remove them when deploying your app. IWhen your app is deployed, Cloud Pub/Sub creates tokens using a private key, then the Google Auth Python library takes care of verifying and decoding the token using Google's public certs, to confirm that the push requests indeed come from Cloud Pub/Sub.
```
lesv: typo `IWhen`
You may feel free to ...  => the mocked private key and public cert file are included for testing purposes only. They SHOULD NOT be included in when your app is deployed.

anguillanneuf: Fixed.

anguillanneuf:
Thanks @lesv. I feel good to merge it now. 
And thanks @engelke @plamut for reviewing my PR earlier.

PR #2098, opened 2019-04-05
Tables Pipeline

sirtorry:
this is a tool not a sample / snippet. need to find home for this outside of python-docs-samples.


sirtorry:
branch has been forked and merged as #2373

PR #2106, opened 2019-04-11
Fix compatibility issues of google_jwt_client.py with python 3.

tomboc:
Original issue: https://groups.google.com/forum/#!topic/google-cloud-endpoints/ORvn8zJ_QQE
When google_jwt_client.py was executed with python 3, the Authorization
header was formatted as, "Authorization: Bearer b'JWT_TOKEN'". The b'
prefix made the token invalid.

This fix is compatible with both python 2 and python 3.


gguuss:
This looks like it was put into the file in another commit. If you resync your fork and it looks like the change is not there, feel free to reopen.

PR #2114, opened 2019-04-16
[Cloud Tasks] Move samples to new folder

averikitsch:
This is to keep samples consistent with other languages.


tasks/create_http_task_test.py
```diff
@@ -21,8 +21,8 @@
 TEST_QUEUE_NAME = os.getenv('TEST_QUEUE_NAME', 'my-appengine-queue')
 
 
-def test_create_task():
-    url = 'https://' + TEST_PROJECT_ID + '.appspot.com/example_task_handler'
```
andrewsg: Why this change in particular?

averikitsch: It can be any endpoint, so I didn't want someone to think that there was an actual app engine handler.

PR #2116, opened 2019-04-19
Order rows client-side when BQ Storage API is used.

tswast:
Parallelism in downloads means the order of rows might not be preserved. (Though it should be when there is only one stream.)

Thought: Do we check for ORDER BY and intentionally only use one stream to download? Does the BQ Storage API respect row order in that case?


tswast:
Filed https://github.com/googleapis/google-cloud-python/issues/7759 to fix this in the client instead.

PR #2119, opened 2019-04-24
Remove deprecated migration scripts

leahecole:
Per request of the composer team, remove these migration scripts


tswast:
Should we add a README file pointing to the in-place upgrade guide?

leahecole:
Yes, we totally should. I'll update the readme now.

tswast:
LGTM!

jbielick:
This script is still the only way to migrate from an old environment to a new one though, right? For example, our composer environment is not VPC-native, private IP, etc. I suppose that's an edge case.

PR #2133, opened 2019-05-01
Update Dataproc samples to use Google Cloud Client Libraries. Handle global region clusters.

aman-ebay:


dataproc/submit_job_to_cluster.py
```diff
@@ -42,222 +60,229 @@ def get_region_from_zone(zone):
...
+         bucket_name,
+         pyspark_file=None,
+         create_new_cluster=True,
+         global_region=True):
+
+    # [START dataproc_get_client]
+    if global_region:
+        region = 'global'
+        # Use the default gRPC global endpoints.
+        client_transport = ''
```
tswast: Usually we use None for unspecified values, but this could work if the client checks for falsiness rather than explicitly checking for None.

aman-ebay: I will push new file versions that eliminate the need for "" or None.

aman-ebay:
I need to work on the code. Will re-open when the code has been updated.

PR #2137, opened 2019-05-02
Add snippet for Collection Group Query

BenWhitehead:
* Bump `google-cloud-firestore` to 1.1.0 which includes
  the new functionality

### DIFF
--- a/firestore/cloud-client/snippets.py
+++ b/firestore/cloud-client/snippets.py
@@ -15,8 +15,9 @@ import datetime
 from time import sleep
 
 from google.cloud import firestore
-from google.cloud.firestore_v1beta1 import ArrayRemove, ArrayUnion
+from google.cloud.firestore_v1 import ArrayRemove, ArrayUnion
 import google.cloud.exceptions
+import pytest
 
 
 def quickstart_new_instance():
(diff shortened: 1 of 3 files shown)


BenWhitehead:
@crwilcox Can I get a review?

firestore/cloud-client/snippets.py
```diff
@@ -815,3 +816,63 @@ def delete_collection(coll_ref, batch_size):
     # [END delete_full_collection]
 
     delete_collection(db.collection(u'cities'), 10)
+
+
+@pytest.mark.skip(reason="Dependant on a composite index being created,"
```
kurtisvg: Why does this need to be here? Aren't the tests in snippet_tests.py?

This should snippet should also probably have tests on it.

firestore/cloud-client/snippets.py
```diff
@@ -815,3 +816,63 @@ def delete_collection(coll_ref, batch_size):
...
+    })
+    cities.document(u'DC').collection(u'landmarks').document().set({
+        u'name': u'National Museum of Nature and Science',
+        u'type': u'museum'
+    })
+    cities.document(u'BJ').collection(u'landmarks').document().set({
+        u'name': u'Jingshan Park',
+        u'type': u'park'
+    })
+    cities.document(u'BJ').collection(u'landmarks').document().set({
```
frankyn: Could you break apart the document ref? It's longer than I'd expect it to be.

```
bj_collection = cities.document(u'BJ').collection(u'landmarks')
bj_collection.document().set( {
  u'name': u'Beijing Ancient Observatory',
  u'type': u'museum'
}
```

BenWhitehead: done.

firestore/cloud-client/snippets.py
BenWhitehead: You're right, I moved it over to the tests file and added an assertion there.

firestore/cloud-client/snippets_test.py
```diff
@@ -255,3 +255,16 @@ def test_listen_for_changes(capsys):
 
 def test_delete_full_collection():
     snippets.delete_full_collection()
+
+
+@pytest.mark.skip(reason="Dependant on a composite index being created,"
```
kurtisvg: I'm still not sure why a skip is being used here. Why can't we integrate the creation of the index into the tests? 

More so, why doesn't the snippet show users how to do these things? Is that information found in a different snippet?

frankyn: +1 to @kurtisvg, if it hinders the usage of the sample then it should be included. 

If it's a timing issue given I/O starts tomorrow, then I'm okay with approving the sample as long as it's fixed afterwards.

BenWhitehead: Creation of the necessary composite index is something that today has to be done through the UI or the Firestore cli. After I/O I'm going to be running through each of the silver languages to make sure there are firestore admin clients and then work on using those to integrate the creation of the index into the setup of the test run.

kurtisvg: Would you please file a bug for this and then add it as `# TODO: b/####` to provide additional context for future readers?

BenWhitehead: Bug created, and TODO added.

BenWhitehead:
@frankyn @kurtisvg Feedback incorporated, can I get your approvals?

Issue #2141, opened 2019-05-06
There are no samples for Data Catalog clients in the repository


ricardosm-cit:
## Describe the issue

Data Catalog is a new product in GCP's Big Data family. There is an initial version for its cloud client library publicly available (https://googleapis.github.io/google-cloud-python/latest/datacatalog/gapic/v1beta1/api.html), but not samples for it in this repository.

As an early adopter of this product, I would like to contribute by creating the `datacatalog/cloud-client` folder in repo's root and adding samples there.

tswast:
Thanks for volunteering to add samples, Ricardo. I agree that `datacatalog/cloud-client` is the best location for the suggested samples.

PR #2148, opened 2019-05-09
Add samples for Data Catalog lookup_entry

ricardosm-cit:
Adding samples to Data Catalog lookup_entry, according to https://github.com/GoogleCloudPlatform/python-docs-samples/issues/2141.


datacatalog/cloud-client/requirements.txt
```diff
@@ -0,0 +1 @@
+google-cloud-datacatalog
```
tswast: We usually pin the version number in our sample code. (There's a bot that comes around to update these files to the latest version)

ricardosm-cit: Done.

datacatalog/cloud-client/lookup_entry_test.py
```diff
@@ -0,0 +1,66 @@
...
+import lookup_entry
+
+PROJECT = os.environ['GCLOUD_PROJECT']
+BIGQUERY_DATASET = os.environ['GCLOUD_BIGQUERY_DATASET']
+BIGQUERY_TABLE = os.environ['GCLOUD_BIGQUERY_TABLE']
+PUBSUB_TOPIC = os.environ['GCLOUD_PUBSUB_TOPIC']
+
+
+def test_lookup_bigquery_dataset():
+    @eventually_consistent.call
```
tswast: Since we aren't creating resources in these tests, what about it is eventually consistent / flakey that it needs this decorator?

datacatalog/cloud-client/lookup_entry_test.py
```diff
@@ -0,0 +1,66 @@
...
+# limitations under the License.
+
+import os
+
+from gcp_devrel.testing import eventually_consistent
+
+import lookup_entry
+
+PROJECT = os.environ['GCLOUD_PROJECT']
+BIGQUERY_DATASET = os.environ['GCLOUD_BIGQUERY_DATASET']
```
tswast: Are these environment variables set in our test config already? I guess they might be if these are the same that are used in BigQuery / Pub/Sub tests?

datacatalog/cloud-client/lookup_entry_test.py
ricardosm-cit: I thought they were included as a standard, to make sure test will be retried in case of flakiness. But I'll remove them.

datacatalog/cloud-client/lookup_entry_test.py
ricardosm-cit: Only GCLOUD_PROJECT is set in https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/testing/test-env.tmpl.sh. There is PUBSUB_TOPIC env var there, but for a different purpose. Data Catalog tests need existing BigQuery and PubSub assets to succeed. In this case, maybe it will be better to create/delete them using a fixture to run such tests. Agree?

tswast: Either create in a fixture or use a public dataset. I recommend a public dataset for BigQuery. I believe there's one (taxi data) for Pub/Sub, too.

ricardosm-cit: Done! Tks.

tswast:
Thanks! I've verified locally that these tests pass for Python 2.7 and Python 3.6.

PR #2158, opened 2019-05-15
Update Dataproc samples.

aman-ebay:
Use Google Cloud Client Libraries.
Handle global region clusters.
Add a single-job inline workflow.


dataproc/list_clusters.py
```diff
@@ -10,49 +10,53 @@
...
-""" Sample command-line program for listing Google Dataproc Clusters
-"""
+Example usage:
+python list_clusters.py --project_id=my-project-id --region=global
 
+"""
 import argparse
-
-import googleapiclient.discovery
+from google.cloud import dataproc_v1
```
tswast: Nit: We group imports for built-ins separately from imports for pip packages.

dataproc/single_job_workflow.py
```diff
@@ -0,0 +1,215 @@
...
+Example Usage to run the inline workflow on a global region managed cluster:
+python submit_job_to_cluster.py --project_id=$PROJECT --gcs_bucket=$BUCKET \
+  --cluster_name=$CLUSTER --zone=$ZONE --global_region
+
+"""
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+import argparse
```
tswast: newline between `__future__` imports and built-ins.

dataproc/single_job_workflow.py
```diff
@@ -0,0 +1,215 @@
...
+import os
+from google.cloud import dataproc_v1
```
tswast: newline between built-ins and pip packages.

dataproc/submit_job_to_cluster.py
```diff
@@ -42,222 +61,230 @@ def get_region_from_zone(zone):
...
-
+def main(project_id,
+         zone,
+         cluster_name,
+         bucket_name,
+         pyspark_file=None,
+         create_new_cluster=True,
+         global_region=True):
+
+    # [START dataproc_get_client]
```
tswast: Thoughts on refactoring this into a shared helper module? I imagine it could get confusing to have the same region tag in multiple files.

dataproc/single_job_workflow.py
```diff
@@ -0,0 +1,215 @@
...
+                            .WorkflowTemplateServiceGrpcTransport(
+                                address="{}-dataproc.googleapis.com:443"
+                                .format(region)))
+        dataproc_workflow_client = dataproc_v1.WorkflowTemplateServiceClient(
+            client_transport
+        )
+    # [END dataproc_get_client]
+
+    try:
+        if pyspark_file:
```
tswast: I'd prefer you move this logic into `get_pyspark_file` itself.

dataproc/single_job_workflow.py
```diff
@@ -0,0 +1,215 @@
...
+
+
+def get_default_pyspark_file():
+    """Gets the PySpark file from this directory."""
+    current_dir = os.path.dirname(os.path.abspath(__file__))
+    f = open(os.path.join(current_dir, DEFAULT_FILENAME), "rb")
+    return f, DEFAULT_FILENAME
+
+
+def get_pyspark_file(filename):
```
tswast: Let's refactor this a little bit so we don't need to pick which function (get_pyspark_file or get_default_pyspark_file) we have to call. Then, the first part of the body of this function becomes:

```
if not filename:
    current_dir = os.path.dirname(os.path.abspath(__file__))
    filename = os.path.join(current_dir, DEFAULT_FILENAME)
```

dataproc/single_job_workflow.py
```diff
@@ -0,0 +1,215 @@
...
+def callback(operation_future):
+    # Reset global when callback returns.
+    global waiting_callback
+    waiting_callback = False
+
+
+def wait_for_workflow_end():
+    """Wait for cluster creation."""
+    print("Waiting for workflow completion ...")
+    print('Workflow and job progress, and job driver output available from: '
```
tswast: Don't mix double-quotes and single quotes for strings in the same file. The only exception is when a string contains a double quote.

I actually recommend running `black` https://black.readthedocs.io/en/stable/ to fix issues like this (but then you'll probably have to clean up some of the line lengths since black doesn't respect the 79 character limit.)

dataproc/list_clusters.py
aman-ebay: Done.

dataproc/single_job_workflow.py
aman-ebay: Done.

dataproc/single_job_workflow.py
aman-ebay: Done. Also moved storage import to end.

dataproc/single_job_workflow.py
aman-ebay: Done.

dataproc/submit_job_to_cluster.py
aman-ebay: Can you clarify? I'm not sure what you are suggesting. AFAIK, the global region is unique to Dataproc and the current get_region function doesn't (can't) parse out the "global" region from a standard zone uri.

dataproc/list_clusters.py
tswast: Did you forget to push your branch? I don't see any commits from today.

dataproc/submit_job_to_cluster.py
tswast: I mean this set of 8 lines or so is repeated several times throughout this PR. We could create a `dataproc_helpers.py` file that contains this as a method and use it from the sample files. (May need to add an empty `__init__.py` to use `from . import dataproc_helpers`)

dataproc/single_job_workflow.py
aman-ebay: Done in single_job_workflow and submit_job_to_cluster

aman-ebay: Done or this file and submit_job_to_cluster.

dataproc/single_job_workflow.py
aman-ebay: Did this is a similar manner to your suggestion in response to previous comment.

dataproc/list_clusters.py
aman-ebay: I committed latest files on 5-17. Please let me know if you don't see the latest changes.

tswast:
FYI @bradmiro (Python sample updates for Dataproc)

dataproc/submit_job_to_cluster.py
tswast: Now that I look at this more closely, the two `dataproc_get_client`s do different things.

I've filed a sample tracker request at devrel/samples/1703 (Googlers only, sorry public readers) and made the single job version a different region tag.

aman-ebay: Looks good. Thanks.

PR #2161, opened 2019-05-16
adding infinite streaming sample to samples folder, unlike indefinite…

blechdom:
… streaming, this sample uses result_end_time to calculate unfinalized audio, and to resend it to maintain context across into the next streaming request

### DIFF
--- a/speech/microphone/transcribe_streaming_infinite.py
+++ b/speech/microphone/transcribe_streaming_infinite.py
@@ -46,10 +46,12 @@ RED = '\033[0;31m'
 GREEN = '\033[0;32m'
 YELLOW = '\033[0;33m'
 
+
 def get_current_time():
     """Return Current Time in MS."""
 
     return int(round(time.time() * 1000))
+
 
 class ResumableMicrophoneStream:
     """Opens a recording stream as a generator yielding the audio chunks."""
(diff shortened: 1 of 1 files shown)


blechdom:
Hi, 
Hoping I can get a reviewer to look at this PR!
Kristin

beccasaurus:
@nnegrey We'd like to share this tutorial with users. It's been shared directly with developers (via the code branch) but we should get this into master and share from there.

Problem: this is missing a test. The [Java](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/speech/cloud-client/src/main/java/com/example/speech/InfiniteStreamRecognize.java) and [Node.js](https://github.com/googleapis/nodejs-speech/blob/master/samples/infiniteStreaming.js) versions _are merged_ and missing tests.

This has recently been tested manually and is working OK for testers.

If the code generally LGTY, can we merge?

Testing a code sample which is a CLI application that listens to the microphone and keeps a stream open and displays transcription results from audio sent via the microphone is non-trivial.

I vote that we...
 - open issues on nodejs-speech, java-docs-samples, and python-docs-samples to add a test to this
 - merge this, as is
 - assign test development – as well as refactoring the code so that it supports being partially embedded on a website to support a tutorial build around this code

Thoughts?

/cc @ecrowdus @telpirion @munkhuushmgl @danaharon

nnegrey:
I don't think it's worth trying to set up testing for it. but happy to merge as is today.

munkhuushmgl:
I was trying to write unit test for it, and I was going to make code to
take more arguments from command line for testing purposes. I know it is a
bit complicated to write , Is that part of the reason why there was no unit
test in the first place?

nnegrey:
Mostly because ideally we test with a microphone (not sure how to start that) and then modifying the sample just to make it testable would further complicate the sample and make it harder for someone to distinguish and understand the relevant parts.

PR #2165, opened 2019-05-18
IAM: added snippet for remove member

DSdatsme:
Example code for removing a member in IAM policy.

### DIFF
--- a/iam/api-client/access.py
+++ b/iam/api-client/access.py
@@ -73,7 +73,8 @@ def modify_policy_add_role(policy, role, member):
 def modify_policy_remove_member(policy, role, member):
     """Removes a  member from a role binding."""
     binding = next(b for b in policy['bindings'] if b['role'] == role)
-    binding['members'].remove(member)
+    if 'members' in binding and member in binding['members']:
+        binding['members'].remove(member)
     print(binding)
     return policy
 # [END iam_modify_policy_remove_member]
--- a/iam/api-client/access.py
+++ b/iam/api-client/access.py
@@ -69,6 +69,16 @@ def modify_policy_add_role(policy, role, member):
 # [END iam_modify_policy_add_role]
 
 
+# [START iam_modify_policy_remove_member]
+def modify_policy_remove_member(policy, role, member):
+    """Removes a  member from a role binding."""
+    binding = next(b for b in policy['bindings'] if b['role'] == role)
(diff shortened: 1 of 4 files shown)


iam/api-client/access.py
```diff
@@ -69,6 +69,16 @@ def modify_policy_add_role(policy, role, member):
 # [END iam_modify_policy_add_role]
 
 
+# [START iam_modify_policy_remove_member]
+def modify_policy_remove_member(policy, role, member):
```
gguuss: Can you add a test for the new function?  Otherwise, looks good.

DSdatsme: Hey @gguuss ,
methods 'modify_policy_remove_member' (by me) and '[modify_policy_add_member](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/iam/api-client/access.py#L48)' (by somebody else) requires an argument 'member' which should be an existing email_id or a service account.
So for writing test cases, should I call [create_service_account()](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/iam/api-client/service_accounts.py#L37) and use it in my methods?
what say?

gguuss: Usually we use the `group:dpebot@google.com` account as a test IAM user, for example:

```
    # Test setting IAM permissions
    MEMBER = "group:dpebot@google.com"
    ROLE = "roles/viewer"
    print(manager.set_iam_permissions(
            service_account_json, project_id, cloud_region, registry_id,
            ROLE, MEMBER))
```

What you're doing in your test looks fine to me though, you're welcome to change to use the member we have used in other tests.

iam/api-client/access_test.py
```diff
@@ -13,17 +13,43 @@
...
+    assert 'etag' in out
+
+    policy = access.modify_policy_add_role(policy, gcp_role, member)
     out, _ = capsys.readouterr()
     assert 'etag' in out
 
-    policy = access.set_policy(project, policy)
+    policy = access.modify_policy_remove_member(policy, gcp_role, member)
     out, _ = capsys.readouterr()
     assert 'etag' in out
```
gguuss: I'm seeing this test failing in Python 2.

gguuss:
I went ahead and pushed a few fixes for the tests to your fork.

PR #2174, opened 2019-05-21
Update detect.py

agnelvishal:
1) I got argument parse error when bucket_name=bucket_name is given
2)blob_list[0] gave me folder name


agnelvishal:
I signed it!

PR #2175, opened 2019-05-21
Add Dataflow script to run templates

davidcavazos:
This adds a helper script to run a Dataflow template from the command line, from Python, or from a Cloud Function HTTP endpoint.

### DIFF
--- a/dataflow/run_template/README.md
+++ b/dataflow/run_template/README.md
@@ -1,0 +1,137 @@
+# Run template
+
+[`main.py`](main.py) - Script to run an [Apache Beam] template on [Google Cloud Dataflow].
+
+The following examples show how to run the [`Word_Count` template], but you can run any other template.
+
+For the `Word_Count` template, we require to pass an `output` Cloud Storage path prefix, and optionally we can pass an `inputFile` Cloud Storage file pattern for the inputs.
+If `inputFile` is not passed, it will take `gs://apache-beam-samples/shakespeare/kinglear.txt` as default.
+
+## Before you begin
+
+1. Install the [Cloud SDK].
+
+1. [Create a new project].
+
+1. [Enable billing].
+
+1. [Enable the APIs](https://console.cloud.google.com/flows/enableapi?apiid=dataflow,compute_component,logging,storage_component,storage_api,bigquery,pubsub,datastore.googleapis.com,cloudfunctions.googleapis.com,cloudresourcemanager.googleapis.com): Dataflow, Compute Engine, Stackdriver Logging, Cloud Storage, Cloud Storage JSON, BigQuery, Pub/Sub, Datastore, Cloud Functions, and Cloud Resource Manager.
+
+1. Setup the Cloud SDK to your GCP project.
+
+   ```bash
+   gcloud init
+   ```
+
+1. Create a Cloud Storage bucket.
+
+   ```bash
+   gsutil mb gs://your-gcs-bucket
+   ```
+
+## Setup
+
+The following instructions will help you prepare your development environment.
+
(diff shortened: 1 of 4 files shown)


dataflow/run_template/main.py
```diff
@@ -0,0 +1,121 @@
...
+    parser.add_argument('--project', help='Google Cloud project ID.')
+    parser.add_argument('--job_name', help='Name for the Dataflow job.')
+    args, argv = parser.parse_known_args()
+
+    # Parse the arguments as a dictionary of type {str: str}.
+    #   Example:  --optionA A1 A2 --optionB B1 --optionA A3
+    #   Produces: {'optionA': 'A1 A2 A3', 'optionB': 'B1'}
+    parameters = {}
+    option = None
+    for arg in argv:
```
andrewsg: This seems complex, is it possible to simplify it by making the command-line tool not support every possible parameter? At the least, argparse needs some custom help text to explain to the user that they can append arbitrary params, and a link to API documentation listing them. But we shouldn't hesitate to restrict the scope and capability of the samples, and link to that documentation in comments only.

andrewsg:
This needs tests to be complete. Check out the test files in similar samples for examples.

dataflow/run_template/README.md
```diff
@@ -0,0 +1,137 @@
...
+1. Navigate to the sample code directory.
+
+   ```bash
+   cd python-docs-samples/dataflow/run_template
+   ```
+
+1. Create a virtual environment and activate it.
+
+   ```bash
+   python3 -m virtualenv env
```
andrewsg: Our other samples use "virtualenv env" here -- perhaps we should use that by default, but if this is Python 3 only and you feel that that will cause difficulties for the average user, we could add a comment suggesting "virtualenv env -p python3" for users on Python 2-by-default platforms (which should become uncommon after the end of this year)

Refer to e.g. https://github.com/GoogleCloudPlatform/python-docs-samples/tree/master/bigtable/quickstart for an example. Also, note that the example I linked uses our README preprocessor; it's not a requirement but you can generate READMEs using README.rst.in if that is helpful for you and will cut down on boilerplate.

dataflow/run_template/main.py
```diff
@@ -0,0 +1,112 @@
...
+        project (str): Google Cloud project ID to run on.
+        job (str): Unique Dataflow job name.
+        template (str): Google Cloud Storage path to Dataflow template.
+        parameters (dict): Dictionary of parameters for the specified template.
+    Returns:
+        The response from the Dataflow service after running the template.
+    """
+    parameters = parameters or {}
+
+    # [START un_template]
```
andrewsg: looks like a typo here, should be dataflow_run_template

dataflow/run_template/requirements.txt
```diff
@@ -0,0 +1 @@
+google-api-python-client
```
andrewsg: Can you anchor this to a specific version? We do have a script that periodically upgrades these, but for testing verifiability it's important to specify the version.

dataflow/run_template/README.md
davidcavazos: I'm changing to `virtualenv env` for now. If at a certain point in the future it makes sense to use the generator I'll switch to it, but I think I'll keep it as is for the time being. Thanks!

dataflow/run_template/main.py
davidcavazos: Good catch, fixed

dataflow/run_template/requirements.txt
davidcavazos: Got it, changed!

PR #2179, opened 2019-05-24
Remove deprecated 'bql' parameter

leahecole:


leahecole:
I have also tested these DAGs in Composer and they function swimmingly with the new parameter

composer/workflows/connections.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2018 Google LLC
```
tswast: No need to update the copyright. It should reflect the year that the file was created.

leahecole: Reverted 👍

PR #2190, opened 2019-06-04
Changed env var to match between app.yaml and main.py in cloud_sql/mysql

fatcat2:
Previously, in the mysql example for cloud sql, there was a discrepancy between the naming of an environment variable in main.py and app.yaml. This PR changes the env var to "CLOUD_SQL_CONNECTION_NAME," matching the naming convention the [Java library uses](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/cloud-sql/mysql/servlet/src/main/java/com/example/cloudsql/ConnectionPoolContextListener.java).


fatcat2:
@kurtisvg

PR #2191, opened 2019-06-04
DLP: Several python example scripts have badly behaved info_types parameter

maximus12793:
Fixing bug where "FIRST_NAME", "LAST_NAME", "EMAIL_ADDRESS" are still present even when user 
specifies info_types. To fix I have converted action='append' to nargs='+' which will preserve defaults but allow user to override rather than append to the default list.


maximus12793:
@andrewsg from discussion earlier.

andrewsg:
Thanks!

PR #2194, opened 2019-06-05
Changed connection driver for ease of use.

pranaynanda:
The current code does not work if tried to implement as is. There are bugs as described in the first comment of this StackOverflow question. https://stackoverflow.com/a/53784015/5056741. To be precise, it raises 'raise InterfaceError("communication error", e)' error which appears to be related to a bug in `postgres+pg8000` driver


pranaynanda:
I signed it!

cloud-sql/postgres/sqlalchemy/main.py
```diff
@@ -36,13 +36,13 @@
...
     # Equivalent URL:
-    # postgres+pg8000://<db_user>:<db_pass>@/<db_name>?host=/cloudsql/<cloud_sql_instance_name>
+    # postgres+psycopg2://<db_user>:<db_pass>@/<db_name>?unix_socket=/cloudsql/<cloud_sql_instance_name>/.s.PGSQL.5432
     sqlalchemy.engine.url.URL(
         drivername='postgresql+psycopg2',
         username=db_user,
         password=db_pass,
         host=cloud_sql_connection_name,
-        database=db_name
+        database=db_name
```
pranaynanda: Eliminate trailing whitespace

cloud-sql/postgres/sqlalchemy/requirements.txt
```diff
@@ -1,3 +1,3 @@
 Flask==1.0.2
 SQLAlchemy==1.2.17
-pg8000==1.13.1
+psycopg2==2.8.2
```
gguuss: Update versions here.

```
Flask==1.1.1
SQLAlchemy==1.3.6
psycopg2==2.8.3
```

gguuss:
@kurtisvg Can you comment on whether we should be using pg8000 or psycopg2?

cloud-sql/postgres/sqlalchemy/main.py
```diff
@@ -36,16 +36,13 @@
 # managing a pool of connections to your database
 db = sqlalchemy.create_engine(
     # Equivalent URL:
-    # postgres+pg8000://<db_user>:<db_pass>@/<db_name>?unix_socket=/cloudsql/<cloud_sql_instance_name>/.s.PGSQL.5432
+    # postgres+psycopg2://<db_user>:<db_pass>@/<db_name>?unix_socket=/cloudsql/<cloud_sql_instance_name>/.s.PGSQL.5432
```
gguuss: Are you sure the parameter is `unix_socket`?  I believe I've seen it as `unix_sock` in some adapters.

pranaynanda:
> @kurtisvg Can you comment on whether we should be using pg8000 or psycopg2?

I tried using pg8000 as is initially. It returned 'raise InterfaceError("communication error", e)'. Changing it to psycopg2 worked. That's when I opened the pull request.

cloud-sql/postgres/sqlalchemy/main.py
gguuss: I'd prefer to [keep the pg8000 adapter and just show the correct parameter](https://github.com/GoogleCloudPlatform/python-docs-samples/pull/2067/files).

pranaynanda: It's cool if it works with the pg8000 adapter. I only changed it because the pg8000 driver wasn't working as per demo and was raising eroors as mentioned at the top. I tested it with psycopg2 instead and it worked. Because the PR is over a month old, I can't find the exact piece of working code that I tested with but I double checked again and this piece works:

```
db = sqlalchemy.create_engine(
    # Equivalent URL:
    'postgres+psycopg2://{}:{}@/{}?host=/cloudsql/{}'.format(db_user, db_pass, db_name, cloud_sql_instance_name)
    )
```

Also StackOverflow says that ['unix_sock' should be 'host'](https://stackoverflow.com/questions/54967660/connect-to-a-database-over-a-unix-socket-using-sqlalchemy).

gguuss: I'll investigate deeper.

pranaynanda: And found this in SQLAlchemy's documentation. 
image

Moreover it appears that pg8000 does not support connecting over Cloud SQL proxy which is a recommended best practice.

gguuss: I'm testing now, thank you btw for taking the time to make the PR :)

kurtisvg:
We use pg8000 specifically because it's a pure python driver and thus it works in many situations where psycopg2 doesn't. The issue you are experiencing is because pg8000 requires the user to specify the suffix `.s.PGSQL.5432`, where as psycopg2 automatically adds this suffix for you.

The suffix was added to the sample in #2235. It looks like the comment is still incorrect however.

cloud-sql/postgres/sqlalchemy/main.py
gguuss: I think you are right, just will confirm.

kurtisvg: The parameters get passed directly to the driver, and thus change depend on which driver you are using. `pg8000` requires `unix_sock`: http://pybrary.net/pg8000/dbapi.html

kurtisvg:
It looks like the parameters are backwards between [mysql](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/f7c2552d0972743446d26f05a0756d58bac923ec/cloud-sql/mysql/sqlalchemy/main.py#L39) and [postgres](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/cloud-sql/postgres/sqlalchemy/main.py#L39).

pranaynanda:
I've changed both main.py and requirements.txt to use pg8000 by default. I've also changed main.py to the correct query statements.

gguuss:
@pranaynanda Thanks again for the PR, I'm thinking it may make sense to leave as pg8000 to simplify getting the requirements installed.

Anecdotally, the sample as configured in Master is working for me and I'm unable to get psycopg2 installed on my dev machines as the sources for postgresql are missing.  What's weird though is that I'm unable to install the dependency pg8000==1.13.2, only pg8000==1.12.5, potentially because I'm testing with Python2. When I first ran the sample, I had the wrong DB name which caused the same exception you were seeing.

pranaynanda:
Sure. Makes sense.

PR #2198, opened 2019-06-06
Improve and fix Cloud Spanner samples that transfer marketing budget

hegemonic:
The samples that transfer part of an album's marketing budget had some issues:

+ `read_write_transaction`: Compared `second_album_budget` with an arbitrary integer, rather than explicitly checking against `transfer_amount`.
+ `write_with_dml_transaction`: Moved money from album 1 to album 2, even though `read_write_transaction` was the other way around. Also retrieved album 1's budget where it should have retrieved album 2's budget.

This change fixes those issues and updates the tests accordingly.

@jsimonweb PTAL

### DIFF
--- a/spanner/cloud-client/snippets.py
+++ b/spanner/cloud-client/snippets.py
@@ -414,7 +414,7 @@ def read_write_transaction(instance_id, database_id):
 
         transfer_amount = 200000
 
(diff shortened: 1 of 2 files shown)


hegemonic:
Friendly ping for @jsimonweb and the repo owners.

PR #2199, opened 2019-06-07
ImageMagick: use separate bucket for blurred images.

ace-n:
@tbpg FYI

Related PR: https://github.com/GoogleCloudPlatform/nodejs-docs-samples/pull/1346
Googlers: See also cl/251939574


functions/imagemagick/main.py
```diff
@@ -73,12 +73,14 @@ def __blur_image(current_blob):
 
     print(f'Image {file_name} was blurred.')
 
-    # Send Blurred image back to the bucket (with a 'blurred-' prefix).
-    # The prefix is necessary to avoid re-invoking the function upon upload.
-    new_file_name = f'blurred-{file_name}'
-    new_blob = current_blob.bucket.blob(new_file_name)
+    # // Upload result to a different bucket, to avoid re-triggering this function.
+    # You can also re-upload it to the same bucket + tell your Cloud Function to
```
engelke: It might be clearer to say "You could instead" instead of "You can also". The current wording looks a bit like saying you should do two things, instead of one or other.

PR #2201, opened 2019-06-07
Cloud Bigtable writes samples

billyjacobson:


bigtable/snippets/writes/write_batch.py
```diff
@@ -0,0 +1,52 @@
...
+    rows[1].set_cell(column_family_id,
+                     "connected_wifi",
+                     1,
+                     timestamp)
+    rows[1].set_cell(column_family_id,
+                     "os_build",
+                     "12145.0.0-rc6",
+                     timestamp)
+
+    table.mutate_rows(rows)
```
tswast: Does `mutate_rows` return anything? If so, I'd like to see a printed value from the response.

If not, will it throw if there's a problem? Add a comment with what errors you might expect.

bigtable/snippets/writes/write_conditionally.py
```diff
@@ -0,0 +1,45 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# [START bigtable_writes_conditional]
+import datetime
+
+from google.cloud import bigtable
+from google.cloud.bigtable.row_filters import ColumnQualifierRegexFilter, \
```
tswast: Nit: Prefer importing modules not classes. i.e. `from google.cloud.bigtable import row_filters` and use `row_filters.ColumnQualifierRegexFilter` in the sample.

bigtable/snippets/writes/write_conditionally.py
```diff
@@ -0,0 +1,45 @@
...
+    row_filter = RowFilterChain(
+        filters=[FamilyNameRegexFilter(column_family_id),
+                 ColumnQualifierRegexFilter('os_build'),
+                 ValueRegexFilter("PQ2A\\..*")])
+    row = table.row(row_key, filter_=row_filter)
+    row.set_cell(column_family_id,
+                 "os_name",
+                 "android",
+                 timestamp)
+    row.commit()
```
tswast: Just as with `mutate_rows`, does `commit` return anything? What will happen if there's an error?

bigtable/snippets/writes/write_increment.py
```diff
@@ -0,0 +1,32 @@
...
+def write_increment(project_id, instance_id, table_id):
+    client = bigtable.Client(project=project_id, admin=True)
+    instance = client.instance(instance_id)
+    table = instance.table(table_id)
+
+    column_family_id = "stats_summary"
+
+    row_key = "phone#4c410523#20190501"
+    row = table.row(row_key, append=True)
+    row.increment_cell_value(column_family_id, "connected_wifi", -1)
```
tswast: What is `-1` in this line? Add a comment explaining it (or use keyword arguments, if it's optional).

bigtable/snippets/writes/write_batch.py
```diff
@@ -0,0 +1,52 @@
...
+
+from google.cloud import bigtable
+
+
+def write_batch(project_id, instance_id, table_id):
+    client = bigtable.Client(project=project_id, admin=True)
+    instance = client.instance(instance_id)
+    table = instance.table(table_id)
+
+    timestamp = datetime.datetime.utcnow()
```
tswast: Question: Any gotchas to know about timestamps? What happens if I use a timestamp in the future or past?

Hopefully this is covered in prose around the samples if it's too much for a comment.

bigtable/snippets/writes/write_batch.py
tswast: Can there be partial failures where only some row writes succeed?

bigtable/snippets/writes/writes_test.py
```diff
@@ -0,0 +1,62 @@
...
+
+import os
+import uuid
+
+from google.cloud import bigtable
+from write_batch import write_batch
+from write_conditionally import write_conditional
+from write_increment import write_increment
+from write_simple import write_simple
+
```
tswast: Nit: Two lines after imports and before code.

bigtable/snippets/writes/writes_test.py
```diff
@@ -0,0 +1,62 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+import uuid
+
+from google.cloud import bigtable
+from write_batch import write_batch
```
tswast: Nit: Separate local module imports from package imports.

```
built-ins

packages

local modules
```

bigtable/snippets/writes/writes_test.py
```diff
@@ -0,0 +1,62 @@
...
+    instance = client.instance(BIGTABLE_INSTANCE)
+
+    table_id = TABLE_ID_PREFIX.format(str(uuid.uuid4())[:16])
+    table = instance.table(table_id)
+    if table.exists():
+        table.delete()
+
+    column_family_id = 'stats_summary'
+    column_families = {column_family_id: None}
+    table.create(column_families=column_families)
```
tswast: Use a pytest fixture with `yield` so that the table is cleaned up even if the tests fail.

https://docs.pytest.org/en/latest/fixture.html#fixture-finalization-executing-teardown-code

bigtable/snippets/writes/write_batch.py
billyjacobson: any timestamps are okay to use

bigtable/snippets/writes/writes_test.py
billyjacobson: Not sure how to do this with what I have. I put `@pytest.fixture(scope="module")` before def tests_writes, but that seems to break things and then I'm not sure what else to do.

bigtable/snippets/writes/write_conditionally.py
billyjacobson: seems buggy

bigtable/snippets/writes/writes_test.py
tswast: That's not how you use fixtures...

An example fixture:

https://github.com/googleapis/google-cloud-python/blob/0e231cab5f066ae5b3e254d0cb7d204e3dac7ebc/bigquery/samples/tests/conftest.py#L58-L68

Note: it's in conftest.py, but could just as easily be in the same file as the tests.

To use, simply add an argument to your test function with the same name as the fixture function.

https://github.com/googleapis/google-cloud-python/blob/0e231cab5f066ae5b3e254d0cb7d204e3dac7ebc/bigquery/samples/tests/test_list_tables.py#L18

billyjacobson: I'm don't really understand this still. These examples all look different. Just committed what I understood

tswast: OK. I just pushed 3233a426a27180d88621b0750914c41c47ba8208 with my requested changes.

The gist of "fixtures" are that they are the equivalent of a `setUp` function in JUnit. If you `yield` from a fixture, the remaining piece of the function is equivalent to `tearDown` in JUnit.

billyjacobson:
Thanks @tswast I understand the fixtures now. It's like you declare the fixture which is a function to return a variable and then inject it by name into your test function so you'll get the yielded variable. Then when the test is over, it can do everything after the yield.

PR #2212, opened 2019-06-11
fix: Updates region to include tasks: datastore_batch_upsert

grant:
We need to include `task1` and `task2` in this sample for it to make sense.
b/111318350


grant:
How do I force Kokoro to run?
This change shouldn't break the tests.

ace-n:
cc @engelke

grant:
Trying to push a file to get the tests to run

vision/cloud-client/document_text/env/.Python
```diff
@@ -0,0 +1 @@
+/System/Library/Frameworks/Python.framework/Versions/2.7/Python
```
engelke: Did you mean to include these env files?

grant: That file is outdated and creds are revoked.

PR #2220, opened 2019-06-13
Functions: improve existing sample

anguillanneuf:
This addresses user feedback that compares our current Hello World [example](https://cloud.google.com/functions/docs/calling/pubsub#functions_calling_pubsub-python) for Cloud Function with Cloud Pub/Sub with that of [Firebase with Cloud Pub/Sub](https://firebase.google.com/docs/functions/pubsub-events). 

~1. `else` clause actually outputs something _else_.~
2. Show how users can access message attributes, publish time, and message IDs.

### DIFF
--- a/functions/helloworld/main.py
+++ b/functions/helloworld/main.py
@@ -83,9 +83,9 @@ def hello_http(request):
 def hello_pubsub(event, context):
     """Background Cloud Function to be triggered by Pub/Sub.
     Args:
-         event (dict):  The dictionary with data specific to this type of event.
-         The `data` field contains the PubsubMessage message. The `attributes`
-         field will contain custom attributes if there are any.
+         event (dict):  The dictionary with data specific to this type of
+         event. The `data` field contains the PubsubMessage message. The
+         `attributes` field will contain custom attributes if there are any.
          context (google.cloud.functions.Context): The Cloud Functions event
          metadata. The `event_id` field contains the Pub/Sub message ID. The
          `timestamp` field contains the publish time.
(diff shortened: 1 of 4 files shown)


functions/helloworld/main.py
```diff
@@ -80,19 +80,22 @@ def hello_http(request):
...
     """
     import base64
 
-    if 'data' in data:
-        name = base64.b64decode(data['data']).decode('utf-8')
+    if 'data' in event:
+        name = base64.b64decode(event['data']).decode('utf-8')
     else:
-        name = 'World'
+        name = 'World! Something went wrong'
```
andrewsg: Is "something went wrong" always true in the case where there's no data, or is it sometimes just a consequence that the user didn't add any? If something really is wrong, can we be more specific for the user's benefit?

anguillanneuf: The goal there is to print something other than "World!" Otherwise the output will always be "Hello World!"

tswast: If something really did go wrong, maybe raise an exception?

ace-n: If that's the case - maybe we should remove this `if-else` logic from this sample? (Though the current logic does allow non-Pub/Sub triggers to invoke the function without issue.)

IMO, we shouldn't be documenting how to handle (relatively rare) GCP-side system failures (on this page at least).

anguillanneuf:
I have decided to just update the sample to show how to access the message ID and publish time. 

"Hello world!" just indicates something is failing. If users see "Hello {something else}!", they can be sure that everything has worked.

andrewsg:
Thanks!

PR #2231, opened 2019-06-19
Pub/Sub: fix all publisher examples

anguillanneuf:
- add `future.result()` to resolve futures to ensure publish success
- update publish with error handler example some more

### DIFF
--- a/pubsub/cloud-client/publisher_test.py
+++ b/pubsub/cloud-client/publisher_test.py
@@ -111,11 +111,7 @@ def test_publish_with_batch_settings(topic, capsys):
 
 
 def test_publish_with_error_handler(topic, capsys):
-
-    with _make_sleep_patch():
-        with pytest.raises(RuntimeError, match='sigil'):
-            publisher.publish_messages_with_error_handler(
-                PROJECT, TOPIC)
+    publisher.publish_messages_with_error_handler(PROJECT, TOPIC)
 
     out, _ = capsys.readouterr()
     assert 'Published' in out
--- a/pubsub/cloud-client/subscriber_test.py
+++ b/pubsub/cloud-client/subscriber_test.py
@@ -169,13 +169,15 @@ def test_update(subscriber_client, subscription, capsys):
 def _publish_messages(publisher_client, topic):
     for n in range(5):
         data = u'Message {}'.format(n).encode('utf-8')
-        publisher_client.publish(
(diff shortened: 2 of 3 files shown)


pubsub/cloud-client/subscriber_test.py
```diff
@@ -169,13 +169,15 @@ def test_update(subscriber_client, subscription, capsys):
...
+        future = publisher_client.publish(
             topic, data=data)
+        future.result()
 
 
 def _publish_messages_with_custom_attributes(publisher_client, topic):
     data = u'Test message'.encode('utf-8')
-    publisher_client.publish(topic, data=data, origin='python-sample')
+    future = publisher_client.publish(topic, data=data, origin='python-sample')
+    future.result()
```
hongalex: Same as above: do we want to print this?

pubsub/cloud-client/subscriber_test.py
```diff
@@ -169,13 +169,15 @@ def test_update(subscriber_client, subscription, capsys):
 def _publish_messages(publisher_client, topic):
     for n in range(5):
         data = u'Message {}'.format(n).encode('utf-8')
-        publisher_client.publish(
+        future = publisher_client.publish(
             topic, data=data)
+        future.result()
```
hongalex: Do we care about printing the result of this?

pubsub/cloud-client/publisher_test.py
```diff
@@ -111,11 +111,7 @@ def test_publish_with_batch_settings(topic, capsys):
 
 
 def test_publish_with_error_handler(topic, capsys):
-
-    with _make_sleep_patch():
```
hongalex: Why is this being removed?

pubsub/cloud-client/subscriber_test.py
anguillanneuf: No we don't.

pubsub/cloud-client/subscriber_test.py
anguillanneuf: Same as above.

pubsub/cloud-client/publisher_test.py
anguillanneuf: We no longer wait indefinitely. We just wait for the dictionary `futures` to run dry now.

PR #2234, opened 2019-06-21
Use print() function in both Python 2 and Python 3

cclauss:
Legacy __print__ statements are syntax errors in Python 3 but __print()__ function works as expected in both Python 2 and Python 3.

### DIFF
--- a/appengine/standard/i18n/i18n_utils.py
+++ b/appengine/standard/i18n/i18n_utils.py
@@ -19,6 +19,7 @@
 The idea of this example, especially for how to translate strings in
 Javascript is originally from an implementation of Django i18n.
 """
+from __future__ import print_function
 
 
 import gettext
@@ -49,7 +50,7 @@ def _get_plural_forms(js_translations):
         for l in js_translations._catalog[''].split('\n'):
             if l.startswith('Plural-Forms:'):
                 plural = l.split(':', 1)[1].strip()
-                print "plural is %s" % plural
+                print("plural is %s" % plural)
     if plural is not None:
         for raw_element in plural.split(';'):
             element = raw_element.strip()
@@ -57,7 +58,7 @@ def _get_plural_forms(js_translations):
                 n_plural = int(element.split('=', 1)[1])
             elif element.startswith('plural='):
                 plural = element.split('=', 1)[1]
-                print "plural is now %s" % plural
+                print("plural is now %s" % plural)
     else:
         n_plural = 2
         plural = '(n == 1) ? 0 : 1'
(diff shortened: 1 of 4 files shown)


appengine/standard/i18n/i18n_utils.py
```diff
@@ -19,7 +19,7 @@
 The idea of this example, especially for how to translate strings in
 Javascript is originally from an implementation of Django i18n.
 """
-
+from __future__ import print_function
```
gguuss: I'm not sure you need this import.

appengine/standard/i18n/i18n_utils.py
```diff
@@ -49,15 +54,15 @@ def _get_plural_forms(js_translations):
         for l in js_translations._catalog[''].split('\n'):
             if l.startswith('Plural-Forms:'):
                 plural = l.split(':', 1)[1].strip()
-                print "plural is %s" % plural
+                print("plural is %s" % plural)
```
gguuss: Prefer string formatter and single quotes:

```
print(`plural is {}`.format(plural))
```

appengine/standard/mail/header.py
```diff
@@ -1,3 +1,4 @@
+from __future__ import print_function
```
gguuss: Not sure this is required

appengine/standard/i18n/i18n_utils.py
cclauss: As discussed in "[__prevent compatibility regressions__](https://docs.python.org/3/howto/pyporting.html#prevent-compatibility-regressions)" this directive _changes the way that Python 2 operates_ to make legacy print statements a syntax error.  This prevents Python 2 developers from adding code that breaks Python 3 compatibility.
```python
python2 -c "from __future__ import print_function ; print 'Hi'"  # SyntaxError
```

appengine/standard/i18n/i18n_utils.py
cclauss: [__python/black__](https://github.com/python/black) (think gofmt for Python) is reformatting code to prefer " over ' but I will make the requested change.

appengine/standard/i18n/i18n_utils.py
gguuss: This is a good idea, it may make sense to standardize this as more people are shifting to Python 3.

appengine/standard/i18n/i18n_utils.py
gguuss: We'll consider this alternative style; I don't personally have a strong opinion either way. Thank you for the pointers, also thanks a ton for the fixes!

gguuss: Regarding python/black and quotes, according to [the style guide](https://google.github.io/styleguide/pyguide.html) it looks like some teams are using [yapf](https://github.com/google/yapf/).

gguuss:
@cclauss Friendly ping, happy to accept your enhancements once you've applied the style we prefer.

PR #2244, opened 2019-06-27
Set redis version to ">=3.1.0"

darrencarlton:
The current version of the Redis Python library is 3.2.1. Instead of hard-coding "3.2.1" and having to update the code example again, I suggest we set this to "greater than or equal to 3.1.0".

I tested the code example with this setting, and it worked just fine.


darrencarlton:
Thanks, Gus

PR #2248, opened 2019-06-27
Make backup/restore file name a parameter

engelke:
Hard coding could confuse readers.

### DIFF
--- a/monitoring/api/v3/alerts-client/snippets.py
+++ b/monitoring/api/v3/alerts-client/snippets.py
@@ -108,15 +108,17 @@ def delete_notification_channels(project_name, channel_ids, force=None):
 
 
 # [START monitoring_alert_backup_policies]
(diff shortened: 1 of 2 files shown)


monitoring/api/v3/alerts-client/snippets.py
```diff
@@ -136,9 +138,11 @@ def default(self, obj):
 # [START monitoring_alert_create_channel]
```
gguuss: /FYI there probably should not be multiple codeblocks here with the same tag, if you remove them, check that there aren't extra end tags around either.

engelke: The multiple tags have been there for a while, and the matching docs seem to be picking them up okay, so I don't want to change them now.

RAbraham:
thanks guys!

PR #2232, opened 2019-06-20
Container Analysis samples

DanSanche:
added Python samples for the [Container Analysis](https://cloud.google.com/container-registry/docs/container-analysis) GA

PR #2258, opened 2019-07-08
Container Analysis samples

DanSanche:
added Python samples for the Container Analysis GA

Re-creating https://github.com/GoogleCloudPlatform/python-docs-samples/pull/2232 as a branch, rather than an external fork


container_registry/container_analysis/README.md
```diff
@@ -0,0 +1,62 @@
...
+1. **Set Environment Variables**
+
+    ```
+    export GCLOUD_PROJECT="YOUR_PROJECT_ID"
+    ```
+
+1. **Run Tests**
+
+    ```
+    python -m pytest ./samples_test.py
```
leahecole: The last nitpicky change I have is to instead encourage users to install nox and run tests on the container_analysis samples with it so that it more closely matches the AUTHORING_GUIDE recommendation. The Python 3.6 nox command for container analysis is `nox -s "py36(sample='./container_registry/container_analysis')"` and I can tell you that it runs successfully :)

PR #2262, opened 2019-07-10
codelabs/functions: create 'Python Powered' codelab

di:
cc @theacodes


codelabs/functions/python_powered/app.py
```diff
@@ -0,0 +1,9 @@
+from main import python_powered
```
theacodes: All of these files need the standard license header.

codelabs/functions/python_powered/main.py
```diff
@@ -0,0 +1,45 @@
+from flask import escape
```
theacodes: Our general code style is to import modules, not members. So `import flask` and use `flask.escape` and `flask.send_file`.

codelabs/functions/python_powered/test_main.py
```diff
@@ -0,0 +1,25 @@
+from unittest import TestCase
+from unittest.mock import Mock
+
+from main import hello_name, hello_world
+
+
+class TestHello(TestCase):
```
theacodes: Hmm. I think you're using this in your codelab as example code, yeah?

If so, this is fine, otherwise, we generally use `pytest` style.

di: > Hmm. I think you're using this in your codelab as example code, yeah?

Yeah. I would have used `pytest` style, but the `conftest.py` in this repo seems incompatible with Python 3?

```
$ python -m pytest
ImportError while loading conftest '.../googlecloudplatform/python-docs-samples/conftest.py'.
../../../conftest.py:17: in <module>
    import mock
E   ModuleNotFoundError: No module named 'mock'
```

(It also saves a step for the codelab user here.)

PR #2270, opened 2019-07-12
Update kubernetespodoperator

leahecole:
Region tag updates and sample changes related to the restructure of the KubernetesPodOperator tutorial. New version visible to internal Googlers in staging.


leahecole:
These failures are similar to what I see in #2131, and I'm trying to figure out if I've been doing something wrong, and how to correct it.

PR #2274, opened 2019-07-18
Revert "Update detect.py"

nnegrey:
Reverts GoogleCloudPlatform/python-docs-samples#2174

The change breaks our tests. I'm guessing the original issue was that the bucket path was passed incorrectly.


nnegrey:
Also fix other broken tests

PR #2275, opened 2019-07-22
Fix "delete_full_collection" snippet

LouisTrezzini:
Query limit is incorrectly set to 10 instead of `batch_size`. 

[Java](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/b9f7749ba2755c9bf3e8ea1ec1fd581a67b6d1b0/firestore/src/main/java/com/example/firestore/snippets/ManageDataSnippets.java#L309-L330) or [Go](https://github.com/GoogleCloudPlatform/golang-samples/blob/862f0c0569fb13c2f2fee5ec20579ce3887736ba/firestore/firestore_snippets/save.go#L261-L298) implementations use the correct batchSize value.

### DIFF
--- a/firestore/cloud-client/snippets.py
+++ b/firestore/cloud-client/snippets.py
@@ -802,7 +802,7 @@ def delete_full_collection():
 
     # [START delete_full_collection]
     def delete_collection(coll_ref, batch_size):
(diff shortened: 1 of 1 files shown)


LouisTrezzini:
Duplicate of #2216

PR #2276, opened 2019-07-23
[BLOCKED] AutoML Tables: Docs samples updated to use new (pending) client

lwander:
@sirtorry PTAL

### DIFF
--- a/tables/automl/automl_tables_dataset.py
+++ b/tables/automl/automl_tables_dataset.py
@@ -1,0 +1,601 @@
+#!/usr/bin/env python
+
+# Copyright 2019 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


sirtorry:
todo: tests

lwander:
@sirtorry do we want tests here? Or was your todo for the tests that were missing from the original new client PR.

lwander:
friendly ping on the tests question

sirtorry:
@lwander these will need tests as well. 
@merla18 @beccasaurus your thoughts on whether we should update these snippets now or wait to update it with the rest.

lwander:
Can you point me at where these tests are written/configured? I couldn't find any for the original doc snippets.

sirtorry:
Will wait for new version before merging.

lwander:
@sirtorry which new version? These samples reflect the currently released version, and we have users who need these docs.

PR #2296, opened 2019-07-24
Update list blobs to use new client.list_blobs() method.

frankyn:


gguuss:
PTAL at the fix in this PR - https://github.com/GoogleCloudPlatform/python-docs-samples/pull/2281

PR #2297, opened 2019-07-24
[IAM] Add client initialization to each access.py sample  

askmeegs:
Initializes a `cloudresourcemanager` client in each snippet.

Reason: received a bug report where a user tried to call these functions on an IAM API client, not a ResourceManager client. This sample code lives in the IAM docs, so not being clear about which API is used here could be confusing.

### DIFF
--- a/iam/api-client/access.py
+++ b/iam/api-client/access.py
@@ -29,6 +29,7 @@ import googleapiclient.discovery
 # [START iam_get_policy]
 def get_policy(project_id):
     """Gets IAM policy for a project."""
+
     # pylint: disable=no-member
     credentials = service_account.Credentials.from_service_account_file(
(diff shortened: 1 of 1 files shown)


tswast:
Thanks! I agree this will be clearer for users who find these snippets in non-ResourceManager contexts.

tswast:
Looks like there's maybe a problem with requirements.txt? https://travis-ci.org/GoogleCloudPlatform/python-docs-samples/builds/563281779?utm_medium=notification&utm_source=email is complaining about missing `oauth2client`.

tswast:
FYI: in the future, only a subset of the tests run on travis when a PR is from a fork. Best to create a new branch on python-docs-samples, instead. (I usually prefix the branch name with either an issue number or my username)

PR #2298, opened 2019-07-24
Update string reported in snippet; update check in remove labels test.

gguuss:


gguuss:
Hush DPE bot, I signed it.

PR #2299, opened 2019-07-25
Update logging for cloud run sample

averikitsch:


run/README.md
```diff
@@ -60,22 +60,29 @@ For more Cloud Run samples beyond Python, see the main list in the [Cloud Run Sa
...
 gcloud builds submit --tag gcr.io/${GOOGLE_CLOUD_PROJECT}/${SAMPLE}
+
+# Deploy to Cloud Run
 gcloud beta run deploy $SAMPLE \
-  # Needed for Manual Logging sample.
-  --set-env-var GOOGLE_CLOUD_PROJECT=${GOOGLE_CLOUD_PROJECT}
+  --set-env-vars GOOGLE_CLOUD_PROJECT=${GOOGLE_CLOUD_PROJECT} \
   --image gcr.io/${GOOGLE_CLOUD_PROJECT}/${SAMPLE}
+
+# Note: GOOGLE_CLOUD_PROJECT env var is needed for Manual Logging sample.
```
grayside: Given the Manual Logging sample does not exist in Python, we can remove this note.

run/pubsub/main.py
```diff
@@ -25,14 +27,16 @@
 @app.route('/', methods=['POST'])
 def index():
     envelope = request.get_json()
+    logging.basicConfig(stream=sys.stdout, level=logging.INFO)
+
     if not envelope:
         msg = 'no Pub/Sub message received'
-        print(f'error: {msg}')
+        logging.error(f'error: {msg}')
```
grayside: To make sure I understand, why does Python need the logging library instead of simply relying on print to stdout?

run/pubsub/main.py
```diff
@@ -41,8 +45,10 @@ def index():
     if isinstance(pubsub_message, dict) and 'data' in pubsub_message:
         name = base64.b64decode(pubsub_message['data']).decode('utf-8').strip()
 
-    print(f'Hello {name}!')
-    return ('', 204)
+    msg = f'Hello {name}!'
+    logging.info(msg)
+
+    return msg
```
grayside: This change diverges the sample from the Node.js and Go samples, which return empty/204 responses. Returning the msg as part of the HTTP response is not used by Pub/Sub. I think this change is fine but we should have a comment indicating it is for testing.

run/pubsub/main.py
averikitsch: I thought this was the standard we would want to set. It doesn't need it though.

averikitsch: Do you prefer just flushing the stdout?

grayside: The goal on this one is to be simple, idiomatic, and that it behaves in Stackdriver as expected. I don't have any requirements on the specifics of the first two.

averikitsch: sg. Reverted to original sample and added a flush for logs.

run/README.md
```diff
@@ -36,7 +36,7 @@ For more Cloud Run samples beyond Python, see the main list in the [Cloud Run Sa
 2. [Build the sample container](https://cloud.google.com/run/docs/building/containers#building_locally_and_pushing_using_docker):
 
     ```
-    export SAMPLE=$sample
+    export SAMPLE=<SAMPLE_DIRECTORY>
```
grayside:
```suggestion
    export SAMPLE=<SAMPLE_NAME>
```

run/README.md
```diff
@@ -36,7 +36,7 @@ For more Cloud Run samples beyond Python, see the main list in the [Cloud Run Sa
...
     cd $SAMPLE
     docker build --tag $sample .
```
grayside:
```suggestion
    docker build --tag $SAMPLE .
```

run/README.md
```diff
@@ -60,21 +60,26 @@ For more Cloud Run samples beyond Python, see the main list in the [Cloud Run Sa
...
     ```
-    export SA_KEY_NAME=my-key-name-123
+    export SA_KEY_PATH=/path/to/service/account/my-key-name-123.json
+
     PORT=8080 && docker run --rm \
         -p 8080:${PORT} -e PORT=${PORT} \
-        -e GOOGLE_APPLICATION_CREDENTIALS=/tmp/keys/${SA_KEY_NAME}.json \
-        -v $GOOGLE_APPLICATION_CREDENTIALS:/tmp/keys/${SA_KEY_NAME}.json:ro \
+        -e GOOGLE_APPLICATION_CREDENTIALS=$SA_KEY_PATH \
+        -v $GOOGLE_APPLICATION_CREDENTIALS:$SA_KEY_PATH:ro \
```
grayside: Possibly this code block needs more comments in general. Let me break down what I had happening originally:

1. SA_KEY_NAME is set to the name of the service account key file as we want it to exist inside the container
2. As a convention, we've decided the /tmp/keys/ directory is where the key will be found inside the container
3. The GOOGLE_APPLICATION_CREDENTIALS env var inside the container is set to look for that file
4. The volume mount is using the $GOOGLE_APPLICATION_CREDENTIALS variable on the host machine.

The result it the local machine's $GOOGLE_APPLICATION_CREDENTIALS is being made available in a central, generic location inside the container.

When we adapt that to the refactor here, the $SA_KEY_PATH variable is a bit odd without a comment, because the path used there is arbitrarily selected as a location inside the container. It's something we can dictate and does not need to relate to the local host.

Hopefully that wasn't too confusing. Based on that explanation, we either:

1. Revert to the original and add sufficient text to clarify.
2. Keep the new version and add sufficient text to clarify.

averikitsch:
@grayside how does this look?

PR #2310, opened 2019-07-31
Fix managed instances demo web server to correctly start/stop CPU loading

djmailhot:
The previous design did not correctly close running Processes meant for CPU loading.  Once CPU loading was started, it would not be able to be stopped.

This design only creates one Process and never closes it.  Instead, it simply sleeps when CPU loading is turned off.


djmailhot:
Now wondering who I should ask for a merge

andrewsg:
Oh, I can merge. Just a sec.

PR #2311, opened 2019-08-01
BigQuery: Use user-based authentication in quickstart.

tswast:


bigquery/cloud-client/simple_app.py
```diff
@@ -15,15 +15,42 @@
...
 # [START bigquery_simple_app_deps]
+import google_auth_oauthlib
 from google.cloud import bigquery
 # [END bigquery_simple_app_deps]
 
 
-def query_stackoverflow():
+def authenticate(client_id, client_secret):
+    # [START bigquery_simple_app_authenticate]
+    # TODO(developer): Set the client ID and client secret to the values
```
shollyman: This TODO requires a lot of background understanding.  Should we link to more detailed documentation about setting up/managing client ID and secret?  This could also be the case where we supply a default client ID for the samples, with commentary about replacing them for your own application to avoid blocking someone on this extra level of setup.

leahecole:
Hello! I wanted to see if this PR is still being thought about/worked on? If so, cool, great, carry on, if not, please close!

tswast:
Definitely still on my mind! I'm hung up a bit because I don't want to force the user to create a client ID and client secret if I don't have to. (For example, pydata-google-auth and pandas-gbq embed a public key.) With "henhouse" this isn't the worst thing, but it's still an extra step.

tswast:
Since this is a pretty big change to the Python BQ quickstart and a bit of an experiment, I'm closing this out. I won't have time to rewrite the quickstart (or follow-up on the inevitable feedback it'll generate) before EOY.

PR #2316, opened 2019-08-06
Add Migration Guide Snippets for Cloud Tasks

averikitsch:


appengine/flexible/tasks/snippets_test.py
```diff
@@ -0,0 +1,107 @@
...
+    result = snippets.delete_queue(
+        TEST_PROJECT_ID, TEST_LOCATION, QUEUE_NAME_1)
+    assert result is None
+
+    result = snippets.delete_queue(
+        TEST_PROJECT_ID, TEST_LOCATION, QUEUE_NAME_2)
+    assert result is None
+
+
+@pytest.mark.order7
```
grant: Nit: It would make sense if `@pytest.mark.order7` came before `@pytest.mark.order8` if possible.

appengine/flexible/tasks/snippets_test.py
```diff
@@ -0,0 +1,107 @@
...
+def test_retry_task():
+    QUEUE_NAME = []
+    for i in range(3):
```
grant: `3` should be a constant.

Like `QUEUE_SIZE = 3`.

Otherwise, we may break the tests if we change L95 without changing L105 vice versa.

appengine/flexible/tasks/snippets_test.py
```diff
@@ -0,0 +1,107 @@
...
+    name = "projects/{}/locations/{}/queues/{}".format(
+        TEST_PROJECT_ID, TEST_LOCATION, QUEUE_NAME[2])
+    result = snippets.retry_task(
+        TEST_PROJECT_ID, TEST_LOCATION, QUEUE_NAME[0], QUEUE_NAME[1],
+        QUEUE_NAME[2])
+    assert name in result.name
+
+    for i in range(3):
+        snippets.delete_queue(
+            TEST_PROJECT_ID, TEST_LOCATION, QUEUE_NAME[i])
```
grant: I think this line could be more readable with named parameters

```py
project=TEST_PROJECT_ID, location=TEST_LOCATION, queue=QUEUE_NAME[i]
```

PR #2321, opened 2019-08-07
Fixing bug in translate beta samples

ecrowdus:
@beccasaurus 

now: translate_text_with_glossary iterates over glossary translations
previously: translate_text_with_glossary iterates over non-glossary translations


translate/cloud-client/hybrid_glossaries/hybrid_tutorial.py
```diff
@@ -0,0 +1,186 @@
+# Copyright 2018 Google Inc. All Rights Reserved.
```
beccasaurus: Looks like unrelated files for a tutorial are here in your master branch that you're sending the Pull Request from.

Would you mind removing everything besides the change to fix this bug? You should also be able to reproduce this bug in a test, ideally this PR would have 2 files changes: 1. the test 2. the sample

ecrowdus: Sorry about that! I removed the other files when I created the PR, but I've since added them back; I forgot that the PR would keep updating with new commits to my forked repo.

I'll add the test now, and remove the other files.

ecrowdus:
Will replicate PR with branching to clean up commit history.

PR #2338, opened 2019-08-13
conversion of notebook for jupyter/Colab

merla18:
conversion of the notebook to support both Jupyter and Colab + bug fixes


gguuss:
Additionally of note, it appears that submitting Jupyter notebooks circumvents our LINT checks, which is potentially bad as well.

PR #2340, opened 2019-08-13
conversion to jupyter/colab

merla18:
plus bug fixes


sirtorry:
I agree with @gguuss, we'll eventually want to set up CI/CD to test our notebooks. However, since this notebook is strictly an improvement over the old version, let's merge it.

PR #2349, opened 2019-08-19
Purge products

dizcology:
samples for new features:
* delete all products in a product set
* delete all products not in any product sets

### DIFF
--- a/vision/cloud-client/product_search/product_in_product_set_management.py
+++ b/vision/cloud-client/product_search/product_in_product_set_management.py
@@ -25,10 +25,12 @@ import argparse
 
 # [START vision_product_search_add_product_to_product_set]
 # [START vision_product_search_remove_product_from_product_set]
+# [START vision_product_search_purge_products_in_product_set]
 from google.cloud import vision
 
 # [END vision_product_search_add_product_to_product_set]
 # [END vision_product_search_remove_product_from_product_set]
+# [END vision_product_search_purge_products_in_product_set]
 
 
 # [START vision_product_search_add_product_to_product_set]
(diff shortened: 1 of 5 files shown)


dizcology:
DO NOT MERGE until updated to `operation.result(timeout=60)`.

vision/cloud-client/product_search/product_management.py
```diff
@@ -181,6 +183,39 @@ def delete_product(project_id, location, product_id):
...
+    operation = client.purge_products(
+        parent=parent,
+        delete_orphan_products=True,
+        # The operation is irreversible and removes multiple products.
+        # The user is required to pass in force=True to actually perform the
+        # purge.
+        # If force is not set to True, the service raises an exception.
+        force=force)
+
+    # TODO: replace with operation.result(timeout=60)
```
munkhuushmgl: Java there is already method for this purpose, java--> ` response.get(60, TimeUnit.SECONDS);` However, C# --> `operation.Result.PollUntilCompleted();` which doesn't contain timeout option.   Do we really have to include this 60 sec timeout for every language?

davidcavazos:
Once the `operation.result(timeout=60)` is in place, LGTM

vision/cloud-client/product_search/product_in_product_set_management.py
```diff
@@ -117,6 +119,45 @@ def remove_product_from_product_set(
...
+    operation = client.purge_products(
+        parent=parent,
+        product_set_purge_config=product_set_purge_config,
+        # The operation is irreversible and removes multiple products.
+        # The user is required to pass in force=True to actually perform the
+        # purge.
+        # If force is not set to True, the service raises an exception.
+        force=force)
+
+    # TODO: replace with operation.result(timeout=60)
```
beccasaurus: Adding a GitHub PR comment for this TODO to track it ~ this comment will disappear once the TODO is fixed :D

vision/cloud-client/product_search/product_management.py
dizcology: specifying the timeout is not required, and does not need to be consistent across languages.  if possible each language should handle this in the language-specific conventional approach.

steffnay:
LGTM

dizcology:
sample tests passing

PR #2358, opened 2019-08-24
Use latest esp docker image with tag 1

qiwzhang:
Otherwise, many users still using very old ESP image 1.16.0

Tag "1" is always points to the latest ESP image


qiwzhang:
It seems that you were seeing the wrong error log.   Here we only have some kube deployment yaml files.  We don't have anything related to Java.  

The only way to test it is to follow this [doc](https://cloud.google.com/endpoints/docs/grpc/get-started-kubernetes-engine).  That doc uses this deployment yaml file.

gguuss:
LGTM as it will match the other service.

PR #2372, opened 2019-08-29
Storage: HMAC key samples

tritone:


storage/cloud-client/hmac.py
```diff
@@ -0,0 +1,164 @@
...
+    """
+    # [START storage_list_hmac_keys]
+    # project_id = 'Your Google Cloud project ID'
+    storage_client = storage.Client(project=project_id)
+    hmac_keys = storage_client.list_hmac_keys(project_id=project_id)
+    print('HMAC Keys:')
+    for hmac_key in hmac_keys:
+        print('The HMAC key metadata is:')
+        # TODO: fix this once field is added; see
+        #  https://github.com/googleapis/google-cloud-python/pull/9115
```
frankyn: You can probably remove this now that library was released with the fix.

storage/cloud-client/requirements.txt
```diff
@@ -1,2 +1,2 @@
 google-cloud-pubsub==0.39.1
-google-cloud-storage==1.17.0
+google-cloud-storage==1.18.0
```
frankyn: 1.19.0

storage/cloud-client/hmac.py
```diff
@@ -0,0 +1,164 @@
...
+    in order to succeed.
+    """
+    # [START storage_delete_hmac_key]
+    # project_id = 'Your Google Cloud project ID'
+    # access_id = 'ID of an inactive HMAC key'
+    storage_client = storage.Client(project=project_id)
+    hmac_key = storage_client.get_hmac_key_metadata(access_id,
+                                                    project_id=project_id)
+    hmac_key.delete()
+    print('The key is deleted, though it may still appear in getHmacKeys()'
```
frankyn: Use `list_hmac_keys()` instead of `getHmacKeys()`

storage/cloud-client/hmac.py
```diff
@@ -0,0 +1,154 @@
...
+
+def get_key(access_id, project_id):
+    """
+    Retrieve the HMACKeyMetadata with the given access id.
+    """
+    # [START storage_get_hmac_key]
+    # project_id = 'Your Google Cloud project ID'
+    # access_id = 'ID of an HMAC key'
+    storage_client = storage.Client(project=project_id)
+    hmac_key = storage_client.get_hmac_key_metadata(access_id,
```
kurtisvg: Prefer not using hanging indents - see #7 under [Style&Linting](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/AUTHORING_GUIDE.md#style--linting).
```suggestion
    hmac_key = storage_client.get_hmac_key_metadata(
        access_id, project_id=project_id)
```

storage/cloud-client/hmac.py
```diff
@@ -0,0 +1,154 @@
...
+
+def activate_key(access_id, project_id):
+    """
+    Activate the HMAC key with the given access ID.
+    """
+    # [START storage_activate_hmac_key]
+    # project_id = 'Your Google Cloud project ID'
+    # access_id = 'ID of an inactive HMAC key'
+    storage_client = storage.Client(project=project_id)
+    hmac_key = storage_client.get_hmac_key_metadata(access_id,
```
kurtisvg: hanging indent -- same as above.

storage/cloud-client/hmac.py
```diff
@@ -0,0 +1,154 @@
...
+
+def deactivate_key(access_id, project_id):
+    """
+    Deactivate the HMAC key with the given access ID.
+    """
+    # [START storage_deactivate_hmac_key]
+    # project_id = 'Your Google Cloud project ID'
+    # access_id = 'ID of an active HMAC key'
+    storage_client = storage.Client(project=project_id)
+    hmac_key = storage_client.get_hmac_key_metadata(access_id,
```
kurtisvg: hanging indent -- same as above.

storage/cloud-client/hmac.py
```diff
@@ -0,0 +1,154 @@
...
+def delete_key(access_id, project_id):
+    """
+    Delete the HMAC key with the given access ID. Key must have state INACTIVE
+    in order to succeed.
+    """
+    # [START storage_delete_hmac_key]
+    # project_id = 'Your Google Cloud project ID'
+    # access_id = 'ID of an inactive HMAC key'
+    storage_client = storage.Client(project=project_id)
+    hmac_key = storage_client.get_hmac_key_metadata(access_id,
```
kurtisvg: hanging indent - same as above.

storage/cloud-client/hmac.py
```diff
@@ -0,0 +1,154 @@
...
+    """
+    # [START storage_create_hmac_key]
+    # project_id = 'Your Google Cloud project ID'
+    # service_account_email = 'Service account used to generate HMAC key'
+    storage_client = storage.Client(project=project_id)
+    hmac_key, secret = storage_client.create_hmac_key(
+                                service_account_email=service_account_email,
+                                project_id=project_id)
+    print('The base64 encoded secret is {}'.format(secret))
+    print('Do not miss that secret, there is no API to recover it.')
```
kurtisvg:
```suggestion
    print('Do not lose this secret, as it cannot be recovered.')
```

storage/cloud-client/hmac.py
```diff
@@ -0,0 +1,154 @@
...
+    """
+    Deactivate the HMAC key with the given access ID.
+    """
+    # [START storage_deactivate_hmac_key]
+    # project_id = 'Your Google Cloud project ID'
+    # access_id = 'ID of an active HMAC key'
+    storage_client = storage.Client(project=project_id)
+    hmac_key = storage_client.get_hmac_key_metadata(access_id,
+                                                    project_id=project_id)
+    hmac_key.state = 'INACTIVE'
```
kurtisvg: What's the significance of this value? Is it always `'INACTIVE'` or `'ACTIVE'` -- does the library provide an ENUM for these values? When should I use which -- perhaps link to a docs page with more context?

storage/cloud-client/hmac_test.py
```diff
@@ -0,0 +1,88 @@
...
+import pytest
+
+import hmac
+
+
+PROJECT_ID = os.environ['GOOGLE_CLOUD_PROJECT']
+SERVICE_ACCOUNT_EMAIL = os.environ['HMAC_KEY_TEST_SERVICE_ACCOUNT']
+
+
+@pytest.fixture()
```
kurtisvg:
```suggestion
@pytest.fixture
```

storage/cloud-client/hmac_test.py
```diff
@@ -0,0 +1,88 @@
...
+def new_hmac_key():
+    """
+    Fixture to create a new HMAC key, and to guarantee all keys are deleted at
+    the end of each test.
+    """
+    hmac_key = hmac.create_key(PROJECT_ID, SERVICE_ACCOUNT_EMAIL)
+    yield hmac_key
+    storage_client = storage.Client(project=PROJECT_ID)
+    hmac_key = storage_client.get_hmac_key_metadata(hmac_key.access_id,
```
kurtisvg: I'm a little confused by this line - is this just updating the metadata in case it's changed in the test? A comment might be helpful to clarify.

storage/cloud-client/hmac.py
```diff
@@ -0,0 +1,154 @@
...
+from google.cloud import storage
+
+
+def list_keys(project_id):
+    """
+    List all HMAC keys associated with the project.
+    """
+    # [START storage_list_hmac_keys]
+    # project_id = 'Your Google Cloud project ID'
+    storage_client = storage.Client(project=project_id)
```
kurtisvg: Does this client need to be closed? If so, we should use the with-resource statement.

storage/cloud-client/hmac_test.py
```diff
@@ -0,0 +1,88 @@
...
+SERVICE_ACCOUNT_EMAIL = os.environ['HMAC_KEY_TEST_SERVICE_ACCOUNT']
+
+
+@pytest.fixture()
+def new_hmac_key():
+    """
+    Fixture to create a new HMAC key, and to guarantee all keys are deleted at
+    the end of each test.
+    """
+    hmac_key = hmac.create_key(PROJECT_ID, SERVICE_ACCOUNT_EMAIL)
```
kurtisvg: This fixture is inconsistently handling fixtures - sometimes it uses the snippet to perform the action and sometimes not. We should be consistent.

I think it makes more sense to avoid using the snippet, so that we don't have to worry about the logs from another snippet inadvertently showing up in the test or logs somewhere.

storage/cloud-client/hmac_test.py
```diff
@@ -0,0 +1,88 @@
...
+    yield hmac_key
+    storage_client = storage.Client(project=PROJECT_ID)
```
kurtisvg: We probably want to make this it's own fixture and scope it to `module` to avoiding initializing a new one every time.

storage/cloud-client/hmac_test.py
```diff
@@ -0,0 +1,88 @@
...
+    hmac_key = storage_client.get_hmac_key_metadata(hmac_key.access_id,
+                                                    project_id=PROJECT_ID)
+    if hmac_key.state == 'DELETED':
+        return
+    if not hmac_key.state == 'INACTIVE':
```
kurtisvg: Can we not delete the key if it isn't `INACTIVE`? If so, should the delete snippet clarify that and perform the same check?

storage/cloud-client/hmac.py
tritone: This language was taken from the canonical sample, see https://github.com/googleapis/nodejs-storage/blob/master/samples/hmacKeyCreate.js#L48 . I would think that consistency is most important, unless we want to change it in all of them?

storage/cloud-client/hmac.py
tritone: Fixed. Have we considered using yapf or another opinionated formatter that will make these decisions automatically?

storage/cloud-client/hmac.py
kurtisvg: I'll leave the final decision up to @frankyn, but my concerns are that:

1. It's currently improper english, which might make it harder for non-native english speakers to understand
2. Saying "there is no API to recover it" seems to imply that there may be some other way to recover it, like a support ticket or something (which I assume it not the case)

storage/cloud-client/hmac.py
kurtisvg: Java has something like this ([google-java-formatter](https://github.com/google/google-java-format)), but I'm not aware of plans for Python. I'm okay with it as long as it's compatible with the google python style guides.

storage/cloud-client/hmac.py
tritone: It's from here: https://github.com/googleapis/google-cloud-python/blob/master/storage/google/cloud/storage/hmac_key.py#L33 . There's no enum in the client library for this (though I think I would have used one probably...).

The canonical sample seems to use the same pattern: https://github.com/googleapis/nodejs-storage/blob/master/samples/hmacKeyActivate.js#L44 . We could add a docs link across all samples if it seems necessary (I don't know which docs should be linked to but I can look into it).

storage/cloud-client/hmac_test.py
tritone: Fixed.

storage/cloud-client/hmac_test.py
tritone: Yep, that's exactly what it is-- the hmac_key object will be out of date if the state has been changed during the test. Will add comment.

storage/cloud-client/hmac.py
tritone: I don't believe it does; there seems to be no method to do this and I don't see any examples of it. See https://github.com/googleapis/google-cloud-python/blob/master/storage/google/cloud/storage/client.py for client source code.

storage/cloud-client/hmac_test.py
tritone: I thought about this a bit-- the downside of not using the snippet is that then it's difficult to test the create_key snippet (because then I can't leverage the teardown in the fixture for the newly created key, so if assertions in the test fail the key will never be deleted). Any thoughts on how to get around this?

storage/cloud-client/hmac_test.py
tritone: Are you referring specifically to storage_client? I could do this, but I guess I wasn't thinking of it as a resource that is expensive to initialize.

storage/cloud-client/hmac_test.py
tritone: Yes, only inactive keys can be deleted. The delete snippet does mention that the access_id should be for an inactive hmac key (I basically just copied what was in the canonical sample, https://github.com/googleapis/nodejs-storage/blob/master/samples/hmacKeyDelete.js ).

storage/cloud-client/hmac_test.py
tritone: Other tests such as https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/storage/cloud-client/acl_test.py and https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/storage/cloud-client/bucket_policy_only_test.py seem to use a similar approach with including storage.client in one function-level fixture.

kurtisvg: Yeah - some of our clients launch threads and tie up other resources (which is why I asked about `close()` elsewhere). This client doesn't look too heavy, but there are a [couple of network calls for oath and an http session being created](https://github.com/googleapis/google-cloud-python/blob/db481bfdd6816d020d99df0d4caa307358ab1141/core/google/cloud/client.py#L126), so it's probably worthwhile to reuse. 

I'm actually not sure if you can nest fixtures, but if it doesn't work and since you don't have to close it, you can probably just move it to a module level variable.

storage/cloud-client/hmac_test.py
kurtisvg: I think adding a comment before the deletion (in the snippet) might be more noticeable for users (I missed it), but up to you.

storage/cloud-client/hmac.py
tritone: It is a google project so I would assume it should be? :) I can run it by others on the team who are opinionated about python stuff.

storage/cloud-client/hmac.py
tritone: I also considered doing this but I'm a little concerned that this will be more confusing to whoever is reading/using the samples. The current version works with just one import (`from google.cloud import storage`) ; in order to use the class attributes the user would have to add `from google.cloud.storage.hmac_key import HMACKeyMetadata` as well. I can go this route if you prefer though.

storage/cloud-client/hmac_test.py
tritone: Good points, thanks for the advice. I'll take your first suggestion.

storage/cloud-client/hmac_test.py
tritone: Cool, I'll try this. Since many tests in this repo use the storage client I would move this to a file called shared_fixtures or something.

tritone: You can nest fixtures, but I decided to go for the module-level variable because it seems pretty safe and it's less confusing to read.

storage/cloud-client/hmac_test.py
tritone: I edited the comment to make it hopefully a bit more obvious.

storage/cloud-client/hmac.py
kurtisvg: Ok, that seems reasonable. I think if we were using the newer style samples where the imports are included it would make more sense, but since in this version imports out of the snippet I can see how it might be more confusing.

PR #2373, opened 2019-08-29
Update the automl pipelines tables_client to use new client lib

lwander:


lwander:
@sirtorry

sirtorry:
@beccasaurus

PR #2378, opened 2019-09-04
fix: MP3 file test for Text-to-Speech SSML addresses

ecrowdus:
[b/140437136](https://buganizer.corp.google.com/issues/140437136)

In addition to fixing the test, I also moved the samples into the `cloud-client` directory.

@nnegrey


texttospeech/cloud-client/ssml_addresses_test.py
```diff
@@ -39,16 +39,11 @@ def test_ssml_to_audio(capsys):
 
     # Assert audio file generated
     ssml_to_audio(input_ssml, 'test_example.mp3')
-    assert os.path.isfile('test_example.mp3')
-
-    # Assert audio file generated correctly
-    assert filecmp.cmp('test_example.mp3',
-                       'resources/expected_example.mp3',
```
nnegrey: Is expected_example.mp3 still needed?

ecrowdus: None of the tests depend on it, but my thinking was that it's nice to have an MP3 file in the repo that demonstrates what the correct output is.

nnegrey: Let's remove it, we try not to have unused files.

ecrowdus: Sounds good--just removed it.

PR #2384, opened 2019-09-05
Acknowledge messages before parsing to avoid infinite loop / retry 

gguuss:
Making a minor switch on the image pulling example to avoid infinite loop if a message can't be processed.


hongalex:
In general, we shouldn't ack messages before they are processed. Depending on the reason why the message can't be processed, we should do one of two things.

If the message is taking too long to process, we should use `ModifyAckDeadline` so that the message is not constantly re-delivered. https://google-cloud.readthedocs.io/en/latest/pubsub/subscriber/api/client.html. Alternatively, you can manually extend the default AckDeadline on a topic.

If the call to write is failing, it should be wrapped in a try/except/finally and retried/cancelled (with `Ack`). This way, we can retry the message at a later time if it's a one off issue, or confirm that the message is corrupt/incorrect.

PR #2385, opened 2019-09-06
Pub/Sub to GCS using Dataflow Tutorial

anguillanneuf:
~~Unable to merge until Apache Beam supports TextIO for Python/Streaming. Currently, only Pub/Sub to BigQuery is supported in Python. Check out their Build-in I/O Transforms table [here](https://beam.apache.org/documentation/io/built-in/).~~
_(Sorry @kir-titievsky I was wrong about this!)_

Streaming Pub/Sub to GCS using Dataflow Python Tutorial. 

~~@davidcavazos I have got lots of string encoding and decoding operations in the code. If you see a way to reduce redundancy, please let me know.~~

I will work on a separate PR to add Stream Pub/Sub to BigQuery using Dataflow.

### DIFF
--- a/pubsub/streaming-analytics/PubSubToGCS.py
+++ b/pubsub/streaming-analytics/PubSubToGCS.py
@@ -1,3 +1,18 @@
+# Copyright 2019 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#       http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# [START pubsub_to_gcs]
 import argparse
 import datetime
 import json
@@ -9,8 +24,9 @@ from apache_beam.options.pipeline_options import PipelineOptions
 
 
 class GroupWindowsIntoBatches(beam.PTransform):
-    """A composite transform that enriches Pub/Sub messages by publish
-    timestamps and then groups them based on their window information.
+    """A composite transform that groups Pub/Sub messages based on publish
+    time and outputs a list of dictionaries, where each contains one message
+    and its publish timestamp.
     """
 
     def __init__(self, window_size):
@@ -19,32 +35,23 @@ class GroupWindowsIntoBatches(beam.PTransform):
 
     def expand(self, pcoll):
         return (pcoll
-                # Assigns window info to each element in the PCollection.
+                # Assigns window info to each Pub/Sub message based on its
+                # publish timestamp.
                 | beam.WindowInto(window.FixedWindows(self.window_size))
-                # Transform each element by adding publish timestamp info.
                 | 'Process Pub/Sub Message' >> (beam.ParDo(AddTimestamps()))
-                # Use a dummy key to group all the elements in this window.
+                # Use a dummy key to group the elements in the same window.
                 | 'Add Dummy Key' >> beam.Map(lambda elem: (None, elem))
                 | 'Groupby' >> beam.GroupByKey()
                 | 'Abandon Dummy Key' >> beam.MapTuple(lambda _, val: val))
 
 
 class AddTimestamps(beam.DoFn):
-    """A distributed processing function that acts on the elements in
-    the PCollection.
-    """
 
     def process(self, element, publish_time=beam.DoFn.TimestampParam):
-        """Enrich each Pub/Sub message with its publish timestamp.
-
-        Args:
-            element (bytes): A Pub/Sub message.
-            publish_time: Default to the publish timestamp returned by the
-            Pub/Sub server that's been bound to the message by Apache Beam
-            at runtime.
-
-        Yields:
-            dict of str: Message body and publish timestamp.
+        """Processes each incoming windowed element by extracting the Pub/Sub
+        message body and its publish timestamp into a dictionary.
+        `publish_time` defaults to the publish timestamp returned by the
+        Pub/Sub server and is accessible by Beam at runtime.
         """
 
         yield {
@@ -60,13 +67,7 @@ class WriteBatchesToGCS(beam.DoFn):
         self.output_path = output_path
 
     def process(self, batch, window=beam.DoFn.WindowParam):
-        """Write one batch per file to a Google Cloud Storage bucket.
-
-        Args:
-            batch (list of dict): Each dictionary contains one message and its
-            publish timestamp.
-            window: Window inforamtion bound to the batch.
-        """
+        """Write one batch per file to a Google Cloud Storage bucket. """
 
         ts_format = '%H:%M'
         window_start = window.start.to_utc_datetime().strftime(ts_format)
@@ -95,9 +96,8 @@ def run(argv=None):
 
     # `save_main_session` is set to true because one or more DoFn's rely on
     # globally imported modules.
-    pipeline_options = PipelineOptions(pipeline_args,
-                                       streaming=True,
-                                       save_main_session=True)
+    pipeline_options = PipelineOptions(
+        pipeline_args, streaming=True, save_main_session=True)
 
     with beam.Pipeline(options=pipeline_options) as pipeline:
         (pipeline
@@ -111,3 +111,4 @@ def run(argv=None):
 if __name__ == '__main__': # noqa
     logging.getLogger().setLevel(logging.INFO)
     run()
+# [END pubsub_to_gcs]
--- a/pubsub/streaming-analytics/PubSubToGCS.py
+++ b/pubsub/streaming-analytics/PubSubToGCS.py
@@ -5,30 +5,69 @@ import logging
 import apache_beam as beam
 import apache_beam.transforms.window as window
 from apache_beam.options.pipeline_options import PipelineOptions
-from apache_beam.options.pipeline_options import SetupOptions
-from apache_beam.options.pipeline_options import StandardOptions
 
 
 class ProcessDoFn(beam.DoFn):
+    """A distributed processing function that acts on individual elements in
+    the PCollection.
+    """
+
     def process(self, element, publish_time=beam.DoFn.TimestampParam):
+        """Enrich each Pub/Sub message with its publish timestamps.
+
+        Args:
+            element (bytestring): Pub/Sub message.
+            publish_time: Default to the publish timestamp from Pub/Sub.
+
+        Yields:
+            str: Message body and publish timestamp.
+        """
         publish_time = datetime.datetime.utcfromtimestamp(
             float(publish_time)).strftime("%Y-%m-%d %H:%M:%S.%f")
         yield element.decode('utf-8') + ' published at ' + publish_time
 
 
-class CustomDoFn(beam.DoFn):
+class OutputDoFn(beam.DoFn):
+
     def __init__(self, output_path):
         self.output_path = output_path
 
     def process(self, batch, window=beam.DoFn.WindowParam):
-        ts_format = '%Y-%m-%d %H:%M:%S'
+        """Write each batch in its own file to a Google Cloud Storage bucket.
+
+        Args:
+            batch (list of bytestring): list of enriched Pub/Sub messages.
+            window: Window info that has been bound to each batch.
+        """
+
+        ts_format = '%H:%M'
         window_start = window.start.to_utc_datetime().strftime(ts_format)
         window_end = window.end.to_utc_datetime().strftime(ts_format)
-        filename = self.output_path + ' ' + window_start + ' to ' + window_end
+        filename = '-'.join([self.output_path, window_start, window_end])
         contents = '\n'.join(str(element) for element in batch)
 
         with beam.io.gcp.gcsio.GcsIO().open(filename=filename, mode='w') as f:
             f.write(contents.encode('utf-8'))
+
+
+class ApplyWindowing(beam.PTransform):
+    """A composite transform that enriches Pub/Sub messages by their publish
+    timestamps and groups them by timed windows.
+    """
+
+    def __init__(self, window_size):
+        # Convert minutes into seconds.
+        self.window_size = int(window_size * 60)
+
+    def expand(self, pcoll):
+        return (pcoll
+                # Assigns each element in the PCollection to a time interval.
+                | beam.WindowInto(window.FixedWindows(self.window_size))
+                # Add publish timestamps to each element.
+                | 'Process Pub/Sub Message' >> beam.ParDo(ProcessDoFn())
+                | 'Add Empty Key' >> beam.Map(lambda elem: (None, elem))
+                | 'Groupby' >> beam.GroupByKey()
+                | 'Remove Key' >> beam.MapTuple(lambda key, val: val))
 
 
 def run(argv=None):
@@ -39,29 +78,27 @@ def run(argv=None):
               '"projects/<PROJECT_NAME>/topics/<TOPIC_NAME>".'))
     parser.add_argument(
         '--window_size',
-        type=int,
-        default=1,
+        type=float,
+        default=1.0,
         help=('Output file\'s window size in number of minutes.'))
     parser.add_argument(
         '--output_path',
         help=('GCS Path of the output file including filename prefix.'))
     known_args, pipeline_args = parser.parse_known_args(argv)
 
-    # One or more DoFn's rely on global context.
-    pipeline_options = PipelineOptions(pipeline_args)
-    pipeline_options.view_as(SetupOptions).save_main_session = True
-    pipeline_options.view_as(StandardOptions).streaming = True
+    # `save_main_session` is True because one or more DoFn's rely on
+    # globally imported modules.
+    pipeline_options = PipelineOptions(pipeline_args,
+                                       streaming=True,
+                                       save_main_session=True)
 
-    with beam.Pipeline(options=pipeline_options) as p:
-        (p
-            | 'Read Pub/Sub Messages' >> beam.io.ReadFromPubSub(
-                topic=known_args.input_topic)
-            | beam.WindowInto(window.FixedWindows(known_args.window_size*60))
-            | beam.ParDo(ProcessDoFn())
-            | beam.Map(lambda e: (None, e))
-            | beam.GroupByKey()
-            | beam.MapTuple(lambda k, v: v)
-            | 'Write to GCS' >> beam.ParDo(CustomDoFn(known_args.output_path)))
+    with beam.Pipeline(options=pipeline_options) as pipeline:
+        (pipeline
+        | 'Read PubSub Messages' >> beam.io.ReadFromPubSub(
+            topic=known_args.input_topic)
+        | 'Window into' >> ApplyWindowing(known_args.window_size)
+        | 'Write to GCS' >> beam.ParDo(OutputDoFn(known_args.output_path)))
+
 
 if __name__ == '__main__': # noqa
     logging.getLogger().setLevel(logging.INFO)
--- a/pubsub/streaming-analytics/PubSubToGCS.py
+++ b/pubsub/streaming-analytics/PubSubToGCS.py
@@ -60,7 +60,7 @@ class AddTimestamps(beam.DoFn):
         """
 
         yield {
-            'message_body': json.loads(element),
+            'message_body': element.decode('utf-8'),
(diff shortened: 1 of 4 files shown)


pubsub/streaming-analytics/PubSubToGCS.py
```diff
@@ -0,0 +1,105 @@
+import argparse
+import datetime
+import logging
+
+import apache_beam as beam
+import apache_beam.transforms.window as window
+from apache_beam.options.pipeline_options import PipelineOptions
+
+
+class ProcessDoFn(beam.DoFn):
```
davidcavazos: Please use a more descriptive name like `AddTimestamps`.

pubsub/streaming-analytics/PubSubToGCS.py
```diff
@@ -0,0 +1,105 @@
...
+        Args:
+            element (bytestring): Pub/Sub message.
+            publish_time: Default to the publish timestamp from Pub/Sub.
+
+        Yields:
+            str: Message body and publish timestamp.
+        """
+        publish_time = datetime.datetime.utcfromtimestamp(
+            float(publish_time)).strftime("%Y-%m-%d %H:%M:%S.%f")
+        yield element.decode('utf-8') + ' published at ' + publish_time
```
davidcavazos: How about making this into a dict? That would also make it necessary to format as string when writing to GCS which will most likely happen in any real world scenario. I think it's also pretty common to use dictionaries as an intermediate data structure, so it might be a good pattern to showcase.

```
yield {
  'message': element.decode('utf-8'),
  'publish_time': datetime.datetime.utcfromtimestamp(float(publish_time)),
}
```

pubsub/streaming-analytics/PubSubToGCS.py
```diff
@@ -0,0 +1,105 @@
...
+
+        Args:
+            batch (list of bytestring): list of enriched Pub/Sub messages.
+            window: Window info that has been bound to each batch.
+        """
+
+        ts_format = '%H:%M'
+        window_start = window.start.to_utc_datetime().strftime(ts_format)
+        window_end = window.end.to_utc_datetime().strftime(ts_format)
+        filename = '-'.join([self.output_path, window_start, window_end])
```
davidcavazos: This would only keep track of hours and minutes. If someone is running this at around 5pm in PDT, it would be around the 23:00 to 00:00 conversion in UTC. Moreover, a streaming pipeline will most likely run for days, or even months or years. I think it would be convenient to keep the entire timestamp of the window_start. We can omit the window_end since it can be inferred from the next window_start.

I would use the format `%Y%M%D-%H%M%S`, but feel free to use a different one.

pubsub/streaming-analytics/PubSubToGCS.py
```diff
@@ -0,0 +1,105 @@
...
+
+        Yields:
+            str: Message body and publish timestamp.
+        """
+        publish_time = datetime.datetime.utcfromtimestamp(
+            float(publish_time)).strftime("%Y-%m-%d %H:%M:%S.%f")
+        yield element.decode('utf-8') + ' published at ' + publish_time
+
+
+class OutputDoFn(beam.DoFn):
```
davidcavazos: Please use a more descriptive name like `WriteBatchesToGCS`

pubsub/streaming-analytics/PubSubToGCS.py
```diff
@@ -0,0 +1,105 @@
...
+        window_start = window.start.to_utc_datetime().strftime(ts_format)
+        window_end = window.end.to_utc_datetime().strftime(ts_format)
+        filename = '-'.join([self.output_path, window_start, window_end])
+        contents = '\n'.join(str(element) for element in batch)
+
+        with beam.io.gcp.gcsio.GcsIO().open(filename=filename, mode='w') as f:
+            f.write(contents.encode('utf-8'))
+
+
+class ApplyWindowing(beam.PTransform):
```
davidcavazos: Please use a more descriptive name like `GroupWindowsIntoBatches`

pubsub/streaming-analytics/PubSubToGCS.py
```diff
@@ -0,0 +1,105 @@
...
+    """
+
+    def __init__(self, window_size):
+        # Convert minutes into seconds.
+        self.window_size = int(window_size * 60)
+
+    def expand(self, pcoll):
+        return (pcoll
+                # Assigns each element in the PCollection to a time interval.
+                | beam.WindowInto(window.FixedWindows(self.window_size))
```
davidcavazos: I would move this out into the outer-most pipeline definition at `run` so users can more easily tweak their windowing preferences.

pubsub/streaming-analytics/PubSubToGCS.py
```diff
@@ -0,0 +1,105 @@
...
+                # Add publish timestamps to each element.
+                | 'Process Pub/Sub Message' >> beam.ParDo(ProcessDoFn())
```
davidcavazos: I would also move this out since it doesn't concern the windowing or batching.

pubsub/streaming-analytics/PubSubToGCS.py
```diff
@@ -0,0 +1,105 @@
...
+                | 'Add Empty Key' >> beam.Map(lambda elem: (None, elem))
```
davidcavazos: Please add a comment explaining that we want to group *all* the elements from a window, so we add an empty (unused) key to every element, and use `GroupByKey` to group them all together.

pubsub/streaming-analytics/PubSubToGCS.py
```diff
@@ -0,0 +1,105 @@
...
+                | 'Groupby' >> beam.GroupByKey()
+                | 'Remove Key' >> beam.MapTuple(lambda key, val: val))
```
davidcavazos: Consider the following if they make sense to you:
* `'Remove Key'` -> `'Remove Empty Key'`
* `lambda key, val: val` -> `lambda _, val: val`

pubsub/streaming-analytics/PubSubToGCS.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+def run(argv=None):
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        '--input_topic',
+        help=('The Cloud Pub/Sub topic to read from.'
```
davidcavazos: The parentheses are not really required since the "multiline string" is within the parentheses from `add_argument` anyways. :)

pubsub/streaming-analytics/PubSubToGCS.py
```diff
@@ -0,0 +1,105 @@
...
+        (pipeline
+         | 'Read PubSub Messages' >> beam.io.ReadFromPubSub(
+             topic=known_args.input_topic)
+         | 'Window into' >> ApplyWindowing(known_args.window_size)
+         | 'Write to GCS' >> beam.ParDo(OutputDoFn(known_args.output_path)))
+
+
+if __name__ == '__main__': # noqa
+    logging.getLogger().setLevel(logging.INFO)
+    run()
```
davidcavazos: Totally optional, but I find it useful to do any `argparse` related stuff here in `__main__` and then pass the arguments in the correct format to `run`. That way `run` could be imported and called from another script, and `__main__` only exists as a command-line interpreter.

pubsub/streaming-analytics/PubSubToGCS.py
```diff
@@ -0,0 +1,105 @@
...
+        Args:
+            batch (list of bytestring): list of enriched Pub/Sub messages.
+            window: Window info that has been bound to each batch.
+        """
+
+        ts_format = '%H:%M'
+        window_start = window.start.to_utc_datetime().strftime(ts_format)
+        window_end = window.end.to_utc_datetime().strftime(ts_format)
+        filename = '-'.join([self.output_path, window_start, window_end])
+        contents = '\n'.join(str(element) for element in batch)
```
davidcavazos: If you do decide to use dictionaries instead of strings, it might make sense to call a `beam.Map` to format it into whatever you like before calling this `DoFn`.

pubsub/streaming-analytics/README.md
```diff
@@ -0,0 +1,171 @@
+# Stream Cloud Pub/Sub with Cloud Dataflow
+
+Sample(s) showing how to use [Google Cloud Pub/Sub] with [Google Cloud Dataflow].
+
+## Before you begin
```
davidcavazos: You can now omit all these verbose instructions. I've (recently) merged them into the [dataflow/README.md](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/dataflow/README.md). You can follow a similar format to this [example](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/dataflow/encryption-keys/README.md).

You might need to enable some extra APIs (like PubSub) and keep some of the sample-specific instructions not covered in that guide (like the cron publisher).

pubsub/streaming-analytics/README.md
```diff
@@ -0,0 +1,171 @@
...
+* [PubSubToGCS.py](PubSubToGCS.py)
+
+The following example will run a streaming pipeline. It will read messages from a Pub/Sub topic, then window them into fixed-sized intervals, and write one file per window into a GCS location.
+
++ `--project`: sets the Google Cloud project ID to run the pipeline on
++ `--inputTopic`: sets the input Pub/Sub topic to read messages from
++ `--output`: sets the output GCS path prefix to write files to
++ `--runner [optional]`: specifies the runner to run the pipeline, defaults to `DirectRunner`
++ `--windowSize [optional]`: specifies the window size in minutes, defaults to 1
++ `--temp_location`: needed for executing the pipeline
```
davidcavazos: Needed only if the runner is set to DataflowRunner (?)

pubsub/streaming-analytics/README.md
```diff
@@ -0,0 +1,171 @@
...
+
+### Google Cloud Pub/Sub to Google Cloud Storage
+
+* [PubSubToGCS.py](PubSubToGCS.py)
+
+The following example will run a streaming pipeline. It will read messages from a Pub/Sub topic, then window them into fixed-sized intervals, and write one file per window into a GCS location.
+
++ `--project`: sets the Google Cloud project ID to run the pipeline on
++ `--inputTopic`: sets the input Pub/Sub topic to read messages from
++ `--output`: sets the output GCS path prefix to write files to
```
davidcavazos: `--output_path`?

pubsub/streaming-analytics/README.md
```diff
@@ -0,0 +1,171 @@
...
+## Streaming Analytics
+
+### Google Cloud Pub/Sub to Google Cloud Storage
+
+* [PubSubToGCS.py](PubSubToGCS.py)
+
+The following example will run a streaming pipeline. It will read messages from a Pub/Sub topic, then window them into fixed-sized intervals, and write one file per window into a GCS location.
+
++ `--project`: sets the Google Cloud project ID to run the pipeline on
++ `--inputTopic`: sets the input Pub/Sub topic to read messages from
```
davidcavazos: `--input_topic`?

pubsub/streaming-analytics/README.md
```diff
@@ -0,0 +1,171 @@
...
++ `--output`: sets the output GCS path prefix to write files to
++ `--runner [optional]`: specifies the runner to run the pipeline, defaults to `DirectRunner`
++ `--windowSize [optional]`: specifies the window size in minutes, defaults to 1
```
davidcavazos: `--window_size`?

pubsub/streaming-analytics/README.md
```diff
@@ -0,0 +1,171 @@
...
+```
+
+## Cleanup
+
+1. Delete the [Google Cloud Scheduler] job.
+    ```bash
+    gcloud scheduler jobs delete publisher-job
+    ```
+
+1. `Ctrl+C` to stop the program in your terminal. Note that this does not actually stop the job if you use `DataflowRunner`. Skip 3 if you use the `DirectRunner`.
```
davidcavazos: Instead of "Skip 3 ...", please add that as a note in that step.

pubsub/streaming-analytics/PubSubToGCS.py
anguillanneuf: I'd keep it as HH:MM for this tutorial. Adding the date is too hard for the eyes to see. I don't imagine people leaving this code to run for hours.

pubsub/streaming-analytics/PubSubToGCS.py
anguillanneuf: The reason I have this transformation here is because I want to show a similar pipeline chart as the Java example, which has been published [here](https://cloud.google.com/pubsub/docs/pubsub-dataflow).

pubsub/streaming-analytics/PubSubToGCS.py
anguillanneuf: Ditto.

pubsub/streaming-analytics/PubSubToGCS.py
anguillanneuf: Oh really? I didn't know that!

pubsub/streaming-analytics/README.md
anguillanneuf: Thanks for the link! Will use a similar format.

pubsub/streaming-analytics/PubSubToGCS.py
anguillanneuf: That could work too! I will keep it as it is.

pubsub/streaming-analytics/PubSubToGCS.py
```diff
@@ -0,0 +1,113 @@
...
+        default=1.0,
+        help='Output file\'s window size in number of minutes.')
+    parser.add_argument(
+        '--output_path',
+        help='GCS Path of the output file including filename prefix.')
+    known_args, pipeline_args = parser.parse_known_args(argv)
+
+    # `save_main_session` is set to true because one or more DoFn's rely on
+    # globally imported modules.
+    pipeline_options = PipelineOptions(pipeline_args,
```
gguuss: Nit, I prefer to avoid the hanging indent, you can still pass lint with this as:

PipelineOptions(
    pipeline_args,
    ...

pubsub/streaming-analytics/README.md
```diff
@@ -0,0 +1,211 @@
+# Stream Cloud Pub/Sub with Cloud Dataflow
```
gguuss: Is this page intended to be the tutorial or is this documentation on Google Cloud DevSite? If the tutorial is the markdown, it might be a better idea to publish this in the Community repo.

pubsub/streaming-analytics/PubSubToGCS.py
```diff
@@ -0,0 +1,113 @@
...
+class AddTimestamps(beam.DoFn):
+    """A distributed processing function that acts on the elements in
+    the PCollection.
+    """
+
+    def process(self, element, publish_time=beam.DoFn.TimestampParam):
+        """Enrich each Pub/Sub message with its publish timestamp.
+
+        Args:
+            element (bytes): A Pub/Sub message.
```
gguuss: Nit. We usually don't add parameter descriptions in our python samples but instead give example inputs.

pubsub/streaming-analytics/PubSubToGCS.py
```diff
@@ -0,0 +1,113 @@
+import argparse
```
gguuss: Missing Copyright header

pubsub/streaming-analytics/PubSubToGCS.py
```diff
@@ -0,0 +1,113 @@
...
+import json
+import logging
+
+import apache_beam as beam
+import apache_beam.transforms.window as window
+from apache_beam.options.pipeline_options import PipelineOptions
+
+
+class GroupWindowsIntoBatches(beam.PTransform):
+    """A composite transform that enriches Pub/Sub messages by publish
```
gguuss: Unless enriches is a Dataflow-specific term, I would just say, "A composite transform that groups Pub/Sub messages into windows based on timestamps added to the messages."

pubsub/streaming-analytics/requirements.txt
```diff
@@ -0,0 +1 @@
+apache-beam==2.15.0
```
gguuss: Missing tests.

pubsub/streaming-analytics/README.md
gguuss: Any reason you're not using README-gen and shipping this as restructured text / generated documentation?

anguillanneuf: I should add region tags next. The tutorial is going to go here: https://cloud.google.com/pubsub/docs/pubsub-dataflow

anguillanneuf: I avoided it because this tutorial does not a standard README. Much of it matches the Java tutorial's README.

pubsub/streaming-analytics/requirements.txt
anguillanneuf: @davidcavazos I couldn't quite figure out how to write tests. I tried a few different ways. Hope to get your help!

davidcavazos: Testing would also be easier if your `run` function had the signature `run(input_topic, output_path, window_size=1.0, pipeline_args=None)` and then `__main__` only parsed the arguments with `argparse`.

Optional, but I would also omit the `=` signs on the `pipeline_args` and break them into their own list element instead. I wander if there's a parsing error or something due to quoting or some other shell language assumption.

```
pipeline_args = [
  '--project', PROJECT,
  '--runner', 'DirectRunner',
  '--temp_location', 'gs://{}/pubsub/PubSubToGCS/temp'.format(BUCKET),
]
```

Also, can we use a local temporary directory or does it have to be GCS? (`'--temp_location', tempfile.mkdtemp()`).

For the `output_path`, you'll also want to make a unique directory so there are no collisions if this test is ran concurrently by more than one user. Something like `gs://{}/pubsub/PubSubToGCS/{}/output'.format(BUCKET, uuid.uuid4().hex)`.

Since this is a streaming job, it will run (and block) forever unless you stop it manually, so I would recommend [running on another Process with a timeout](https://dreamix.eu/blog/webothers/timeout-function-in-python-3). Do we need the `time.sleep(WAIT_TIME)`?

You probably also want to delete all the GCS resources created if the test succeeded.

As for your error, it's failing to decode the JSON messages in `AddTimestamps`. Could it be that you're passing arbitrary text rather than a JSON encoded message? Depending on your use case, maybe just converting the message into a string might be enough.

pubsub/streaming-analytics/PubSubToGCS_test.py
```diff
@@ -0,0 +1,93 @@
...
+    # Times out the streaming pipeline after 90 seconds.
+    pipeline_process.join(timeout=90)
+    # Immediately kills the publish process after the pipeline shuts down.
+    publish_process.join(timeout=0)
+
+    pipeline_process.terminate()
+    publish_process.terminate()
+
+    # Check for output files on GCS.
+    assert sp.call(['gsutil', 'ls', 'gs://{}/pubsub/'.format(BUCKET)]) == 0
```
gguuss: Test LGTM, consider using the client library instead of subprocess if you make another change, consider a todo if you want to bookmark it:

```python
    from google.cloud import storage

...

    storage_client = storage.Client()
    blobs = storage_client.list_blobs('gs://{}/pubsub/'.format(BUCKET))
	assert blobs.length > 0

    bucket = storage_client.get_bucket(bucket_name)
    bucket.delete()
```

anguillanneuf: Thanks @gguuss! Hmm, when I tried this locally, my test failed at `storage_client = storage.Client()` (possibly due to library version mismatch). 
```bash
    super(Connection, self).__init__(client, client_info)
TypeError: __init__() takes 2 positional arguments but 3 were given
```

gguuss: Don't worry about this one in this PR, probably better saved for later (optionally too, there's a tradeoff in needing more libraries) as a TODO or Issue.

pubsub/streaming-analytics/PubSubToGCS_test.py
```diff
@@ -0,0 +1,101 @@
...
+    storage_client = storage.Client()
+    bucket = storage_client.get_bucket(BUCKET)
+    blobs = list(bucket.list_blobs(prefix='pubsub/{}/output'.format(UUID)))
+    assert len(blobs) > 0
+
+    # Clean up. Delete topic. Delete files.
+    publisher_client.delete_topic(topic_path)
+
+    for blob in blobs:
+        bucket.delete_blob(blob.name)
```
davidcavazos: Should we also delete the `gs://BUCKET/pubsub/UUID` directory? It won't be used for anything else and every run will keep accumulating them.

PR #2386, opened 2019-09-09
Code adjustments to ensure that snippets for docs have all necessary info

mdejong3:
I'm working on the following two docs: 
https://cloud.google.com/iam/docs/creating-managing-service-accounts
https://cloud.google.com/iam/docs/creating-managing-service-account-keys

The python code snippets for these docs don't always capture the variables and imports necessary to run the code. To fix this, I made the following changes:
-Moved variable definitions to ensure that they would be captured in each code snippet
-Added additional region tags around imports

I may need help with linting--I couldn't figure out how to get Nox to work on my machine.


iam/api-client/service_account_keys.py
```diff
@@ -22,23 +22,31 @@
...
+# [END iam_create_key]
+# [END iam_list_keys]
+# [END iam_delete_key]
 
 
 # [START iam_create_key]
 def create_key(service_account_email):
     """Creates a key for a service account."""
 
     # pylint: disable=no-member
```
andrewsg: I'm not sure this disable=no-member directive is required anymore. Could you please delete it (in all cases) and re-run lint and see if an error occurs? If deleting it does cause an error to occur, we can re-add it.

iam/api-client/service_account_keys.py
```diff
@@ -22,23 +22,31 @@
...
+    credentials = service_account.Credentials.from_service_account_file(
```
andrewsg: It seems strange that we have to define this at all since Cloud Client libraries handle this automatically. Are we certain that discovery is the right library to use and that discovery can't be coaxed to do this automatically?

iam/api-client/service_account_keys.py
mdejong3: Will do!

iam/api-client/service_account_keys.py
andrewsg: I was hoping it would have a Cloud library somewhere but maybe it doesn't. At any rate, we can assume that's accurate for now. Thanks.

andrewsg:
For some reason the CI is only running lint, not the actual tests. I will try to run them locally before approving.

iam/api-client/service_accounts.py
```diff
@@ -73,27 +87,40 @@ def rename_service_account(email, new_display_name):
...
+
+    my_service_account = service.projects().serviceAccounts().get(
         name=resource).execute()
 
     # Then you can update the display name
-    service_account['displayName'] = new_display_name
-    service_account = service.projects().serviceAccounts().update(
+    my_service_account['displayName'] = new_display_name
+    my_service_account = service.projects().serviceAccounts().update(
         name=resource, body=service_account).execute()
```
andrewsg: This line is throwing an error in the test suite I ran manually and I believe it is because you need to rename "body=service_account" to "body=my_service_account". It's alarming that this test didn't run in CI, it must be disabled for some reason.

iam/api-client/service_accounts.py
```diff
@@ -73,27 +87,40 @@ def rename_service_account(email, new_display_name):
...
 
     print('Updated display name for {} to: {}'.format(
-        service_account['email'], service_account['displayName']))
-    return service_account
+        service_account['email'], my_service_account['displayName']))
```
andrewsg: Same problem here, need to rename service_account['email'] to my_service_account

PR #2388, opened 2019-09-11
Updating functions/billing/main.py - Code sample references

markmirch:
Fixed the code sample tags to be correct and added some very small error checking.

### DIFF
--- a/functions/billing/main.py
+++ b/functions/billing/main.py
@@ -16,15 +16,14 @@
 # [START functions_billing_stop]
 import base64
 import json
-# [END functions_billing_stop]
 import os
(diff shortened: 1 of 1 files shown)


grant:
So this PR looks like it is:
- Fixing a bad region tag
- Adding a missing import `os` for the region tag `functions_billing_stop`.

markmirch:
> So this PR looks like it is:
> 
> * Fixing a bad region tag
> * Adding a missing import `os` for the region tag `functions_billing_stop`.

Yep, that's exactly right! Thanks for the clarification.

PR #2398, opened 2019-09-13
Pub/Sub: max_rpc_timeout should be <10 mins

anguillanneuf:
If `max_rpc_timeout` is set to be the same as `total_timeout`, retrying isn't happening either. Updating it to be 10s.


hongalex:
I think we discussed this offline that we should leave it the way it is for now.

PR #2402, opened 2019-09-16
Add bigquery_kms_key Dataflow sample

davidcavazos:
Creating a Dataflow sample to show how to use Customer-Managed Encryption Keys (CMEK) for the [Using customer-managed encryption keys](https://cloud.google.com/dataflow/docs/guides/customer-managed-encryption-keys) docs page.

### DIFF
--- a/dataflow/encryption-keys/README.md
+++ b/dataflow/encryption-keys/README.md
@@ -1,0 +1,197 @@
+# Using customer-managed encryption keys
+
+[![Open in Cloud Shell](http://gstatic.com/cloudssh/images/open-btn.svg)](https://console.cloud.google.com/cloudshell/editor)
+
+This sample demonstrate how to use
+[cryptographic encryption keys](https://cloud.google.com/kms/)
+for the I/O connectors in an
+[Apache Beam](https://beam.apache.org) pipeline.
+For more information, see the
+[Using customer-managed encryption keys](https://cloud.google.com/dataflow/docs/guides/customer-managed-encryption-keys)
+docs page.
+
+## Before you begin
+
+Follow the
+[Getting started with Google Cloud Dataflow](../README.md)
+page, and make sure you have a Google Cloud project with billing enabled
+and a *service account JSON key* set up in your `GOOGLE_APPLICATION_CREDENTIALS` environment variable.
+Additionally, for this sample you need the following:
+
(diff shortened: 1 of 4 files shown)


dataflow/encryption-keys/README.md
```diff
@@ -0,0 +1,197 @@
...
+     --keyring $KMS_KEYRING --purpose encryption
+   ```
+
+   > *Note:* Although you can destroy the
+   > [*key version material*](https://cloud.google.com/kms/docs/destroy-restore),
+   > you [cannot delete keys and key rings](https://cloud.google.com/kms/docs/object-hierarchy#lifetime).
+   > Key rings and keys do not have billable costs or quota limitations,
+   > so their continued existence does not impact costs or production limits.
+
+1. Grant Encrypter/Decrypter permissions to the *Dataflow*, *Compute Engine*, and *BigQuery* accounts.
```
shollyman: We should try to do a better job explaining what this means.  Are there pages we can link to that explain this notion of "service accounts created by a service" vs those that are user created?  If not, can we at least indicate that's what's going on here, as just calling them "dataflow/compute/bigquery account" is ambiguous/confusing.

dataflow/encryption-keys/bigquery_kms_key.py
```diff
@@ -0,0 +1,88 @@
...
+                query=query,
+                use_standard_sql=True,
+                kms_key=kms_key,
+            ))
+            | 'Write to BigQuery with KMS key' >>
+            beam.io.WriteToBigQuery(
+                output_bigquery_table,
+                schema=schema,
+                write_disposition=beam.io.BigQueryDisposition.WRITE_TRUNCATE,
+                kms_key=kms_key,
```
shollyman: Interesting.  I would have expected some kind of bigquery namespace e.g. beam.io.*, does beam do some cleverness to collapse this or is it just a very big namespace with all the enums for all sources/sinks?

davidcavazos: Beam does some cleverness in the `__init__.py` files for every directory level. They import anything from certain modules. It is in fact located at `apache_beam.io.gcp.bigquery. WriteToBigQuery`, but it is collapsed (and thus accessible) simply as `apache_beam.io.WriteToBigQuery` :)

dataflow/encryption-keys/README.md
davidcavazos: Good point, I'll add some more descriptions and look out for any useful links as well.

PR #2403, opened 2019-09-16
Pub/Sub: improve pub.py

anguillanneuf:
Continuously prints out "Published x messages" when `pub.py` is running. This would give the user some indication that the program is running / the publisher is retrying publish / and not hanging. 

@kir-titievsky does this look like a reasonable solution? 

Open Question: 
The client library doesn't throw a Service Unavailable error until total RPC timeout is reached (default to 600s/10mins). We need to update the client library so it can fail fast in that scenario.

### DIFF
--- a/pubsub/cloud-client/quickstart/pub.py
+++ b/pubsub/cloud-client/quickstart/pub.py
@@ -57,10 +57,10 @@ def pub(project_id, topic_name):
     api_future.add_done_callback(get_callback(api_future, data, ref))
 
     # Keep the main thread from exiting while the message future
-    # gets resovled in the background.
+    # gets resolved in the background.
     while api_future.running():
         time.sleep(0.5)
-        print("Published {} message.".format(ref["num_messages"]))
+        print("Published {} message(s).".format(ref["num_messages"]))
 
 
 if __name__ == '__main__':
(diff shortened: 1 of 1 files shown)


pubsub/cloud-client/quickstart/pub.py
```diff
@@ -39,24 +39,28 @@ def callback(api_future):
...
+
     # When you publish a message, the client returns a future.
     api_future = client.publish(topic_path, data=data)
-    api_future.add_done_callback(get_callback(api_future, data))
+    api_future.add_done_callback(get_callback(api_future, data, ref))
 
-    # Keep the main thread from exiting until background message
-    # is processed.
+    # Keep the main thread from exiting while the message future
+    # gets resovled in the background.
```
hongalex: nit: typo "resolved"

pubsub/cloud-client/quickstart/pub.py
```diff
@@ -39,24 +39,28 @@ def callback(api_future):
...
     while api_future.running():
-        time.sleep(0.1)
+        time.sleep(0.5)
+        print("Published {} message.".format(ref["num_messages"]))
```
hongalex: nit: s/message/message(s)

PR #2404, opened 2019-09-17
Fixup sample for list_hmac_keys

tritone:
Fix data printed to match canonical sample.

### DIFF
--- a/storage/cloud-client/hmac.py
+++ b/storage/cloud-client/hmac.py
@@ -29,14 +29,8 @@ def list_keys(project_id):
     hmac_keys = storage_client.list_hmac_keys(project_id=project_id)
     print('HMAC Keys:')
     for hmac_key in hmac_keys:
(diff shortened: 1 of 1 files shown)


storage/cloud-client/hmac.py
```diff
@@ -29,14 +29,9 @@ def list_keys(project_id):
     hmac_keys = storage_client.list_hmac_keys(project_id=project_id)
     print('HMAC Keys:')
     for hmac_key in hmac_keys:
-        print('The HMAC key metadata is:')
-        print('Key ID: {}'.format(hmac_key.id))
+        print('Service Account Email: {}'.format(
```
frankyn: Email is missing from the other samples. I missed that metadata field in Ruby and saw it was missing here as well. Only additional ask I have.

tritone: Added to the other samples. Let me know if you see anything else!

PR #2410, opened 2019-09-23
Add service account enable/disable snippets

mdejong3:
Adding service account enable/disable snippets to go along with this doc addition: https://critique.corp.google.com/#review/267282634


mdejong3:
Also, is there a way to only include relevant commits in a PR? This one is cluttered with commits that are already merged.

PR #2413, opened 2019-09-24
BigQuery: Update pandas/bqstorage samples to latest library changes.

tswast:
* Show how to use BQ Storage API with pandas-gbq and magics.
* Update tests to actually run BQ Storage API on smaller IRS 990 table.

### DIFF
--- a/bigquery/pandas-gbq-migration/requirements.txt
+++ b/bigquery/pandas-gbq-migration/requirements.txt
@@ -1,2 +1,5 @@
-google-cloud-bigquery[pandas,pyarrow]==1.9.0
-pandas-gbq==0.9.0
+google-cloud-bigquery==1.20.0
+google-cloud-bigquery-storage==0.7.0
+pandas==0.25.1
+pandas-gbq==0.11.0
+pyarrow==0.14.1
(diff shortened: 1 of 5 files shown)


bigquery/pandas-gbq-migration/requirements.txt
```diff
@@ -1,2 +1,5 @@
-google-cloud-bigquery[pandas,pyarrow]==1.9.0
-pandas-gbq==0.9.0
+google-cloud-bigquery==1.20.0
+google-cloud-bigquery-storage==0.7.0
+pandas==0.25.1
+pandas-gbq==0.11.0
+pyarrow==0.14.1
```
shollyman: The pyarrow req still feels odd to me.  Should we add a comment if there's an alternate option, or start expressing a harder dependency via one of the other libraries?

tswast: I think at some point we'll make `pyarrow` a hard dependency in `google-cloud-bigquery-storage` (in fact, it already is in the conda package), but probably not until after filtering support is launched for the arrow wire format.

PR #2417, opened 2019-09-25
iam: remove extra spaces after region tag

melaniedejong:
There were a few extra spaces after the [START iam_enable_service_account] region tag.


melaniedejong:
@googlebot I signed it!

PR #2434, opened 2019-10-03
Fix style / fix lint / update requirements

gguuss:
* Taking a stab at manual fixes on stale requirements and tests


gguuss:
Tests are timing out, splitting into separate PRs.

PR #2437, opened 2019-10-04
App engine upversion

gguuss:
Updates app engine versions in split PR.


gguuss:
Tests are green, take a look soon as these will conflict with virtually any PR and take a long time to run.

appengine/standard_python37/django/requirements.txt
```diff
@@ -1,2 +1,2 @@
-Django==2.1.11
+Django==2.2.5
```
Chetchaiyan: In appengine/standard-python37/django/

I found problem when upgrade Django from 2.1.11 to 2.2.5 as shown in this update.
The error state that 

_django.core.exceptions.ImproperlyConfigured: mysqlclient 1.3.13 or newer is required; you have 0.9.3._

Which shown the problem coming from PyMySQL 0.9.3

Can you give me advise on fixing or avoiding this error?

gguuss: @Chetchaiyan I think you just need to update your `mysqlclient` version in your project requirements to be `mysqlclient==1.4.4` or run `pip install --upgrade mysqlclient`. You're confusing this library with PyMySQL, which is in fact version 0.9.3.

Feel free to follow up offline with me, email is github`<at>`gusclass`<dot>`com

billbodapolypore: @Chetchaiyan I am having the same issue.

@gguuss I tried both of your suggestions:

- Adding `mysqlclient==1.4.4` to the requirements.txt then running `pip install -r requirements.txt` 
- Or running `pip install --upgrade mysqlclient`

For both of your suggestions I got these errors:

- `ERROR: Command errored out with exit status 1`
- ...
- `MySQLdb/_mysql.c(29): fatal error C1083: Cannot open include file: 'mysql.h': No such file or directory`
- ...
- `ERROR: Failed building wheel for mysqlclient`

However, when I simply change back `Django==2.1.11` everything runs exactly as it is described in [this Getting Started](https://cloud.google.com/python/django/appengine) guide that I am following.  Therefore, I was able to successfully finish the guide and achieve my learning objective but I'm concerned another beginner behind me may not get to completion and get frustrated with this same error.  It seems to me that something else needs to be addressed in the ` appengine/standard_python37/django/` example when upgrading to Django to 2.2.5 but I'm too novice at Python and Django to know / understand what else needs to be changed.

davidpoblador: The skeleton of the project `requirements.txt` makes the assumption that we are going to be using PyMySQL, which is incompatible with Django 2.2.5. I suggest either changing the requirement to use `mysqlclient` instead or wait for a PyMySQL version bump.

PR #2450, opened 2019-10-07
Dataflow client library

davidcavazos:
* Use client libraries instead of `subprocess`
* Updated README formatting
* Updated requirements

### DIFF
--- a/dataflow/run_template/main_test.py
+++ b/dataflow/run_template/main_test.py
@@ -12,14 +12,20 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# To run the tests:
+#   nox -s "lint(sample='./dataflow/run_template')"
+#   nox -s "py27(sample='./dataflow/run_template')"
+#   nox -s "py36(sample='./dataflow/run_template')"
+
 import flask
 import json
 import os
 import pytest
-import subprocess as sp
 import time
 
 from datetime import datetime
+from googleapiclient.discovery import build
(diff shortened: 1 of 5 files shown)


dataflow/run_template/main_test.py
```diff
@@ -86,6 +106,16 @@ def test_run_template_json(app):
...
+def dataflow_jobs_cancel(job_id):
+    # Wait time until a job can be cancelled, as a best effort.
+    # If it fails to be cancelled, the job will run for ~8 minutes.
+    time.sleep(5)  # seconds
+    request = dataflow.projects().jobs().update(
+        projectId=PROJECT,
+        jobId=job_id,
+        body={'requestedState': 'JOB_STATE_CANCELLED'}
+    )
+    request.execute()
```
gguuss: In the subprocess approach, you check for success, can you check here or is this "fire and forget"?  Probably fine either way as it's a small job.

davidcavazos: Yes, after closer inspection that shouldn't actually be the case. If launching the template fails it will raise an exception which will cause the test to fail.

The job can only be cancelled after it changes status from not started into ready or something like that which takes a couple seconds (usually). That's why we're sleeping for 5 seconds. However, if for some reason it takes longer and it cannot be cancelled, the job will still run for 5-8 minutes which is not a big deal, but that shouldn't cause the test to fail.

So we're cancelling as a best effort to try to minimize resources used, so I think we're fine in not checking the output of the cancel.

PR #2456, opened 2019-10-08
removed 'speech_transcribe_infinite_streaming_init' region tag

bradmiro:
Deleted region tag as it didn't make sense in the context of the other samples


alixhami:
This PR is pretty confusing. It looks like a lot more happened than just a region tag removal. Are you sure this didn't undo a bunch of other unrelated commits?

PR #2463, opened 2019-10-09
feat: Add serverless scope sample

grant:
Add a serverless scope sample.

http://b/142060621

```
nox -s "lint(sample='./functions/tips')"
```

### DIFF
--- a/functions/tips/main.py
+++ b/functions/tips/main.py
@@ -53,13 +53,13 @@ def function_specific_computation():
 
 
 def light_computation():
-    numbers = list(range(1, 9))
+    numbers = list(range(1, 10))
     return reduce(lambda x, t: t + x, numbers)
 
 
 def heavy_computation():
     # Multiplication is more computationally expensive than addition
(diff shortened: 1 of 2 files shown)


grant:
Should be ready for review.

functions/tips/main.py
```diff
@@ -50,13 +52,52 @@ def function_specific_computation():
...
 
 # Declared at cold-start, but only initialized if/when the function executes
 lazy_global = None
 
+# [START functions_tips_scopes]
+# [START run_tips_global_scope]
+# Global (instance-wide) scope
+# This computation runs at instance cold-start
+instance_var = heavy_computation()
+# [END functions_tips_global_scope]
```
averikitsch: Can you wrap this code in one region tag not two separate chunk?

averikitsch: Also it looks like you put this code in the middle of another region tag.

grant: Yikes. Thanks for catching that.
The latest diff should be cleaner.

PR #2467, opened 2019-10-10
Asset: Update Real Time Feed API unit tests

cwxie-google:
We changed our feature recently, which need a pubsub topic to create resource.

Update the all CRUD tests.


gguuss:
Looks like I ran the wrong tests :/

gguuss:
But they're still passing!

PR #2471, opened 2019-10-10
Add to comment to clarify kubernetespodoperator

leahecole:
Ping me for internal bug reference


leahecole:
Oof I'll figure out this failure sometime next week maybe if I have time. Low priority, @tswast

tswast:
Oof. Test failure is a strange one. Sounds like a problem with installing `apache-airflow[gcp_api]==1.10.3` https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/composer/workflows/requirements.txt

tswast:
Sounds like we need to pin the Flask or maybe just update to latest patch release of Airflow. https://github.com/puckel/docker-airflow/issues/395

PR #2473, opened 2019-10-11
Fix encoding of object names.

houglum:
Per the docs, the "/" character should not be escaped in the object name
portion of the signed URL.  Also, Python 2 and 3 differ on whether the
tilde character should be escaped by the url library's `quote` method,
so in order for our sample to work on both Python2 and Python3, we
should always specify the tilde as a "safe" character. For reference, we
already fixed this same bug in gsutil a few months ago:
https://github.com/GoogleCloudPlatform/gsutil/blob/4531874e3a3d45ca642fc9c9fa438ba58c8f1494/gslib/commands/signurl.py#L543

### DIFF
--- a/storage/signed_urls/generate_signed_urls.py
+++ b/storage/signed_urls/generate_signed_urls.py
@@ -30,6 +30,7 @@ import hashlib
 import sys
 
 # pip install six
+import six
 from six.moves.urllib.parse import quote
 
 # [START storage_signed_url_signer]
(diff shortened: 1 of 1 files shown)


houglum:
Context: I saw this script included in our docs at https://cloud.google.com/storage/docs/access-control/signing-urls-manually, and was trying to use it to reproduce the issue mentioned in https://issuetracker.google.com/issues/142546499, when I noticed that forward slashes in object names were being encoded.  It's mentioned in both our docs and the [AWS S3 docs here](https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-query-string-auth.html) that you should not escape forward slashes in the object name.

houglum:
Summoning @frankyn for review

frankyn:
@houglum can you ping TW to republish documentation?

houglum:
Done! CC'd you on the email.

PR #2476, opened 2019-10-15
Add jinja2 example

montao:
Create example which uses jinja2 template at the url `/about` similar to the hello_world project but using a jinja2 template.


andrewsg:
This repository is used for sample code that is directly imported by our Google Cloud documentation at cloud.google.com/docs. I'm afraid we can't directly integrate completely samples that aren't related to our docs in most cases. Thank you, though!

PR #2480, opened 2019-10-17
Use deploy redirector

jamesward:


gguuss:
Looks like this might be a [duplicate of this PR](https://github.com/GoogleCloudPlatform/python-docs-samples/pull/2484/files). Otherwise, LGTM.

leahecole:
these test failures are not caused by the modification of this MD file, so I'm merging :)

PR #2489, opened 2019-10-21
Add set_endpoint samples

nnegrey:


automl/beta/data_residency.py
```diff
@@ -0,0 +1,38 @@
...
+# limitations under the License.
+
+
+def data_residency(project_id):
+    """Set a regional data endpoint using data residency"""
+    # [START automl_client_library]
+    # [START automl_data_residency]
+    from google.cloud import automl_v1beta1 as automl
+
+    client_options = {'api_endpoint': 'automl.googleapis.com:443'}
```
dizcology: Need to update to an actual regional endpoint here and in the other samples.

nnegrey: So at the very least, this is the valid `default` endpoint.

nnegrey: Done

PR #2490, opened 2019-10-21
table parsing and key value parsing samples

dizcology:
two samples:
- table parsing
- key value parsing


document/cloud-client/parse_key_value.py
```diff
@@ -0,0 +1,120 @@
+#!/usr/bin/env python
+
+# Copyright 2019 Google LLC. All Rights Reserved.
```
nnegrey: License should just be: `# Copyright 2019 Google LLC`

Without the `.` period after LLC
And drop the `All Rights Reserved.`

document/cloud-client/parse_key_value.py
```diff
@@ -0,0 +1,120 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import argparse
+
+
+# [START document_parse_key_value]
```
nnegrey: Should we put the start just above the `import` lines and below the function comment?

Then we also started adding the defaults for the parameters as commented out variables?

document/cloud-client/parse_key_value.py
dizcology: done.

document/cloud-client/parse_key_value.py
dizcology: updated, please take a look.

dizcology:
replaced by https://github.com/GoogleCloudPlatform/python-docs-samples/pull/2551

PR #2497, opened 2019-10-25
Add Set Endpoint Samples

nnegrey:


automl/beta/set_endpoint.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+def set_endpoint(project_id):
+    """Change your endpoint"""
+    # [START automl_set_endpoint]
+    from google.cloud import automl_v1beta1 as automl
+
+    # You must first create a dataset, using the `eu` endpoint, before you can
+    # call other operations such as: list, get, import, delete, etc.
+    client_options = {'api_endpoint': 'eu-automl.googleapis.com:443'}
```
dizcology: Are there other endpoints the user can select here?  If so please consider linking to the doc page that contains that information.

automl/beta/set_endpoint.py
```diff
@@ -0,0 +1,39 @@
...
+
+    # A resource that represents Google Cloud Platform location.
+    # project_id = 'YOUR_PROJECT_ID'
+    project_location = client.location_path(project_id, 'eu')
+    # [END automl_set_endpoint]
+
+    # List all the datasets available
+    # Note: Create a dataset in `eu`, before calling `list_datasets`.
+    response = client.list_datasets(
+        project_location, '')
```
dizcology: what is the `''` argument here?  if it is required please pass it in as named argument.

language/cloud-client/v1/set_endpoint_test.py
```diff
@@ -0,0 +1,22 @@
...
+# limitations under the License.
+
+import set_endpoint
+
+
+def test_set_endpoint(capsys):
+    set_endpoint.set_endpoint()
+
+    out, _ = capsys.readouterr()
+    assert 'Sentiment' in out
```
dizcology: in this case the test does not really give a way of confirming that the endpoint has been set correctly, right?

automl/beta/set_endpoint.py
nnegrey: Annoyingly, it is required. 
I'll have to look, I think it is a filter field.

language/cloud-client/v1/set_endpoint_test.py
nnegrey: Technically no, but if it doesn't throw an error then it is working.

automl/beta/set_endpoint.py
nnegrey: No, just eu

PR #2501, opened 2019-10-28
fix problems with automl docs

alefhsousa:
Today when we try to use the function `batch_predict` follow the docs we receive and error saying: `the paramaters should be a pandas.Dataframe` it’s happens because the first parameter of the function `batch_predict` is a pandas.Dataframe. To solve this problem we need to use de named parameters of python.


alefhsousa:
@googlebot I signed it!

sirtorry:
@leahecole please approve and merge. thank you.

leahecole:
@sirtorry please tap the techwriter if they need to do a refresh of the page

PR #2502, opened 2019-10-28
fix: add bucket-name as required arg to v4 snippets

jkwlui:


frankyn:
Spoke with @jkwlui offline: Agreed that's the issue and context for this PR.

Issue #2483, opened 2019-10-17
Verify correct usage for changed Resources parameter


gguuss:
While upgrading the client library, I noticed that one of the expected parameters in:

    composer/workflows/kubernetes_pod_operator.py 

Was causing failures.

You can see [the "fix" I added in this commit](https://github.com/GoogleCloudPlatform/python-docs-samples/pull/2436/commits/e26cd5f31be07d4ffc0e9421b3c18a912336d5e7). 

This issue is to track determining whether this is the right approach or to replace the hack with the correct (expected) input.

leahecole:
Debugging note - this PR in Airflow is what caused the problem - https://github.com/apache/airflow/pull/4551

Now investigating if merits a change in our sample, or if it actually should be a change in Airflow. Stay tuned while I figure that out!

leahecole:
Update - it's an us problem! See the linked PR (#2506) and the clarifying comment with instructions for how to run in < airflow 1.10.4

PR #2506, opened 2019-10-31
Fix kubernetespodoperator resources param

leahecole:
Addresses #2483


leahecole:
If preferred, I can separate out the commit to requirements.txt. I originally had them together in case it was related, but later figured out it's not. LMK what's best.

PR #2547, opened 2019-11-18
Fix pubsub topic name format

engelke:
Apparently the patch API tolerated a bad format name until a few weeks ago, so this error was uncaught until now.


leahecole:
That's what happens with a product in beta! Thanks for the quick fix :)

PR #2554, opened 2019-11-18
Adding Cloud run system package sample

dinagraves:
Creating python sample for https://cloud.google.com/run/docs/tutorials/system-packages.

This sample Cloud Run service demonstrates using a customized Dockerfile to configure system packages whose command-line utilities are used as part of serving HTTP requests.

### DIFF
--- a/run/system-package/Dockerfile
+++ b/run/system-package/Dockerfile
@@ -1,0 +1,42 @@
+# Copyright 2019 Google, LLC.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Use the official Python image.
+# https://hub.docker.com/_/python
+FROM python:3.7
+
+# [START run_system_package_ubuntu]
+RUN apt-get update -y && apt-get install -y \
+  graphviz \
+  && apt-get clean
+# [END run_system_package_ubuntu]
+
+# Copy application dependency manifests to the container image.
+# Copying this separately prevents re-running pip install on every code change.
+COPY requirements.txt .
+
+# Install production dependencies.
+RUN pip install -r requirements.txt
+
+# Copy local code to the container image.
+ENV APP_HOME /app
+WORKDIR $APP_HOME
+COPY . .
+
+# Run the web service on container startup. 
+# Use gunicorn webserver with one worker process and 8 threads.
+# For environments with multiple CPU cores, increase the number of workers
+# to be equal to the cores available.
+CMD exec gunicorn --bind :$PORT --workers 1 --threads 8 main:app
+
--- a/run/system-package/README.md
+++ b/run/system-package/README.md
@@ -32,7 +32,7 @@ _Note: you may need to install `pytest` using `pip install pytest`._
 
 ## Deploy
 
-```
+```sh
 # Set an environment variable with your GCP Project ID
 export GOOGLE_CLOUD_PROJECT=<PROJECT_ID>
 
--- a/run/system-package/README.md
+++ b/run/system-package/README.md
@@ -1,0 +1,44 @@
+# Cloud Run System Package Sample
+
+This sample shows how to use a CLI tool installed as a system package as part of a web service.
+
+Use it with the [Using system packages tutorial](https://cloud.google.com/run/docs/tutorials/system-packages).
+
+
+[![Run in Google Cloud][run_img]][run_link]
+
+[run_img]: https://storage.googleapis.com/cloudrun/button.svg
+[run_link]: https://console.cloud.google.com/cloudshell/editor?shellonly=true&cloudshell_image=gcr.io/cloudrun/button&cloudshell_git_repo=https://github.com/GoogleCloudPlatform/python-docs-samples&cloudshell_working_dir=run/system-package
+
+## Build
+
+```
+docker build --tag graphviz:python .
+```
+
+## Run Locally
+
+```
+docker run --rm -p 9090:8080 graphviz:python
+```
+
+## Test
+
+```
+pytest
+```
+
+_Note: you may need to install `pytest` using `pip install pytest`._
+
+## Deploy
+
+```
+# Set an environment variable with your GCP Project ID
+export GOOGLE_CLOUD_PROJECT=<PROJECT_ID>
+
+# Submit a build using Google Cloud Build
+gcloud builds submit --tag gcr.io/${GOOGLE_CLOUD_PROJECT}/graphviz
+
+# Deploy to Cloud Run
+gcloud beta run deploy graphviz --image gcr.io/${GOOGLE_CLOUD_PROJECT}/graphviz
+```
--- a/run/system-package/main.py
+++ b/run/system-package/main.py
@@ -40,7 +40,7 @@ def index():
 
         # If no graphviz definition or bad graphviz def, return 400
         if "syntax" in str(e):
-            return f"Bad Request: {e}", 400
+            return "Bad Request: {}".format(e), 400
 
         return "Internal Server Error", 500
 
--- a/run/system-package/main.py
+++ b/run/system-package/main.py
@@ -18,6 +18,15 @@ import subprocess
 import sys
 
 
+# Uses local `dot` binary from Graphviz: https://graphviz.gitlab.io
+
+# NOTE:
+# Local installation may not be the same as container installation.
+# To run locally, run `brew install graphviz`
+# and update PACKAGE_PATH to the output of `which dot`.
+PACKAGE_PATH = '/usr/bin/dot'
+
+
 app = Flask(__name__)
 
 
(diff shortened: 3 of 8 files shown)


run/system-package/Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+# limitations under the License.
+
+# Use the official Python image.
+# https://hub.docker.com/_/python
+FROM python:3.7
+
+# [START run_system_package_ubuntu]
+RUN apt-get update -y && apt-get install -y \
+  graphviz \
+  && apt-get clean
```
grant: Bash commands should not be escaped with `\` unnecessarily (i.e. between `install` and `graphviz`).

Would this work and be more readible?

```sh
RUN apt-get update -y
RUN apt-get install -y graphviz
RUN apt-get clean
```

run/system-package/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+```
+pytest
+```
+
+_Note: you may need to install `pytest` using `pip install pytest`._
+
+## Deploy
+
+```
```
grant: \`\`\`sh

run/system-package/main.py
```diff
@@ -0,0 +1,72 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from flask import Flask, make_response, request
+import os
+import subprocess
+import sys
+
+app = Flask(__name__)
```
grant: nit: I believe python style is usually 2 newlines between blocks like this, or at least after the statement (1 on each side here).

https://github.com/GoogleCloudPlatform/python-docs-samples/search?p=1&q=app+%3D+Flask%28__name__%29&type=&utf8=%E2%9C%93

run/system-package/main.py
```diff
@@ -0,0 +1,72 @@
...
+
+# [START run_system_package_handler]
+@app.route('/diagram.png', methods=['GET'])
+def index():
+    try:
+        image = create_diagram(request.args.get('dot'))
+
```
grant: nit: Newline not needed here.

run/system-package/main.py
```diff
@@ -0,0 +1,72 @@
...
+        response.headers.set('Content-Type', 'image/png')
+        return response
+
+    except Exception as e:
+        print(f'error: {e}')
+
+        # Flush the stdout to avoid log buffering.
+        sys.stdout.flush()
+
+        if 'syntax' in str(e):
```
grant: This strange conditional could use a comment.

run/system-package/main.py
```diff
@@ -0,0 +1,72 @@
...
+            return f'Bad Request: {e}', 400
+
+        return 'Internal Server Error', 500
+# [END run_system_package_handler]
+
+
+# [START run_system_package_exec]
+def create_diagram(dot):
```
grant: What is `dot`? Are we using Graphviz's DOT language?
Can we add a link somewhere in this sample?

i.e.
https://www.graphviz.org/doc/info/lang.html

run/system-package/main.py
```diff
@@ -0,0 +1,72 @@
...
+    if not dot:
+        raise Exception('syntax: no graphviz definition provided')
+
+    dot_args = ['/usr/bin/dot',  # Command to run
```
grant: Where is `/usr/bin/dot` coming from? It's not on my machine. Perhaps add a comment around this.

run/system-package/main.py
```diff
@@ -0,0 +1,72 @@
...
+from flask import Flask, make_response, request
+import os
+import subprocess
+import sys
+
+app = Flask(__name__)
+
+# [START run_system_package_handler]
+@app.route('/diagram.png', methods=['GET'])
+def index():
```
grant: A comment as to what this method does would be nice.

For example, we seem to expect a `dot` query parameter in our URL. What is the shape of that query param? What is an example? Is there a link to what this type of data structure?

run/system-package/main_test.py
```diff
@@ -0,0 +1,56 @@
...
+    r = client.get('/diagram.png?dot=digraph')
+    assert r.status_code == 400
+
+
+@mock.patch('main.subprocess.run',
+            mock.MagicMock(return_value=SubprocessOject('Image')))
+def test_good_dot_parameter(client):
+    r = client.get(
+        '/diagram.png?dot=digraph G { A -> {B, C, D} -> {F} }')
+    assert r.status_code == 200
```
grant: It's great that we have tests, however this is not a strong assertion. Our program could do nothing and still respond with a `200`.

Can we test things like the `Content-Type` and the content of the response is not empty or has some value?

run/system-package/Dockerfile
busunkim96: It's recommended to reduce the number of separate `RUN`s where it makes sense to minimize the number of layers. https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#minimize-the-number-of-layers

grant: That makes sense.

However, I'd still not put `graphviz` on it's own line:

```sh
RUN apt-get update -y \
  apt-get install -y graphviz \
  apt-get clean
```

dinagraves: Currently it matches the snippet from the [go sample](https://github.com/GoogleCloudPlatform/golang-samples/blob/master/run/system_package/Dockerfile).  Should I update both?

run/system-package/main.py
dinagraves: ack

run/system-package/main.py
dinagraves: LMK if this is clearer!

run/system-package/main_test.py
dinagraves: This is hard b/c I don't have the package installed on my computer - so I wanted to mock the package.

run/system-package/main.py
dinagraves: Added a comment - is it clearer?

run/system-package/main.py
dinagraves: Added a link - LMK if if you think it needs more info!

run/system-package/Dockerfile
grayside: While conciseness is nice, the reason most of the samples use a dedicated line for the package is it makes the diffs much cleaner to have one package per line for the future.

run/system-package/main_test.py
grayside: The way this has been handled in Java, Go, and Node is to add the graphviz package to teh containers used for testing.

dinagraves: I see - so far all the python tests for the other samples I've worked on could be run locally without any special dependencies.  @grant Do you know how to include the package for testing?

run/system-package/main.py
grant: I still don't understand what this method does by reading the comment.

```
Uses the graphviz DOT language
https://www.graphviz.org/doc/info/lang.html
```

From reading the code, it seems like this method takes a HTTP GET request with query param `dot` and returns a png with the rendered DOT diagram in a HTTP response.

I think it would be clearer if the comment stated something more along that.

run/system-package/main.py
grant: I can't see the link.

run/system-package/main_test.py
grant: >  @grant Do you know how to include the package for testing?

Add `graphviz==0.13.2` to `requirements.txt`?

I don't know how the other languages test, but we should just follow their tests, but for python.

run/system-package/main.py
grant: I am still seeing `bash: /usr/bin/dot: No such file or directory`.

I think it would be important to state the dependency of this snippet. Maybe look at the other language samples.



```py
# Uses local `dot` binary from Graphviz: https://graphviz.gitlab.io 
dot_args = ['/usr/bin/dot',
  ...
```

run/system-package/main_test.py
dinagraves: The issue there is that would be a python module.  The point is that it's a system package installed in the container - it should run a binary.

grant: Sorry, I really don't know what we mean here. Maybe ask @grayside since he seems to know what is being asked and wrote the other samples?

dinagraves: Talked to Averi and she said she installed graphviz locally for Java , so I did the same for Python.  It has a different location path when you run brew install vs how it's installed in the container, but I put the path in a global variable to easily be able to switch for local tests.

run/system-package/main.py
dinagraves: Updated.

run/system-package/main.py
dinagraves: This is now at the top of the file with a more verbose explanation of where it comes from, how to install locally, and how to find the filepath.

run/system-package/main.py
dinagraves: Link is to the graphviz page - now at top of file.

run/system-package/Dockerfile
dinagraves: Going to resolve this and keep it consistent with the other language samples.

run/system-package/main_test.py
dinagraves: Currently blocked by https://github.com/googleapis/testing-infra-docker/pull/45

busunkim96: I believe graphviz is now available in the published image. Can a stronger assertion be made here now? Or is this thread irrelevant now?

dinagraves: Updated!

run/system-package/main.py
```diff
@@ -0,0 +1,85 @@
...
+    # Takes an HTTP GET request with query param dot and
+    # returns a png with the rendered DOT diagram in a HTTP response.
+    try:
+        image = create_diagram(request.args.get("dot"))
+        response = make_response(image)
+        response.headers.set("Content-Type", "image/png")
+        return response
+
+    except Exception as e:
+        print(f"error: {e}")
```
busunkim96: This style of format string is not supported in Python < 3.6
```suggestion
        print("error: {}".format(e))
```

run/system-package/main.py
```diff
@@ -0,0 +1,85 @@
...
+
+        # Flush the stdout to avoid log buffering.
+        sys.stdout.flush()
+
+        # If no graphviz definition or bad graphviz def, return 400
+        if "syntax" in str(e):
+            return f"Bad Request: {e}", 400
```
busunkim96:
```suggestion
            return "Bad Request: {}".format(e), 400
```

busunkim96:
This sample should not run with 2.7. Issue filed for the incomplete skip in #2596

PR #2558, opened 2019-11-19
fix: Use different versions of pytest for python 2 and python3

### LINKED ISSUE #2556: appgengine fusion tests failing
Hey! It seems like this PR #2428 broke the appengine fusion tests awhile ago - it's having trouble finding that version of pytest for the python 2 tests. Given the sunset of Python 2, do you have opinions on how best to resolve this? Feel free to set up time with me if you want to look at this together.

crwilcox:
Fix #2556 by providing a valid pytest version for python2.

### DIFF
--- a/container_registry/container_analysis/requirements.txt
+++ b/container_registry/container_analysis/requirements.txt
@@ -1,6 +1,7 @@
 google-cloud-pubsub==1.0.0
 google-cloud-containeranalysis==0.3.0
 grafeas==0.1.0
-pytest==5.1.3
+pytest==5.3.0; python_version > "3.0"
+pytest==4.6.6; python_version < "3.0"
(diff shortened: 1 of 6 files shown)


leahecole:
I'm lookign through the kokorofailures - the pubsub one is failing in fusion rn as well, so I'm not particularly concerned about that. You didn't touch BQ, so that seems like a fluke, but run I'm not sure about, because you did edit those files, and it's passing in fusion. Any thoughts @kurtisvg ?

kurtisvg:
It's passing travis, which is using the diff testing to only test the relevant packages. The other tests might be flakey, or failing from conflicts with other CI jobs (since we only have 1 testing project).

leahecole:
Fine by me too.

PR #2566, opened 2019-11-20
Dialogflow example that streams in through mic, out through speaker

dalequark:
This example shows how to stream audio in to Dialogflow through the microphone, capture Dialogflow's audio response, and play that response out through a speaker.


kurtisvg:
Please make sure it's passing the tests in the `continuous-integration/travis-ci/pr` status check.

dialogflow/cloud-client/mic_stream_audio_response.py
```diff
@@ -0,0 +1,123 @@
...
+            voice=dialogflow.types.VoiceSelectionParams(ssml_gender=dialogflow.enums.SsmlVoiceGender.SSML_VOICE_GENDER_FEMALE))
+        output_audio_config = dialogflow.types.OutputAudioConfig(
+            audio_encoding=dialogflow.enums.OutputAudioEncoding.OUTPUT_AUDIO_ENCODING_LINEAR_16,
+            sample_rate_hertz=SAMPLE_RATE,
+            synthesize_speech_config=speech_config)
+        query_input = dialogflow.types.QueryInput(audio_config=input_audio_config)
+
+        # The first request contains the configuration.
+        yield dialogflow.types.StreamingDetectIntentRequest(
+            session=session_path, query_input=query_input, output_audio_config=output_audio_config)
```
nnegrey: Move all the initial request building to a different method:
```
def _build_initial_request(self):
        """The first request should always contain the configuration"""
        session_path = self.session_client.session_path(
            project=self.project_id, session=self.session_id
        )

        input_audio_config = dialogflow.types.InputAudioConfig(
            audio_encoding=dialogflow.enums.AudioEncoding.AUDIO_ENCODING_LINEAR_16,
            language_code=self.language_code,
            sample_rate_hertz=SAMPLE_RATE,
        )
        query_input = dialogflow.types.QueryInput(audio_config=input_audio_config)

        voice = dialogflow.types.VoiceSelectionParams(
            ssml_gender=dialogflow.enums.SsmlVoiceGender.SSML_VOICE_GENDER_FEMALE
        )
        speech_config = dialogflow.types.SynthesizeSpeechConfig(voice=voice)
        output_audio_config = dialogflow.types.OutputAudioConfig(
            audio_encoding=dialogflow.enums.OutputAudioEncoding.OUTPUT_AUDIO_ENCODING_LINEAR_16,
            sample_rate_hertz=SAMPLE_RATE,
            synthesize_speech_config=speech_config,
        )

        return dialogflow.types.StreamingDetectIntentRequest(
            session=session_path,
            query_input=query_input,
            output_audio_config=output_audio_config,
        )
```

dialogflow/cloud-client/mic_stream_audio_response.py
```diff
@@ -0,0 +1,123 @@
...
+import simpleaudio as sa
+import argparse
+import uuid
+
+# Audio recording parameters
+SAMPLE_RATE = 16000
+CHUNK_SIZE = int(SAMPLE_RATE / 10)
+
+
+def grab_intent(projectId, sessionId, languageCode):
```
nnegrey: Should this be a class instead?
```
class Stream:
    """Start stream from microphone input to dialogflow API"""

    def __init__(self, project_id, session_id, language_code):
        self.project_id = project_id
        self.session_id = session_id
        self.language_code = language_code
        self.session_client = dialogflow.SessionsClient()
        self.final_request_received = False
```

dialogflow/cloud-client/mic_stream_audio_response.py
```diff
@@ -0,0 +1,123 @@
...
+# limitations under the License.
+
+"""Dialogflow API streams user input through the microphone and
+speaks voice responses through the speaker.
+
+Examples:
+  python mic_stream_audio_response.py
+  python mic_stream_audio_response.py--project-id PROJECT_ID
+"""
+
```
nnegrey: Add a region tag for docs import
`# [START dialogflow_microphone_streaming]`

dialogflow/cloud-client/mic_stream_audio_response.py
```diff
@@ -0,0 +1,123 @@
...
+def play_audio(audio):
+    audio_obj = sa.play_buffer(audio, 1, 2, SAMPLE_RATE)
+    audio_obj.wait_done()
+
+def main():
+    while(True):
+        response = grab_intent(args.project_id, args.session_id, args.language_code)
+        play_audio(response.output_audio)
+
```
nnegrey: Make it copy pastable from docs:

Maybe something like:
```
# TODO: developer uncomment and replace these variables with your own
# project_id = 'YOUR_PROJECT_ID'
# session_id = uuid.uuid4()
# language_code = 'en-US'
# main(project_id, session_id, langauge_code)
# [END dialogflow_microphone_streaming]
```

dialogflow/cloud-client/mic_stream_audio_response.py
```diff
@@ -0,0 +1,123 @@
...
+        # The first request contains the configuration.
+        yield dialogflow.types.StreamingDetectIntentRequest(
+            session=session_path, query_input=query_input, output_audio_config=output_audio_config)
+
+        while True:
+            if final_request_received:
+                print("received final request")
+                input_stream.close()
+                print("closed stream")
```
nnegrey: Can remove these print statements

dialogflow/cloud-client/mic_stream_audio_response.py
```diff
@@ -0,0 +1,123 @@
...
+
+    while True:
+        print('=' * 20)
+        requests = __request_generator()
+        responses = session_client.streaming_detect_intent(requests)
+
+        for response in responses:
+            print(f'Intermediate transcription result: {response.recognition_result.transcript}')
+            if response.recognition_result.is_final:
+                final_request_received = True
```
nnegrey: Switch this to have a little less noisy output:
```
if response.recognition_result.is_final:
                    print(
                        f"\nFinal transcription: {response.recognition_result.transcript}"
                    )
                    self.final_request_received = True
                elif not self.final_request_received:
                    print(
                        f"Intermediate transcription: {response.recognition_result.transcript}"
                    )
```

dialogflow/cloud-client/mic_stream_audio_response.py
```diff
@@ -0,0 +1,123 @@
...
+                return
+            if input_stream.is_active():
+                content = input_stream.read(CHUNK_SIZE, exception_on_overflow = False)
+                yield dialogflow.types.StreamingDetectIntentRequest(input_audio=content)
+        
+
+
+
+    while True:
+        print('=' * 20)
```
nnegrey: If switching to a class, add this here to reset it: `self.final_request_received = False`

dialogflow/cloud-client/mic_stream_audio_response.py
```diff
@@ -0,0 +1,123 @@
...
+        requests = __request_generator()
+        responses = session_client.streaming_detect_intent(requests)
+
+        for response in responses:
+            print(f'Intermediate transcription result: {response.recognition_result.transcript}')
+            if response.recognition_result.is_final:
+                final_request_received = True
+            if response.query_result.query_text:
+                print(f'Fullfilment Text: {response.query_result.fulfillment_text}')
+                print(f'Intent: {response.query_result.intent}')
```
nnegrey: Can print the intent display_name

dialogflow/cloud-client/mic_stream_audio_response.py
```diff
@@ -0,0 +1,123 @@
...
+
+def grab_intent(projectId, sessionId, languageCode):
+    """Start stream from microphone input to dialogflow API"""
+   
+    session_client = dialogflow.SessionsClient()
+    # Audio output stream
+    
+    final_request_received = False
+
+    def __request_generator():
```
nnegrey: Should there only be 1 underscore to mark private?

dialogflow/cloud-client/mic_stream_audio_response.py
nnegrey: Actually these print statements need to be changed anyway, I think the `f"{}" is only a python 3 thing and as of today, we still support python 2. 

So 
```
                    print(
                        '\nFinal transcription: {}'.format(
                            response.recognition_result.transcript
                        )
                    )
```

dalequark: I think it's actually very useful to see all of the intermediate recognize results, although the output is a little noisy. Very helpful for debugging.

nnegrey: Oh I was just removing the ones that didn't have any transcript results. They all still print out though.

dalequark:
Made some formatting updates and took most of Noah's suggestions except adding a new class. I found that because of the interdependency between the microphone stream and the dialogflow stream, adding a class would have been difficult.

dialogflow/cloud-client/mic_stream_audio_response.py
```diff
@@ -0,0 +1,161 @@
...
+
+# OUTPUT
+# ====================
+# Intermediate transcription: Saint Francis
+# Intermediate transcription: San Francisco
+# Intermediate transcription: San Francisco
+
+# Final transcription: a room in San Francisco
+# Fullfilment Text: What date?
+# Intent: room.reservation
```
nnegrey: Oh my bad you can remove this output

dalequark:
@kurtisvg - this example essentially extends others by adding input from the microphone and output through a speaker. Failure is most likely to happen as a result of the user's hardware. Not sure how we can write tests for this. Please advise.

nnegrey:
I have one initial idea that I haven't thought through very much so not sure how viable or how much refactoring it would take. 

It would be to not use the microphone as part of the tests, but try to split out which parts are for the microphone and which parts call the Dialogflow api, then just have the test send bytes to the Dialogflow method and verify that works. But I'm not sure if that would over complicate the sample.

kurtisvg:
> @kurtisvg - this example essentially extends others by adding input from the microphone and output through a speaker. Failure is most likely to happen as a result of the user's hardware. Not sure how we can write tests for this. Please advise.

Is there some way we can mock microphone input and provide testing data to ensure the code works correctly?

Tests are necessary because as dependencies are updated we need to know if the snippet still works or not. Unless the code is ran and verified we aren't sure if it's broken or not.

wibimaster:
> > @kurtisvg - this example essentially extends others by adding input from the microphone and output through a speaker. Failure is most likely to happen as a result of the user's hardware. Not sure how we can write tests for this. Please advise.
> 
> Is there some way we can mock microphone input and provide testing data to ensure the code works correctly?
> 
> Tests are necessary because as dependencies are updated we need to know if the snippet still works or not. Unless the code is ran and verified we aren't sure if it's broken or not.

You can write some pieces of silence frame ?

```python
sample_rate = 16000
frame_duration = 10  # ms
frame = b'\x00\x00' * int(sample_rate * frame_duration / 1000)
```

or include a wav file.

Examples on https://github.com/wiseman/py-webrtcvad for both methods :)

tmatsuo:
@dalequark 
Are you still working on this?

tmatsuo:
FYI @dalequark If I get no response by the end of April, I'll close this PR.

tmatsuo:
Closing

Issue #2561, opened 2019-11-20
pubsub tests failing in fusion


leahecole:
Hi Alex, it's looking like pubsub tests are failing for the repo in fusion (I'll send you a link) - could you please take a look and see if changes need to be made?

andrewferlitsch:
@hongalex , @leahecole  - if this is a test failing in fusion, then shouldn't the failing be added as bug in buganizer and not here in issues?

hongalex:
This is passing again as of 11/28. Closing.

PR #2567, opened 2019-11-20
Have to specify flask library in app.yaml

engelke:


leahecole:
@anguillanneuf @hongalex could this have been what caused the test problems?

anguillanneuf:
@leahecole Which tests?

anguillanneuf:
@engelke Is `app.yaml` for the deployed app and `requirements.txt` mainly for the local testing app?

leahecole:
@anguillanneuf the tests that were previously failing in #2561 - the ones that are in fusion

anguillanneuf:
@engelke `library` doesn't look to be a valid element in the Python 3.7 runtime for the Standard Environment: https://cloud.google.com/appengine/docs/standard/python3/config/appref

engelke:
This code sample is for the Python 2.7 App Engine runtime, not the 3.7 one. The 2.7 runtime needs third party libraries specified in yaml, while the 3.7 one just uses requirements.txt. I think this whole folder should only be tested under Python 2.7 (and thought that the configuration used to do it that way).

anguillanneuf:
@leahecole This sample is for Python 2.7, could we turn off all the Python 3 tests here?

leahecole:
Happy to merge this given that it is a 2.7 only example and it passes the 2.7 CI. I'll open an issue to see if we can have kokoro run only 2.7 on this sample

PR #2569, opened 2019-11-21
Add vision ocr set endpoint samples

nnegrey:


vision/cloud-client/detect/set_endpoint.py
```diff
@@ -0,0 +1,42 @@
...
+        image_uri='gs://cloud-samples-data/vision/text/screen.jpg')
+    image = vision.types.Image(source=image_source)
+
+    response = client.text_detection(image=image)
+
+    print('Texts:')
+    for text in response.text_annotations:
+        print('{}'.format(text.description))
+
+        vertices = (['({},{})'.format(vertex.x, vertex.y)
```
davidcavazos: Nit: no need for the extra parentheses since the square brackets from the list comprehension already "serve as the parentheses". The parser knows that a `[` opened and can be closed at any other line just like a `(`.

vision/cloud-client/detect/set_endpoint.py
```diff
@@ -0,0 +1,42 @@
...
+
+def set_endpoint():
+    """Change your endpoint"""
+    # [START vision_set_endpoint]
+    from google.cloud import vision
+
+    client_options = {'api_endpoint': 'eu-vision.googleapis.com:443'}
+
+    client = vision.ImageAnnotatorClient(client_options=client_options)
+    # [END vision_set_endpoint]
```
davidcavazos: In Go, and probably other languages, we have to include the entire sample. So I'm also including the part below but added some comments to try to make it explicit that the relevant part is to create the client only.

```go
// Create a client with a custom endpoint.
...

// Use the client with custom endpoint to detect texts on an image.
...
```

Do you think we should keep the code below in this sample as well for consistency? That would give users a runnable piece of code that does something.

vision/cloud-client/detect/set_endpoint.py
```diff
@@ -0,0 +1,42 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+def set_endpoint():
+    """Change your endpoint"""
+    # [START vision_set_endpoint]
+    from google.cloud import vision
+
+    client_options = {'api_endpoint': 'eu-vision.googleapis.com:443'}
```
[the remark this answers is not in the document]
nnegrey: This is the only supported endpoint

vision/cloud-client/detect/set_endpoint.py
nnegrey: No, this is going on a special docs page and they don't want to show anything more than the above, I just have code below so that we can have a test.

davidcavazos: Got it, let me change it in the Go sample to return the client and do the rest in the test then.

vision/cloud-client/detect/set_endpoint.py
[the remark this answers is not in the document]
davidcavazos: Are there plans to support any others? What's the difference from not supplying this? I think I still prefer to have it as a variable, or do you think it makes more sense to hardcode it?

nnegrey:
The tests seems super flaky due to 2 problems:
1) Google Cloud Storage seems to rate limit how quickly the API can access these files
2) The API or library doesn't throw an error rather it returns an error:
```
error {
  code: 3
  message: "The URL does not appear to be accessible by us. Please double check or download the content and pass it in."
}
```
The samples don't do any checks for an error message, rather they just don't print results and it looks like everything succeeded when it did not.

Seems to be somewhat random to which tests are allowed to access the files. 

I suggest just deleting the `http` tests as they are just an additional test to the GCS test and calling the same sample code, which makes it a test of the API and not a sample test. 
@kurtisvg, @leahecole

kurtisvg:
> The API or library doesn't throw an error rather it returns an error:

This definitely seems like this is a bug that should be filed against the client library. 

In the mean time, I think it would be reasonable to consider updating the samples to add this check in, since users might also experience this error and might not know to look for them. 

> I suggest just deleting the `http` tests as they are just an additional test to the GCS test and calling the same sample code, which makes it a test of the API and not a sample test.
> @kurtisvg, @leahecole

Seems reasonable reason to me 👍

leahecole:
+1 to the opening a bug against the client library, and +1 to removing tests that are testing an API and not actually a sample. That's just duplicate effort w/ the client library

nnegrey:
- Opened a bug on the client library
- Opened a bug on myself to add the error checking to the samples
- Deleted the samples that do HTTP tests for this PR to get the tests passing.

PR #2580, opened 2019-12-02
Update DLP samples to use dlp_v2 client.

maximus12793:
Per b/143006235, update doc samples to use dlp_v2 client. Requirements specify 0.12.1 but code samples seem to be using old client. We want to upgrade to keep consistency with Source: https://googleapis.dev/python/dlp/latest/index.html.

### DIFF
--- a/dlp/deid.py
+++ b/dlp/deid.py
@@ -20,8 +20,9 @@ import argparse
 
 
 # [START dlp_deidentify_masking]
-def deidentify_with_mask(project, string, info_types, masking_character=None,
-                         number_to_mask=0):
+def deidentify_with_mask(
+    project, string, info_types, masking_character=None, number_to_mask=0
+):
     """Uses the Data Loss Prevention API to deidentify sensitive data in a
     string by masking it with a character.
     Args:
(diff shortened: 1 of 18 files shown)


kurtisvg:
Error:
```
==================================== ERRORS ====================================
__________________ ERROR at teardown of test_inspect_bigquery __________________
  File "/tmpfs/src/github/python-docs-samples/dlp/inspect_content_test.py", line 65, in bucket
    blob.delete()
  File "/tmpfs/src/github/python-docs-samples/.nox/py36-3-6-sample-dlp/lib/python3.6/site-packages/google/cloud/storage/blob.py", line 511, in delete
    self.name, client=client, generation=self.generation
  File "/tmpfs/src/github/python-docs-samples/.nox/py36-3-6-sample-dlp/lib/python3.6/site-packages/google/cloud/storage/bucket.py", line 1002, in delete_blob
    _target_object=None,
  File "/tmpfs/src/github/python-docs-samples/.nox/py36-3-6-sample-dlp/lib/python3.6/site-packages/google/cloud/_http.py", line 393, in api_request
    raise exceptions.from_http_response(response)
google.api_core.exceptions.NotFound: 404 DELETE https://www.googleapis.com/storage/v1/b/python-docs-samples-tests-dlp-python-client-test/o/test.txt?generation=1575313792548155: No such object: python-docs-samples-tests-dlp-python-client-test/test.txt
```

kurtisvg:
Looking at the source, I don't think it is related to the `test_inspect_bigquery`. This just happens to be the last test in the file, and so the tear-down for the bucket fixture is being triggered then. 

My guess is that this is a race condition caused by running the dlp tests in multiple places in the same project - you can probably avoid it by adding a try/except for `google.api_core.exceptions.NotFound: 404 DELETE` around the `blob.delete()` on L65.

maximus12793:
That does not seem to be the issue. But I have added the try/except around the blob delete. The tests seem to show that some of the pub/sub topic_ids do not exist. Also its not clear why these other tests are running given that they do not have dependencies on dlp. This makes the check in process pretty difficult as 3-5 seem to be failing at any given point in time.

kurtisvg:
You can ignore the other tests for now - there's some churn we are working on sorting out.

kurtisvg:
Opened #2585 for tracking tests.

kurtisvg:
Overriding merge - tests are flaky and being tracked separate from this version bump

Issue #2564, opened 2019-11-20
Add a run test flag, to re-run samples.


nnegrey:
Sometimes there is a problem with the resources used rather than the code itself, rather than making up a change to get tests to run again, adding a run flag would be nice.

leahecole:
I think there is a kokoro force run flag, but I'm not sure if it works as expected yet. @kurtisvg do you know?

kurtisvg:
So for Kokoro tests, you can use `kokoro: run` or `kokoro: force-run` labels. This doesn't apply to Travis though.

busunkim96:
I've seen Ben push empty commits to retrigger CI for Node repositories.
```
git commit --allow-empty -m "trigger ci"
git push
```

I believe if you have write permissions you can also re-run Travis builds using the UI.

kurtisvg:
+1 to using the UI - that's how I typically trigger rebuilds with travis.

kurtisvg:
Marking as fixed - the required tests are now on Kokoro.

Issue #2563, opened 2019-11-20
Run all tests not just a subset


nnegrey:
I noticed on a few of my PRs, it was only running a subset of the tests and not all of them. 
While I've run them all locally there could be something I have set correctly here but not for everyone else that would be good to catch.

leahecole:
Is this in direct conflict with #2559 ?

nnegrey:
Yes and no, like I want all the test files I change to run. 
Like say it found 47 tests, it runs like 19 of them.

leahecole:
Is "it" in this case travis?

nnegrey:
Yea, so moving to Kokoro might solve this anyway.

busunkim96:
The logic to determine if something's "changed" might be borked then. :thinking: Good to know.

nnegrey:
Maybe?
Like here is the output: `collected 45 items / 26 deselected / 19 selected   `

kurtisvg:
New tests should run py.test in each folder, so this should be covered.

Issue #2559, opened 2019-11-19
Run tests only for APIs with changes


busunkim96:
Opening a PR should only result in test runs for APIs with changes.

google-cloud-python has a script to do something similar.

busunkim96:
On a second pass, it seems the noxfile has some logic to do this, but it doesn't seem to be working correctly in practice.

```py
def _get_changed_files():
    """Returns a list of files changed for this pull request / push.

    If running on a public CI like Travis or Circle this is used to only
    run tests/lint for changed files.
    """
    if not ci_diff_helper:
        return None


    try:
        config = ci_diff_helper.get_config()
    except OSError:  # Not on CI.
        return None


    changed_files = ci_diff_helper.get_changed_files('HEAD', config.base)


    changed_files = set([
        './{}'.format(filename) for filename in changed_files])


    return changed_files
```

kurtisvg:
Fixed in #2582

PR #2582, opened 2019-12-03
Consolidate Kokoro tests by python language. 

kurtisvg:
Requires #2581 to be merged first.


kurtisvg:
Fixes #2564 - all tests will be on Kokoro, so we can use the labels to trigger
Fixes #2563 - should run all python tests in the project

kurtisvg:
Fixes #2559 - presubmit tests will only run on projects containing changes

PR #2587, opened 2019-12-06
chore: run black on all files

busunkim96:
The result of `black .` run from the root of the repository.

Black had trouble with one file `composer/data/python2_script.py` because of a Python 2 style print statement. I manually changed it (also noted below).


composer/data/python2_script.py
```diff
@@ -12,4 +12,4 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-print 'Output from Python 2.'  # noqa
+print("Output from Python 2.")  # noqa
```
busunkim96: The only manual change in this PR.

gguuss: I believe we prefer single-quotes to double-quotes.

busunkim96:
Would it be possible to revisit the standards in the authoring guide?

Black doesn't allow for the single/double quotes to be configured (see https://github.com/psf/black#strings). I *think* it will make it easier in the long run for folks to be able to run a code formatter to achieve a consistent style.

gguuss:
> Would it be possible to revisit the standards in the authoring guide?
> 
> Black doesn't allow for the single/double quotes to be configured (see [psf/black#strings](https://github.com/psf/black#strings)). I _think_ it will make it easier in the long run for folks to be able to run a code formatter to achieve a consistent style.

K.

[YAPF](https://github.com/google/yapf/) is another code formatter that teams at Google use and that is recommended in the [Google Python style guide](https://google.github.io/styleguide/pyguide.html).

kurtisvg:
> > Would it be possible to revisit the standards in the authoring guide?
> > Black doesn't allow for the single/double quotes to be configured (see [psf/black#strings](https://github.com/psf/black#strings)). I _think_ it will make it easier in the long run for folks to be able to run a code formatter to achieve a consistent style.
> 
> K.
> 
> [YAPF](https://github.com/google/yapf/) is another code formatter that teams at Google use and that is recommended in the [Google Python style guide](https://google.github.io/styleguide/pyguide.html).

I think we defaulted to black because it is what the client libraries team has already established using. Perhaps we need to resist this in the next meeting.

busunkim96:
:+1: It sounds like this needs additional discussion. I'll leave this open for now.

busunkim96:
Closing as we've decided that black will be an optional tool contributors can use to format their code.

PR #2593, opened 2019-12-10
Remove black from noxfile-template

busunkim96:
Removing black from lint check until we've had more discussion on what we'd like our style guide to be.


gguuss:
LGTM, thank you!

busunkim96:
Squashing and merging as Leah is traveling all day and Kurtis is OOO. :grimacing: We should revisit permissions on this repo.

PR #2481, opened 2019-10-17
Automl refactor

nnegrey:
Ditch this PR: https://github.com/GoogleCloudPlatform/python-docs-samples/pull/2470
So that the samples go in the right place from the get-go.

Once merged, I will update docs and delete the branches

PR #2594, opened 2019-12-10
Add filter defaults for language/automl

texasmichelle:
Add filter defaults for list_datasets and list_models
Remove unnecessary filter from list_model_evaluations and
display_evaluation

b/130171300

/cc @andrewferlitsch


sirtorry:
should we be adding this to https://github.com/GoogleCloudPlatform/python-docs-samples/pull/2481?

PR #2602, opened 2019-12-11
[Storage] Split samples

frankyn:
Split Storage samples for usability which will include required imports to run samples.
I did not split sample tests.


frankyn:
Thank you!!

PR #2604, opened 2019-12-12
Pub/Sub: remove infinite while loops in subscriber examples

anguillanneuf:
The infinite while loops in the subscriber examples suppress errors such as `403 User not authorized to perform this action`. This PR 1). updates all `subscriber.subscribe(...)` by explicitly resolving the streaming pull future:

```python
streaming_pull_future = subscriber.subscribe(...)
streaming_pull_future.result()
```

2). adds teardown code after `yield` for test fixtures.

### DIFF
--- a/pubsub/cloud-client/publisher_test.py
+++ b/pubsub/cloud-client/publisher_test.py
@@ -25,7 +25,8 @@ import publisher
 
 UUID = uuid.uuid4().hex
 PROJECT = os.environ["GCLOUD_PROJECT"]
-TOPIC = "publisher-test-topic-" + UUID
+TOPIC_ONE = "publisher-test-topic-one-" + UUID
+TOPIC_TWO = "publisher-test-topic-two-" + UUID
 
 
 @pytest.fixture
@@ -34,8 +35,8 @@ def client():
 
 
 @pytest.fixture
-def topic(client):
-    topic_path = client.topic_path(PROJECT, TOPIC)
+def topic_one(client):
+    topic_path = client.topic_path(PROJECT, TOPIC_ONE)
 
     try:
         response = client.get_topic(topic_path)
@@ -43,6 +44,20 @@ def topic(client):
         response = client.create_topic(topic_path)
 
     yield response.name
+
+
+@pytest.fixture
+def topic_two(client):
+    topic_path = client.topic_path(PROJECT, TOPIC_TWO)
+
+    try:
+        response = client.get_topic(topic_path)
+    except:  # noqa
+        response = client.create_topic(topic_path)
+
+    yield response.name
+
+    client.delete_topic(response.name)
 
 
 def _make_sleep_patch():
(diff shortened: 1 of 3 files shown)


pubsub/cloud-client/publisher_test.py
```diff
@@ -45,6 +46,20 @@ def topic(client):
     yield response.name
```
davidcavazos: How about adding a TODO here to delete the topic afterwards. It would be nice to have an explanation of why it's not possible at the moment.

pubsub/cloud-client/publisher_test.py
```diff
@@ -34,8 +35,8 @@ def client():
 
 
 @pytest.fixture
-def topic(client):
-    topic_path = client.topic_path(PROJECT, TOPIC)
+def topic_one(client):
+    topic_path = client.topic_path(PROJECT, TOPIC_ONE)
 
     try:
         response = client.get_topic(topic_path)
```
davidcavazos: Nit: maybe change `response` to `topic` to make the name more explicit?

pubsub/cloud-client/publisher_test.py
```diff
@@ -45,6 +46,20 @@ def topic(client):
     yield response.name
 
 
+@pytest.fixture
+def topic_two(client):
```
davidcavazos: When is `topic_one` used, and when is `topic_two` used? Can we change their names to something that gives us a better idea of what they're for?

pubsub/cloud-client/subscriber.py
```diff
@@ -161,16 +161,16 @@ def update_subscription(project_id, subscription_name, endpoint):
...
     """Receives messages from a pull subscription."""
     # [START pubsub_subscriber_async_pull]
     # [START pubsub_quickstart_subscriber]
-    import time
-
     from google.cloud import pubsub_v1
 
     # TODO project_id = "Your Google Cloud Project ID"
     # TODO subscription_name = "Your Pub/Sub subscription name"
+    # TODO timeout = "How long the subscriber should listen for messages in
```
davidcavazos: The idea of these TODOs is to give users a concrete example of how these variables should look like if they uncomment them. I personally omit the `TODO` word, but that's more on a consistency with the product. I guess the timeout is a number, so it should be something like:

```
# TODO timeout = 5.0  # How long the subscriber should listen for messages (seconds)
```

pubsub/cloud-client/subscriber.py
```diff
@@ -183,27 +183,33 @@ def callback(message):
...
+        subscription_path, callback=callback
+    )
+    print("Listening for messages on {}..\n".format(subscription_path))
 
-    # The subscriber is non-blocking. We must keep the main thread from
-    # exiting to allow it to process messages asynchronously in the background.
-    print("Listening for messages on {}".format(subscription_path))
-    while True:
-        time.sleep(60)
+    # result() will block indefinitely if `timeout` is not set, unless
```
davidcavazos: result() *in a future* will block ...

pubsub/cloud-client/publisher_test.py
anguillanneuf: Done.

pubsub/cloud-client/publisher_test.py
anguillanneuf: Done.

pubsub/cloud-client/publisher_test.py
anguillanneuf: Done.

pubsub/cloud-client/subscriber.py
anguillanneuf: Done.

pubsub/cloud-client/subscriber.py
anguillanneuf: Done.

PR #2608, opened 2019-12-12
automl: add base dataset samples for automl ga

nnegrey:


automl/cloud-client/delete_dataset.py
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env python
```
kurtisvg: This only makes sense if you plan on executing the script - there isn't a main in this script, so this should be removed.

kurtisvg: (same for the other scripts)

automl/cloud-client/export_dataset.py
```diff
@@ -0,0 +1,40 @@
...
+
+def export_dataset(project_id, dataset_id, gcs_uri):
+    """Export a dataset."""
+    # [START automl_export_dataset]
+    from google.cloud import automl
+
+    # TODO(developer): Uncomment and set the following variables
+    # project_id = 'YOUR_PROJECT_ID'
+    # dataset_id = 'YOUR_DATASET_ID'
+    # gcs_uri = 'gs://BUCKET_ID/path_to_export/'
```
kurtisvg:
```suggestion
    # gcs_uri = 'gs://YOUR_BUCKET_ID/path/to/export/'
```

automl/cloud-client/get_dataset.py
```diff
@@ -0,0 +1,102 @@
...
+        "Text classification dataset metadata: {}".format(
+            dataset.text_classification_dataset_metadata
+        )
+    )
+    # [END automl_language_text_classification_get_dataset]
+
+    # [START automl_translate_get_dataset]
+    print("Translation dataset metadata:")
+    print(
+        u"\tsource_language_code: {}".format(
```
kurtisvg: I think we can probably drop python2 support at this point, but if not please use `from future import unicode_literals` instead.

automl/cloud-client/get_dataset_test.py
```diff
@@ -0,0 +1,28 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+
+import get_dataset
+
+PROJECT_ID = os.environ["GCLOUD_PROJECT"]
```
kurtisvg: nit: Strong preference to a wrapper to throw an error if this isn't set (like we have in Java)

automl/cloud-client/get_dataset_test.py
```diff
@@ -0,0 +1,28 @@
...
+DATASET_ID = "TEN4058147884539838464"
```
kurtisvg: nit: Strong preference that this should be an environment variables, which a wrapper explaining this resource should be created/set beforehand

automl/cloud-client/import_dataset_test.py
```diff
@@ -0,0 +1,63 @@
...
+import import_dataset
+
+PROJECT_ID = os.environ["GCLOUD_PROJECT"]
+BUCKET_ID = "{}-lcm".format(PROJECT_ID)
+DATASET_ID = "TEN4058147884539838464"
+
+
+@pytest.fixture(scope="function")
+def create_dataset():
+    from google.cloud import automl
```
kurtisvg: nit: for tests, these imports can remain at the top

automl/cloud-client/get_dataset.py
nnegrey: Can we switch to inline?
```
print(f"\tsource_language_code: {dataset.translation_dataset_metadata.source_language_code}")
```

nnegrey: errr actually if Kokoro is running 2.7 still, can't do the `f` prefix

automl/cloud-client/get_dataset_test.py
nnegrey: Agreed, though I believe the error occurs even further up.
https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/conftest.py#L20

nnegrey: Python will throw an error if it isn't set.

automl/cloud-client/get_dataset.py
kurtisvg: fstrings is different than unicode literals, but require python >=3.6

nnegrey:
@kurtisvg, these should be good to go now.

PR #2612, opened 2019-12-13
Automl ga vision classification

nnegrey:


automl/cloud-client/vision_classification_deploy_model_node_count_test.py
```diff
@@ -0,0 +1,37 @@
+# Copyright 2018 Google LLC
```
gguuss: Nit. Copyright date?

automl/cloud-client/vision_classification_predict_test.py
```diff
@@ -0,0 +1,43 @@
+# Copyright 2018 Google LLC
```
gguuss: Nit. copyright

PR #2613, opened 2019-12-13
Automl ga base samples

nnegrey:


automl/cloud-client/get_operation_status_test.py
```diff
@@ -0,0 +1,42 @@
...
+
+@pytest.fixture(scope="function")
+def get_operation_id():
+    client = automl.AutoMlClient()
+    project_location = client.location_path(PROJECT_ID, "us-central1")
+    response = client.transport._operations_client.list_operations(
+        project_location, ""
+    )
+    operation_id = ""
+    for operation in response:
```
kurtisvg: I don't follow this `for` loop. Is the goal to just get the first operation? Any reason you can't use something like:
```python
operation_id = response[0].name
```

automl/cloud-client/get_operation_status_test.py
```diff
@@ -0,0 +1,42 @@
...
+from google.cloud import automl
+import pytest
+
+import get_operation_status
+
+PROJECT_ID = os.environ["AUTOML_PROJECT_ID"]
+
+
+@pytest.fixture(scope="function")
+def get_operation_id():
```
kurtisvg: nit: Usually fixture names are just the name of the thing
```suggestion
def operation_id():
```

automl/cloud-client/get_operation_status_test.py
```diff
@@ -0,0 +1,42 @@
...
+    client = automl.AutoMlClient()
+    project_location = client.location_path(PROJECT_ID, "us-central1")
+    response = client.transport._operations_client.list_operations(
```
kurtisvg: This looks like it's doing something wrong - why is this calling an underscored method? Does the higher level client not provide a function for the endpoint?

automl/cloud-client/get_operation_status_test.py
nnegrey: I'll put a a note, but if you try to access the iterator directly you get this error:
```
TypeError: 'GRPCIterator' object is not subscriptable
```
And you can't use `next()` as it throws this error:
```
TypeError: 'GRPCIterator' object is not an iterator
```
Then I didn't see a way to access one item directly.

automl/cloud-client/get_operation_status_test.py
nnegrey: The python client library doesn't seem to give you direct access to the operations client, while Java and Node.js do:

Java:
`client.getOperationsClient()`
Node.js
`client.operationsClient`

nnegrey:
@kurtisvg, tests outside this PR are running and failing like different products (`storage` and `dialogflow`), is this behavior intended?

(I'll look at dialogflow, but mostly wondering about other tests running)

nnegrey:
dialogflow was just on the old tests so updating the branch should fix those.

nnegrey:
Well ok one test Failed: google.api_core.exceptions.ResourceExhausted: 429 There are too many import data operations are running.

nnegrey:
Alright, looks like I've fixed the tests. 
@kurtisvg, mind taking a look again?

automl/cloud-client/get_operation_status_test.py
kurtisvg: nit: Did we file a bug against the client library for this?

automl/cloud-client/get_operation_status_test.py
kurtisvg: Does `response.pages.next()` work? 

Reference: [page_iterator.pages](https://googleapis.dev/python/google-api-core/latest/page_iterator.html#google.api_core.page_iterator.GRPCIterator.pages), [page_iterator.next](https://googleapis.dev/python/google-api-core/latest/page_iterator.html#google.api_core.page_iterator.Page.next)

nnegrey: `response.pages` throws:
```
ValueError: ('Iterator has already started', <google.api_core.page_iterator.GRPCIterator object at 0x1117f2610>)
```

But I can get it to work by calling `pages` on the request first so that it doesn't start the iterator.
```
generator = client.transport._operations_client.list_operations(project_location, "").pages
page = next(generator)
operation = page.next()
yield operation.name
```

automl/cloud-client/get_operation_status_test.py
nnegrey: Oh, yes, filed now.

automl/cloud-client/get_operation_status_test.py
kurtisvg: Hmm okay, that makes 0 sense to me but that makes me less unhappy than the for loop.

PR #2616, opened 2019-12-13
Add second action in a comment to ensure correct syntax

pheonixblade9:
The actions syntax was confusing for a customer, so adding a dummy entry to make sure the syntax for usage is clearer to avoid this in the future.

### DIFF
--- a/dlp/risk.py
+++ b/dlp/risk.py
@@ -94,7 +94,10 @@ def numerical_risk_analysis(
     }
 
     # Tell the API where to send a notification when the job is complete.
-    actions = [{"pub_sub": {"topic": "{}/topics/{}".format(parent, topic_id)}}]
+    actions = [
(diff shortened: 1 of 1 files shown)


dlp/risk.py
```diff
@@ -94,7 +94,10 @@ def callback(message):
     }
 
     # Tell the API where to send a notification when the job is complete.
-    actions = [{"pub_sub": {"topic": "{}/topics/{}".format(parent, topic_id)}}]
+    actions = [
+        {'pub_sub': {'topic': '{}/topics/{}'.format(parent, topic_id)}},
+       #  {'save_findings': {'output_config': {'table': {'project_id': 'dlapi-test','dataset_id':'ste1','table_id': 'pytest1'}}}
```
busunkim96: :wave: I'd prefer this be uncommented so that it's exercised by the tests.

PR #2629, opened 2019-12-17
feat: dataproc quickstart sample added and create_cluster updated

bradmiro:
added a quickstart sample and updated create_cluster


dataproc/create_cluster.py
```diff
@@ -14,22 +14,25 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# This sample walks a user through creating a Cloud Dataproc cluster using
+# the Python client library.
+
+# [START dataproc_create_cluster]
```
andrewsg: I think the TODO(developer) uncomment and set idiom is more applicable to when the region tag is inside the function and therefore the function signature is unset. If the region tag is outside of the function, then the user will see the variable both in the function signature and in the comments. It's not a big deal but if you can think of another way to communicate that the user is responsible for filling those in (perhaps by moving the comments outside of the function definition to the top level, or something) that might help.

dataproc/quickstart/quickstart.py
```diff
@@ -0,0 +1,125 @@
...
+from google.cloud import storage
+
+
+def quickstart(project_id, region, cluster_name, job_file_path):
+    """This quickstart sample walks a user through creating a Cloud Dataproc
+       cluster, submitting a PySpark job from Google Cloud Storage to the
+       cluster, reading the output of the job and deleting the cluster, all
+       using the Python client library.
+    """
+    # TODO(developer): Uncomment and set the following variables.
```
andrewsg: Same comment here re: the TODO tag. Again, feel free to pass on these, they're not a blocker.

dataproc/quickstart/quickstart.py
```diff
@@ -0,0 +1,125 @@
...
+    # Termimal states for a job.
+    terminal_states = {
+        dataproc.types.JobStatus.ERROR,
+        dataproc.types.JobStatus.CANCELLED,
+        dataproc.types.JobStatus.DONE
+    }
+
+    # Create a timeout such that the job gets cancelled if not in a
+    # terminal state after a fixed period of time.
+    timeout = 600000
```
andrewsg: Please specify the unit here, either in a modified variable name or in a comment.

dataproc/quickstart/quickstart.py
```diff
@@ -0,0 +1,125 @@
...
+
+    print('Job {} finished with state {}:\n{}'.format(
+        job_id,
+        job_response.status.State.Name(job_response.status.state),
+        output))
+
+    # Delete the cluster once the job has terminated.
+    cluster_client.delete_cluster(project_id, region, cluster_name).result()
+
+    print('Cluster {} successfully deleted.'.format(cluster_name))
```
andrewsg: In the create_cluster_test it looked like cluster_client.delete_cluster was asynchronous and we had to wait for the operation to complete to verify success. Why not do that here? This print statement could fire and pretend to be a success even if the delete call eventually failed silently.

bradmiro: This is also an asynchronous call that blocks until completed, I just short-handed it into a single line. However, I'll split it into two lines as I did with `create_cluster` to make this a bit clearer.

andrewsg: Oh, whoops! I should have read it more carefully. But, sure, for clarity two lines here seems like a good idea.

PR #2639, opened 2019-12-19
Cloud Healthcare API beta FHIR updates

noerog:
Split create_resource into 3 components and update region tags/tests:

create_patient
create_encounter
create_observation

In tests, create a patient fixture to use in different test methods.

Update README.

Add file and tests for executeBundle.

Add conditions for each conditional method (conditionalUpdate, conditionalPatch, conditionalDelete)


noerog:
Can this be merged if all tests except 3.7 are passing?

gguuss:
@noerog Is there a reason you don't want / need Python 3.7 support?

noerog:
Not really, but if it passes on our main/required tests I'd rather submit the PR and fix 3.7 later because the test failure makes no sense to me right now.

gguuss:
I can take a look to see if there's something you can do to fix 3.7. I'd prefer to avoid committing something that breaks on the latest Python.

gguuss:
Looks like a race condition?   I'll see if re-running gets 'em green.

gguuss:
It appears that in Python 3.x there's a problem, I'll investigate further.

gguuss:
@crwilcox Any ideas on why these tests are passing locally but failing on CI?

noerog:
Just an fyi: now 3.7, which was failing a week ago, is passing, but 3.5 is failing, even though no code has changed.

leahecole:
taking a look

healthcare/api-client/fhir/fhir_resources_test.py
```diff
@@ -28,6 +28,7 @@
 project_id = os.environ["GOOGLE_CLOUD_PROJECT"]
```
leahecole: nit - GCLOUD_PROJECT is preferred, afaik

healthcare/api-client/fhir/fhir_resources_test.py
```diff
@@ -28,6 +28,7 @@
...
 service_account_json = os.environ["GOOGLE_APPLICATION_CREDENTIALS"]
 
+bundle = os.path.join(os.path.dirname(__file__), "resources/execute_bundle.json")
 dataset_id = "test_dataset_{}".format(int(time.time()))
```
leahecole: I'd recommend using the UUID module instead - I think it's possible that the resources run by the tests in parallel are created at the same time and are colliding and that's what's causing various 404s

healthcare/api-client/fhir/fhir_resources_test.py
leahecole: nit: GCLOUD_PROJECT is preferred, afaik

healthcare/api-client/fhir/fhir_resources_test.py
```diff
@@ -53,36 +54,55 @@ def test_fhir_store():
...
         project_id,
         cloud_region,
         dataset_id,
         fhir_store_id,
         resource_type,
+        patient_resource_id,
     )
 
+
+def test_CRUD_patient(test_dataset, test_fhir_store, capsys):
```
leahecole: This should be separated into four tests - one for each CRUD operation

healthcare/api-client/fhir/fhir_resources_test.py
```diff
@@ -102,40 +122,59 @@ def test_CRUD_search_resource(test_dataset, test_fhir_store, capsys):
...
+    assert "Got Patient resource" in out
+    # update_resource test
+    assert "Updated Patient resource" in out
+    # patch_resource test
+    assert "Patched Patient resource" in out
+    # delete_resource test
+    assert "Deleted Patient resource" in out
+
+
+def test_resource_versions(test_dataset, test_fhir_store, test_patient, capsys):
```
leahecole: I'd recommend splitting this up into multiple tests as well

healthcare/api-client/fhir/fhir_resources_test.py
```diff
@@ -169,58 +208,89 @@ def test_CRUD_search_resource(test_dataset, test_fhir_store, capsys):
...
+
+    # list_resource_history test
+    assert "History for Patient resource" in out
+    # get_resource_history test
+    assert "Got history for Patient resource" in out
+    # delete_resource_purge test
+    assert "Deleted versions of Patient resource" in out
+
+
+def test_conditions_resource(test_dataset, test_fhir_store, test_patient, capsys):
```
leahecole: aaaand here as well (splitting up into multiple)

healthcare/api-client/fhir/fhir_resources_test.py
```diff
@@ -169,58 +208,89 @@ def test_CRUD_search_resource(test_dataset, test_fhir_store, capsys):
...
+    print(out)
+    # conditional_patch_resource test
+    assert "Conditionally patched" in out
+    # conditional_update_resource test
+    assert "Conditionally updated" in out
+    # conditional_delete_resource test
+    assert "Conditionally deleted" in out
+
 
+def test_search_resources(test_dataset, test_fhir_store, test_patient, capsys):
```
leahecole: and here

healthcare/api-client/fhir/fhir_resources_test.py
noerog: Good point, thanks. Done.

healthcare/api-client/fhir/fhir_resources_test.py
noerog: Done.

healthcare/api-client/fhir/fhir_resources_test.py
noerog: I would prefer to keep them in one because they all rely on each other (update has to happen before list which has to happen before get) rather than creating more fixtures. Does that sound OK?

healthcare/api-client/fhir/fhir_resources_test.py
noerog: Done.

healthcare/api-client/fhir/fhir_resources_test.py
noerog: Done.

healthcare/api-client/fhir/fhir_resources_test.py
noerog: Done.

noerog:
I suspect the conditional update/patch errors were caused by there being duplicate FHIR resources. According to the FHIR spec for conditional update and patch, if there are multiple versions of a resource that are detected, a 412 error is returned: https://www.hl7.org/fhir/http.html#cond-update / https://www.hl7.org/fhir/http.html#patch

noerog:
Healthcare tests are all passing but it seems that /automl/ tests are failing which causes the whole suite to fail.

healthcare/api-client/fhir/fhir_resources_test.py
```diff
@@ -91,124 +116,287 @@ def test_CRUD_search_resource(test_dataset, test_fhir_store, capsys):
...
 
-    fhir_resources.update_resource(
+    out, _ = capsys.readouterr()
+
+    print(out)
+
+    assert "Got Patient resource" in out
+
+
+@pytest.mark.skip(reason='flaky test sometimes returns 403 errors, need to investigate')
```
leahecole: I'm cool with a skip as long as an issue gets opened up in the repo referencing this

healthcare/api-client/fhir/fhir_resources_test.py
```diff
@@ -91,124 +116,287 @@ def test_CRUD_search_resource(test_dataset, test_fhir_store, capsys):
...
+def test_conditional_patch_resource(test_dataset, test_fhir_store, test_patient, capsys):
+    # The conditional method tests use an Observation, so we have to create an
```
leahecole: these comments are informative - thank you!

healthcare/api-client/fhir/fhir_resources_test.py
```diff
@@ -91,124 +116,287 @@ def test_CRUD_search_resource(test_dataset, test_fhir_store, capsys):
...
+    fhir_resources.conditional_patch_resource(
+        service_account_json,
+        base_url,
+        project_id,
+        cloud_region,
+        dataset_id,
+        fhir_store_id,
+    )
+
+    # In accordance with the FHIR spec, if conditional patch or conditional update
```
leahecole: 👏 again with the helpful comments 🎉

healthcare/api-client/fhir/fhir_resources_test.py
```diff
@@ -91,124 +116,287 @@ def test_CRUD_search_resource(test_dataset, test_fhir_store, capsys):
...
+    )
+
+    out, _ = capsys.readouterr()
+
+    print(out)
+
+    assert "Conditionally patched" in out
+
+
+@pytest.mark.skip(reason='flaky test sometimes returns 412 errors, need to investigate')
```
leahecole: same as above - want to see an issue opened for this investigation (ideally with you assigned to this and the other :) )

PR #2641, opened 2019-12-21
Update datalabeling samples to use test endpoint.

kurtisvg:
Update datalabeling samples to use test endpoint. These samples trigger real human action, so we want to use the test endpoint to avoid that from happening.

### DIFF
--- a/datalabeling/create_annotation_spec_set_test.py
+++ b/datalabeling/create_annotation_spec_set_test.py
@@ -18,6 +18,7 @@ import os
 
 import create_annotation_spec_set
 from google.cloud import datalabeling_v1beta1 as datalabeling
+from google.api_core.client_options import ClientOptions
 import pytest
 
 PROJECT_ID = os.getenv('GCLOUD_PROJECT')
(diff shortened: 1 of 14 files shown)


datalabeling/create_annotation_spec_set_test.py
```diff
@@ -33,4 +34,11 @@ def test_create_annotation_spec_set(capsys):
     # Delete the created annotation spec set.
     annotation_spec_set_name = response.name
     client = datalabeling.DataLabelingServiceClient()
+
+    # If provided, use a provided test endpoint - this will prevent tests on
+    # this snippet from triggering any action by a real human
+    if 'DATALABELING_ENDPOINT' in os.environ:
+        opts = ClientOptions(api_endpoint=os.getenv('DATALABELING_ENDPOINT'))
+        client = datalabeling.DataLabelingServiceClient(client_options=opts)
```
crwilcox: would it make sense to move the other client creation to an else so we aren't creating a client twice?

kurtisvg: Normally yes, but in this case we want the part that is shown on the docs to look normal (which is why we hide it in the region tags)

PR #2659, opened 2019-12-24
Update CODEOWNERS

busunkim96:
@GoogleCloudPlatform/python-samples-owners Could you do a first pass through this list? After that I will open it up to individuals listed below.

This was roughly my thought process:
Inspect the recent history:
* Is there a DPE who has made product-specific updates in the last year?
  * YES: Assign as owner.
  * NO: Is there a Tech Writer who has made product-specific sample updates in the last year?
     * YES: Assign as owner.
     * NO: Leave unassigned.


.github/CODEOWNERS
```diff
@@ -1,27 +1,64 @@
...
-
-# Torry Yang is the primary maintainer of the Tables samples.
-/tables/                  @sirtorry
+*                       @GoogleCloudPlatform/python-samples-owners
 
-# Tianzi Cai is the primary maintainer of the Pub/Sub samples.
-/pubsub/                  @anguillanneuf
+/auth/                          @busunkim96
+/automl/                        @nnegrey
+/bigquery/                      @tswast
```
gguuss: Maybe replace with Seth Hollyman as Tim is moving on to another team.

.github/CODEOWNERS
```diff
@@ -1,27 +1,64 @@
...
+/kubernetes_engine/     
+/language/                      @nnegrey
+/logging/           
+/memorystore/
+/ml_engine/                     @alecglassford      
+/monitoring/            
+/notebooks/                     @alixhami
+/opencensus/
+/profiler/                      @jqll
+/pubsub/                        @anguillanneuf
```
gguuss: If we support multiple owners, maybe add hongalex here?

.github/CODEOWNERS
```diff
@@ -1,27 +1,64 @@
...
+/iot/                           @gguuss
+/jobs/
+/kms/                           @DanSanche
+/kubernetes_engine/     
+/language/                      @nnegrey
+/logging/           
+/memorystore/
+/ml_engine/                     @alecglassford      
+/monitoring/            
+/notebooks/                     @alixhami
```
gguuss: Torry also has contributed here.

.github/CODEOWNERS
```diff
@@ -1,27 +1,64 @@
...
-
-# Gus Class is the primary maintainer of the IoT samples.
-iot/*                     @gguuss
-
-# Alix Hamilton is the primary maintainer of the Jupyter notebook samples.
-notebooks/*               @alixhami
-
-# Torry Yang is the primary maintainer of the Tables samples.
-/tables/                  @sirtorry
+*                       @GoogleCloudPlatform/python-samples-owners
```
kurtisvg: One issue we'll have to resolve at some point is how to handle reviewers from two different sets we discussed at the last meeting:

> ```
> # Order is important; the last matching pattern takes the most
> # precedence. When someone opens a pull request that only
> # modifies JS files, only @js-owner and not the global
> # owner(s) will be requested for a review.
> *.js    @js-owner
> ```

So this means that currently any changes to just the samples won't trigger/require a review from the "python samples readability" set. 

One work around to this might be to d something like this for each entry:
```
/cloud-sql/                     @kvg @GoogleCloudPlatform/python-samples-owners
```

.github/CODEOWNERS
```diff
@@ -1,27 +1,64 @@
...
 
-# Tianzi Cai is the primary maintainer of the Pub/Sub samples.
-/pubsub/                  @anguillanneuf
+/auth/                          @busunkim96
+/automl/                        @nnegrey
+/bigquery/                      @tswast
+/bigquery_storage/              @tswast
+/bigtable/                      @billyjacobson
+/blog/
```
kurtisvg: Should we comment out folders that don't have specific owners? I'm not sure if leaving it blank explicitly changes it to "no-owner"

.github/CODEOWNERS
```diff
@@ -1,27 +1,64 @@
...
+/run/                           @averikitsch
+/scheduler/                     @averikitsch        
+/spanner/                       @mmmarklu
+/speech/                        @nnegrey
+/storage/                       @crwilcox
+/tables/automl/                 @sirtorry
+/tasks/                         @averikitsch
+/texttospeech/                  @nnegrey   
+/third_party/apache-airflow/    @leahecole
+/trace/                         @engelke
```
leahecole: We're going to have some incoming trace samples with #2675 - the team has given me dukeng and ziweizhao as the github usernames of folks who could be these codeowners

.github/CODEOWNERS
busunkim96: Good catch, that seems to be the case, I've added python-samples-owners as an owner for every folder, so there are no longer blank folders.

.github/CODEOWNERS
busunkim96: Done

.github/CODEOWNERS
busunkim96: Done

.github/CODEOWNERS
busunkim96: Added

.github/CODEOWNERS
busunkim96: Added

.github/CODEOWNERS
busunkim96: Replaced mentions of tswast with shollyman

PR #2660, opened 2019-12-26
Update lazy globals snippet to use an LRU cache

orf:
This Python snippet in the docs seems rather unpythonic and generally strange. One of the better ways of doing this lazy-evaluation of potentially expensive functions is to use [functools.lru_cache](https://docs.python.org/3/library/functools.html#functools.lru_cache). If the function accepts no arguments then it's equivalent to:

```
some_global = None

def my_cached_function():
    if some_global is None:
        some_global = expensive_function()
    return some_global
```

Using an LRU cache here is probably a bit better than advocating using a bunch of global variables and conditionally initializing them.


functions/tips/main.py
```diff
@@ -90,11 +90,15 @@ def scope_demo(request):
...
 # [START run_tips_global_lazy]
+from functools import lru_cache
+
 # Always initialized (at cold-start)
 non_lazy_global = file_wide_computation()
 
-# Declared at cold-start, but only initialized if/when the function executes
-lazy_global = None
+# Only initialized if/when the function executes
+@lru_cache()
```
tmatsuo, line 99: About the comment:

How about?

```
# The return value of this function will be cached for future invocation. 
```

functions/tips/main.py
```diff
@@ -90,11 +90,15 @@ def scope_demo(request):
...
+def lazy_computation():
+    return file_wide_computation()
```
tmatsuo, line 101: It seems that originally it was `function_specific_computation()`?

functions/tips/main.py
```diff
@@ -108,13 +112,11 @@ def lazy_globals(request):
         Response object using `make_response`
         <http://flask.pocoo.org/docs/1.0/api/#flask.Flask.make_response>.
     """
-    global lazy_global, non_lazy_global
-
+    
     # This value is initialized only if (and when) the function is called
-    if not lazy_global:
-        lazy_global = function_specific_computation()
+    lazy_value = lazy_computation()
```
tmatsuo, line 117: Do you need to have this local variable?

functions/tips/main.py
```diff
@@ -108,13 +112,11 @@ def lazy_globals(request):
...
 
-    return 'Lazy: {}, non-lazy: {}.'.format(lazy_global, non_lazy_global)
+    return 'Lazy: {}, non-lazy: {}.'.format(lazy_value, non_lazy_global)
```
tmatsuo, line 119: Can't you just do:

```python
# The return value of this function will be cached for future invocation.
@lru_cache
def function_specific_computation():
    return 1
# ...
# ...
# and then in this function:
    return 'Lazy: {}, non-lazy: {}.'.format(function_specific_computation(), non_lazy_global)
```

functions/tips/main.py
```diff
@@ -90,11 +90,15 @@ def scope_demo(request):
...
 
-# Declared at cold-start, but only initialized if/when the function executes
-lazy_global = None
+# Only initialized if/when the function executes
+@lru_cache()
+def lazy_computation():
+    return file_wide_computation()
 
 
 def lazy_globals(request):
```
tmatsuo, line 104: I think the name of the function doesn't reflect what it is once we merge this PR.

How about `global_and_cached_value` or something?

tmatsuo:
No activities for a while, closing. Feel free to re-open.

PR #2664, opened 2019-12-27
Add initial Secret Manager samples

sethvargo:
Add initial Secret Manager samples


sethvargo:
Can someone please enable the Secret Manager API on the test project(s):

```
gcloud services enable secretmanager.googleapis.com
```

kurtisvg:
> Can someone please enable the Secret Manager API on the test project(s):
> 
> ```
> gcloud services enable secretmanager.googleapis.com
> ```

👍

secretmanager/api-client/access_secret_version.py
```diff
@@ -0,0 +1,61 @@
+#!/usr/bin/env python
+
+# Copyright 2019 Google, Inc
```
kurtisvg:
```suggestion
# Copyright 2019 Google LLC
```

(same elsewhere)

secretmanager/api-client/create_secret.py
```diff
@@ -0,0 +1,61 @@
...
+    # Create the Secret Manager client.
+    client = secretmanager.SecretManagerServiceClient()
+
+    # Build the resource name of the parent project.
+    parent = client.project_path(project_id)
+
+    # Create the secret.
+    response = client.create_secret(parent, secret_id, {
+        'replication': {
+            'automatic': {},
```
kurtisvg: Is this supposed to be an empty dict?

secretmanager/api-client/quickstart_test.py
```diff
@@ -0,0 +1,21 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+
+# import os
+
+# import quickstart
+
+
+# def test_quickstart():
+#   os.environ['SECRETMANAGER_SECRET_ID'] = 'todo'
```
kurtisvg: Should this test still be disabled?

secretmanager/api-client/requirements.txt
```diff
@@ -0,0 +1 @@
+google-cloud-secret-manager>=0.1
```
kurtisvg: Please pin a specific version

secretmanager/api-client/update_secret.py
```diff
@@ -0,0 +1,54 @@
...
+    name = client.secret_path(project_id, secret_id)
+
+    # Update the secret.
+    secret = {'name': name, 'labels': {'secretmanager': 'rocks'}}
+    update_mask = {'paths': ['labels']}
+    response = client.update_secret(secret, update_mask)
+
+    # Print the new secret name.
+    print('Updated secret: {}'.format(response.name))
+# [END secretmanager_update_secret]
```
kurtisvg: nit: Indentation

secretmanager/api-client/access_secret_version.py
sethvargo: Hmm - both storage and kms (which is where I copied the license headers from) use Inc. Is this a recent change to use LLC? Should the canonical example (storage) be updated?

secretmanager/api-client/create_secret.py
sethvargo: Yes. The API is a bit weird here because replication is a oneOf automatic or user_managed. user_managed has options where automatic does not.

secretmanager/api-client/access_secret_version.py
kurtisvg: IIRC the change occurred in 2018, but it's only applied to new files. 

We do have some new canonical and updates to the contributing guide in the pipeline.

Issue #1788, opened 2018-10-24
[FIrestore Samples] Update tests to be self-contained and verified for correctness.


frankyn:
## In which file did you encounter the issue?

https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/firestore/cloud-client/snippets_test.py

## Describe the issue

Tests aren't self contained and also not being verified for correctness. At the moment, tests only run the samples and verify that they execute without failure. Although, tests do provide some level of support using this method, they aren't verifying state correctness of Firestore post after testing a given sample.

gguuss:
This may be fixed in [the updated tests](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/firestore/cloud-client/snippets_test.py) from @BenWhitehead.

BenWhitehead:
There are a few older tests that need to have assertions added to them.

crwilcox:
@BenWhitehead do we have the individual tests we should fix?

BenWhitehead:
@crwilcox I think with your updated bootstrapping for the tests that should hopefully transparently take care of this.

@frankyn Do you have specific tests that we should double check to make sure the fix from @crwilcox addresses things?

andrewferlitsch:
@crwilcox PTAL

frankyn:
Apologies, I missed this in email.

The tests are not validating correctness and only making sure samples execute without failures.

Here's an example of a test being verified: https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/firestore/cloud-client/snippets_test.py#L74-L80

Here's an example of test not being verified: https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/firestore/cloud-client/snippets_test.py#L70-L71

leahecole:
I've been in touch with @BenWhitehead recently, and it is still in the pipeline.

meredithslota:
A partial fix was implemented in https://github.com/GoogleCloudPlatform/python-docs-samples/pull/2667 but work remains (see last comment from @frankyn).

PR #2667, opened 2019-12-30
fix: Improve Firestore tests by separating test runs, improving cleanup

### LINKED ISSUE #2529: firestore: `listen_multiple` test is flakey
https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/firestore/cloud-client/snippets.py#L679

no changes

## Describe the issue

The `listen_multiple` firestore sample is flakey, likely due to not waiting appropriately for all of the async events to happen before the assertions.
To fix this flakeyness, someone with good python multithreading knowledge will be needed to add the correct await conditions.

crwilcox:
Fixes #2529, Possibly #1788 

- Add test to cleanup distributed counters
- Separate Collections for each test run: Add a monkeypatched version of the Firestore Client so we can add a unique string to each collection name.
- Use stream instead of get. Get is deprecated in the API
- Moved all delete scenarios to end of testing


leahecole:
@crwilcox it looks like SF appears sometimes, but also sometimes there are additional items in the set. It seems like these tests may be getting in the way of each other - is there a way to make them independent?

crwilcox:
Currently looking at some of the other tests as well. They are also flaky. Will update this once I have hardened them all

crwilcox:
@frankyn #1788 may also be resolved by this work. I have made each run create new resources to there are no longer collisions between runs.

crwilcox:
🎉

PR #2671, opened 2019-12-30
fix: translate test

### LINKED ISSUE #2670: translate tests failing
looks like the translate model was updated in the back end.

leahecole:
Fixes #2670


leahecole:
Given that I didn't change the thing that's failing Python 2, I'm guessing it's been a problem there for awhile. Given Python 2 EOL, are we cool with merging it anyway @busunkim96 ? If not, I can figure out a creative workaround for this non ASCII letter.

busunkim96:
I'd prefer the samples and tests work with 2.7 until we explicitly drop 2.7 support in the latest library version.

It looks like the encoding needs to be declared at the top of the file:

https://github.com/GoogleCloudPlatform/python-docs-samples/blob/7907b968f12190ecbbc8b5dda6b6c71643a23081/translate/cloud-client/quickstart_test.py#L1-L14

PR #2687, opened 2020-01-07
Add voice selection by name

texasmichelle:
Add name parameter based on user request.

b/143877133


busunkim96:
It looks like 2.7 is failing because `html` doesn't exist in the standard library for Python 2.7 https://docs.python.org/3/library/html.html


```
______ ERROR collecting texttospeech/cloud-client/ssml_addresses_test.py _______
ImportError while importing test module '/tmpfs/src/github/python-docs-samples/texttospeech/cloud-client/ssml_addresses_test.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
ssml_addresses_test.py:15: in <module>
    from ssml_addresses import text_to_ssml
ssml_addresses.py:19: in <module>
    import html
E   ImportError: No module named html
- generated xml file: /tmpfs/src/github/python-docs-samples/texttospeech/cloud-client/sponge_log.xml -
```

texasmichelle:
If these tests are running repo-wide, would it make sense to add `future` to testing/requirements.txt instead?

texasmichelle:
Per @leahecole, best to keep `future` in this product area for now since the issue hasn't cropped up in other products and py2 will be disappearing eventually.

PR #2690, opened 2020-01-07
fix continuous lint

leahecole:


leahecole:
For anyone who is looking in hindsight, this sample is for the python 3.7 runtime explicitly, so we're ok with it failing 3.5 checks.

PR #2692, opened 2020-01-08
Cleanup bigtable python examples

billyjacobson:
Use new row types for mutations
Update bigtable version in requirements
Delete table after tests
Change bigtable cluster variable to bigtable instance for consistency
Create and delete quickstart table during test


bigtable/quickstart/requirements.txt
```diff
@@ -1,2 +1,2 @@
-google-cloud-bigtable==1.0.0
-google-cloud-core==1.0.3
+google-cloud-bigtable==1.2.1
+google-cloud-core==1.1.0
```
crwilcox: is the explicit dependency on core needed for something? Bigtable specifies   "google-cloud-core >= 1.0.3, < 2.0dev",

leahecole:
I made a change this morning that should fix the iam error (https://github.com/GoogleCloudPlatform/python-docs-samples/blob/4fb6a3951204464d19baf5e0ed89252a77e20f39/iam/api-client/access.py#L29) might just need to do some merging w/ upstream.

PR #2696, opened 2020-01-08
dialogflow: use unique names for all resources for tests to avoid col…

nnegrey:
…lision


nnegrey:
I think the problem is way back when these tests were written they all sorta rely on one machine running them at a time, now with 4 in parallel some of the requests are expecting to list one resource but find multiple and that is one problem. 

I'm breaking the tests up into individual files and have proper setup teardown rather than rely on them running in order which some do.

nnegrey:
Alright looks like I need to request a quota increase, it caps out at 60 requests per minute and with 4 different versions running at once that is quite soon.

dialogflow/cloud-client/create_document_test.py
```diff
@@ -0,0 +1,57 @@
...
+
+PROJECT_ID = os.getenv('GCLOUD_PROJECT')
+KNOWLEDGE_BASE_NAME = 'knowledge_' \
+                      + datetime.datetime.now().strftime("%Y%m%d%H%M%S")
+DOCUMENT_DISPLAY_NAME = 'test_document_' \
+                     + datetime.datetime.now().strftime("%Y%m%d%H%M%S")
+pytest.KNOWLEDGE_BASE_ID = None
+
+
+@pytest.fixture(scope="function", autouse=True)
```
nnegrey: @gguuss, while I wait for the quota increase. 

How do you feel about these types of tests? 

I was gonna modify the rest and separate them into individual files in different PRs, but the current Knowledge and Document tests were checking that only 1 thing existing, so I at least had to update those for the 4 python versions.

gguuss: I'm fine for checking that all the response parameters are there but adding more detailed checks, particularly in ML APIs, can make the checks more brittle.

gguuss:
@nnegrey I think you should be able to merge when green, just lmk if you need me to take a look again.

nnegrey:
Getting errors about the long running client:
```
AttributeError: module 'google.longrunning.operations_pb2' has no attribute 'OperationsStub'
```
Let me see if something changed in the client or something.

nnegrey:
I believe these tests failed cus the old tests ran over night and caused some issues cus they fail on multiple run times of python.

nnegrey:
I have a couple PRs ready to fix those oddities, but this one should go first.

PR #2697, opened 2020-01-08
Update Mac setup guide

### LINKED ISSUE #2693: Mac setup

MAC_SETUP.md

No

## Describe the issue

These instructions don't work on Catalina (10.15.2). If someone knows of a workaround, it would be great to add it here. Otherwise, we should add an acknowledgement so that users don't waste their time.

When I try to install pyenv, I get this message:
```
$ brew install pyenv
Warning: You are using macOS 10.15.
We do not provide support for this pre-release version.
You will encounter build failures with some formulae.
```

texasmichelle:
Fixes #2693

Add instructions for Catalina
Move installation verification after installation of specific versions
Update latest python versions and add commands for each
Add console label to code blocks

### DIFF
--- a/MAC_SETUP.md
+++ b/MAC_SETUP.md
@@ -94,7 +94,7 @@ test their code.
 
     ```console
     $ which python
-    /Users/tswast/.pyenv/shims/python
+    ~/.pyenv/shims/python
     ```
 
 
--- a/MAC_SETUP.md
+++ b/MAC_SETUP.md
@@ -10,15 +10,18 @@ test their code.
 
 ## Before you begin
 
-1.  [Optional] Install [homebrew](https://brew.sh/).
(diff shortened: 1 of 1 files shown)


MAC_SETUP.md
```diff
@@ -37,80 +40,85 @@ test their code.
...
-1. Compile the necessary Python versions with pyenv. Use the latest release
-   of the versions you wish to test against.  A list of available versions
-   is available on [python.org](https://www.python.org/doc/versions/)
+1.  After you have installed a python version through pyenv,
+    verify that you are now using the pyenv Python shim.
 
-    As of December 9, 2019 latest Python versions are:
+    ```console
+    $ which python
+    /Users/tswast/.pyenv/shims/python
```
gguuss: Maybe use `~/.pyenv/shims/python` to avoid having a username in here

PR #2698, opened 2020-01-09
Make texttospeech quickstart compatible with microgenerator library (DO NOT MERGE)

busunkim96:
(1) Request dictionaries are passed to functions. This was converted automatically using a script provided by the microgenerator.
(2) Types and enums are all available at `google.cloud.texttospeech`, so `google.cloud.texttospeech.enums` and `google.cloud.types` are unneeded.

### DIFF
--- a/texttospeech/cloud-client/quickstart.py
+++ b/texttospeech/cloud-client/quickstart.py
@@ -48,7 +48,7 @@ def run_quickstart():
 
     # Perform the text-to-speech request on the text input with the selected
     # voice parameters and audio file type
-    response = client.synthesize_speech(synthesis_input, voice, audio_config)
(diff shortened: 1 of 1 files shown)


texttospeech/cloud-client/quickstart.py
```diff
@@ -28,27 +28,27 @@ def run_quickstart():
     Note: ssml must be well-formed according to:
         https://www.w3.org/TR/speech-synthesis/
     """
-    from google.cloud import texttospeech
+    from google.cloud import texttospeech_v1 as texttospeech
```
busunkim96: Temporary change because enums are missing in `texttospeech`

PR #2705, opened 2020-01-10
fix: get bounds for blocks instead of pages

### LINKED ISSUE #2702: Issue in vision api document text detection example
https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/vision/cloud-client/document_text/doctext.py

No.

## Describe the issue
Here are lines 88, 89 of this file:
```python
        if (feature == FeatureType.PAGE):
            bounds.append(block.bounding_box)
```
Here, in case the feature requested is the page - the block's bounding box is added, instead of the page's. This needs to be changed to page.bounding_box.

busunkim96:
Closes #2702

### DIFF
--- a/vision/cloud-client/document_text/doctext.py
+++ b/vision/cloud-client/document_text/doctext.py
@@ -86,7 +86,7 @@ def get_document_bounds(image_file, feature):
                 bounds.append(block.bounding_box)
 
         if (feature == FeatureType.PAGE):
-            bounds.append(block.bounding_box)
+            bounds.append(page.bounding_box)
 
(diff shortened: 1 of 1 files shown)


leahecole:
lgtm once tests pass

busunkim96:
@leahecole Yep I just took another look at the API and I'm not sure this needs to be changed.

leahecole:
Seems like it doesn't, and if Noah (or another ML APIs person he nominates) takes a look and deems it fit, then we can close this and the issue :)

busunkim96:
@leahecole Would you mind taking another look?

PR #2707, opened 2020-01-10
bigtable: read and filter snippets

billyjacobson:


bigtable/snippets/filters/filter_snippets.py
```diff
@@ -0,0 +1,357 @@
...
+# [END bigtable_filters_limit_value_regex]
+# [END bigtable_filters_limit_timestamp_range]
+# [END bigtable_filters_limit_block_all]
+# [END bigtable_filters_limit_pass_all]
+# [END bigtable_filters_modify_strip_value]
+# [END bigtable_filters_modify_apply_label]
+# [END bigtable_filters_composing_chain]
+# [END bigtable_filters_composing_interleave]
+# [END bigtable_filters_composing_condition]
+from google.cloud.bigtable.row_filters import ApplyLabelFilter, \
```
crwilcox: are all of these imports needed by each sample? Is there a way to separate them so it is a smaller subset?

Also, I think it is still the plan to split snippets to be separate files. Perhaps that would make this more straightforward?

bigtable/snippets/reads/read_snippets.py
```diff
@@ -0,0 +1,177 @@
...
+# [END bigtable_reads_row]
+# [END bigtable_reads_row_partial]
+# [END bigtable_reads_rows]
+# [END bigtable_reads_row_range]
+# [END bigtable_reads_row_ranges]
+# [END bigtable_reads_prefix]
+# [END bigtable_reads_filter]
+
+# [START bigtable_reads_row]
+from google.cloud.bigtable.row_filters import ColumnQualifierRegexFilter, \
```
crwilcox: this import is included in bigtable_reads_row but doesn't seem to be used for this snippet?

bigtable/snippets/reads/read_snippets.py
```diff
@@ -0,0 +1,177 @@
...
+
+
+# [START bigtable_reads_row]
+# [START bigtable_reads_row_partial]
+# [START bigtable_reads_rows]
+# [START bigtable_reads_row_range]
+# [START bigtable_reads_row_ranges]
+# [START bigtable_reads_prefix]
+# [START bigtable_reads_filter]
+def print_row(row):
```
crwilcox: Would this function be useful in the library? I see it used across two snippets.

bigtable/snippets/filters/filters_test.py
```diff
@@ -0,0 +1,214 @@
...
+
+
+import os
+import uuid
+
+import datetime
+import pytest
+from google.cloud import bigtable
+
+from filter_snippets import filter_composing_chain, \
```
crwilcox: Typically in python, with this import count, I would see `import filter_snippets` and then use as `filter_snippets.filter_composing_chain`. This isn't a hard and fast rule though. Is there a style preference for this in samples?

bigtable/snippets/filters/filter_snippets.py
billyjacobson: Ok, so the issue I was having is that the reading and printing functions are exactly the same, so I didn't want to have that code duplicated throughout in case any of it needed to change. If there is a way to split the imports, I can do that so each snippet only gets the necessary ones.

bigtable/snippets/reads/read_snippets.py
billyjacobson: Only for running snippets like these. So perhaps, but probably not necessary.

bigtable/snippets/filters/filters_test.py
billyjacobson: I just went with this because I thought it was cleaner, I can change if you'd prefer

bigtable/snippets/filters/filter_snippets.py
crwilcox: If you want to punt this for a bit we can, but the end goal is to have 1 sample: 1 file. At that point the imports will split. For an example of of what I mean, you can look at the storage samples.

PR #2711, opened 2020-01-10
fix: Update MAC_SETUP to describe using builtin venv

crwilcox:
@billyjacobson and I discussed the venv section of this setup guide. Folks use two methods and we were only showing one. Introducing a section showing how to use `venv`


crwilcox:
Incorporated @billyjacobson suggestions.

crwilcox:
Also, realized that we have no mention of nox. So I added a small bit at the bottom with common commands

MAC_SETUP.md
```diff
@@ -127,36 +123,15 @@ source venv-name/bin/activate
 
 Typically you will name the venv `venv`, or `venv38` for a python 3.8 venv.
 
+## Nox
```
billyjacobson: This info is in contributing. I'd prefer we don't repeat this info, but just include the mac specific information and link to the contributing doc section

MAC_SETUP.md
```diff
@@ -102,22 +102,18 @@ test their code.
...
 
-In either case, you can create and source virtual environments, `python` will point to this virtual environment, not a global python install. Each virtual environment can have its own set of packages that can be different from others.
-
-
-### Option 1: Using builtin support, venv
-Python ships with support of its own to create virtual environments. You can [create a virtual environment](https://docs.python.org/3/library/venv.html).
+To create a virtual environment, run `python3 -m venv`.
 
 ```
 cd python-docs-samples
```
billyjacobson: perhaps putting venv-name in square brackets or uppercase to indicate it's something you fill in

MAC_SETUP.md
```diff
@@ -97,19 +101,46 @@ test their code.
...
+
+This will make python and python3 point to Python 3.8.1. python2 will go to 2.7.17. You can also further specify versions, such as python3.6 to use that version.
+
+## Python virtual environments
+Using Virtual Environments is fairly common across the Python community. While there is built in support in python for creating environments, there are also other tools that the community has embraced. Option 1 relies on python itself, Option 2 uses an additional tool that some users may prefer.
+
+In either case, you can create and source virtual environments, `python` will point to this virtual environment, not a global python install. Each virtual environment can have its own set of packages that can be different from others.
+
+
+### Option 1: Using builtin support, venv
```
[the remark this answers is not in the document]
kurtisvg: I disagree with this (or at least the related changes). I prefer using `pyenv-virtualenv` and think it's the simpler option to have pyenv manage my virtualenvs - especially if you are unfamiliar with the concept of virtualenvs to begin with. 

If we are going to only recommend one way I'd prefer it stay pyenv.

MAC_SETUP.md
```diff
@@ -97,35 +101,37 @@ test their code.
...
+This will make python and python3 point to Python 3.8.1. python2 will use 2.7.17. You can also further specify versions, such as python3.6 to use that version.
 
-    ```console
-    cd ~/src/python-docs-samples
-    ```
+## Python virtual environments
+Using [Virtual Environments](https://docs.python.org/3/library/venv.html) prevents inadvertent modifications to your global python install. Once created and sourced, calls to `python` will uses this virtual environment, not a global python install. Each virtual environment can have its own set of packages that can be different from others.
 
-1.  Create a virtualenv using `pyenv virtualenv`.
+To create a virtual environment, run `python3 -m venv`.
```
billyjacobson: You don't need to include this command twice. Just leave it in code block

MAC_SETUP.md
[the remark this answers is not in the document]
crwilcox: I added it back. What I did change is I didn't refer to them as options. using the built in is quicker and I put it first. I present pyenv-virtualenv as an additional option

MAC_SETUP.md
crwilcox: I didn't see this info in contributing? Maybe I am just overlooking it?

texasmichelle: It's in the Authoring Guide, which is linked to from CONTRIBUTING.

MAC_SETUP.md
```diff
@@ -97,35 +101,76 @@ test their code.
...
+## Managing python versions using Pyenv global
+Pyenv allows you to configure the priority order for your python installs.
 
-## Using pyenv and pyenv-virtualenv to manage your Python versions
+```
+pyenv global 3.8.1 3.7.6 3.6.10 3.5.9 2.7.17
+```
 
-1.  Change to the desired source directory.
+This will make python and python3 point to Python 3.8.1. python2 will use 2.7.17. You can also further specify versions, such as python3.6 to use that version.
```
texasmichelle: This is a good explanation of something that wasn't entirely clear to me before 👍

MAC_SETUP.md
```diff
@@ -97,35 +101,76 @@ test their code.
...
 
-    ```console
-    cd ~/src/python-docs-samples
-    ```
+## Python virtual environments
+Using [Virtual Environments](https://docs.python.org/3/library/venv.html) prevents inadvertent modifications to your global python install. Once created and sourced, calls to `python` will uses this virtual environment, not a global python install. Each virtual environment can have its own set of packages that can be different from others.
```
texasmichelle: s/uses/use/
Nit: max line length of 79 chars.

MAC_SETUP.md
texasmichelle: I'd also like to add a link to the first mention of nox a few lines above, since this doc is the first time I ever heard of the tool and I initially thought it was a typo for "now".

MAC_SETUP.md
```diff
@@ -97,35 +101,76 @@ test their code.
...
+    pyenv local python-docs-samples 3.8.1 3.7.6 3.6.10 3.5.9 2.7.17	
+    ```	
+
+1.  Now, when you `cd` into the source directory or a subdirectory within it,	
+    pyenv will make your virtualenv the default Python. Since you specified	
+    more than one version, it will also add binaries like `python36` and	
+    `python27` to your PATH, which nox uses when picking Python interpreters.	
+
+1.  Add `.python-version` to your [global gitignore	
+    file](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore),
```
texasmichelle: Nit: this rendered link looks a bit weird separated by a newline, although that could be my markdown rendering tool. If the line is too long, you could put the link at the bottom and reference it here with a nickname (e.g. the last line of [this markdown](https://raw.githubusercontent.com/googleapis/nodejs-translate/master/CONTRIBUTING.md)).

crwilcox: Thanks for finding that. It looked good in my previewer, but I combined links to one line just to be safe.

Issue #2704, opened 2020-01-09
In vision samples add error check, since the error isn't thrown by default


nnegrey:
PR this first came up: https://github.com/GoogleCloudPlatform/python-docs-samples/pull/2569

Perform an error check on all sample code for something like this: 
```
error {
  code: 3
  message: "The URL does not appear to be accessible by us. Please double check or download the content and pass it in."
}
```
Then republish docs

nnegrey:
Another interesting note, not all methods have an error, they just return no results.

nnegrey:
Methods that have no error:
1. object_localization - (beta and GA)

Methods that will throw an error:
1. batch_annotate_files (beta)

PR #2725, opened 2020-01-15
vision: update samples to throw errors if one occurs

nnegrey:
Context: https://github.com/GoogleCloudPlatform/python-docs-samples/issues/2704

Added errors to all snippets except those that either:
- have no error: object_localization - (beta and GA)
- that will throw an error: batch_annotate_files (beta)


nnegrey:
Yea, this is more a quick stop-gap because the client library's partial gapic layers aren't throwing the errors by default (added a bug on the library), but in the meantime we want to at least notify users if something went wrong rather than silently fail. 

I also didn't want to add a whole large error check sections to each and every sample to check all the possible error codes that the API could throw.

nnegrey:
@gguuss, worth adding a link to this page?
https://cloud.google.com/apis/design/errors

leahecole:
I'm just reassigning to gus since he has owner privileges and already was looking at this :)

gguuss:
@nnegrey Yes! That's a great idea, linking to the errors page in the thrown exception would be awesome.

PR #2728, opened 2020-01-16
re-write authoring guide, add links to other documentation

dmahugh:
Note that because many sections were added or completely re-written, the diff view is not helpful. This PR adds topics covered in the Java guide (for consistency), adds links to Google Cloud documentation where appropriate, and links to specific examples in existing samples where appropriate. Some sections have been removed to avoid duplication and simplify maintenance - for example, instead of showing the license header to use, it now links to the instructions in the LICENSE file for how to insert a license header.

Feedback welcome!

### DIFF
--- a/AUTHORING_GUIDE.md
+++ b/AUTHORING_GUIDE.md
@@ -1,7 +1,7 @@
 # Python Sample Authoring Guide
 
 We're happy you want to write a Python sample! Like a lot of Pythonistas,
-we're opinationed and fussy. This guide is a reference for the format and
+we're opinioned and fussy. This guide is a reference for the format and
 style expected of samples contributed to the
 [python-docs-samples](https://github.com/GoogleCloudPlatform/python-docs-samples)
 repo. The guidelines below are intended to ensure that all Python samples
@@ -48,12 +48,7 @@ started with using a service or API — it should be in a _quickstart_ folder.
 
 ### Python Version
 
-As of this writing (early 2020), the default Python version for samples is
-Python 3.7. All samples should pass tests on 3.7, and support for 3.8 is recommended.
-
-Support for Python 2.7 is no longer required for new samples. If you’re
-making a change to an existing sample that supports Python 2.7, the preferred
-approach is to maintain compatibility with 2.7.
+Samples should support Python 3.6, 3.7, and 3.8.
 
 If the API or service your sample works with has specific Python version
 requirements different from those mentioned above, the sample should support
--- a/AUTHORING_GUIDE.md
+++ b/AUTHORING_GUIDE.md
@@ -333,12 +333,19 @@ is expected. Strive to verify the content of the output rather than the syntax.
 For example, the test might verify that a string is included in the output,
 without taking a dependency on where that string occurs in the output.
 
-### Using nox
+### Running tests
+
 
 Automated testing for samples in `python-docs-samples` is managed by
 [nox](https://nox.readthedocs.io). Nox allows us to run a variety of tests,
 including the flake8 linter, Python 2.7, Python 3.x, and App Engine tests,
 as well as automated README generation.
+
+__Note:__ As a temporary workaround, each project currently uses first
+`noxfile-template.py` found in a parent folder above the current sample. In 
+order to simulate this locally, you need to copy + rename the parent
+`noxfile-template.py` as `noxfile.py` in the folder of the project you want to
+run tests.
 
 To use nox, install it globally with `pip`:
 
@@ -346,45 +353,26 @@ To use nox, install it globally with `pip`:
 $ pip install nox
 ```
 
-Nox automatically discovers all samples in the repository and generates
-three types of sessions for each sample:
-
-1. A test sessions (`gae`, `py27` and `py36`) for running the system tests
-against a specific Python version.
-2. `lint` sessions for running the style linter.
-3. `readmegen` sessions for regenerating READMEs.
-
-Because nox generates all of these sessions, it's often useful to filter down
-by just the sample you're working on. For example, if you just want to see
-which sessions are available for storage samples:
-
+To run style checks on your samples:
 ```console
-$ nox -k storage -l
-* gae(sample='./appengine/standard/storage/api-client')
-* gae(sample='./appengine/standard/storage/appengine-client')
-* lint(sample='./appengine/flexible/storage')
-* lint(sample='./appengine/standard/storage/api-client')
-* lint(sample='./appengine/standard/storage/appengine-client')
-* lint(sample='./storage/api')
-* lint(sample='./storage/cloud-client')
-* lint(sample='./storage/transfer_service')
-* py27(sample='./appengine/flexible/storage')
-* py27(sample='./storage/api')
-* py27(sample='./storage/cloud-client')
-* py35(sample='./appengine/flexible/storage')
-* py35(sample='./storage/api')
-* py35(sample='./storage/cloud-client')
-* readmegen(sample='./storage/api')
-* readmegen(sample='./storage/cloud-client')
-* readmegen(sample='./storage/transfer_service')
+nox -s lint
 ```
 
-Now you can use nox to run a specific session, for example, if you want to
-lint the storage cloud-client samples:
-
+To run tests with a python version, use the correct `py-3.*` sessions:
 ```console
-$ nox -s "lint(sample='./storage/cloud-client')"
+nox -s py-3.6
 ```
+
+To run a specific file:
+```console
+nox -s py-3.7 -- snippets_test.py
+```
+
+To run a specific test from a specific following:
+```console
+nox -s py-3.7 -- snippets_test.py:test_list_blobs
+```
+
 
 ### Test Environment Setup
 
--- a/AUTHORING_GUIDE.md
+++ b/AUTHORING_GUIDE.md
@@ -1,212 +1,122 @@
 # Python Sample Authoring Guide
 
-We're happy you want to write a Python sample! Like a lot of Pythonistas, we're
-opinationed and fussy. This guide intends to be a reference for the format and
-style expected of samples that live in
-[python-docs-samples](https://github.com/GoogleCloudPlatform/python-docs-samples).
+We're happy you want to write a Python sample! Like a lot of Pythonistas,
+we're opinationed and fussy. This guide is a reference for the format and
+style expected of samples contributed to the
+[python-docs-samples](https://github.com/GoogleCloudPlatform/python-docs-samples)
+repo. The guidelines below are intended to ensure that all Python samples
+meet the following goals:
 
-## Canonical sample
+* **Copy-paste-runnable.** A developer should be able to copy and paste the
+code into their own environment and run it with as few modifications as
+possible.
+* **Teach through code.** Each sample should demonstrate best practices for
+interacting with Google Cloud libraries, APIs, or services.
+* **Idiomatic.** Each sample should follow widely accepted Python best
+practices as covered below.
 
-The [Cloud Storage Samples](https://github.com/GoogleCloudPlatform/python-docs-samples/tree/master/storage/cloud-client)
-are great examples of what we expect from samples. Take a look at some of the files within to gain better understanding of the patterns we use.
+## Sample Guidelines
 
-## The basics
+This section covers guidelines for Python samples. Note that
+[Testing Guidelines](#testing-guidelines) are covered separately below.
 
-No matter what, all samples must:
+### Folder Location
 
-1. Have a license header.
-1. Pass lint.
-1. Be either a web application or a runnable console application.
-1. Have a `requirements.txt` containing all of its third-party dependencies. All
-requirements must be pinned.
-1. Work in Python 2.7, 3.6, and 3.7. App Engine Standard is exempt as it
-   only supports 2.7. Our default version is currently Python 3.6.
-1. Have tests.
+Each sample should be in a folder under the top-level folder of
+[python-docs-samples](https://github.com/GoogleCloudPlatform/python-docs-samples)
+that corresponds to the Google Cloud service or API used by the sample.
+For example, a sample demonstrating how to work with BigTable should be
+in a subfolder under the
+[python-docs-samples/bigtable](https://github.com/GoogleCloudPlatform/python-docs-samples/tree/master/bigtable)
+folder.
 
-## Style & linting
+Conceptually related samples under a service or API should be grouped into
+a subfolder. For example, App Engine Standard samples are under the
+[appengine/standard](https://github.com/GoogleCloudPlatform/python-docs-samples/tree/master/appengine/standard)
+folder, and App Engine Flex samples are under the
+[appengine/flexible](https://github.com/GoogleCloudPlatform/python-docs-samples/tree/master/appengine/flexible)
+folder.
 
-We follow [pep8](https://www.python.org/dev/peps/pep-0008/) and the
-*external* [Google Python Style
-Guide](https://google.github.io/styleguide/pyguide.html), which we verify with
-[flake8](https://pypi.python.org/pypi/flake8). In general:
+If your sample is a set of discrete code snippets that each demonstrate a
+single operation, these should be grouped into a `snippets` folder. For
+example, see the snippets in the
+[bigtable/snippets/writes](https://github.com/GoogleCloudPlatform/python-docs-samples/tree/master/bigtable/snippets/writes)
+folder.
 
-1.  4 spaces.
-1.  `CamelCase` only for classes, `snake_case` elsewhere.
-1.  `_` for private variables, members, functions, and methods only. Samples
-  should generally have very few private items.
-1.  `CAP_WITH_UNDERSCORES` for constants.
-1.  `mixedCase` is only acceptable when interface with code that doesn't follow
-  our style guide.
-1.  79-character line limit.
-1.  Imports should be in three sections separated by a blank line - standard
-    library, third-party, and package-local. Sample will have very few
-    package-local imports.
+If your sample is a quickstart — intended to demonstrate how to quickly get
+started with using a service or API — it should be in a _quickstart_ folder.
 
-See [Automated tools](#automated-tools) for information on how to run the lint
-checker.
+### Python Version
 
-Beyond PEP8, there are several idioms and style nits we prefer.
+As of this writing (early 2020), the default Python version for samples is
+Python 3.7. All samples should pass tests on 3.7, and support for 3.8 is recommended.
 
-1. Use single quotes (`'`) except for docstrings (which use `"""`).
-1. Typically import modules over members, for example
-   `from gcloud import datastore`
-   instead of `from gcloud.datastore import Client`. Although you should use
-   your best judgment, for example
-   `from oauth2client.contrib.flask_util import UserOAuth2`
-   and `from oauth2client.client import GoogleCredentials` are both totally
-   fine.
-1. Never alias imports unless there is a name collision.
-1. Use `.format()` over `%` for string interpolation.
-1. Generally put a blank line above control statements that start new indented
-   blocks, for example:
+Support for Python 2.7 is no longer required for new samples. If you’re
+making a change to an existing sample that supports Python 2.7, the preferred
+approach is to maintain compatibility with 2.7.
 
-    ```python
-    # Good
-    do_stuff()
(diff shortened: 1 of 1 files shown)


dmahugh:
CC @texasmichelle as an FYI

AUTHORING_GUIDE.md
```diff
@@ -1,207 +1,59 @@
...
-and docstrings only as needed to further explain. Always introduce functions and
-variables before they are used. Prefer less indirection. Prefer imperative
-programming as it is easier to understand.
-
-### Shebang
-
-If, and only if, your sample application is a command-line application, then
-include a shebang as the first line. Separate the shebang from the rest of
-the application with a blank line. The shebang should always be:
+We're happy you want to write a Python sample! Like a lot of Pythonistas, we're opinationed and fussy. This guide is a reference for the format and style expected of samples contributed to the [python-docs-samples](https://github.com/GoogleCloudPlatform/python-docs-samples) repo. The guidelines below are intended to assure that all Python samples meet the following goals:
```
texasmichelle: s/assure/ensure/
Nit: max 79 char line breaks

AUTHORING_GUIDE.md
```diff
@@ -1,207 +1,59 @@
...
 
-Demonstrates how to create a simple echo API as well as how to use the
-various authentication methods available.
+All Python samples should follow the best practices defined in the [PEP 8 style guide](https://www.python.org/dev/peps/pep-0008/) and the [Google Python Style Guide](http://google.github.io/styleguide/pyguide.html). The automated linting process for Python samples uses [flake8](http://flake8.pycqa.org/en/latest/) to verify conformance to common Python coding standards, so the use of flake8 is recommended.
 
-For more information, see the README.md under /appengine/flexible and the
-documentation at https://cloud.google.com/appengine/docs/flexible.
-"""
-```
+If you prefer to use [pylint](https://www.pylint.org/), that’s fine. But note that Python samples for this repo are not required to conform to pylint’s default settings outside the scope of PEP 8, such as the “too many arguments” or “too many local variables” warnings.
```
texasmichelle: s/that's fine. But//

AUTHORING_GUIDE.md
```diff
@@ -299,198 +151,124 @@ If documenting primitive types, be sure to note if they have a particular set
...
 
-Because all the tests here are system tests, you'll need to have a Google
-Cloud project with billing enabled. Once you have this configured, you'll
-need to set environment variables for the tests to be able to use your project
-and its resources. See `testing/test-env.tmpl.sh` for a list of all environment
-variables used by all tests. Not every test needs all of these variables.
+Because all tests are system tests that use live resources, running tests requires a Google Cloud project with billing enabled, as covered under [Creating and Managing Projects](https://cloud.google.com/resource-manager/docs/creating-managing-projects).
 
-#### Google Cloud Storage resources
+Once you your project created and configured, you'll need to set environment variables to identify the project and resources to be used by tests. See [testing/test-env.tmpl.sh](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/testing/test-env.tmpl.sh) for a list of all environment variables used by all tests. Not every test needs all of these variables.
```
texasmichelle: s/Once you/Once you have/

AUTHORING_GUIDE.md
```diff
@@ -299,198 +151,124 @@ If documenting primitive types, be sure to note if they have a particular set
...
-The testing of `python-docs-samples` is managed by
-[nox](https://nox.readthedocs.io). Nox allows us to run a variety of tests,
-including the linter, Python 2.7, Python 3, App Engine, and automatic README
-generation.
+Automated testing for samples in `python-docs-samples` is managed by [nox](https://nox.readthedocs.io). Nox allows us to run a variety of tests, including the flake8 linter, Python 2.7, Python 3.x, and App Engine tests, as well as automated README generation.
 
 To use nox, install it globally with `pip`:
 
-    $ pip install nox
+```
```
texasmichelle: nit: specify the code block as console type, e.g. ` ```console`.

AUTHORING_GUIDE.md
dmahugh: Missed this one, resolving in next commit.

AUTHORING_GUIDE.md
```diff
@@ -1,207 +1,117 @@
...
+[bigtable/snippets/writes](https://github.com/GoogleCloudPlatform/python-docs-samples/tree/master/bigtable/snippets/writes)
+folder.
+
+If your sample is a quickstart — intended to demonstrate how to quickly get
+started with using a service or API — it should be in a _quickstart_ folder.
+
+## Python Version
+
+As of this writing (early 2020), the default Python version for samples is
+Python 3.6. All samples should pass tests on 3.6, and support for 3.7 and
```
leahecole: I would be inclined to say it's 3.7 at this point - but what do other folks think?

AUTHORING_GUIDE.md
```diff
@@ -281,216 +191,203 @@ Args:
...
+* Whenever possible, tests should avoid taking a dependency on the precise
+output of APIs. For example, if a test is intended to verify a JSON payload
+returned from an endpoint, it should only check for the existence of the
+expected keys and values, and the test should continue to work correctly
+if other values are added to the response in a future version of the API.
+* Samples that use App Engine Standard should use the [App Engine
+testbed](https://cloud.google.com/appengine/docs/standard/python/refdocs/google.appengine.ext.testbed)
+for system testing, as shown in [this
+example](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/appengine/standard/localtesting/datastore_test.py).
+* All tests should be independent of one another and order-independent.
```
leahecole: WOOP WOOP SAY IT LOUDER FOR THE PEOPLE IN THE BACK

AUTHORING_GUIDE.md
```diff
@@ -281,216 +191,203 @@ Args:
...
+Samples should include tests to verify that the sample runs correctly and
+generates the intended output. Follow these guidelines while writing your
+tests:
+
+* Use [pytest](https://docs.pytest.org/en/latest/)-style tests and plain
+asserts. Don't use `unittest`-style tests or `assertX` methods.
+* Whenever possible, tests should avoid taking a dependency on the precise
+output of APIs. For example, if a test is intended to verify a JSON payload
+returned from an endpoint, it should only check for the existence of the
+expected keys and values, and the test should continue to work correctly
```
leahecole: What's the difference between the precise output vs. expected keys and values? (I'm guessing it's order, but I want to double check)

AUTHORING_GUIDE.md
```diff
@@ -281,216 +191,203 @@ Args:
...
+* All tests should be independent of one another and order-independent.
 
-Again, descriptive names prevent you from having to exhaustively describe
-every parameter.
+## Arrange, Act, Assert
 
-Some samples demonstrate multiple functions. You should use *subparsers* to
-handle this, for example:
+Tests for samples should follow the “Arrange, Act, Assert” structure for
+unit tests:
```
leahecole: Unit or system?

AUTHORING_GUIDE.md
```diff
@@ -281,216 +191,203 @@ Args:
...
-if __name__ == '__main__':
-    parser = argparse.ArgumentParser(
-        description=__doc__,
-        formatter_class=argparse.RawDescriptionHelpFormatter)
-    parser.add_argument('bucket_name', help='Your cloud storage bucket.')
+* _Arrange_ — create and configure the components required for the test.
+Avoid nesting; prioritize readability and simplicity over efficiency.
+* _Act_ — execute the code to be tested, such as sending a request and
+receiving a response.
+* _Assert_ — verify that the test results match what is expected.
```
leahecole: Consider adding a quick example here

AUTHORING_GUIDE.md
```diff
@@ -281,216 +191,203 @@ Args:
...
 
-    subparsers.add_parser('list', help=list_blobs.__doc__)
+Whenever possible, tests should run against the live production version of
+cloud APIs and resources. This will assure that any breaking changes in those
+resources are identified by the tests.
 
-    upload_parser = subparsers.add_parser('upload', help=upload_blob.__doc__)
-    upload_parser.add_argument('source_file_name')
-    upload_parser.add_argument('destination_blob_name')
+Creating mocks for external resources is optional, and not recommended in
```
leahecole: I'd like to add a bit here that says the exception generally is for long running operations (like AutoML). Best practice I've seen is to begin the operation, assert that the operation exists and is running, and then cancel the operation.

AUTHORING_GUIDE.md
```diff
@@ -281,216 +191,203 @@ Args:
...
-    if args.command == 'list':
-        list_blobs(args.bucket_name)
-    elif args.command == 'upload':
-        upload_blob(
-            args.bucket_name,
-            args.source_file_name,
-            args.destination_blob_name)
-```
+When tests need temporary resources (such as a temp file or folder), they
+should create reasonable names for these resources with a UUID attached to
```
leahecole: We should explicitly state using the UUID module as opposed to using random number generation

AUTHORING_GUIDE.md
```diff
@@ -281,216 +191,203 @@ Args:
...
+without taking a dependency on where that string occurs in the output.
 
-```python
-if __name__ == '__main__':
-    # This is used when running locally. Gunicorn is used to run the
-    # application on Google App Engine. See entrypoint in app.yaml.
-    app.run(host='127.0.0.1', port=8080, debug=True)
-```
-## Writing tests
+## Using nox
```
leahecole: This section may change a little bit if we integrate @kurtisvg 's propsed noxfile changes #2806

AUTHORING_GUIDE.md
```diff
@@ -281,216 +191,203 @@ Args:
...
+requires a Google Cloud project with billing enabled, as covered under
+[Creating and Managing Projects](https://cloud.google.com/resource-manager/docs/creating-managing-projects).
+
+Once you have your project created and configured, you'll need to set environment
+variables to identify the project and resources to be used by tests. See
+[testing/test-env.tmpl.sh](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/testing/test-env.tmpl.sh)
+for a list of all environment variables used by all tests. Not every test
+needs all of these variables.
+
+## Google Cloud Storage Resources
```
leahecole: This section might need an update, but I'm curious to hear other opinions. I feel like each sample may require resources more than just GCS - and they tend to specify the variables you need to fill in. I'm not sure that the ./resources section is kept up to date. I could be wrong though - so let's hear what the rest of the owners think!

AUTHORING_GUIDE.md
kurtisvg: I think we have a couple different options. that there probably needs to be some more discussion around (for new versions and significant changes):
1. Target the latest stable python version 
2. Target the current most popular python version (and adjust every ~6 months)?
3. Target the earliest version the client libraries support

AUTHORING_GUIDE.md
```diff
@@ -1,207 +1,117 @@
...
+
+* **Copy-paste-runnable.** A developer should be able to copy and paste the
+code into their own environment and run it with as few modifications as
+possible.
+* **Teach through code.** Each sample should demonstrate best practices for
+interacting with Google Cloud libraries, APIs, or services.
+* **Idiomatic.** Each sample should follow widely accepted Python best
+practices as covered below.
+
+# Sample Guidelines
```
kurtisvg: nit: Usually a single `#` indicates title, and then headers are `##` or more. Not sure if this matters for Github's rendering (as it looks fine as is), but may cause issues later if we decide to autogen a table of contents or something.

AUTHORING_GUIDE.md
```diff
@@ -1,207 +1,117 @@
...
+[python-docs-samples](https://github.com/GoogleCloudPlatform/python-docs-samples)
+that corresponds to the Google Cloud service or API used by the sample.
+For example, a sample demonstrating how to work with BigTable should be
+in a subfolder under the
+[python-docs-samples/bigtable](https://github.com/GoogleCloudPlatform/python-docs-samples/tree/master/bigtable)
+folder.
+
+Conceptually related samples under a service or API should be grouped into
+a subfolder. For example, App Engine Standard samples are under the
+[appengine/standard](https://github.com/GoogleCloudPlatform/python-docs-samples/tree/master/appengine/standard)
```
kurtisvg: Should we sort based on runtime version as well? for example, `appengine/standard/python3.7`?

AUTHORING_GUIDE.md
```diff
@@ -1,207 +1,117 @@
...
-Demonstrates how to create a simple echo API as well as how to use the
-various authentication methods available.
-
-For more information, see the README.md under /appengine/flexible and the
-documentation at https://cloud.google.com/appengine/docs/flexible.
-"""
-```
-
-### Functions & classes
+## Functions and Classes
```
kurtisvg: This section seems to contain several things that might be better to break up into smaller sections so its easier to link/reference:
* Prefer descriptive function names
* rules around documenting arguements
* Mentions types, but I think we should embrace the py3 type annotations `def hello_world(name: string):`

AUTHORING_GUIDE.md
kurtisvg: For ML APIs in particular, the results may not be consistent. In these situations we should test that the call was successfully, but not necessarily that the data returned was consistent.

AUTHORING_GUIDE.md
```diff
@@ -281,216 +191,203 @@ Args:
...
+if other values are added to the response in a future version of the API.
+* Samples that use App Engine Standard should use the [App Engine
+testbed](https://cloud.google.com/appengine/docs/standard/python/refdocs/google.appengine.ext.testbed)
+for system testing, as shown in [this
+example](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/appengine/standard/localtesting/datastore_test.py).
+* All tests should be independent of one another and order-independent.
 
-Again, descriptive names prevent you from having to exhaustively describe
-every parameter.
+## Arrange, Act, Assert
```
kurtisvg: This is actually a rule of thumb for the snippets themselves. Most snippets should use the follow pattern:

1. Arrange - construct the API request and build the different components needed to configure the call
2. Act - send the request to the server and wait for a response
3. Assert - verify the response is a success, and act upon it (usually printing some results to show the call was successful)

AUTHORING_GUIDE.md
kurtisvg: I'd like this wording to be stronger - Mocks should be strongly discouraged, as they don't verify the validity of the sample against the API.

AUTHORING_GUIDE.md
```diff
@@ -281,216 +191,203 @@ Args:
...
-        formatter_class=argparse.RawDescriptionHelpFormatter)
-    parser.add_argument('bucket_name', help='Your cloud storage bucket.')
+* _Arrange_ — create and configure the components required for the test.
+Avoid nesting; prioritize readability and simplicity over efficiency.
+* _Act_ — execute the code to be tested, such as sending a request and
+receiving a response.
+* _Assert_ — verify that the test results match what is expected.
 
-    subparsers = parser.add_subparsers(dest='command')
+## External Resources
```
kurtisvg: Also, any external resources that are required before the test (e.g. a Cloud SQL instance) should be pass in via an environment variable. This should be limited to infrastructure - if there is some data that need to be on that instance, the test should create it as part of the test (and clean it up when complete).

AUTHORING_GUIDE.md
kurtisvg: Worth calling out that we run our tests in parallel, and the tests should be able to pass when run in parallel as well.

AUTHORING_GUIDE.md
kurtisvg: +1 this needs to be generalized to infrastructure resources as a whole.

AUTHORING_GUIDE.md
dmahugh: Addressed in commit [30673e7](https://github.com/GoogleCloudPlatform/python-docs-samples/pull/2728/commits/30673e7853619a8bbf193a93eb41c9a65061303d).

AUTHORING_GUIDE.md
dmahugh: I've updated to say 3.7 for now, and I agree that we need to discuss/decide a general strategy for Python versions going forward. Addressed in commit [751c8e2](https://github.com/GoogleCloudPlatform/python-docs-samples/pull/2728/commits/751c8e27c591a489d2cb47477b0a971d41fcf36a)

AUTHORING_GUIDE.md
dmahugh: Added explicit requirement for running tests in parallel, in commit [751c8e2](https://github.com/GoogleCloudPlatform/python-docs-samples/pull/2728/commits/751c8e27c591a489d2cb47477b0a971d41fcf36a)

AUTHORING_GUIDE.md
dmahugh: Clarified that dependency on order should be avoided, and added guidance on just verifying API call was successful in some cases. Addressed in commit [751c8e2](https://github.com/GoogleCloudPlatform/python-docs-samples/pull/2728/commits/751c8e27c591a489d2cb47477b0a971d41fcf36a)

AUTHORING_GUIDE.md
dmahugh: Added explicit guidance about use of ```uuid``` module. Addressed in commit [751c8e2](https://github.com/GoogleCloudPlatform/python-docs-samples/pull/2728/commits/751c8e27c591a489d2cb47477b0a971d41fcf36a)

AUTHORING_GUIDE.md
dmahugh: Added note about use of mocks strongly discouraged, tests should verify sample against the actual APIs. Addressed in commit [751c8e2](https://github.com/GoogleCloudPlatform/python-docs-samples/pull/2728/commits/751c8e27c591a489d2cb47477b0a971d41fcf36a)

AUTHORING_GUIDE.md
```diff
@@ -281,216 +188,212 @@ Args:
...
-handle this, for example:
+### Region Tags
 
-```python
-if __name__ == '__main__':
-    parser = argparse.ArgumentParser(
-        description=__doc__,
-        formatter_class=argparse.RawDescriptionHelpFormatter)
-    parser.add_argument('bucket_name', help='Your cloud storage bucket.')
+Sample code may be integrated into Google Cloud Documentation through the use
```
nnegrey: May not be true across the board, but seems like our tracking tools suggest if the samples aren't in docs that we should be deleting them? 

Love to hear what other folks are suggesting.

AUTHORING_GUIDE.md
```diff
@@ -1,207 +1,114 @@
...
-available in readmes or documentation.
+If you prefer to use [pylint](https://www.pylint.org/), note that Python
+samples for this repo are not required to conform to pylint’s default settings
+outside the scope of PEP 8, such as the “too many arguments” or “too many
+local variables” warnings.
 
-Here's a simple docstring for a command-line application with straightforward
-usage:
+The use of [Black](https://pypi.org/project/black/) to standardize code
+formatting and simplify diffs is recommended, but optional.
```
nnegrey: Would it be helpful to put that they should adjust the defaults of black to use `black -l 79` so that it matches PEP8?

AUTHORING_GUIDE.md
```diff
@@ -1,207 +1,114 @@
...
-Demonstrates how to create a simple echo API as well as how to use the
-various authentication methods available.
-
-For more information, see the README.md under /appengine/flexible and the
-documentation at https://cloud.google.com/appengine/docs/flexible.
-"""
-```
-
-### Functions & classes
+### Functions and Classes
```
nnegrey: Should we add default args to this section as an additional option?

```
def list_blobs(bucket_name="YOUR_BUCKET_NAME"):
```

AUTHORING_GUIDE.md
crwilcox: I would prefer we try to match what is being used on google-cloud-python/ python-{product} as to not be divergent. A few flake8 rules are ignored (https://github.com/googleapis/python-storage/blob/master/.flake8) and black is used with defaults.

kurtisvg: +1 to consistency with the client repos

PR #2730, opened 2020-01-16
samples(storage): IAM conditions samples

jkwlui:
Existing Bucket/IAM samples are updated to use:
 - `requested_policy_version=3` to request policy with conditions
 - `policy["rolename"]` is deprecated, use `policy.bindings` to access bindings instead.

Added sample to show how to add a conditional binding to an existing Bucket's policy.


jkwlui:
The tests passed previously when re-run individually.

PR #2734, opened 2020-01-17
fix: added cli functionality to dataproc quickstart example

bradmiro:
The region tag is intentionally placed to include the CLI logic.

### DIFF
--- a/dataproc/quickstart/quickstart.py
+++ b/dataproc/quickstart/quickstart.py
@@ -15,14 +15,15 @@
 # limitations under the License.
 
 # [START dataproc_quickstart]
-"""This quickstart sample walks a user through creating a Cloud Dataproc
-    cluster, submitting a PySpark job from Google Cloud Storage to the
-    cluster, reading the output of the job and deleting the cluster, all
-    using the Python client library.
+"""
+This quickstart sample walks a user through creating a Cloud Dataproc
+cluster, submitting a PySpark job from Google Cloud Storage to the
+cluster, reading the output of the job and deleting the cluster, all
+using the Python client library.
 
-    Usage:
-        python quickstart.py --project_id <PROJECT_ID> --region <REGION> \
-            --cluster_name <CLUSTER_NAME> --job_file_path <GCS_JOB_FILE_PATH>
+Usage:
+    python quickstart.py --project_id <PROJECT_ID> --region <REGION> \
+        --cluster_name <CLUSTER_NAME> --job_file_path <GCS_JOB_FILE_PATH>
 """
 
 import argparse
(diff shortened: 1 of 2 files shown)


dataproc/quickstart/quickstart.py
```diff
@@ -15,25 +15,24 @@
 # limitations under the License.
 
 # [START dataproc_quickstart]
+"""This quickstart sample walks a user through creating a Cloud Dataproc
+    cluster, submitting a PySpark job from Google Cloud Storage to the
+    cluster, reading the output of the job and deleting the cluster, all
+    using the Python client library.
+
+    Usage:
+        python3 quickstart.py --project_id <PROJECT_ID> --region <REGION> \
```
kurtisvg: Is this a python3 only sample? I _think_ the official policy is we are still supporting 2.7 atm.

dataproc/quickstart/quickstart.py
```diff
@@ -15,25 +15,24 @@
 # limitations under the License.
 
 # [START dataproc_quickstart]
+"""This quickstart sample walks a user through creating a Cloud Dataproc
```
kurtisvg:
```suggestion
"""
This quickstart sample walks a user through creating a Cloud Dataproc
```

dataproc/quickstart/quickstart.py
```diff
@@ -125,4 +124,23 @@ def quickstart(project_id, region, cluster_name, job_file_path):
     operation.result()
 
     print('Cluster {} successfully deleted.'.format(cluster_name))
-    # [END dataproc_quickstart]
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(
```
kurtisvg: This isn't tested at all - if we are going to add CLI functionality, we should have some tests verifying that it still works. 

Alternatively, I'd rather avoid CLI and instead give the user somewhere at the top of the code to set the variables instead.

dataproc/quickstart/quickstart.py
```diff
@@ -125,4 +124,23 @@ def quickstart(project_id, region, cluster_name, job_file_path):
...
+        description=__doc__,
+        formatter_class=argparse.RawDescriptionHelpFormatter,
+    )
+    parser.add_argument('--project_id', type=str,
```
kurtisvg: nit: No hanging indents
```suggestion
    parser.add_argument(
      '--project_id', type=str,
```

dataproc/quickstart/quickstart.py
bradmiro: I can add a test for this, but the motivation for CLI is to make this a runnable tool as-is without needing to modify code.

PR #2738, opened 2020-01-17
Dialogflow session entity

nnegrey:


nnegrey:
Ooops pushed too many PRs at once and hit a quota.

nnegrey:
So turns out we might end up deleting a lot of the Dialogflow samples as they aren't currently in docs. Going to close these PRs and wait for a decision then re-open if the samples end up not being deleted.

PR #2740, opened 2020-01-20
Fix bigtable tests

leahecole:
Ignore til it's not a holiday! I'm doing some work today in anticipation for an upcoming vacation. 

@billyjacobson, I just updated it to call the environment var in the testing


billyjacobson:
I added a "BIGTABLE_INSTANCE" variable to the encrypted file with environment variables. Not sure if it didn't get picked up or wasn't done correctly. 

I changed it because "BIGTABLE_CLUSTER" is incorrect since it's talking about the instance, and it's inconsistent with other languages, which makes maintaining this code more difficult 

@leahecole

leahecole:
reverted, but made sure binary file was updated with that change, and that seems to have fixed things

leahecole:
OH WAIT it doesn't run bigtable tests, only testing. hmm one sec ill trigger it

leahecole:
Alright @billyjacobson - I got your KeyError fixed but we're exceeding quota. Is there anything we can do to refactor for that, or should we just request more quota?

Also, I'll remove those placeholder comments and fix the lint before we merge :)

leahecole:
@kurtisvg and @billyjacobson - when you take a look at the review, if you look at the failed build on the previous commit, you can see that all of the affected bigtable tests passed - on this run of the build, it just ran against `testing/` - the only change i made was to remove the comments that triggered the bigtable tests in the first place.

PR #2743, opened 2020-01-22
vision: move published samples into master

texasmichelle:
Cherry-picked from tmp-generated-samples branch.

Add samples and tests.
Add missing attribute mime_type.
Fix encoding issue in py2.


texasmichelle:
Interesting test failure: `Service account HMAC key limit reached`

nnegrey:
Yea, I got that on one of my PRs from the `storage` samples.

vision/cloud-client/detect/vision_async_batch_annotate_images.py
```diff
@@ -0,0 +1,99 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright 2019 Google LLC
```
nnegrey: I think we want to bump to 2020 since it'll technically be a new file for the master branch.

vision/cloud-client/detect/vision_async_batch_annotate_images.py
```diff
@@ -0,0 +1,99 @@
+# -*- coding: utf-8 -*-
```
nnegrey: I also think we stopped adding these from comments on my last PR

vision/cloud-client/detect/vision_async_batch_annotate_images.py
```diff
@@ -0,0 +1,99 @@
...
+# To install the latest published package dependency, execute the following:
+#   pip install google-cloud-vision
+
+# sample-metadata
+#   title: Async Batch Image Annotation
+#   description: Perform async batch image annotation
+#   usage: python3 samples/v1/vision_async_batch_annotate_images.py \
+#     [--input_image_uri \
+#     "gs://cloud-samples-data/vision/label/wakeupcat.jpg"] \
+#     [--output_uri "gs://your-bucket/prefix/"]
```
nnegrey: Should we drop all this now that it technically isn't auto-generated?

vision/cloud-client/detect/vision_async_batch_annotate_images.py
```diff
@@ -0,0 +1,99 @@
...
+
+    client = vision_v1.ImageAnnotatorClient()
+
+    # input_image_uri = 'gs://cloud-samples-data/vision/label/wakeupcat.jpg'
+    # output_uri = 'gs://your-bucket/prefix/'
+
+    if isinstance(input_image_uri, six.binary_type):
+        input_image_uri = input_image_uri.decode("utf-8")
+    if isinstance(output_uri, six.binary_type):
+        output_uri = output_uri.decode("utf-8")
```
nnegrey: I think we can just delete these lines.

vision/cloud-client/detect/vision_async_batch_annotate_images.py
```diff
@@ -0,0 +1,99 @@
...
+    )
+    parser.add_argument(
+        "--output_uri", type=str, default="gs://your-bucket/prefix/")
+    args = parser.parse_args()
+
+    sample_async_batch_annotate_images(args.input_image_uri, args.output_uri)
+
+
+if __name__ == "__main__":
+    main()
```
nnegrey: Should we remove CLI too?

vision/cloud-client/detect/vision_async_batch_annotate_images.py
texasmichelle: Removed

vision/cloud-client/detect/vision_async_batch_annotate_images.py
texasmichelle: Done.

vision/cloud-client/detect/vision_async_batch_annotate_images.py
texasmichelle: Done.

vision/cloud-client/detect/vision_async_batch_annotate_images.py
texasmichelle: Done.

vision/cloud-client/detect/vision_async_batch_annotate_images.py
texasmichelle: The canonical sample referenced in the authoring guide has been removed. I think this functionality is kind of nice, but I'd like to see a documented example if there's a different standard for this repo. There's no reference to this in the updates to AUTHORING (#2728).

texasmichelle: Are you suggesting removing `def main():` and moving its contents into `if __name__ == "__main__":`?

nnegrey: Remove lines 82-99 so no CLI at all. 

I think someone is working on a new canonical

texasmichelle: Done

vision/cloud-client/detect/vision_async_batch_annotate_images.py
```diff
@@ -0,0 +1,60 @@
...
+
+    # input_image_uri = "gs://cloud-samples-data/vision/label/wakeupcat.jpg"
+    # output_uri = "gs://your-bucket/prefix/"
+
+    source = {"image_uri": input_image_uri}
+    image = {"source": source}
+    type_ = enums.Feature.Type.LABEL_DETECTION
+    features_element = {"type": type_}
+    type_2 = enums.Feature.Type.IMAGE_PROPERTIES
+    features_element_2 = {"type": type_2}
```
kurtisvg: nit: Can we inline these types? Doesn't look like they are being used elsewhere. 

Same with feature_element(s) preferably

kurtisvg: (same in other samples)

vision/cloud-client/detect/vision_async_batch_annotate_images.py
```diff
@@ -0,0 +1,60 @@
...
+def sample_async_batch_annotate_images(
+    input_image_uri="gs://cloud-samples-data/vision/label/wakeupcat.jpg",
+    output_uri="gs://your-bucket/prefix/",
+):
+    """Perform async batch image annotation"""
+    # [START vision_async_batch_annotate_images_core]
+
+    client = vision_v1.ImageAnnotatorClient()
+
+    # input_image_uri = "gs://cloud-samples-data/vision/label/wakeupcat.jpg"
```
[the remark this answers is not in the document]
kurtisvg: (same in other samples)

vision/cloud-client/detect/vision_async_batch_annotate_images_test.py
```diff
@@ -0,0 +1,53 @@
...
+        input_image_uri=input_image_uri, output_uri=GCS_DESTINATION_URI
+    )
+
+    out, _ = capsys.readouterr()
+
+    assert "Output written to GCS" in out
+    assert len(list(bucket.list_blobs(prefix=OUTPUT_PREFIX))) > 0
+
+    for blob in bucket.list_blobs(prefix=OUTPUT_PREFIX):
+        blob.delete()
```
kurtisvg: These buckets should be fixtures, so they can be removed even if an error occurs in the test

vision/cloud-client/detect/vision_batch_annotate_files.py
```diff
@@ -0,0 +1,77 @@
...
+    type_ = enums.Feature.Type.DOCUMENT_TEXT_DETECTION
+    features_element = {"type": type_}
+    features = [features_element]
+
+    # The service can process up to 5 pages per document file. Here we specify
+    # the first, second, and last page of the document to be processed.
+    pages_element = 1
+    pages_element_2 = 2
+    pages_element_3 = -1
+    pages = [pages_element, pages_element_2, pages_element_3]
```
kurtisvg: nit: inline these values as well?

kurtisvg:
In the future, you can use the "Request Review" button to whoever is assigned to let them know you are ready for review. This will ensure it gets past my filters :)

vision/cloud-client/detect/vision_async_batch_annotate_images.py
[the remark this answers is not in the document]
dizcology: the function def line won’t show up in the docs.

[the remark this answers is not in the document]
nnegrey: These were auto-generated, so there are 2 region tags, 
`vision_async_batch_annotate_images`
and 
`vision_async_batch_annotate_images_core` 
Depends on which one is used.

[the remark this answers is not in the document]
kurtisvg: That's fair enough, can we settle on one way to show them?

vision/cloud-client/detect/vision_async_batch_annotate_images.py
texasmichelle: Done

vision/cloud-client/detect/vision_async_batch_annotate_images_test.py
texasmichelle: Done

vision/cloud-client/detect/vision_batch_annotate_files.py
texasmichelle: Done

vision/cloud-client/detect/vision_async_batch_annotate_images.py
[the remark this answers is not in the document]
texasmichelle: The defaults were in the CLI functionality that I removed. As a comment, the user can't run as-is without editing. As function parameters, it won't show up in the docs. How do other samples handle this?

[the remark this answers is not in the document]
kurtisvg: There are two region tags here - `vision_async_batch_annotate_images` and `vision_async_batch_annotate_images_core `. If the first is used, then the default parameters are shown and the comment isn't needed. If the second is used, then the default parameters aren't needed but the comment is. 

Do we need both region tags? Why are we linking the sample differently from two different regions? I would like to avoid having the same values in two places if possible. 

Other samples usually use one pattern or the other. My preference is for using the outter region tags and removing the comments - showing imports is useful and it's preferable to have the default values being checked by the compiler.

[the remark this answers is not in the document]
texasmichelle: I verified that nothing live is pointing to the inner region tag before removing. That's a good point about compile-time verification.

texasmichelle:
That failure looks unrelated to these changes. Is there a command to re-run tests?

kurtisvg:
> That failure looks unrelated to these changes. Is there a command to re-run tests?

You can apply the `kokoro:run` or `kokoro:force-run` to trigger new builds. (as Noah did here)

texasmichelle:
Word on the street is that `kokoro:run` often doesn't work.

Is there anything outstanding? I think this is ready for review.

vision/cloud-client/detect/vision_batch_annotate_files.py
```diff
@@ -0,0 +1,61 @@
...
+from google.cloud.vision_v1 import enums
+import io
+
+
+def sample_batch_annotate_files(file_path="resources/kafka.pdf"):
+    """
+    Perform batch file annotation
+
+    Args:
+      file_path Path to local pdf file, e.g. /path/document.pdf
```
nnegrey: Should this be dropped since there is now a default above?

texasmichelle: Even with defaults, it's still good style to have docstrings for all sorts of reasons, like bubbling up in error messages or providing clear context. Or are we going for the bare minimum inside docs?

nnegrey: In the past our snippets on docs go for the bare minimum, but that's usually because the region tags didn't include the doc strings and now with it outside the function def maybe it should have them. 

I'm also waiting on the new style guide rules for Python.
So... Idk! 

This default in the function def might be more helpful `path/to/your/document.pdf` otherwise, we'll get folks saying they don't have that file.

texasmichelle: @kurtisvg docstrings or no?

That file is included in the resources directory, which is the point of the default. Do you think that will still be problematic for some users?

kurtisvg: I lean slightly towards no (or minimal) docstring because it often feels too verbose. A single line explaining the action the snippet takes seems the most valuable.

RE: the file in the resources directory - the goal is to optimize for folks copy-pasting from c.g.c, which means they would need to supply their own file. With that in mind `path/to/your/document.pdf` makes the most sense because it's more obvious they need to change it.

vision/cloud-client/detect/vision_batch_annotate_files_gcs.py
```diff
@@ -0,0 +1,57 @@
...
+# limitations under the License.
+
+# [START vision_batch_annotate_files_gcs]
+
+from google.cloud import vision_v1
+from google.cloud.vision_v1 import enums
+
+
+def sample_batch_annotate_files(
+    storage_uri="gs://cloud-samples-data/vision/document_understanding/kafka.pdf",
```
kurtisvg: nit: Is this a publicly accessible bucket or should this be in the something like `"gs://YOUR-PROJECT-ID/path/to/your/file.pdf"`

texasmichelle: Yes, this is publicly accessible, confirmed with a non-corp account.

PR #2744, opened 2020-01-22
Add max-line-length to nox lint.

kurtisvg:
Makes `nox -s lint` compatible with `nox -s blacken`'s defaults.


noxfile-template.py
```diff
@@ -125,6 +125,7 @@ def _setup_appengine_sdk(session):
     "--exclude",
     ".nox,.cache,env,lib,generated_pb2,*_pb2.py,*_pb2_grpc.py",
     "--ignore=E121,E123,E126,E203, E226,E24,E266,E501,E704,W503,W504,I100,I201,I202",
+    "--max-line-length=88",
```
busunkim96: Could you also change this in https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/noxfile.py?

kurtisvg: I can, but I don't think we are actively using that file and I plan to delete it in another PR to reduce confusion.

PR #2747, opened 2020-01-23
speech: fixing flaky auto punctuation test

czahedi:


czahedi:
Fixed my commit username for future commits, but the CLA was triggered by having the wrong name to begin with

nnegrey:
@czahedi, looks like some tests are failing.

Just look at the speech ones if you have access, otherwise we can sync on Tuesday and go through this info.

PR #2748, opened 2020-01-23
Adding environment port variable

dinagraves:
To run locally, the PORT environment variable needs to be set.

### DIFF
--- a/run/hello-broken/README.md
+++ b/run/hello-broken/README.md
@@ -18,7 +18,7 @@ docker build --tag hello-broken:python .
 ## Run Locally
 
 ```
-docker run --rm -p 9090:8080 hello-broken:python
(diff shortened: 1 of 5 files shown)


averikitsch:
Does this mean we should remove the `if os.getenv("PORT") else 8080` from the main.py?

dinagraves:
> Does this mean we should remove the `if os.getenv("PORT") else 8080` from the main.py?

No, this is only for if you want to run the docker image locally.  Otherwise the container won't start!   

You can still run the python file outside of a docker image, in which case, you need the getenv, eg: `python main.py`

leahecole:
Once this passes the appropriate CI, I'll merge it in! I know that this uses f strings and we previously said that was okay for it to fail earlier python versions. :)

PR #2749, opened 2020-01-28
Add Secret Manager IAM samples

sethvargo:


sethvargo:
The python 3.5 failure seems to be related to quota. @kurtisvg can you please re-run?

secretmanager/api-client/snippets_test.py
```diff
@@ -43,6 +45,11 @@ def project_id():
     return os.environ['GCLOUD_PROJECT']
 
 
+@pytest.fixture()
+def iam_user():
+    return 'user:sethvargo@google.com'
```
kurtisvg: Is this supposed to be there? If it's a fixture it should probably generate one somehow, or else probably better off as just a constant.

sethvargo: Is there a good default service account to use? I couldn't find an obvious one, and I don't have permissions to create a service account on this project.

kurtisvg: I can make one for this project, but if it's a resource the user needs to set up ahead of time to run the tests it should be passed via an environment variable.

sethvargo: Okay, so if we wanted to set it as `GCLOUD_SECRETS_SERVICE_ACCOUNT`... where would we configure that in koroko?

sethvargo: @kurtisvg  Okay, so if we wanted to set it as GCLOUD_SECRETS_SERVICE_ACCOUNT... where would we configure that in koroko?

kurtisvg: Done.

sethvargo: Hey @kurtisvg - Koroko seems to say that it's not set. Do I need do do something special to pick up the change?

kurtisvg: I've added them to the secret-tar in this [commit](https://github.com/GoogleCloudPlatform/python-docs-samples/pull/2749/commits/a52879834c482db5790af4ff199cf474f6ab529a). You'll want to make sure you pull that commit before your next rebase/force-push.

PR #2768, opened 2020-02-04
Translate: migrate published samples

texasmichelle:
Migrate from tmp-generated-samples branch 615c08e
Remove boilerplate
Update copyright date
Blacken
Remove unused imports


translate/cloud-client/translate_v3_detect_language.py
```diff
@@ -0,0 +1,45 @@
...
+# limitations under the License.
+
+# [START translate_v3_detect_language]
+from google.cloud import translate
+
+
+def sample_detect_language(project_id="YOUR_PROJECT_ID"):
+    """Detecting the language of a text string."""
+
+    text = "Hello, world!"
```
nnegrey: Should we inline this?

I think it should at least come after the `parent` line 26

PR #2771, opened 2020-02-05
Add GAE billing limit sample

ace-n:
**Why no tests?**
It's been tested manually, and we're trying to merge ASAP. Will fast-follow with tests once this is merged.


kurtisvg:
Overriding - functions is a 3.7+ environment only

PR #2774, opened 2020-02-05
Add GAE billing tests + fix region tags

### LINKED ISSUE #2772: Add tests to GAE billing GCF sample
The GAE billing GCF sample was merged in #2771 without tests (in the interest of time).

We should add tests. 😃

ace-n:
Fixes #2772


ace-n:
Chatted offline with @kurtisvg - since GCF only supports Python 3.7, we can disable `functions/*` tests for the other versions (2.7, 3.5, 3.6)

.kokoro/tests/run_tests.sh
```diff
@@ -76,6 +76,16 @@ for file in **/requirements.txt; do
         fi
     fi
 
+    # Skip unsupported Python versions for Cloud Functions
+    # (Some GCF samples' dependencies don't support them)
+    if [[ "$file" == "functions/"* ]]; then
+      PYTHON_VERSION="$(python --version 2>&1)"
```
tmatsuo, line 82: FYI, this code is completely wrong.

This code checks the default Python version, rather than Python version in the nox session.

This resulted in disabling all the tests under functions directory. #3767 removed this code block then our tests started to fail.

PR #2775, opened 2020-02-05
.kokoro: publish test log to Build Cop Bot

tbpg:
The Build Cop Bot will open/close issues for failing/passing tests.


.kokoro/tests/run_tests.sh
```diff
@@ -90,6 +90,33 @@ for file in **/requirements.txt; do
     nox -s "$RUN_TESTS_SESSION"
     EXIT=$?
 
+    # If we're running system tests, send the test log to the Build Cop Bot.
+    # See https://github.com/googleapis/repo-automation-bots/tree/master/packages/buildcop.
+    if [[ $KOKORO_BUILD_ARTIFACTS_SUBDIR = *"system-tests"* ]]; then
```
leahecole: Is this system-tests directory something that's on the kokoro end or is it searching for it here? 

If searching here, it won't find it - we restructured the kokoro config files to be broken down by python version and then into continuous, periodic, presubmit, and lint, unlike before when we had system tests by product

tbpg: Good catch. Switched to continuous. Let me know if you'd prefer periodic.

leahecole: Sweet. They're all going to fail right away - we're slowly but surely resolving all failures within.

.kokoro/tests/run_tests.sh
```diff
@@ -100,6 +100,33 @@ for file in **/requirements.txt; do
     nox -s "$RUN_TESTS_SESSION"
     EXIT=$?
 
+    # If this is a continuous build, send the test log to the Build Cop Bot.
+    # See https://github.com/googleapis/repo-automation-bots/tree/master/packages/buildcop.
+    if [[ $KOKORO_BUILD_ARTIFACTS_SUBDIR = *"continuous"* ]]; then
+      # Use the service account with access to the repo-automation-bots project.
+      gcloud auth activate-service-account --key-file $GOOGLE_APPLICATION_CREDENTIALS
+      gcloud config set project repo-automation-bots
```
kurtisvg: Would it be better to use `-project` flag on the command below instead? Hopefully none of the samples are relying on gcloud to detect projects, but it might be safer to minimize side effects

tbpg: Good idea! Switched.

PR #2793, opened 2020-02-07
Fix monitoring tests

### LINKED ISSUE #2792: monitoring tests are failing w/ quota exceeded
I'm cleaning it up and the tests now

leahecole:
fixes #2792


leahecole:
Still working on figuring out cleanup - realizing my initial strategy is not valid for parallelization - I need to take a break and look again later

leahecole:
Ugh the alerts tests are failing because of our parallel test runs, and the cloud-client failure is flaky and is marked as flaky, but still being considered a failure. I'm a little stuck here and taking another break (maybe a really long one) and if other folks have input please let me know!

leahecole:
(just reverted to a state where more tests were passing than failing) - These tests do pass locally but just don't work when run in parallel - I'm out of clever ideas for now

tmatsuo:
@leahecole FYI the alert-client tests will be fixed with #3285

leahecole:
Closing this because I think @tmatsuo is looking into this, and I'd lost track of it completely

PR #2830, opened 2020-02-10
fix: vision product search tests to call setup and teardown and use uuid

nnegrey:
Fix: https://github.com/GoogleCloudPlatform/python-docs-samples/issues/2825
Fix: https://github.com/GoogleCloudPlatform/python-docs-samples/issues/2826
Fix: https://github.com/GoogleCloudPlatform/python-docs-samples/issues/2827

Note: because there are 4 runtimes being tested, I expect to find that import product set to fail as it uses a static .csv file hosted on GCS that has a set product id, I have a feeling I'll have to dynamically create and upload that file then delete it afterward.


nnegrey:
@leahecole, we need 2 approving reviews now?

busunkim96:
@nnegrey We recently bumped it to two (one for the codeowner and one from someone in python-samples-owners). But in this case I think you are the codeowner for vision, so there was no second reviewer tagged.

leahecole:
^yup what Bu Sun said.

PR #2835, opened 2020-02-10
fix: vision automl - delete unused files that have moved

nnegrey:
Fix: https://github.com/GoogleCloudPlatform/python-docs-samples/issues/2823
Fix: https://github.com/GoogleCloudPlatform/python-docs-samples/issues/2824


nnegrey:
@kurtisvg, do we have docker installed on kokoro?
I guess some other tests are trying to use it and the sample is in docs. 
https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/vision/automl/edge_container_predict/automl_vision_edge_container_predict_test.py#L61

nnegrey:
Was blocked by: https://github.com/googleapis/testing-infra-docker/pull/51

PR #2844, opened 2020-02-11
Fix incorrect data submission

dhercher:
The configuration to send composer needs to be inside the data block rather than the JSON body itself


dhercher:
@googlebot I signed it!

grant:
Looks fine but the author @leahecole should approve.

leahecole:
@dhercher are you running this from Google cloud functions? I just ran it this morning following [this tutorial](https://cloud.google.com/composer/docs/how-to/using/triggering-with-gcf#triggering-with-gcf-creating-a-function-python) (which the code was written for) and it worked fine - could you tell me more about what you're trying to do and the desired outcome?

PR #2893, opened 2020-02-12
convert cancel to fixture

leahecole:


leahecole:
@kurtisvg is this what you're thinking? also FYI @davidcavazos Kurtis reached out to me offline after I'd merged your PR about using a fixture instead, so I took a stab at it since it was on me for merging it

kurtisvg:
No, you actually want to combine it with `unique_job_name`. So something like this:

```python
@pytest.fixture(scope="function")
def dataflow_job_id(label):
   job_id = datetime.now().strftime('{}-%Y%m%d-%H%M%S-{}'.format(  label, uuid.uuid4().hex))

  yield job_id

  cancel_operation(job_id)
```

Then use it in the test like this:
```python
def test_run_template_http_data(app, dataflow_job_id):
```

So what this does is every test that has `dataflow_job_id` get a unique id, and then. the fixture cleans up that operation (but after the test is done - so it always happens, regardless of errors/fails).


The second thing that looks problematic is the infinite poll for a certain condition:
```python
        while state != 'JOB_STATE_RUNNING':
            job = dataflow.projects().jobs().get(
                projectId=PROJECT,
                jobId=job_id
            ).execute()
            state = job['currentState']
            time.sleep(1)
```

If the job crashes and never reaches `JOB_STATE_RUNNING`, this will poll forever and ever and the test will never finish. In general with waiting for a thing to happen, there usually should be some upper bound. Something like: 

```python
total_attempts = 60 
while state != 'JOB_STATE_RUNNING':
     total_attempts -= 1
     if total_attempts < 0:
         raise Exception(f"Unable to cancel job with id: {job_id}")
      job = dataflow.projects().jobs().get(
                projectId=PROJECT,
                jobId=job_id
            ).execute()
            state = job['currentState']

            time.sleep(1)
```

davidcavazos:
Thanks for the pro `pytest` moves here :)

leahecole:
I made some progress on this today - I'm passing an argument to the fixture the way I want to be, but don't quite have cancellation working properly. @davidcavazos I'm going to set up some time with you to ask some questions about Dataflow + how the resources behave before I go down the rabbit hole.

leahecole:
DONE! I learned SO much about pytest and Dataflow so thank you @davidcavazos for your help there. 

Things I did

- I added a parametrized fixture that creates the name w/ UUID for the dataflow jobs. 
- The next thing I targeted was the cancellation function - unfortunately, the dataflow cancel function requires the ID and not the name, so I had to make a helper that gets the UUID from the name with UUID 
- added assertions that check that the response from the application contains the name of the dataflow job we made
- added backoff to the cancellation instead of a while loop that could be infinite

leahecole:
The tests that are broken are unrelated to my PR

dataflow/run_template/main_test.py
```diff
@@ -44,29 +46,72 @@ def app():
...
+    # cancel the Dataflow job after running the test
+    dataflow_jobs_cancel(job_name)
+
+
+# Takes in a Dataflow job name and returns its job ID
+def get_job_id_from_name(job_name):
+    # list the 100 most recent Dataflow jobs
+    jobs_request = dataflow.projects().jobs().list(
+        projectId=PROJECT,
+        pageSize=50  # only return the 50 most recent results - our job is likely to be in here. If the job is not found, first try increasing this number
```
davidcavazos, line 68: Maybe also add `filter` to only get active jobs?

dataflow/run_template/main_test.py
```diff
@@ -44,29 +46,72 @@ def app():
...
+    yield job_name
+
+    # cancel the Dataflow job after running the test
+    dataflow_jobs_cancel(job_name)
+
+
+# Takes in a Dataflow job name and returns its job ID
+def get_job_id_from_name(job_name):
+    # list the 100 most recent Dataflow jobs
+    jobs_request = dataflow.projects().jobs().list(
```
davidcavazos, line 66: Nit: can we add a link comment to the REST API docs?

https://cloud.google.com/dataflow/docs/reference/rest/v1b3/projects.jobs/list

dataflow/run_template/main_test.py
```diff
@@ -44,29 +46,72 @@ def app():
...
-def test_run_template_python_empty_args(app):
+# We retry the cancel operation a few times until the job is in a state where it can be cancelled
+@backoff.on_exception(backoff.expo, HttpError, max_time=240)
+def dataflow_jobs_cancel(job_name):
+    # to cancel a dataflow job, we need its ID, not its name
+    job_id = get_job_id_from_name(job_name)
+
+    if job_id:
+        # Cancel the Dataflow job if it exists. If it doesn't, job_id will be equal to None.
+        request = dataflow.projects().jobs().update(
```
davidcavazos, line 89: Nit: can we add a comment with a link to the docs as well?

https://cloud.google.com/dataflow/docs/reference/rest/v1b3/projects.jobs/update

dataflow/run_template/main_test.py
leahecole: good idea

dataflow/run_template/main_test.py
leahecole: definitely

davidcavazos:
Looking good, can you add a link to the `update` REST API method as well? Other than that, looks good as long as the tests pass (or fail without being these changes fault :)

leahecole:
> Looking good, can you add a link to the `update` REST API method as well? Other than that, looks good as long as the tests pass (or fail without being these changes' fault :)

done

dataflow/run_template/main_test.py
leahecole: done

PR #2918, opened 2020-02-14
datalabeling: ensure all tests use test endpoint

nnegrey:
Fix: https://github.com/GoogleCloudPlatform/python-docs-samples/issues/2805
Fix: https://github.com/GoogleCloudPlatform/python-docs-samples/issues/2967


nnegrey:
Yea, basically most of the files and tests were setting this, but these were not so when some tests called these methods they happened on the liver service not the test one. 

The product team provides the test endpoint.

nnegrey:
@gguuss ping

nnegrey:
Status: it appears there has been a change in the API, investigating.

PR #2920, opened 2020-02-14
fix: address bigquery/bqml test failures 

shollyman:
The BQML tests use a non-unique dataset ID for multiple examples and
testing currently triggers a lot of concurrent creation/deletions of
said dataset. Switch to a dataset that leverages uuid to avoid
invocations stomping on one another, which also necessitates
parameterizing much of the SQL.

There's also an issue with the pandas import currently, possibly due to
recent changes in panda. This change also pins pandas to 0.22 and
doesn't rely on the dependency being expressed as an extra through
google-cloud-bigquery.

### DIFF
--- a/bigquery/bqml/data_scientist_tutorial_test.py
+++ b/bigquery/bqml/data_scientist_tutorial_test.py
@@ -20,7 +20,7 @@ import uuid
 
 # [START bqml_data_scientist_tutorial_import_and_client]
 client = bigquery.Client()
-# We use a unique dataset ID for this example to avoid collisions with 
+# We use a unique dataset ID for this example to avoid collisions with
 # other invocations of this tutorial.  In practice, you could leverage
 # a persistent dataset and not create/destroy it with each invocation.
 dataset_id = "bqml_tutorial_{}".format(str(uuid.uuid4().hex))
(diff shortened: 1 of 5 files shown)


bigquery/bqml/requirements.txt
```diff
@@ -1,3 +1,4 @@
-google-cloud-bigquery[pandas]==1.20.0
+pandas==0.22
+google-cloud-bigquery>=1.24.0
```
kurtisvg: Can we explicitly pin this? 
```suggestion
google-cloud-bigquery==1.24.0
```

bigquery/bqml/data_scientist_tutorial_test.py
```diff
@@ -16,29 +16,34 @@
...
 # [END bqml_data_scientist_tutorial_import_and_client]
 import pytest
+import uuid
 
 # [START bqml_data_scientist_tutorial_import_and_client]
 client = bigquery.Client()
+# We use a unique dataset ID for this example to avoid collisions with
+# other invocations of this tutorial.  In practice, you could leverage
+# a persistent dataset and not create/destroy it with each invocation.
+dataset_id = "bqml_tutorial_{}".format(str(uuid.uuid4().hex))
```
kurtisvg: Would it be better to move this outside the region tag?

shollyman: It would be better, but much of the subsequent SQL requires rewriting to use the generated ID so I thought it aid understanding why all the SQL has python format strings.

PR #2921, opened 2020-02-15
Added config for using Memorystore (Redis) with App Engine standard P…

michaelawyu:
…ython 2 runtime


michaelawyu:
This sample is originally shared among 4 environments: GCE, GKE, App Engine (Standard/Python 3) and App Engine (Flexible). Some of the dependencies are neither compatible with Python 2 nor needed in the old environment, hence the exclusion from Python 2 tests.

memorystore/redis/gae_standard_py2_deployment/app.yaml
```diff
@@ -0,0 +1,38 @@
+# Copyright 2019 Google LLC.
```
busunkim96: Since these are new files, could you bump the copyright year to 2020?

PR #2936, opened 2020-02-19
dialogflow: make flaky test more generic

nnegrey:


nnegrey:
Fix: https://github.com/GoogleCloudPlatform/python-docs-samples/issues/2898

nnegrey:
I think the datetime seconds is not quite granular enough for the 4 python versions in parallel, switching to UUID. And where required cutting strings to length of 30 characters.

nnegrey:
and quota issues again.

PR #2948, opened 2020-02-20
remove misleading raise_for_status

TAOXUY:
When the response has error, the error detail is not exposed because of raise_for_status, so remove it and show the detailed message.

For instance, the detailed error is UNAUTHENTICATED:Method doesn't allow unregistered callers (callers without established identity). Please use API Key or other form of API consumer identity to call this API.
but we only got requests.exceptions.HTTPError: 401 Client Error: Unauthorized for url.


endpoints/getting-started/clients/google-jwt-client.py
```diff
@@ -69,7 +69,6 @@ def make_jwt_request(signed_jwt, url='https://your-endpoint.com'):
     }
     response = requests.get(url, headers=headers)
 
-    response.raise_for_status()
```
busunkim96: @DanSanche Does `response.text` need to be returned or is it alright to do the `print()` inside the function?

The `print()` could be moved into the function. 

```python
print(response.text)
response.raise_for_status()
# [END endpoints_jwt_request]
return response.text
```

DanSanche: I'm forgetting some of the context around this sample, but I think printing should be fine. That code block looks good to me

busunkim96: @TAOXUY would you mind making that change?

endpoints/getting-started/clients/google-jwt-client.py
```diff
@@ -69,8 +69,7 @@ def make_jwt_request(signed_jwt, url='https://your-endpoint.com'):
...
-    return response.text
+    print(response.text)
 # [END endpoints_jwt_request]
```
busunkim96: @TAOXUY Thank you! Could you also add a `response.raise_for_status()`after the `print()`?

TAOXUY:
Please take a look

leahecole:
@Daniel-Sanche can you take another look at this please?

TAOXUY:
@busunkim96, please take a look.

endpoints/getting-started/clients/google-jwt-client.py
```diff
@@ -68,9 +68,8 @@ def make_jwt_request(signed_jwt, url='https://your-endpoint.com'):
         'content-type': 'application/json'
     }
     response = requests.get(url, headers=headers)
+    print(response.status_code, response.content)
```
busunkim96, line 71:
```suggestion
    print(response.status_code, response.content)
    response.raise_for_status()
```

Hi - could you re-add `raise_for_status()` so an exception is thrown if the request fails?

TAOXUY: Done

TAOXUY:
@busunkim96, please take a look.

busunkim96:
Thank you for the contribution @TAOXUY!

PR #2987, opened 2020-02-25
automl: update broken beta tests to use fake data

nnegrey:
Fix: https://github.com/GoogleCloudPlatform/python-docs-samples/issues/2907
Fix: https://github.com/GoogleCloudPlatform/python-docs-samples/issues/2906


nnegrey:
Oh yea, there is some other sample in there that relies on having docker available, which I think got added to the Kokoro environment recently, but not sure if that was released. 

```
FileNotFoundError: [Errno 2] No such file or directory: 'docker': 'docker'
```

@kurtisvg, this PR: https://github.com/googleapis/testing-infra-docker/pull/51
Have those changes taken affect and should now work here?

kurtisvg:
@nnegrey Should be, but I triggered the tests to confirm

busunkim96:
@kurtisvg @nnegrey  I triggered a re-build after those changes but there were some issues with the image (missing Python 2.7 😞) so I manually rolled back to the previous version. Will investigate today.

busunkim96:
Docker images have been republished and look good. Re-triggering tests

nnegrey:
Theses files have been deleted since they were out of date.

PR #3026, opened 2020-03-04
Pin django to a specific version 

leahecole:
another part of splitting up #2651


appengine/standard/django/requirements.txt
```diff
@@ -1,2 +1,2 @@
 PyMySQL==0.9.3
-Django<2.0.0,>=1.11.8
+Django==1.11.28 # needs to stay under 2.0.0
```
kurtisvg: nit: can we add a reason? So future us know why we pinned?

leahecole: done - good nit

PR #3027, opened 2020-03-04
splitting up #2651 part 3/3 - bigquery

leahecole:


leahecole:
Resolved all other failures - this is because datalab is incompatible with Python 3.7 rn. How do we feel about merging anyways? https://github.com/googledatalab/pydatalab/issues/711

PR #3029, opened 2020-03-04
fix: Use unique resources for storage snippets.

### LINKED ISSUE #2980: storage.cloud-client.acl_test: test_remove_bucket_default_owner failed
storage.cloud-client.acl_test: test_remove_bucket_default_owner failed
buildID: [TODO: set buildID]
buildURL: [Build Status](https://source.cloud.google.com/results/invocations/5a70ae44-571c-4809-ac84-84972e6313d4), [Sponge](http://sponge2/5a70ae44-571c-4809-ac84-84972e6313d4)
status: failed

crwilcox:
Fixes #2980, #2977, #2976, #2940, #2927, #2913 by separating buckets used in parallel runs and cleaning up resources


crwilcox:
The storage tests will still show failures for HMAC, possible any other flakes. This PR should stop a category of flakes though.

crwilcox:
Test Runs have HMAC Failures only :)

storage/cloud-client/acl_test.py
```diff
@@ -39,37 +36,29 @@
...
-    """Yields a bucket that resets its acl after the test completes."""
-    bucket = storage.Client().bucket(BUCKET)
-    acl = google.cloud.storage.acl.BucketACL(bucket)
-    object_default_acl = google.cloud.storage.acl.DefaultObjectACL(bucket)
-    acl.reload()
-    object_default_acl.reload()
-    time.sleep(1)  # bucket ops rate limited 1 update per second
+    """Yields a bucket that is deleted after the test completes."""
+    bucket = None
+    while bucket is None or bucket.exists():
```
leahecole: Would existence imply that a test somehow somewhere didn't clean up after itself?

storage/cloud-client/acl_test.py
```diff
@@ -39,37 +36,29 @@
...
+        bucket_name = "acl-test-{}".format(uuid.uuid4())
```
leahecole: nit should this be snake case esp because the test blob is too?

storage/cloud-client/snippets_test.py
```diff
@@ -41,122 +42,133 @@
...
     out, _ = capsys.readouterr()
     assert "Removed labels" in out
 
 
+@pytest.fixture(scope="module")
+def test_bucket():
+    """Yields a bucket that is deleted after the test completes."""
+    bucket = None
+    while bucket is None or bucket.exists():
+        bucket_name = "storage-snippets-test-{}".format(uuid.uuid4())
```
leahecole: same nit re snake case

storage/cloud-client/acl_test.py
crwilcox: This is mostly to avoid UUID collisions. I think python forms them with timestamps so, in theory, the parallel cases could try to make the same buckets

storage/cloud-client/acl_test.py
crwilcox: They are named like this elsewhere so I just repeated the pattern (see bucket lock)

crwilcox: I just confirmed in the project. It seems we have pretty well used '-' elsewhere, so going to avoid starting a new trend.

busunkim96:
@crwilcox Are you planning on force merging this and adding the HMAC test fixes in a separate PR?

crwilcox:
Verified that test failures are expected and all in HMAC. Will make another PR for that once I get a chance to diagnose what is going there.

PR #3037, opened 2020-03-05
fix: Reuse HMAC key as we have a limit of 5

### LINKED ISSUE #2822: storage.cloud-client.hmac_samples_test: test_create_key failed
storage.cloud-client.hmac_samples_test: test_create_key failed
buildID: 2b3daa3ff1daa16903d67e8ea2d4e4e11b51ad0e
buildURL: https://source.cloud.google.com/results/invocations/bbb3728b-dbdb-44ef-bf68-f0844abbadb2
status: failed

crwilcox:
We are limited to a low number (5) hmac keys for a single project. To avoid exhausting this under test this will reuse that key and depend on the test order (a bit) to keep our count down. Fixes #2822


storage/cloud-client/hmac_samples_test.py
```diff
@@ -35,8 +36,7 @@
 SERVICE_ACCOUNT_EMAIL = os.environ["HMAC_KEY_TEST_SERVICE_ACCOUNT"]
 STORAGE_CLIENT = storage.Client(project=PROJECT_ID)
 
-
-@pytest.fixture
+@pytest.fixture(scope="module")
```
frankyn: Does this correct the fixture once for the entire test suite? If `test_delete_key` is ran first then it would be deleted for all.

crwilcox: With the way we have things configured, within the module these will run in order. Yes, if delete is run first, the key will be gone for our tests. So this change makes the order significant.

crwilcox:
Python 2.7 Failure isn't HMAC 👍

PR #3043, opened 2020-03-05
automl: move video object tracking samples out of branch

nnegrey:


automl/beta/video_object_tracking_create_dataset_test.py
```diff
@@ -0,0 +1,50 @@
...
+
+import video_object_tracking_create_dataset
+
+PROJECT_ID = os.environ["AUTOML_PROJECT_ID"]
+pytest.DATASET_ID = None
+
+
+@pytest.fixture(scope="function", autouse=True)
+def teardown():
+    yield
```
kurtisvg: Is it possible to delete by dataset name? It would be preferable if the fixture would return the name w/ uuid and then clean it up afterwards. 

(I feel very iffy about adding an attribute to a third-party module)

nnegrey: No, you have to use the ID which is created by the API. Yea, I've spent time looking for a better solution, I just haven't found one yet, beside using globals. 

One solution was to use a class based path and just set that. I'm open to ideas.

kurtisvg: I think it's better to just have a regular global over assigning it to third party module - otherwise your solution seems good given the constraints.

PR #3044, opened 2020-03-05
automl: move samples into beta set

nnegrey:


nnegrey:
@busunkim96 friendly WFH ping (no rush)

PR #3071, opened 2020-03-10
Clarifying comment for batch requests

czahedi:


czahedi:
Re: Torry's nit, that can be a backlog concern but for now we're just trying to head off the customer confusion about the way batch requests work.

Re: Noah's nit, I've asked the technical writer to wait until this Python PR AND the Node PR go through to rebase the docs.

PR #3078, opened 2020-03-10
Remove Language sample unused region_tag comments

jsimonweb:


gguuss:
@jsimonweb recent changes removed python 2.7 from the kokoro configs (except for GAE), maybe updating your branch will resolve the failing builds

PR #3094, opened 2020-03-12
DO NOT MERGE vision: v1 logo detection samples

czahedi:
v1 logo detection samples. Service not rolled out yet so DO NOT MERGE

### DIFF
--- a/video/cloud-client/analyze/video_detect_logo.py
+++ b/video/cloud-client/analyze/video_detect_logo.py
@@ -12,22 +12,25 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# [START video_detect_logo_gcs_beta]
+import io
 
-from google.cloud import videointelligence_v1p3beta1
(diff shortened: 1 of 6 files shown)


video/cloud-client/analyze/video_detect_logo.py
```diff
@@ -14,16 +14,16 @@
 
 import io
 
-# [START video_detect_logo_beta]
+# [START video_detect_logo]
```
nnegrey: So we try not to do this as it will break docs. Prefer to create a new GA file rather than rename beta files.

czahedi: Ah dang good point. Do we then follow up and delete the beta files?

czahedi:
Closing will rebase with appropriate names

czahedi:
See https://github.com/GoogleCloudPlatform/python-docs-samples/pull/3098

PR #3101, opened 2020-03-16
spanner: Add Cloud Spanner Backup samples

larkee:


spanner/cloud-client/backup_sample.py
```diff
@@ -0,0 +1,273 @@
+# Copyright 2019 Google Inc. All Rights Reserved.
```
skuruppu:
```suggestion
# Copyright 2020 Google Inc. All Rights Reserved.
```

spanner/cloud-client/backup_sample.py
```diff
@@ -0,0 +1,273 @@
...
+                metadata.progress.progress_percent))
+# [END spanner_list_backup_operations]
+
+
+# [START spanner_list_database_operations]
+def list_database_operations(instance_id):
+    spanner_client = spanner.Client()
+    instance = spanner_client.instance(instance_id)
+
+    # List the progress of restore
```
skuruppu:
```suggestion
    # List the progress of restore.
```

spanner/cloud-client/backup_sample.py
```diff
@@ -0,0 +1,273 @@
...
+
+
+# [START spanner_delete_backup]
+def delete_backup(instance_id, backup_id):
+    spanner_client = spanner.Client()
+    instance = spanner_client.instance(instance_id)
+    backup = instance.backup(backup_id)
+    backup.reload()
+
+    # Wait for databases that reference this backup to finish optimizing
```
skuruppu:
```suggestion
    # Wait for databases that reference this backup to finish optimizing.
```

spanner/cloud-client/backup_sample_test.py
```diff
@@ -0,0 +1,107 @@
+# Copyright 2019 Google Inc. All Rights Reserved.
```
skuruppu:
```suggestion
# Copyright 2020 Google Inc. All Rights Reserved.
```

spanner/cloud-client/backup_sample_test.py
```diff
@@ -0,0 +1,107 @@
...
+
+def test_cancel_backup(capsys):
+    backup_sample.cancel_backup(INSTANCE_ID, DATABASE_ID, BACKUP_ID)
+    out, _ = capsys.readouterr()
+    cancel_success = "Backup creation was successfully cancelled." in out
+    cancel_failure = (
+        ("Backup was created before the cancel completed." in out) and
+        ("Backup deleted." in out)
+    )
+    assert cancel_success or cancel_failure
```
skuruppu: We don't have tests for the list samples. Is that intentional?

larkee: Yes. Originally it was because I was worried about them being flaky. Now that the tests are run in a new instance it makes sense to have them again. I'll add them tomorrow.

spanner/cloud-client/backup_sample_test.py
```diff
@@ -0,0 +1,125 @@
...
+        "All backups with backup name containing \"users\":\n"
+        "All backups with database name containing \"bank\":\n"
+        "All backups with expire_time before \"2019-10-18T02:56:53Z\":\n"
+        "All backups with backup size more than 1000 bytes:\n"
+        "All backups created after \"2019-10-18T02:56:53Z\" and are READY:\n"
+    ) in out
+    assert (
+        "All backups created after \"2019-10-18T02:56:53Z\" and are READY:\n" +
+        backup_name
+    ) in out
```
skuruppu: I recommend adding a check to see if there's at least one backup returned. Otherwise it could be that we printed out the message but it didn't retrieve any backups.

larkee: `backup_name` is constructed above and cannot be an empty string. If no backups are returned, the assertion would fail.

spanner/cloud-client/backup_sample.py
```diff
@@ -0,0 +1,273 @@
...
+    filter_ = (
+        "(metadata.database:{}) AND "
+        "(metadata.@type:type.googleapis.com/"
+        "google.spanner.admin.database.v1.CreateBackupMetadata)"
+    ).format(database_id)
+    operations = instance.list_backup_operations(filter_=filter_)
+    for op in operations:
+        metadata = op.metadata
+        # List the pending backups on the instance.
+        if metadata.progress.progress_percent < 100:
```
skuruppu: I realized that this actually can't be tested in the samples because we always wait for backups to complete. So we should remove this check and say "List all create backup operations on this database."

That also reminded me that there's no test for this below. Is that intentional?

spanner/cloud-client/backup_sample.py
```diff
@@ -0,0 +1,273 @@
...
+    instance = spanner_client.instance(instance_id)
+
+    # List all backups.
+    print("All backups:")
+    for backup in instance.list_backups():
+        print(backup.name)
+
+    # List all backups that contain a name.
+    print("All backups with backup name containing \"users\":")
+    for backup in instance.list_backups(filter_="name:users"):
```
hengfengli: I got a question that does this `list_backups` return any results? I wonder whether or not there is a backup with `name:users`.

spanner/cloud-client/backup_sample.py
```diff
@@ -0,0 +1,273 @@
...
+        print(backup.name)
+
+    # List all backups for a database that contains a name.
+    print("All backups with database name containing \"bank\":")
```
hengfengli: same here. Do we have a database name with `bank`?

spanner/cloud-client/backup_sample.py
skuruppu: Yeah it probably wouldn't because we can't control what people would put in the command line. For Node.js, I'm filtering on the backup ID passed into the function but seems like Python doesn't have that.

skuruppu: Ok so Node.js, PHP and Go are more or less consistent. I've asked Knut to do the same in Java. The changes we ended up making are:

1. Get the backup ID as input to listBackups and filter on name:backupId.
2. Database filter was changed to filter on the databaseId given as input.
3. Expire time filter was changed to expire_time < (time now + 30 days).
4. size_bytes > X where X depends on the size of the database you're creating in other parts of the test. PHP is using X = 500 and Node.js is using X = 100.
5. create_time >= (time now - 1 day)

Hope it's ok to make these changes.

spanner/cloud-client/backup_sample.py
```diff
@@ -0,0 +1,276 @@
...
+        print(backup.name)
+
+    # List backups that were created after a timestamp that are also ready.
+    create_time = datetime.utcnow().replace(microsecond=0) - timedelta(days=1)
+    print("All backups created after \"{}-{}-{}T{}:{}:{}Z\" and are READY:".format(
+        *create_time.timetuple()))
+    for backup in instance.list_backups(
+            filter_="create_time >= \"{}-{}-{}T{}:{}:{}Z\" AND state:READY".format(
+                *create_time.timetuple())):
+        print(backup.name)
```
skuruppu: One more thing. We have a pagination sample in the other languages. Was that every here? I may have missed that in the review.

PR #3102, opened 2020-03-16
Adds example parameters 

gguuss:
Adds parameters to IoT Manager sample for appearance in documentation. Note service account path is unused so it is not documented.


iot/api-client/manager/manager.py
```diff
@@ -115,6 +121,12 @@ def create_es256_device(
     """Create a new device with the given id, using ES256 for
     authentication."""
     # [START iot_create_es_device]
+    # project_id = 'YOUR_PROJECT_ID'
+    # cloud_region = 'us-central1'
+    # registry_id = 'your-registry-id'
+    # device_id = 'your-device-id'
+    # certificate_file = 'path/to/certificate.pem'
```
busunkim96: `public_key_file`?

gguuss:
Friendly ping @leahecole: will I need to restructure my parameters, includecode blocks, and comments to match new changes in python-docs-samples style or can we avoid this change as this is not a new sample?

leahecole:
iirc, we haven't made a concrete decision on it. Noah proposed it as part of Doug's authoring guide restructure, but that hasn't been set in stone yet. Node seems to do both - have the defaults and also have the comments for docs. Is it a huge headache to add default values?

gguuss:
@leahecole I would prefer to minimize the changes to this and other IoT core samples because they all depend on each other e.g. the IoT MQTT and HTTP client samples use the manager sample to bootstrap tests, the codelabs use the MQTT samples, and so on. I'm not sure how many things it will break now or in the future to refactor these samples and would prefer to avoid unnecessary change.

leahecole:
that would qualify as a headache, so I'd say it's good enough for now!

gguuss:
@busunkim96 can I get a +1, it appears @hongalex lost access to this repo.

PR #3107, opened 2020-03-17
feat: add remove conditional binding sample

frankyn:


storage/cloud-client/storage_remove_bucket_conditional_iam_binding.py
```diff
@@ -0,0 +1,71 @@
...
+    # Set the policy's version to 3 to use condition in bindings.
+    policy.version = 3
+
+    policy.bindings = [
+        binding
+        for binding in policy.bindings
+        if not (
+            binding["role"] == role
+            and binding.get("condition")
+            and binding["condition"]
```
jkwlui: can just simplify these lines to
```python
and binding.get("condition") == ...
```

storage/cloud-client/storage_remove_bucket_conditional_iam_binding.py
```diff
@@ -0,0 +1,71 @@
...
+            == {
```
jkwlui: Define a variable above.

```python
condition = {
  "title": title,
  ..
}
```

then rewrite as 
```python
and binding.get("condition") == condition
```

storage/cloud-client/storage_remove_bucket_conditional_iam_binding.py
```diff
@@ -0,0 +1,71 @@
...
+import sys
+
+# [START storage_remove_bucket_conditional_iam_binding]
+from google.cloud import storage
+
+
+def remove_bucket_conditional_iam_binding(
+    bucket_name, role, title, description, expression
+):
+    """Add a conditional IAM binding to a bucket's IAM policy."""
```
jkwlui: Add -> Remove

storage/cloud-client/storage_remove_bucket_conditional_iam_binding.py
```diff
@@ -0,0 +1,67 @@
...
+    policy.bindings = [
+        binding
+        for binding in policy.bindings
+        if not (binding["role"] == role and binding.get("condition") == condition)
+    ]
+
+    bucket.set_iam_policy(policy)
+
+    print("Conditional Binding was removed.")
+
```
jkwlui: spaces

frankyn: added by formater

storage/cloud-client/iam_test.py
```diff
@@ -22,24 +22,27 @@
...
-    while bucket is None or bucket.exists():
+    while bucket is None or not bucket.exists():
+        storage_client = storage.Client()
         bucket_name = "test-iam-{}".format(uuid.uuid4())
-        bucket = storage.Client().bucket(bucket_name)
-    bucket.create()
-    bucket.iam_configuration.uniform_bucket_level_access_enabled = True
-    bucket.patch()
+        bucket = storage_client.bucket(bucket_name)
+        bucket.iam_configuration.uniform_bucket_level_access_enabled = True
```
crwilcox: This shouldn't be indented. There is a possible bucket collision and this may be manipulating that instead.

At minimum the creation needs to be outside of the while loop.

storage/cloud-client/uniform_bucket_level_access_test.py
```diff
@@ -27,10 +27,10 @@
 def bucket():
     """Yields a bucket that is deleted after the test completes."""
     bucket = None
-    while bucket is None or bucket.exists():
+    while bucket is None or not bucket.exists():
```
crwilcox: Same comment as above. This is all to handle the possibility that the bucket by the generated name exists. To avoid test flaking, we want a new unique bucket.

PR #3116, opened 2020-03-18
chore: only run 2.7 tests for GAE

busunkim96:
This runs 2.7 tests only for GAE Standard samples.

The second commit intentionally touched files in every directory to trigger CI. The real changes to the noxfiles and kokoro are in the [first commit](https://github.com/GoogleCloudPlatform/python-docs-samples/pull/3116/commits/15e30db3ab93255e3c4f44db71108cabdf311b2c). I will revert the spurious License header commit before merging. :)


averikitsch:
We probably don't need Py2.7 testing anymore and should remove all py2.7 GAE samples as well. @andrewsg

andrewsg:
We'll continue to offer Python 2 on GAE for a very long time, so we can't remove those samples yet. As for Py2 tests on platforms other than GAE, it would be helpful to me if we can continue to run them so we know when Py2 compatibility breaks, and then disable them piecemeal, but I understand if that is not possible.

At any rate, a large number of people run Py2 on our platform so we should keep that in mind.

busunkim96:
@andrewsg We're getting to drop 2.7 support and move to the microgenerator next quarter. We will be *very clear* when we drop 2.7 support (python_requires in setup.py, major version bumps, and notes in changelogs and READMEs). Continuing to run the Python 2 tests on all the samples prevents folks from using modern language features, so I think this is a good time.

Do others have thoughts? @GoogleCloudPlatform/python-samples-owners

appengine/flexible/django_cloudsql/mysite/settings.py
```diff
@@ -1,6 +1,6 @@
 # Copyright 2015 Google Inc.
 #
-# Licensed under the Apache License, Version 2.0 (the "License");
+# Licensed under the APAche License, Version 2.0 (the "License");
```
gguuss: Looks like some of these were changed to be capitolized.

busunkim96:
Removed the extra commit to trigger CI.

PR #2959, opened 2020-02-21
Add SubscriberClient.close() to examples

pradn:

PR #3118, opened 2020-03-19
Pub/Sub: add SubscriberClient.close() to examples

anguillanneuf:
New PR based on @pradn's #2959. 

Close subscriber clients after use either explicitly by calling the `close()` method or implicitly by wrapping them in a context manager.

### DIFF
--- a/pubsub/cloud-client/iam.py
+++ b/pubsub/cloud-client/iam.py
@@ -51,6 +51,8 @@ def get_subscription_policy(project, subscription_name):
     print("Policy for subscription {}:".format(subscription_path))
     for binding in policy.bindings:
         print("Role: {}, Members: {}".format(binding.role, binding.members))
+
+    client.close()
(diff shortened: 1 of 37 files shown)


anguillanneuf:
Thanks for reviewing @gguuss! That is intended.

pubsub/cloud-client/iam.py
```diff
@@ -51,6 +51,8 @@ def get_subscription_policy(project, subscription_name):
     print("Policy for subscription {}:".format(subscription_path))
     for binding in policy.bindings:
         print("Role: {}, Members: {}".format(binding.role, binding.members))
+
+    client.close()
```
pradn: Now that I think more about this, it's probably best to use the "with" syntax wherever possible. ie: every place where the subscriber client and the close call are in the same function block.

anguillanneuf: Acknowledged.

PR #3119, opened 2020-03-19
bigtable: Handle dev instances and use storage utilization in metric scaler

billyjacobson:


bigtable/metricscaler/metricscaler.py
```diff
@@ -45,6 +46,25 @@ def get_cpu_load():
...
+          float: The most recent Cloud Bigtable storage utilization metric
+    """
+    # [START bigtable_metric_scaler_storage_utilization]
+    client = monitoring_v3.MetricServiceClient()
+    utilization_query = query.Query(client,
+                                    project=PROJECT,
+                                    metric_type='bigtable.googleapis.com/'
+                                                'cluster/storage_utilization',
+                                    minutes=5)
+    time_series = list(utilization_query)
```
crwilcox: If we only read the first entry, we can avoid processing the full iterable. I have 4 entries when I run locally.
```suggestion
    time_series = next(utilization_query.iter())
```

bigtable/metricscaler/metricscaler_test.py
```diff
@@ -35,6 +36,8 @@
 INSTANCE_ID_RANGE = 10000
 BIGTABLE_INSTANCE = INSTANCE_ID_FORMAT.format(
     random.randrange(INSTANCE_ID_RANGE))
+BIGTABLE_DEV_INSTANCE = INSTANCE_ID_FORMAT.format(
+    random.randrange(INSTANCE_ID_RANGE))
```
tmatsuo, line 40: We are running multiple builds simultaneously.  What happens when two builds are using the same value? Can you do `INSTANCE_ID_FORMAT.format(uuid.uuid4())`?

billyjacobson: good idea

tmatsuo: Oops sorry, we're getting `Invalid id for collection instances : Length should be between [6,33], but found 54 'metric-scale-test-2dbdbbc5-5a3f-49be-86f4-4645bcc099f2'"`

Maybe use the first part of the uuid, or something to make it shorter?

tmatsuo: @billyjacobson I think `uuid.uuid4().hex` works.

tmatsuo:
Error:

google.api_core.exceptions.InvalidArgument: 400 Error in field 'clusters' : Cluster Id : Invalid id for collection clusters : Length should be between [6,30], but found 33 'metric-scale-test-bb01b83e-8fad-4'

billyjacobson:
@tmatsuo tests are fixed and green :)

PR #3126, opened 2020-03-21
Bumped Django to resolve CVE-2020-7471

jackton1:


jackton1:
@googlebot I signed it!

PR #3136, opened 2020-03-24
fix: adjust bigquery/cloud-client to allow concurrent testing

shollyman:
Natality test was overly stateful.  Change to optimistic dataset creation and elide truncation mode to allow for concurrent overwriting.

Fixes: #2850


shollyman:
> LGTM when the tests pass. Optional change suggestion - I've been asked to avoid reusing resources in testing and instead use unique resources per-test (e.g. create a database with UUID in the name).

In this case, I'm generally more okay with datasets being longlived because there's a higher expectation of persistence by users, and the dataset itself has no cost (resources inside are a different story).

bigquery/cloud-client/requirements.txt
```diff
@@ -1,5 +1,6 @@
 google-cloud-bigquery[pandas]==1.20.0
 google-auth-oauthlib==0.4.0
-ipython==7.8.0
+# ipython 5.9.0 is the last LTS release that supports python 2.7.
```
leahecole: can you instead use the python_version syntax? it'll instead be

ipython==5.9.0; python_version < "3.0"
ipython==7.8.0; python_version > "3.0"

bigquery/cloud-client/requirements.txt
```diff
@@ -1,5 +1,6 @@
...
+ipython==5.9.0
 matplotlib==3.1.1
```
leahecole: looks like we'll have to do the same for matplotlib too - it'll be
matplotlib==2.2.5; python_version < "3.0"
matplotlib==3.1.1; python_version > "3.0"

bigquery/cloud-client/requirements.txt
leahecole: We've just turned off 2.7 testing for everything but GAE, so I think we can remove this commented out version of ipython

shollyman: done

bigquery/cloud-client/authorized_view_tutorial_test.py
```diff
@@ -23,24 +23,21 @@ def client():
...
-
-def test_authorized_view_tutorial(client, to_delete):
+def remove_authz_views(client, dataset_ref):
+    try:
+        ds = client.get_dataset(dataset_ref)
+        ds.access_entries = [entry for entry in ds.access_entries
+                             if entry.entity_type != 'view']
+        client.update_dataset(ds, ['access_entries'])
+    except Exception:
+        pass
```
leahecole, line 33: nit: can we add a print statement at minimum here or catch a specific exception?

shollyman:
Abandoning this one for a different approach.

PR #3137, opened 2020-03-25
Pub/Sub: update publish with batch settings sample

anguillanneuf:
The previous sample resolves each publish future as each message is published, effectively skipping batching. 

The new sample uses the non-blocking function `add_done_callback()` on publish futures instead of `result()` in the main thread.

### DIFF
--- a/pubsub/cloud-client/publisher.py
+++ b/pubsub/cloud-client/publisher.py
@@ -183,8 +183,8 @@ def publish_messages_with_batch_settings(project_id, topic_name):
     # Configure the batch to publish as soon as there is one kilobyte
     # of data or one second has passed.
     batch_settings = pubsub_v1.types.BatchSettings(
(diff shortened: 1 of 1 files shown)


pubsub/cloud-client/publisher.py
```diff
@@ -183,17 +183,24 @@ def publish_messages_with_batch_settings(project_id, topic_name):
     # Configure the batch to publish as soon as there is one kilobyte
     # of data or one second has passed.
     batch_settings = pubsub_v1.types.BatchSettings(
-        max_bytes=1024, max_latency=1  # One kilobyte  # One second
+        max_bytes=1024,
```
hongalex: Can we put the default setting in a comment here? Some users have expressed that this would be useful.

anguillanneuf: Good idea. Done!

PR #3144, opened 2020-03-26
blunderbuss: Remove dmahugh

kurtisvg:
Doug's taking a break from python-docs-samples for a bit.


leahecole:
You're welcome back anytime @dmahugh ! I'll remove your name as someone to bug on other pages too.

Issue #3221, opened 2020-03-29
PyMySQL doesn't work?


Bartleby2718:
## In which file did you encounter the issue?

https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/appengine/standard_python37/django/mysite/settings.py

### Did you change the file? If so, how?

Switch from `PyMySQL` to `mysqlclient`

## Describe the issue

When I deploy to GCP using `gcloud app deploy`, I get the 502 error after deployment. I checked the error with `gcloud app logs tail -s default`, and the error message was `mysqlclient 1.3.13 or newer is required; you have 0.9.3.`, where 0.9.3 is the version of PyMySQL version in the current `requirements.txt`. I switched to `mysqlclient`, and now the error is gone. Also, now that you're no longer using PyMySQL, you can now change the Django version to 2.2+.

PR #3222, opened 2020-03-29
Switch from PyMySQL to mysqlclient

Bartleby2718:
Now the 502 error is gone.


tmatsuo:
@Bartleby2718 

Thank you for your contribution. I assume you encountered 502 errors without this PR. Can you file an issue first? I want to see the repro. Also this sample doesn't have test, so I think we'll need to add some tests first before merging this change in.

Bartleby2718:
@tmatsuo Before creating this PR, I did file an issue (#3221), which was resolved by #3248. I will now close this PR. I didn't know the maintainers wanted to ["continue PyMySQL support"](https://github.com/GoogleCloudPlatform/python-docs-samples/pull/3248#issue-397197327) when I filed the issue. Thank you for the comment, though!

PR #3255, opened 2020-04-02
Remove oauth2client from IoT and Functions samples

busunkim96:
There is one more firebase sample using oauth2client that will need a bit more work.


functions/billing/main.py
```diff
@@ -80,7 +79,6 @@ def stop_billing(data, context):
         'cloudbilling',
         'v1',
         cache_discovery=False,
-        credentials=GoogleCredentials.get_application_default()
```
busunkim96, line 83: Application Default credentials are used by default if no credentials are passed.

PR #3277, opened 2020-04-03
Add GCF bearer token sample

ace-n:
~**N.B:** this sample is both simple and hard to (properly) test, so I didn't include tests here.~

(Properly testing it would likely require deploying it to GCF itself.)


functions/security/main.py
```diff
@@ -0,0 +1,42 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# [START functions_bearer_token]
+import requests
+
+# TODO<developer>: set these values
+REGION = 'us-central1'
+PROJECT_ID = 'my-project'
+RECEIVING_FUNCTION = 'my-function'
```
grant, line 21: There is only one function in this file. Rename to `FUNCTION_NAME`?

functions/security/main_test.py
```diff
@@ -0,0 +1,37 @@
...
+    requestsMock.get.side_effect = [
+        Response(u'some-token'),
+        Response(u'function-done')
+    ]
+
+    res = main.calling_function(flask.request)
+
+    second_headers = requestsMock.get.call_args_list[0][1]
+    assert second_headers == {'headers': {'Metadata-Flavor': 'Google'}}
+    assert 'function-done' == res
```
grant, line 37: This is a yoda condition.
We should `assert res == 'function-done'`.

functions/security/main_test.py
```diff
@@ -0,0 +1,37 @@
+# Copyright 2019 Google LLC
```
gguuss, line 1: New file copyright 2020?

functions/security/main.py
ace-n: This `calling_function` is supposed to call a second "receiving function" with the token, wait for results (via an HTTP response), and return those results to its caller.

grant: This is fine. Would be nice to have a comment or a better function name rather than `my-function`.

PR #3281, opened 2020-04-06
cdn: add sign_cookie example with tests

mpwarres:
This PR expands the existing snippet for Cloud CDN Signed URLs to also illustrate Signed Cookies, a feature currently in closed alpha. This feature extends Cloud CDN's [Signed URL](https://cloud.google.com/cdn/docs/using-signed-urls) functionality.


mpwarres:
@elithrar FYI

cdn/snippets.py
```diff
@@ -71,6 +71,41 @@ def sign_url(url, key_name, base64_key, expiration_time):
 # [END sign_url]
 
 
+# [START sign_cookie]
```
kurtisvg, line 74: These region tags aren't in the proper format (should be `product_sign_cookie`).

elithrar:
@kurtisvg This will go GA shortly (within a week) and we are attempting to sync this up with the need to be in the public docs for launch day.

Is there a better process to handle cases like this? We are not concerned about folks seeing this PR ahead of being merged.

kurtisvg:
@elithrar Most folks use GOB to stage samples when not publicly available, but if you aren't concerned about public knowledge this is fine. 

We won't be able to merge it into master until the feature is publicly available, but you can ping me when that happens and I'll take care of it.

cdn/snippets.py
mpwarres: Done.

mpwarres:
PTAL again:
- Added code example and tests for Signed URL prefix.
- Updated CODEOWNERS entry for /cdn/
- Updated region tags.

Thanks!

cdn/snippets.py
```diff
@@ -94,8 +174,48 @@ def sign_url(url, key_name, base64_key, expiration_time):
             type=lambda d: datetime.datetime.utcfromtimestamp(float(d)),
             help='Expiration time expessed as seconds since the epoch.')
 
+    sign_url_prefix_parser = subparsers.add_parser(
+            'sign-url-prefix',
+            help="Sign a URL prefix to grant temporary authorized access.")
+    sign_url_prefix_parser.add_argument(
+            'url', help='The request URL.')
+    sign_cookie_parser.add_argument(
```
tmatsuo, line 182: Lint fails here FYI

tmatsuo:
@mpwarres 

In case you don't have access to the build logs. Here is the lint output:

```
./snippets.py:182:5: F821 undefined name 'sign_cookie_parser'
    sign_cookie_parser.add_argument(
    ^
```

It seems like `sign_cookie_parser` is defined way below.

cdn/snippets.py
mpwarres: Thanks! Fixed.

cdn/snippets.py
```diff
@@ -68,9 +68,87 @@ def sign_url(url, key_name, base64_key, expiration_time):
...
+    Returns:
+        Returns the Cloud-CDN-Cookie value based on the specified configuration.
+    """
+    encoded_url_prefix = base64.urlsafe_b64encode(
+            url_prefix.strip().encode('utf-8')).decode('utf-8')
+    epoch = datetime.datetime.utcfromtimestamp(0)
+    expiration_timestamp = int((expiration_time - epoch).total_seconds())
+    decoded_key = base64.urlsafe_b64decode(base64_key)
+
+    policy_pattern = u'URLPrefix={encoded_url_prefix}:Expires={expires}:KeyName={key_name}'
```
tmatsuo, line 136: JFYI, from python 3.6, you can use [f-string](https://www.python.org/dev/peps/pep-0498/)

Then it will be
```python
policy = (
    f'URLPrefix={encoded_url_prefix}:Expires={expiration_timestamp}:'
    f'KeyName={key_name}'
)

```

mpwarres: TIL--thanks!

PR #3291, opened 2020-04-07
[iam] fix: reuse the same custom role for read/update tests

### LINKED ISSUE #2885: iam.api-client.custom_roles_test: test_list_roles failed
iam.api-client.custom_roles_test: test_list_roles failed
buildID: 4aa16fd61986dad63823af87e70fdd9f549604a4
buildURL: https://source.cloud.google.com/results/invocations/5ab7025a-38dd-480d-b56b-b89161fd76a4
status: failed

### LINKED ISSUE #2886: iam.api-client.custom_roles_test: test_delete_role failed
iam.api-client.custom_roles_test: test_delete_role failed
buildID: 4aa16fd61986dad63823af87e70fdd9f549604a4
buildURL: https://source.cloud.google.com/results/invocations/5ab7025a-38dd-480d-b56b-b89161fd76a4
status: failed

tmatsuo:
fixes #2885
fixes #2886

This PR will reduce the number of temporary custom roles to half.
Not a fundamental fix, but it will definitely mitigate the above issues.


iam/api-client/custom_roles_test.py
```diff
@@ -33,23 +33,21 @@
...
+            GCLOUD_PROJECT,
+            CUSTOM_ROLE_TITLE,
+            CUSTOM_ROLE_DESCRIPTION,
+            CUSTOM_ROLE_PERMISSIONS,
+            CUSTOM_ROLE_STAGE,
+        )
     except googleapiclient.errors.HttpError:
-        print("Custom role already deleted.")
-
+        # Ignore error since we just reuse the same custom role.
```
kurtisvg, line 47: Can we make sure that we are ignoring the right error? Perhaps verify the message is correct and the snippet isn't failing for another reason?

tmatsuo: Sure! I'll make a change

tmatsuo:
@kurtisvg @leahecole PTAL

tmatsuo:
lol the other test failed. I filed https://github.com/GoogleCloudPlatform/python-docs-samples/issues/3292

This is unreleated, so I'll just re-run the test.

iam/api-client/custom_roles_test.py
```diff
@@ -33,22 +33,23 @@
...
+            CUSTOM_ROLE_TITLE,
+            CUSTOM_ROLE_DESCRIPTION,
+            CUSTOM_ROLE_PERMISSIONS,
+            CUSTOM_ROLE_STAGE,
+        )
+    except googleapiclient.errors.HttpError as e:
+        if "HttpError 409" not in str(e):
+            raise e
+        # Ignore error since we just reuse the same custom role.
+        pass
```
leahecole, line 50: Instead of a pass, please make this an informational print statement

tmatsuo: It's almost always not printed, but I added it :)

tmatsuo:
Ironically we reached the 300 limit because of the lots of builds due to PRs for fixing this very issue.

I think we need to only run create/delete tests (which consume the quota) in the nightly builds. I'll add some changes for it in this PR later today.

tmatsuo:
Still getting "Maximum number of roles reached. Maximum is: 300"

More flakes I fixed, less likely this test will pass ...

tmatsuo:
I'm running the test in question 100 times locally for repro.

tmatsuo:
@leahecole Now the test passes! I addressed your comment, so I'm going to dismiss your review and merge this because it's a great chance :)

PR #3299, opened 2020-04-07
[iot] fix: mitigate flaky tests

### LINKED ISSUE #3129: iot.api-client.manager.manager_test: test_add_config_unauth_device failed
This test failed!

To configure my behavior, see [the Build Cop Bot documentation](https://github.com/googleapis/repo-automation-bots/tree/master/packages/buildcop).

If I'm commenting on this issue too often, add the `buildcop: quiet` label and
I will stop commenting.

---

commit: ac54e6e24fa7e126a627520e91e425107ff149ea
buildURL: [Build Status](https://source.cloud.google.com/results/invocations/66a85dbf-8b14-43ff-8ead-8abe95383829), [Sponge](http://sponge2/66a85dbf-8b14-43ff-8ead-8abe95383829)
status: failed

tmatsuo:
fixes #3129 

* use uuid1 instead of time for resource name
* correctly use the pytest marker
* mark another test as flaky


iot/api-client/manager/manager_test.py
```diff
@@ -33,7 +32,7 @@
 es_cert_path = 'resources/ec_public.pem'
 rsa_cert_path = 'resources/rsa_cert.pem'
 rsa_private_path = 'resources/rsa_private.pem'  # Must match rsa_cert
-topic_id = 'test-device-events-{}'.format(int(time.time()))
+topic_id = 'test-device-events-{}'.format(uuid.uuid1())
```
crwilcox, line 35: is uuid1 sufficiently random? uuid1 is a hash of machine address and time. I think it also has another identifier, but it may collide. I think we use uuid4 in a lot of the other tests for this reason.

tmatsuo: I thought the original author was thinking about the case where they need to manually delete the leaked resources, so I thought we should use time based uuid1, but whatever, I can use uuid4. Another commit is on the way.

tmatsuo:
@crwilcox Thanks! PTAL

PR #3311, opened 2020-04-08
[monitoring] testing: mitigate 409 conflicts

### LINKED ISSUE #2971: monitoring.api.v3.alerts-client.snippets_test: test_enable_alert_policies failed

tmatsuo:
fixes #2971

### DIFF
--- a/monitoring/api/v3/alerts-client/snippets_test.py
+++ b/monitoring/api/v3/alerts-client/snippets_test.py
@@ -118,20 +118,23 @@ def test_list_alert_policies(capsys, pochan):
     assert pochan.alert_policy.display_name in out
 
 
-@pytest.mark.flaky(rerun_filter=delay_on_aborted, max_runs=5)
 def test_enable_alert_policies(capsys, pochan):
+    @retry(wait_exponential_multiplier=1000, wait_exponential_max=10000,
+           stop_max_attempt_number=5, retry_on_exception=retry_if_aborted)
+    def invoke_sample(val):
+        snippets.enable_alert_policies(pochan.project_name, val)
     # These sleep calls are for mitigating the following error:
     # "409 Too many concurrent edits to the project configuration.
     # Please try again."
     # Having multiple projects will void these `sleep()` calls.
     # See also #3310
     time.sleep(2)
-    snippets.enable_alert_policies(pochan.project_name, True)
+    invoke_sample(True)
     out, _ = capsys.readouterr()
     assert "Enabled {0}".format(pochan.project_name) in out
 
     time.sleep(2)
-    snippets.enable_alert_policies(pochan.project_name, False)
+    invoke_sample(False)
     out, _ = capsys.readouterr()
     assert "Disabled {}".format(pochan.project_name) in out
 
--- a/monitoring/api/v3/alerts-client/snippets_test.py
+++ b/monitoring/api/v3/alerts-client/snippets_test.py
@@ -118,25 +118,24 @@ def test_list_alert_policies(capsys, pochan):
     assert pochan.alert_policy.display_name in out
 
 
+@pytest.mark.flaky(rerun_filter=delay_on_aborted, max_runs=5)
 def test_enable_alert_policies(capsys, pochan):
-    @retry(wait_exponential_multiplier=1000, wait_exponential_max=10000,
-           stop_max_attempt_number=5, retry_on_exception=retry_if_aborted)
-    def invoke_sample(val):
-        snippets.enable_alert_policies(pochan.project_name, val)
     # These sleep calls are for mitigating the following error:
     # "409 Too many concurrent edits to the project configuration.
     # Please try again."
     # Having multiple projects will void these `sleep()` calls.
     # See also #3310
     time.sleep(2)
-    invoke_sample(True)
+    snippets.enable_alert_policies(pochan.project_name, True)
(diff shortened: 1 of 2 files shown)


monitoring/api/v3/alerts-client/snippets_test.py
```diff
@@ -102,15 +103,23 @@ def test_enable_alert_policies(capsys, pochan):
     def invoke_sample(val):
         snippets.enable_alert_policies(pochan.project_name, val)
 
+    # These sleep calls are for mitigating the following error:
+    # "409 Too many concurrent edits to the project configuration.
+    # Please try again."
+    # TODO(#3310): Having multiple projects will void these `sleep()` calls
```
kurtisvg, line 110: nit: I prefer not putting TODO(issues) in code - IMO the issue should have a one way link to the code in question, but if we decide to close this issue for different reason, another commit is required to remove this TODO

monitoring/api/v3/alerts-client/snippets_test.py
```diff
@@ -102,15 +103,23 @@ def test_enable_alert_policies(capsys, pochan):
...
+    time.sleep(2)
     invoke_sample(False)
+    time.sleep(2)
```
kurtisvg, line 113: Do we need to run the snippets twice? Why isn't once sufficient enough?

tmatsuo:
ugh, the setup is seeing `google.api_core.exceptions.ServiceUnavailable: 503 Deadline Exceeded`.
Now it's retrying on that error too.

monitoring/api/v3/alerts-client/snippets_test.py
tmatsuo: OK

monitoring/api/v3/alerts-client/snippets_test.py
tmatsuo: Ask original author ;)

My understanding is that, the first call will disable it correctly, and the second call is for `assert "already disabled" in out`?

I'm just preserving the original test.

tmatsuo:
@kurtisvg Thanks! I just removed the TODOs. PTAL

monitoring/api/v3/alerts-client/snippets_test.py
kurtisvg: If original test is flaky, why not replace it instead of trying to preserve the original?

tmatsuo: Removing one call won't help much, but I can remove some.

tmatsuo:
@kurtisvg Thanks, I removed 2 API calls. PTAL

monitoring/api/v3/alerts-client/snippets_test.py
```diff
@@ -126,6 +131,12 @@ def invoke_sample():
         snippets.replace_notification_channels(
             pochan.project_name, alert_policy_id, [notification_channel_id])
 
+    # This sleep call is for mitigating the following error:
+    # "409 Too many concurrent edits to the project configuration.
+    # Please try again."
+    # Having multiple projects will void this `sleep()` call.
+    # See also #3310
+    time.sleep(2)
```
kurtisvg, line 139: Do we still need the sleep if we are only calling the snippet once? 

How does the sleep differ from the @retry annotation on `invoke_sample`? Would it be better to just retry the entire test?

tmatsuo: > Do we still need the sleep if we are only calling the snippet once?
I think it's still safe to have sleep because if we have n builds simultaneously, there will be 2*n api calls in a short period of time.

Retrying entire test sounds good. Maybe I can mark them as flaky with some delay.

kurtisvg: Flakey can do something like this:

```python
def delay_on_exception(err, *args):
    if issubclass(err[0], ProductCrashedError):
       time.sleep(1)
       return True # will retry
    return False

@flaky(rerun_filter=delay_on_exception)
def test_something_else():
```

tmatsuo:
@kurtisvg 
I just mark them as `flaky` with some delay.
The test passes locally. A little concern is that the test depends on the default "enabled" state.

PTAL

tmatsuo:
The log says:
```
test_enable_alert_policies failed and was not selected for rerun.
```
I think my implementation for rerun_filter is wrong, looking.

tmatsuo:
The delete test is failing with the concurrent error. Maybe we can have longer wait.

tmatsuo:
This is a tough one. With the new commits, we have longer wait time and better teardown function.

tmatsuo:
The error is:

```
Policy projects/python-docs-samples-tests/alertPolicies/10275611718925515819 is already enabled
```

I think we should keep using retrying for this test.

tmatsuo:
I understand what's happening. The first `disable` call failed (but it suceeded on the server side), then the following retry failed with assertion.

Maybe we can just allow both messages.

PR #3278, opened 2020-04-03
[jobs] testing: reduce test flakiness

tmatsuo:
fixes #2945
fixes #2818

There were 8 tests which has `time.sleep(10)` for waiting data
propagation or whatever. I chose to use `@eventually_consistent.call`
instead of sleep in those tests.

Bonus point is now the tests are significantly faster because it doesn't
always wait for 10 seconds.

The downside is it will take a long time on failure. We may want to
specify the retry count on the eventualy consistent decorator once the
following issue is fixed:

https://github.com/GoogleCloudPlatform/python-repo-tools/issues/25

PR #3317, opened 2020-04-09
[jobs] testing: add smaller tries to eventually_consistent.call

### LINKED ISSUE #3316: [jobs] testing: add `tries` argument on `@eventually_consistent.call` decorator
I started to use `@eventually_consistent.call` in 
https://github.com/GoogleCloudPlatform/python-docs-samples/pull/3278

Now gcp-devrel-py-tools has a new release 0.0.16 and it accepts arguments on the decorator. It is better to provide smaller `tries` argument (default = 10).

tmatsuo:
fixes #3316 

followup on #3278

### DIFF
--- a/jobs/v3/api_client/auto_complete_sample_test.py
+++ b/jobs/v3/api_client/auto_complete_sample_test.py
@@ -28,7 +28,7 @@ def company_name():
 
 
 def test_auto_complete_sample(company_name, capsys):
-    @eventually_consistent.call
+    @eventually_consistent.call(tries=5)
     def _():
         auto_complete_sample.run_sample(company_name)
         out, _ = capsys.readouterr()
--- a/jobs/v3/api_client/commute_search_sample_test.py
+++ b/jobs/v3/api_client/commute_search_sample_test.py
@@ -28,7 +28,7 @@ def company_name():
 
 
 def test_commute_search_sample(company_name, capsys):
-    @eventually_consistent.call
+    @eventually_consistent.call(tries=5)
     def _():
         commute_search_sample.run_sample(company_name)
         out, _ = capsys.readouterr()
--- a/jobs/v3/api_client/custom_attribute_sample_test.py
+++ b/jobs/v3/api_client/custom_attribute_sample_test.py
@@ -28,7 +28,7 @@ def create_data():
 
 
 def test_custom_attribute_sample(create_data, capsys):
-    @eventually_consistent.call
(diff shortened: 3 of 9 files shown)


jobs/v3/api_client/auto_complete_sample_test.py
```diff
@@ -28,7 +28,7 @@ def company_name():
 
 
 def test_auto_complete_sample(company_name, capsys):
-    @eventually_consistent.call
+    @eventually_consistent.call(tries=5)
```
kurtisvg, line 31: Can we switch to `flaky` or `retry` over `eventually_consistent `?

tmatsuo: @kurtisvg Thanks for the comment. To be honest, we have different opinions.

I think eventually_consistent has some benefit. It gives semantics.

It marks the tests as eventually consistent, which means it is a correct thing to retry.
I quite don't like to use `flaky` or `retry` for eventually consistent tests, because we can not distinguish eventually consistent tests from usual flakes.

Also, it is a great shortcut for eventually consistent issues with exponential back off. There's no need for defining helper callback.

So, for eventually consistent tests, I strongly *want* to stick with `eventually_consistent`.

tmatsuo:
The error:

    assert re.search('COMPANY_ID', out)
AssertionError: assert None
 +  where None = <function search at 0x7fce75338e18>('COMPANY_ID', "Retrying due to eventual consistency.\n{'metadata': {'requestId': '880e787a-99f3-48ea-8d79-a2623e95d1be:APAb7IR0Cpj1XDbIgvr/m71HC7iIXvI+nQ=='}}\n")

It seems like the test is failing correctly? The API response seems different from usual one.

tmatsuo:
Filed: https://github.com/GoogleCloudPlatform/python-docs-samples/issues/3318

It passes 100 times locally. I'd just re-run the test for now.

jobs/v3/api_client/auto_complete_sample_test.py
tmatsuo: So, if we ultimately decide to eliminate gcp-devrel-py-tools, I'll replace `eventually_consistent` with `retrying`. However, I'll stick to eventally_consistent in this PR.

Can you file an issue for eliminating gcp-devrel-py-tools so that we can discuss further there?

tmatsuo:
We decided to remove gcp-devrel-py-tools from the repo, closing.

PR #3319, opened 2020-04-09
fix: use `crypto_key_path` instead of `crypto_key_path_path`

busunkim96:
`crypto_key_path_path` was added to the library in error and will be removed in a future version of the library


irvifa:
Hi, wondering why kokoro build is only failing on python 3.6 here.

busunkim96:
@irvifa It looks like a resource collision from the 3.7 and 3.6 tests running concurrently.
```
google.api_core.exceptions.FailedPrecondition: 400 The request cannot be fulfilled. Resource projects/python-docs-samples-tests/locations/global/keyRings/kms-samples/cryptoKeys/symmetric/cryptoKeyVersions/1 has value DESTROY_SCHEDULED in field crypto_key_version.state.
```

tmatsuo:
@busunkim96 If you saw resource conflicts between builds, can you file an issue so that we'll be able to fix later?

PR #3328, opened 2020-04-10
switch to regular flaky in dlp

leahecole:
partial fix for #3375


tmatsuo:
FYI, there is `@pytest.mark.flaky` pytest marker, and we should use it.

Anyways, I'm planning to take over the issues around dlp (because there are other issues), so maybe we can close this issue.

tmatsuo:
@leahecole Closing this, but feel free to reopen.

PR #3329, opened 2020-04-10
switch to regular flaky in dns

leahecole:


leahecole:
looks like resource collision. i'll take a look.

PR #3333, opened 2020-04-10
switch to regular flaky in monitoring

leahecole:


monitoring/api/v3/api-client/list_resources_test.py
```diff
@@ -56,7 +56,7 @@ def test_list_metrics(client, capsys):
         client, PROJECT_RESOURCE, METRIC)
     stdout, _ = capsys.readouterr()
     regex = re.compile(
-        u'Delta CPU', re.I)
+        u'Delta vCPU', re.I)
```
tmatsuo, line 59: Nit: I think it's better to remove `vCPU` part (only check 'Delta')? But we can do it in another PR

PR #3339, opened 2020-04-10
Switch continuous tests to only test changed files.

### LINKED ISSUE #3314: testing: only run tests which has code changes in continuous builds
In presubmit builds, we only run tests which has code changes. I think we should do the similar thing in continuous builds too.

This will significantly reduce the build time, and avoid consuming cloud resources.
For example, this will significantly mitigate issues like: https://github.com/GoogleCloudPlatform/python-docs-samples/pull/3291

kurtisvg:
- Switches continuous tests to only test files changed from the last commit 
- Switches buildcop to only report fails from the periodic build

Fixes #3314

### DIFF
--- a/.kokoro/tests/run_tests.sh
+++ b/.kokoro/tests/run_tests.sh
@@ -20,11 +20,16 @@ set -eo pipefail
 # Enables `**` to include files nested inside sub-folders
 shopt -s globstar
 
-# `--only-changed` will only run tests on projects container changes from the master branch.
-if [[ $* == *--only-diff* ]]; then
-    ONLY_DIFF="true"
-else
-    ONLY_DIFF="false"
+DIFF_FROM=""
+
+# `--only-diff-master will only run tests on project changes from the master branch.
+if [[ $* == *--only-diff-master* ]]; then
+    DIFF_FROM="origin/master.."
+fi
+
+# `--only-diff-master will only run tests on project changes from the previous commit.
+if [[ $* == *--only-diff-head* ]]; then
+    DIFF_FROM="HEAD~.."
 fi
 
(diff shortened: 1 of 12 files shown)


.kokoro/tests/run_tests.sh
```diff
@@ -67,8 +72,8 @@ for file in **/requirements.txt; do
     cd "$file"
 
     # If $DIFF_ONLY is true, skip projects without changes.
-    if [[ "$ONLY_DIFF" = "true" ]]; then
-        git diff --quiet origin/master.. .
+    if [[ "DIFF_FROM" != "" ]]; then
```
tmatsuo, line 75: `${DIFF_FROM}` instead of `DIFF_FROM`

* need `$`
* nit: also better to use braces

.kokoro/tests/run_tests.sh
```diff
@@ -67,8 +72,8 @@ for file in **/requirements.txt; do
...
+        git diff --quiet "$DIFF_FROM" .
```
tmatsuo, line 76: Nit (you can ignore for now): better to use braces `"${DIFF_FROM}"`

.kokoro/tests/run_tests.sh
kurtisvg: Good catch, thank you

kurtisvg:
[Presubmit check still workin](https://source.cloud.google.com/results/invocations/19a0c1b8-3d6c-44b9-a3b5-66f16169d4ca/log)

kurtisvg:
[Verified continuous tests are working](https://source.cloud.google.com/results/invocations/5faa4984-0e17-4395-932d-040d5b87930d/targets)

PR #3382, opened 2020-04-13
Update Data Catalog samples to V1

mesmacosta:
With release of V1 library, we must update the samples to instruct users to use the supported version.

[ x] Tests pass
[ n/a] Appropriate changes to README are included in PR
[ n/a] API's need to be enabled to test (tell us)
[ n/a] Environment Variables need to be set (ask us to set them)


tmatsuo:
Merging

PR #3383, opened 2020-04-13
vision: increase timeout for tests

nnegrey:
Fix: https://github.com/GoogleCloudPlatform/python-docs-samples/issues/2955
Fix: https://github.com/GoogleCloudPlatform/python-docs-samples/issues/2992


vision/cloud-client/product_search/product_search_test.py
```diff
@@ -30,6 +31,7 @@
 FILTER = 'style=womens'
 
 
+@flaky(max_runs=5, min_passes=1)
```
tmatsuo, line 34: If flaky makes the test pass, I'm quite happy to merge the PR because this flake is a different issue from the timeout.

However, it looks like this test is testing the result of the Vision API. Does it make sense to only test if the API call itself succeed? I'm totally fine if you just file a new issue and fix it with another PR.

nnegrey: From the output of the failures, it seems to be the API call isn't returning any results. So the backend seems to be flaky sometimes, I couldn't reproduce locally though, so I figured this should help.

tmatsuo: Got it.

PR #3385, opened 2020-04-13
video: use flaky on flaky timeouts

nnegrey:
Fix: https://github.com/GoogleCloudPlatform/python-docs-samples/issues/3346
Fix: https://github.com/GoogleCloudPlatform/python-docs-samples/issues/3345
Fix: https://github.com/GoogleCloudPlatform/python-docs-samples/issues/3344
Fix: https://github.com/GoogleCloudPlatform/python-docs-samples/issues/3298
Fix: https://github.com/GoogleCloudPlatform/python-docs-samples/issues/2934


nnegrey:
@frankyn, I got an error when trying to delete the bucket after the test that the Bucket is not empty. 
But the code uses:
```
bucket.delete(force=True)
```
Any ideas?

frankyn:
Hi @nnegrey how many objects are the tests working with and do video tests share the bucket with other samples? 

It's possible to use force delete and still have objects if for example a Temporary Hold or Event-based Hold are set on objects. Force doesn't account for these type of objects. I can't tell from the error which type of failure it is though.

nnegrey:
Each test is supposed to create a temporary bucket and each test should only be creating maybe a handful of output files. The failing test only creates one though. 
```
# Create a temporary bucket to store annotation output.
    bucket_name = str(uuid.uuid1())
    storage_client = storage.Client()
    bucket = storage_client.create_bucket(bucket_name)
    yield bucket
    # Teardown.
    bucket.delete(force=True)
```

nnegrey:
Might've been a weird flaky on the API or storage as I've never seen that error before and this test has been around for a while. Going to merge to fix several bugs. If it comes back up, I'm sure Build Cop will let me know and I'll investigate more from there.

PR #3414, opened 2020-04-16
Correct path for the warmup call

oittaa:


oittaa:
@googlebot I signed it!

appengine/standard_python37/warmup/main_test.py
```diff
@@ -28,5 +28,5 @@ def test_warmup():
     main.app.testing = True
     client = main.app.test_client()
 
-    r = client.get('/')
+    r = client.get('/_ah/warmup')
```
busunkim96, line 31: For other reviewers: this is the `warmup` route in `main.py`

https://github.com/GoogleCloudPlatform/python-docs-samples/blob/00e73771765623a8e7011d29c980ce439f1277e6/appengine/standard_python37/warmup/main.py#L27-L30

PR #3423, opened 2020-04-17
remove samples/tests that aren't on devsite, incl. localized docs

Strykrol:
Remove automl samples (+tests) that aren't on cloud.google.com in en/localized as part of snippet clean up work.


Strykrol:
@jsimonweb & @nnegrey  - Full sample application deletions here. Code isn't on cgc at all (incl. localizations) 

May have use for other sample- project.... For now submitting PR as-is.

PR #3426, opened 2020-04-17
fix: address issues with concurrent BigQuery tests

shollyman:
This PR changes how testing is done for the BQ tests which now run multiple instances concurrently.

Basic premise is we inject new values into the code under test, and use the template syntax to elide the replacement from the embedded representations.  the test harness takes care of deleting resources.


leahecole:
I'm here for it, but I tend to have less strong opinions about this than some of the other folks. @tmatsuo has done a LOT of refactoring tests recently, so I want his eyes too to double check

bigquery/cloud-client/authorized_view_tutorial_test.py
```diff
@@ -14,6 +14,7 @@
 
 from google.cloud import bigquery
 import pytest
+import uuid
```
tmatsuo, line 17: Can you do the following?

```python
import uuid

from google.cloud import bigquery
import pytest

import authorized_view_tutorial
```

Usually imports should be 1) standard libs 2) 3rd party libs and 3) package local libs, separated by an empty line.

bigquery/cloud-client/authorized_view_tutorial_test.py
```diff
@@ -24,24 +25,24 @@ def client():
...
     doomed = []
     yield doomed
     for item in doomed:
-        if isinstance(item, (bigquery.Dataset, bigquery.DatasetReference)):
-            client.delete_dataset(item, delete_contents=True)
-        elif isinstance(item, (bigquery.Table, bigquery.TableReference)):
-            client.delete_table(item)
-        else:
-            item.delete()
+        client.delete_dataset(item, delete_contents=True)
```
tmatsuo, line 32: Is this client automatically retrying transient errors? If not, what happens if delete fails?

bigquery/cloud-client/natality_tutorial_test.py
```diff
@@ -13,30 +13,33 @@
 # limitations under the License.
 
 from google.cloud import bigquery
-from google.cloud import exceptions
+import pytest
```
tmatsuo, line 16: Can you do as follows?

```python
import uuid

from google.cloud import bigquery
import pytest

import natality_tutorial
```

bigquery/cloud-client/natality_tutorial_test.py
```diff
@@ -13,30 +13,33 @@
...
-def test_natality_tutorial():
-    client = bigquery.Client()
-    dataset_ref = client.dataset('natality_regression')
-    assert not dataset_exists(dataset_ref, client)
+@pytest.fixture
+def datasets_to_delete(client):
+    doomed = []
+    yield doomed
+    for item in doomed:
+        client.delete_dataset(item, delete_contents=True)
```
tmatsuo, line 32: Same comment as above

bigquery/cloud-client/quickstart_test.py
```diff
@@ -13,8 +13,8 @@
 # limitations under the License.
 
 from google.cloud import bigquery
-from google.cloud.exceptions import NotFound
 import pytest
+import uuid
```
tmatsuo, line 17: ditto

shollyman:
> Looks good overall.
> 
> If the client library doesn't automatically retry transient failures, I'd retry few times upon such exceptions.
> I found that [backoff](https://pypi.org/project/backoff/) is a decent library for this purpose.
> 
> The code looks like:
> 
> ```python
>     @backoff.on_exception(backoff.expo, max_time=60)
>     def delete_dataset():
>         try:
>             client.delete_dataset(item, delete_contents=True)
>         except NotFound as e:
>             # We ignore this case.                                                                                                                                 
>             print("The dataset doesn't exist: detail: {}".format(str(e)))
>     delete_dataset()
> ```
> 
> If the client library already retries transient errors, just ignore my comment.

Yeah, BQ uses a default retry policy.

https://googleapis.dev/python/bigquery/latest/generated/google.cloud.bigquery.retry.DEFAULT_RETRY.html

bigquery/cloud-client/natality_tutorial_test.py
shollyman: Done

bigquery/cloud-client/quickstart_test.py
shollyman: Done

bigquery/cloud-client/authorized_view_tutorial_test.py
shollyman: Yes, commented elsewhere this library has a default retry policy.

PR #3429, opened 2020-04-18
GCF: add multipart file-persistence note

ace-n:
Given that we added this note in the [Java sample](https://github.com/GoogleCloudPlatform/java-docs-samples/pull/2688), I thought it was worth adding here.


functions/http/main.py
```diff
@@ -85,6 +85,9 @@ def parse_multipart(request):
     # This code will process each file uploaded
     files = request.files.to_dict()
     for file_name, file in files.items():
+        # Note: GCF may not persist saved files across invocations.
+        # Persistent files must be kept in other locations
+        # (such as Cloud Storage buckets).
```
tmatsuo, line 90: This is better but this sample code is still using the wrong way?
Can you also add something like "In this sample however, we use local filesystem for simplicity"?

tmatsuo: Or even, can you show one of the recommended way?

ace-n: Not all files are necessarily persistent - for example, maybe you have a function that processes image files with Cloud Vision API and saves the vision API's output _(as opposed to the image itself)_ somewhere.

Maybe `persistent` is an ambiguous term here that could be replaced with something clearer? e.g:

```
Note: GCF may not keep files saved locally between invocations.
If you want to preserve the uploaded files, you should save them
to another location (such as a Cloud Storage bucket).
```

tmatsuo: Thanks, your new suggestion looks better to me.

tmatsuo: @ace-n So are you going to update the comment?

ace-n: Done - sorry about the delay!

PR #3433, opened 2020-04-18
Update python code sample to include request method field.

slz250:
Note that the client lib doesn't seem to have been updated yet despite the tracking software stating otherwise. Will look into this.


tmatsuo:
@slz250 Is this PR ready to merge?

slz250:
Sorry about this. I'm not too familiar with git and am having some trouble removing a merge commit. My revert command ended up sending review requests to too many people. I then tried to do a git rebase based on this [SO answer](https://stackoverflow.com/a/51400593/4335969) but it looks like for a merge commit, I'll have to drop all of the commits that is being merged. 

This seems like more effort than it's worth so I'll just make the changes in a new PR.

PR #3445, opened 2020-04-20
Adding noxfile for Cloud Run

dinagraves:
Runs the tests for all the sub-directories of the run directory.

### DIFF
--- a/run/noxfile.py
+++ b/run/noxfile.py
@@ -1,0 +1,194 @@
+# Copyright 2019 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from __future__ import print_function
+
+import os
+from pathlib import Path
+
+import nox
+
+
+# DO NOT EDIT - automatically generated.
+# All versions used to tested samples.
+ALL_VERSIONS = ["2.7", "3.6", "3.7", "3.8"]
+
+# Any default versions that should be ignored.
+IGNORED_VERSIONS = ["2.7"]
+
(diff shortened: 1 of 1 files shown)


averikitsch:
I'm not an expert for noxfiles, so I default to @busunkim96

tmatsuo:
@dinagraves Looking at the [build log](https://source.cloud.google.com/results/invocations/11cbbdd4-806f-4da0-bd2c-23930b32dfd2/targets/cloud-devrel%2Fpython-docs-samples%2Fpython3.7%2Fpresubmit/log), it seems that no tests were executed.

By reading [run_tests.sh](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/.kokoro/tests/run_tests.sh), maybe you want to rename `noxfile.py` to `noxfile-template.py` then the test runner will copy the right file for you.

Adding @kurtisvg  for authority.

kurtisvg:
No tests were run because there where no changes in the actual samples.

@tmatsuo is correct in that you need to have the file named "noxfile-template.py" and that noxfile will be used to run samples in child directories. 

However this noxfile doesn't seem to configure versions or add any functionality not in the root noxfile-template.py. What's the goal of adding this noxfile to the repo?

dinagraves:
@kurtisvg Perhaps I misunderstood the instructions.  My impression was that every product directory needed its own noxfile:
https://chat.google.com/room/AAAAnhOHkh0/H7dVDJF4FJM

And in the authoring guide: https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/AUTHORING_GUIDE.md#running-tests

Can we not have a copy of it in the repo that we can use to run the tests locally rather than having do a copy and paste process every time we create a new sample? 

I have a sample that I want to submit, but needed the noxfile to run the tests.

kurtisvg:
> @kurtisvg Perhaps I misunderstood the instructions. My impression was that every product directory needed its own noxfile:
> https://chat.google.com/room/AAAAnhOHkh0/H7dVDJF4FJM
> 
> And in the authoring guide: https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/AUTHORING_GUIDE.md#running-tests
> 
> Can we not have a copy of it in the repo that we can use to run the tests locally rather than having do a copy and paste process every time we create a new sample?
> 
> I have a sample that I want to submit, but needed the noxfile to run the tests.

We are in a (hopefully) temporary state of having the noxfile-template from a parent pom being used to run the tests. The end goal is to have synth-tool generate a noxfile.py for each sample automatically.

PR #3498, opened 2020-04-24
fix: use DELETE FROM for consistency

skuruppu:
Being consistent with [docs](https://cloud.google.com/spanner/docs/dml-syntax#delete_examples).


skuruppu:
I believe the failing tests are due to https://github.com/googleapis/python-spanner/issues/68. @larkee may ask you for some advice on this topic, hope that's ok @tmatsuo.

tmatsuo:
@skuruppu Actually, the error is different.
There are two kind of timeout (or DeadlineExceededError)

1. create_backup() ends up with a DeadlineExceededError.
2. operation.result(timeout) ends up with timeout.

https://github.com/googleapis/python-spanner/issues/68 is for fixing 1.

The errors on this PR is the issue 2 above. I think this is a transient failure on the server side.

tmatsuo:
@skuruppu I updated the branch, and will merge once the test passes.

tmatsuo:
LOL now the both tests are failing with `google.api_core.exceptions.DeadlineExceeded: 504 Deadline Exceeded`, which will be fixed by https://github.com/googleapis/python-spanner/issues/68

PR #3547, opened 2020-04-24
Markdown tutorial

dinagraves:
Sample for https://cloud.google.com/run/docs/tutorials/secure-services

Includes end-to-end testing. 

Update:

This PR has a custom `noxfile.py` at `run/markdown-preview/noxfile.py`.

Currently this custom noxfile is hijacking py-3.7 test for running `cloud_run` test, but eventually we'll move to `cloud_run` nox session.

In the `cloud_run` nox session, I needed to specify the test file `e2e_test.py` in order to exclude tests in subdirs (renderer and editor). This directory hierarchy is for our test shell script to correctly detect whether it should run the test or not according to the changed files.

@kurtisvg @busunkim96 Can you take a look at the `noxfile.py`?


run/markdown-preview/renderer/main.py
```diff
@@ -0,0 +1,34 @@
...
+import os
+
+
+app = Flask(__name__)
+
+
+@app.route("/", methods=["POST"])
+def index():
+    data = request.get_data(as_text=True)
+    return markdown.markdown(data)
```
averikitsch, line 26: nit: comment what `markdown.markdown()` is doing.

averikitsch, line 26: You also need to sanitize this HTML

run/markdown-preview/editor/render.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+def new_request(data):
+    """
+    new_request creates a new HTTP request with IAM ID Token credential.
+    This token is automatically handled by private Cloud Run (fully managed)
+    and Cloud Functions.
+    """
+
+    url = os.environ.get("EDITOR_UPSTREAM_RENDER_URL")
```
averikitsch, line 28: We should check that url is set

run/markdown-preview/README.md
```diff
@@ -0,0 +1,6 @@
+# Markdown Preview
+
+This sample application consists of two services: a "markdown editor" and a separate "markdown renderer".
+
+Read more about how to deploy and work with these services in https://cloud.google.com/run/docs/tutorials/secure-services.
+
```
averikitsch, line 6: We should mention the env vars used for this sample

run/markdown-preview/renderer/main.py
dinagraves: Done - using bleach

dinagraves: Added comment

run/markdown-preview/README.md
dinagraves: Added. LMK if it's not clear or if you think it should be in its own "environment variables" section

run/markdown-preview/renderer/main.py
averikitsch: bleach.clean should be called on the html outputted from markdown.markdown.

run/markdown-preview/editor/requirements.txt
```diff
@@ -0,0 +1,4 @@
+Flask==1.1.1
+pytest==5.3.2; python_version > "3.0"
```
averikitsch, line 2: Is this needed if using a separate requirements file? Same for renderer.

run/markdown-preview/e2e_test_setup.yaml
```diff
@@ -0,0 +1,57 @@
+steps:
```
averikitsch, line 1: nit: format this file similar to https://github.com/GoogleCloudPlatform/cloud-code-samples/blob/master/python/.ci/cloud-run-python-hello-world.cloudbuild.yaml

dinagraves: Took a stab at it. LMK if you want any changes.

run/markdown-preview/renderer/main.py
dinagraves: I think it works either way - cleaning before sending to markdown should remove any harmful tags.  I'm not sure how aggressive bleach is and whether it removes anything we want from the markdown html?  It passes the test though so am running it on the markdown html.

run/markdown-preview/editor/requirements.txt
dinagraves: Good call - fixed this.  Probably should do the rest of the samples as well.

tmatsuo:
@dinagraves Can you use `uuid.uuid4().hex` instead of random int? If it's too long, use slice `uuid.uuid4().hex[:5]`.

run/markdown-preview/e2e_test.py
```diff
@@ -0,0 +1,91 @@
...
+import pytest
+
+
+@pytest.fixture()
+def services():
+    # Unique suffix to create distinct service names
+    suffix = uuid.uuid4().hex
+
+    # Build and Deploy Cloud Run Services
+    subprocess.run(
```
kurtisvg, line 29: We probably need to make sure [`check=True`](https://docs.python.org/3/library/subprocess.html#subprocess.run) or that we check the error codes manually when these commands return.

run/markdown-preview/e2e_test_setup.yaml
```diff
@@ -0,0 +1,59 @@
...
```
kurtisvg, line 1: This probably needs a license at the top

run/markdown-preview/e2e_test.py
kurtisvg: (same for all these calls below)

run/markdown-preview/editor/requirements-test.txt
```diff
@@ -0,0 +1,2 @@
+pytest==5.3.2; python_version > "3.0"
+pytest==5.3.2; python_version < "3.0"
```
kurtisvg, line 2: Only using Python 3.8, no need to clarify this any more. 
```suggestion
pytest==5.3.2
```

run/markdown-preview/noxfile.py
```diff
@@ -0,0 +1,186 @@
...
+
+import nox
+
+
+# DO NOT EDIT - automatically generated.
+# All versions used to tested samples.
+ALL_VERSIONS = ["2.7", "3.6", "3.7", "3.8"]
+
+# Any default versions that should be ignored.
+IGNORED_VERSIONS = ["2.7", "3.6", "3.8"]
```
kurtisvg, line 28: The container is running 3.8-slim, so we probably only want to test on 3.8?
```suggestion
IGNORED_VERSIONS = ["2.7", "3.6", "3.7"]
```

run/markdown-preview/renderer/requirements-test.txt
```diff
@@ -0,0 +1,2 @@
+pytest==5.3.2; python_version > "3.0"
+pytest==5.3.2; python_version < "3.0"
```
kurtisvg, line 2:
```suggestion
pytest==5.3.2
```

run/markdown-preview/requirements-test.txt
```diff
@@ -0,0 +1,4 @@
+pytest==5.3.2; python_version > "3.0"
+pytest==5.3.2; python_version < "3.0"
```
kurtisvg, line 2:
```suggestion
pytest==5.3.2
```

run/markdown-preview/e2e_test.py
```diff
@@ -0,0 +1,91 @@
...
+# limitations under the License.
+
+import json
+import subprocess
+from urllib import request
+import uuid
+
+import pytest
+
+
```
kurtisvg, line 22: nit: Since this is a somewhat uncommon test structure (or at least at this point), would it be helpful to add a comment towards the top of this file (or maybe on the test itself) explaining what the goal of this test is and why's it located here? 

It may be helpful to other readers that don't have the context on the situation.

run/markdown-preview/noxfile.py
dinagraves: We're actually hijacking the 3.7 tests, and running the cloud_run session instead.  If we update here, we would have this in the py() function as well:
```
if session.python == '3.7':
        cloud_run(session)
```

kurtisvg: Oh ok - do we intended to commit using this or can we add the Cloud Run check to Kokoro before?

run/markdown-preview/e2e_test.py
dinagraves: Added a comment. LMK if it needs more clarification.

run/markdown-preview/noxfile.py
dinagraves: What do you think is easier to maintain? @tmatsuo for his opinion as well.

tmatsuo: I'm ok with either way, with a slight preference of merging this as it is.

kurtisvg: Ok - SGTM.

tmatsuo:
I guess the CLA bot is not happy with some commits done by Dine on the github UI, which has `Dina Graves Portman <55413432+dinagraves@users.noreply.github.com>`.

tmatsuo:
@googlebot I consent.

PR #3564, opened 2020-04-27
functions/slack: use env vars + consistent var naming

ace-n:
**Do not merge** until @labtopia has made the necessary docs changes.

Part of b/145664680


ace-n:
@labtopia has LGTMed internally. ✅ [cl/308870671]

functions/slack/main.py
```diff
@@ -14,21 +14,16 @@
...
-with open('config.json', 'r') as f:
-    data = f.read()
-config = json.loads(data)
-
 
 kgsearch = googleapiclient.discovery.build(
     'kgsearch',
     'v1',
-    developerKey=os.environ.get('API_KEY') or config.get('KG_API_KEY'),
+    developerKey=os.environ['KG_API_KEY'],
```
busunkim96, line 26: This will result in a `KeyError` if this isn't in the environment. 

Is passing a API key the recommended way to use this API? If there are credentials in the environment you can do `googleapiclient.discovery.build(...)` and the library will use ADC

ace-n: This API _does_ require an API key. (I got an error when removing the `developerKey` parameter.)

PR #3578, opened 2020-04-29
 docs(storage): add samples for lifer cycle and versioning 

HemangChothani:
Added Samples For

-  Enable, Disable Versioning

-  Enable,Disable life cycle management

-  Set blob metadata


leahecole:
@frankyn or @crwilcox I'd like your eyes on this before I do the final stamp - just want to make sure it looks good from a storage standpoint

PR #3581, opened 2020-04-29
Background function: return on completion

ace-n:


functions/helloworld/main.py
```diff
@@ -98,6 +98,7 @@ def hello_pubsub(event, context):
     else:
         name = 'World'
     print('Hello {}!'.format(name))
+    return
```
tmatsuo, line 101: I think this doesn't change anything. I mean, if you don't have the `return`, it will return None.

ace-n: Agreed. I'm not sure [the docs](https://cloud.google.com/functions/docs/writing/background#functions_background_terminate-python) are 100% correct, but this PR is blocking the Java launch.

I've filed b/155338265 to track this; @tmatsuo feel free to contribute info for Python! 🙂

tmatsuo:
I don't think we need to merge this. Closing. Feel free to reopen if I'm wrong.

PR #3655, opened 2020-04-30
[storage] fix: use a different bucket for requester_pays_test

### LINKED ISSUE #3654: storage.cloud-client.requester_pays_test should use a different bucket
Currently storage.cloud-client.requester_pays_test is using the same Storage bucket as other tests.
If the test failed to disable it, many other tests will fail because of this.

tmatsuo:
fixes #3654

@frankyn FYI

This PR also has small changes in `scripts/decrypt-secrets.sh` and `scripts/encrypt-secrets.sh` to always use `sha256` for the digest.


storage/cloud-client/requester_pays_test.py
```diff
@@ -23,7 +23,14 @@
 import storage_download_file_requester_pays
 import storage_get_requester_pays_status
 
-BUCKET = os.environ["CLOUD_STORAGE_BUCKET"]
+
+# We should not use the same bucket as other tests are using.
+# First look at REQUESTER_PAYS_TEST_BUCKET and fall back to
+# ${CLOUD_STORAGE_BUCKET} + '-requester-pays-test'.
+BUCKET = os.environ.get(
+    "REQUESTER_PAYS_TEST_BUCKET",
```
busunkim96, line 31: Should this be added to https://github.com/GoogleCloudPlatform/python-docs-samples/blob/bb83cea15029f50e8f82c6c2d50c971b35b48c33/testing/test-env.tmpl.sh?

tmatsuo: Good catch, I'll add

storage/cloud-client/README_TEST.md
```diff
@@ -0,0 +1,16 @@
+
```
kurtisvg, line 1: Why `README_TEST`? If users need this to run the tests, we should mention it in the main README

storage/cloud-client/requester_pays_test.py
```diff
@@ -23,7 +23,14 @@
 import storage_download_file_requester_pays
 import storage_get_requester_pays_status
 
-BUCKET = os.environ["CLOUD_STORAGE_BUCKET"]
+
+# We should not use the same bucket as other tests are using.
+# First look at REQUESTER_PAYS_TEST_BUCKET and fall back to
```
kurtisvg, line 28: Is falling back to a default really necessary? Can't we just have an explicit check?

storage/cloud-client/README_TEST.md
tmatsuo: Sure, I can do it

storage/cloud-client/requester_pays_test.py
tmatsuo: That works too, but it requires us to add the environment variable in the Kokoro config. I can clean up later. Does it work for you?

tmatsuo:
@busunkim96 @kurtisvg Thanks, PTAL

storage/cloud-client/requester_pays_test.py
kurtisvg: You don't need to do it as part of the kokoro config, you can add it to the [secrets tar](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/02131911100d8c9ac94e1a4b2dc18a0eb3c09f7e/.kokoro/tests/run_tests.sh#L42).

You can use the `decrypt-secrets.sh` to decrypt the current secrets, add that secret in, and use `encrypt-secrets.sh` to re-encrypt. Then just commit the new tar and the secret will be included in the run.

leahecole: +1 to Kurtis's comment - I think going for a fallback is starting to go down the rabbit hole of all of the checks we could be doing

tmatsuo: TIL about the secrets.tar :) Will do

tmatsuo:
@kurtisvg OK, now we have REQUESTER_PAYS_TEST_BUCKET in the secret file. I also updated the decrypt/encrypt scripts to use sha256. We'll see how it works in the Kokoro builds.

PTAL

PR #3667, opened 2020-05-01
fix: bump ipython version to address compilation issues with py3.8

shollyman:
Fixes: #3658
Fixes: #3659
Fixes: #3660


tmatsuo:
@shollyman Thanks for the quick fix!
Unfortunately, now we still don't run python-3.8 presubmit, but I confirmed that it's passing on my workstation. LGTM

tmatsuo:
I don't know why, but there's no jobs queued in Kokoro. Let me update the branch, so that Kokoro will run the test again. I'll merge once the tests pass.

PR #3668, opened 2020-05-01
Test labeling: appengine/flexible

ace-n:
@leahecole FYI - please merge **prior to** your planned fixit.


tmatsuo:
@ace-n  
What's your motivation for switching to unittest class?

ace-n:
None in particular - should I be using a different class here?

tmatsuo:
Per offline chat, we decided not to use the class style tests, closing.

PR #3678, opened 2020-05-04
chore: add a script for injecting region tags

tmatsuo:
Adding a script for modifying the junit result xml.

If the test doesn't have any properties, it's a no-op. If the test has any properties, this script will inject the region tags into the testname.


tmatsuo:
Aha, the `sponge_log.xml` files in the test_data directory didn't get added.

tmatsuo:
@busunkim96 

If we will need this script, yes, but I hope not. I mean I'm proposing a different design that will require us to only add `property` field to the test.

tmatsuo:
@ace-n Agreed to read the `property` field in the junit.xml, so I'm closing this PR.

PR #3690, opened 2020-05-05
Update and add Cloud KMS samples

sethvargo:
This updates the Cloud KMS samples to match the format from the other 6 languages. It also updates the samples to note the workaround for https://github.com/googleapis/gapic-generator-python/issues/364.

I dropped all "runnable" samples except the quickstart. Only about half of the previous samples included them, and most were broken. I don't think KMS developers are running samples on the command line. Furthermore, KMS often deals in bytes (not strings), which requires developers to base64 encode/decode the data when using it on the CLI. This leads to confusion when developers try to use the API directly, since the API accepts raw bytes. The top questions on Python+KMS on Stackoverflow are entirely around this confusion.

There's also a bug in the new noxfile template, which I fixed in here.


sethvargo:
CI failures seem to be quota related.

kms/api-client/quickstart.py
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 
-# Copyright 2017 Google, Inc
+# Copyright 2020 Google LLC
```
tmatsuo, line 3: Here and there
Please don't update the copyright year. 
I think it is ok to keep the copyright declaration.
If you really want to add Google LLC bit, you can do it, but you have to keep the original year and the original name.

kms/api-client/snippets_test.py
```diff
@@ -1,6 +1,4 @@
-#!/usr/bin/env python
-
-# Copyright 2017 Google, Inc
+# Copyright 2020 Google LLC
```
tmatsuo, line 1: ditto

kms/api-client/snippets_test.py
```diff
@@ -13,226 +11,410 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 
+import hashlib
+import os
+import pytest
```
tmatsuo, line 16: `pytest` is a 3rd party library, so please place it in the 2nd block.

```python
import hashlib
import os
import time
import uuid

import pytest

import create... # local imports
```

kms/api-client/quickstart.py
tmatsuo: I'm not a lawyer, so If you unsure, consult OSPO

kms/api-client/snippets_test.py
sethvargo: Done

kms/api-client/snippets_test.py
sethvargo: Done

kms/api-client/quickstart.py
sethvargo: Done

PR #3691, opened 2020-05-05
Update logging example to retrieve the default handler

darrencarlton:
This change makes the code sample usable for App Engine apps as well as other environments.

When a Logging client asks for the default handler, Logging returns a handler that is appropriate for the environment the client is running in. When the code is running in App Engine, AppEngineHandler is returned, and this is needed to enable key logging features in App Engine, such as correlating app logs with request logs.

Here's info about get_default_handler():
https://googleapis.dev/python/logging/latest/client.html

This change also updates branding to "Cloud Logging".


logging/cloud-client/handler.py
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 
-# Copyright 2016 Google Inc. All Rights Reserved.
+# Copyright 2020 Google Inc. All Rights Reserved.
```
tmatsuo, line 3: Don't update the copyright year

kurtisvg: +1

PR #3699, opened 2020-05-06
Gae python version update

engelke:
App Engine standard for Python 3.7 samples all now use Python 3.8. Folder name changed, replacing the _python37 suffix with just _python3.


crwilcox:
```
./main.py:43:1: E302 expected 2 blank lines, found 1
```

Issue #3310, opened 2020-04-08
Consider having multiple projects for testing


tmatsuo:
For some tests, we ultimately need to have multiple projects for testing.

* "409 Too many concurrent edits to the project configuration. Please try again"
* Maximum number of custom roles (300) reached

I created a label `needs-multiple-projects` which indicates that the issue needs multiple projects for the fundamental fix.

tmatsuo:
@kurtisvg @crwilcox @leahecole If you know any issues which needs multiple projects for the fundamental fix, please add the label `needs-multiple-projects`. Once you go through your knowledge, feel free to unassign yourself.

tmatsuo:
For implementation, I think we should use [gimmeproj](https://github.com/GoogleCloudPlatform/golang-samples/tree/master/testing/gimmeproj). Example uses in [ruby](https://github.com/GoogleCloudPlatform/ruby-docs-samples/tree/master/spec) and [go](https://github.com/GoogleCloudPlatform/golang-samples/blob/master/testing/kokoro/system_tests.sh#L132-L145) 

To use gimmeproj, I think we'll probably need to
1) create a new project for meta project
2) try to run all the tests on the meta project, figure out what configs/operations/resource files are needed besides enabling APIs, and write them down (for example, TEST-REQUIREMENTS.md in each directory).
3) maybe standardize on the resource files (or even, write a shell script for setup)
4) then we can finally create pool projects with all the needed stuff

tmatsuo:
An easier option.

Maybe we can have few more cloud projects with suffixes like `-py36`, `-py37`, etc.

We can have a common library in this repo, probably in the `lib` folder or somewhere. We can add that directory to `sys.path` in the `noxfile.py` so that you can just do something like:

```python

import test_helper


project_id = test_helper.get_project_id_from_kokoro_job_name()
# you'll get `xxx-py36` for the py-36 build and `xxxx-py37` for the py-37 build, etc.
```

This way, it is very easy for us to only opt in some of the tests.

tmatsuo:
Came up with an easier option.

Rather than having a library, we can just override the envvar `GOOGLE_CLOUD_PROJECT` in `noxfile.py`.

tmatsuo:
Created 3 new projects.

- python-docs-samples-tests-py36
- python-docs-samples-tests-py37
- python-docs-samples-tests-py38

I'll start using those projects with monitoring tests.

tmatsuo:
It's live now.

PR #3700, opened 2020-05-06
testing: multi project noxfile-template.py

tmatsuo:
This is a part of https://github.com/GoogleCloudPlatform/python-docs-samples/issues/3310#issuecomment-624917137

This PR also adds `cloud-run` session.


tmatsuo:
Note: Python 3.8 build should fail because the service account has no permission on my personal project.

tmatsuo:
Oh I have type in the project names, also there's no python3.8 presubmit.
Now python3.7 build should fail.

tmatsuo:
The python3.7 build failed as expected.

tmatsuo:
[py-3.6 successful build](https://source.cloud.google.com/results/invocations/97cace3a-2fed-4f7d-adbf-84a2c256c20d) with our main project

[py-3.7 failed build](https://source.cloud.google.com/results/invocations/9924ed53-8c0d-4658-a344-6a0632652de8) with my personal project

I'll add the permission and run the test again.

tmatsuo:
Both builds succeeded.
[py-3.6](https://source.cloud.google.com/results/invocations/983325d9-2909-4ea1-8c7e-2dab80256cff)
[py-3.7](https://source.cloud.google.com/results/invocations/fdb511fa-9e80-4c50-93d2-55f2813fe7fc)

monitoring/api/v3/alerts-client/noxfile.py
```diff
@@ -0,0 +1,198 @@
...
+        post_install(session)
+
+    session.run(
+        "pytest",
+        *(PYTEST_COMMON_ARGS + session.posargs),
+        # Pytest will return 5 when no tests are collected. This can happen
+        # on travis where slow and flaky tests are excluded.
+        # See http://doc.pytest.org/en/latest/_modules/_pytest/main.html
+        success_codes=[0, 5],
+        env=get_pytest_env_vars()
```
kurtisvg, line 151: Does this override the current environment or is it just set addition?

monitoring/api/v3/alerts-client/noxfile.py
```diff
@@ -0,0 +1,198 @@
...
+import os
+from pathlib import Path
+
+import nox
+
+# Proposed TEST_CONFIG.
+# You can have a file `testconfig` or something and it will be injected into
+# the generated noxfile.py.
+TEST_CONFIG = {
+    'use_multi_project': True,
```
kurtisvg, line 26: Rather than `use_multi_project: bool`, could we make this something like `override_project_variable: str` and pass in an env var to reference instead?

So something like:
```python
TEST_CONFIG = {
    `override_project_variable`: "VERSION_PYTHON_SAMPLES_TEST",
    # ...
} 

# ...
ret = {}
if TEST_CONFIG.get('override_project_variable'):
    ret[GOOGLE_CLOUD_PROJECT] = os.environ.get(TEST_CONFIG['override_project_variable'])

```

I think something like this might be preferable because:
1. We can customize the project per build, and don't have to regenerate all of the noxfiles when we make a change
2. It lets us keep the project id used private

tmatsuo: Good call for using an envvar.

Maybe we can just use the an envvar `BUILD_SPECIFIC_GCLOUD_PROJECT`, and keep the config simple.

```python
TEST_CONFIG = {
    'use_multi_project': True,
    ...
}

MULTI_PROJECT_OVERRIDE = "BUILD_SPECIFIC_GCLOUD_PROJECT"

    # then use it when the env var exists.
    if TEST_CONFIG.get('use_multi_project'):
        ret[GOOGLE_CLOUD_PROJECT] = os.environ.get(
            MULTI_PROJECT_OVERRIDE, os.environ.get('GOOGLE_CLOUD_PROJECT'))

```

Unless we want to change that environment variable name (`BUILD_SPECIFIC_GCLOUD_PROJECT`).

monitoring/api/v3/alerts-client/noxfile.py
tmatsuo: I think it'll set addition, but override the same keys.

monitoring/api/v3/alerts-client/noxfile.py
tmatsuo: Maybe config key should be `use_build_specific_project`? but I think you get the point.

kurtisvg: That seems okay - I do feel like a little more flexibility might be nice, but I can't think of a specific example off the top of my head. 

 I do think we should avoid falling back to `GOOGLE_CLOUD_PROJECT` if they've opted in - I'd rather throw an error to make it clear something is wrong.

kurtisvg: RE: using a `str` over `bool`, we may need to specify two different fall back projects for two different projects. With `true` we can only have one back up per build - by specifying and env var we'll have more freedom to have more fallback projects.

One example of this is currently the AutoML tests use a different project that the AutoML team maintains. We could specify one env var for the AutoML project, and still supply a different one for monitoring tests.

tmatsuo: It seems like automl tests are using "AUTOML_PROJECT_ID".

Can we just treat them as a different thing?

tmatsuo:
@kurtisvg @busunkim96 

I modified the `noxfile-template.py` and added the default `.test_config.py`.
PTAL

tmatsuo:
Now I have a file named `.test_config.py` to "hide" it from other tooling. Like hiding from coverage tool, etc.

But I also think it's very attractive to use a valid file name as python module. For example, if the file name is `test_config.py`, we can easily do the following in the `noxfile.py`:

```python

try:
    from test_config import TEST_CONFIG_OVERRIDE
except ImportError:
    TEST_CONFIG_OVERRIDE = {}
```
This is very attractive because there's no need to manipulate the noxfile-template at all.

WDYT @kurtisvg @busunkim96 

Also, we may not need to copy the `noxfile-template.py` at all how about to use nox's `-f` option?

monitoring/api/v3/alerts-client/noxfile.py
kurtisvg: We could, but it them means we have to expect the user to know to set `"AUTOML_PROJECT_ID"`. If we keep it wider, the samples could consistently refer to `"GOOGLE_CLOUD_PROJECT"` which would be more consistent (and hopefully intuitive) to other users, while the automl samples could take advantage of this noxfile feature to still customize the project.

test_config.py
```diff
@@ -0,0 +1,51 @@
...
+    #     'cloud_run' # Test session for Cloud Run application.
+    'opt_in_sessions': [],
+
+    # Only relevant for the `cloud_run` session. Specify the file
+    # names for your e2e test.
+    'cloud_run_e2e_test_files': ['e2e_test.py'],
+
+    # If set to True, the test will install the library from the root
+    # of the repository.
+    'install_library_from_source': False,
```
kurtisvg, line 38: I _think_ this functionality should be configurable per build - not _per location_? Sometimes we want the same project to be tested one way and sometimes another, so I think this config should only be an environment variable.

noxfile-template.py
```diff
@@ -116,6 +177,52 @@ def py(session):
...
+        # Activate service account
+        key_file = os.environ['GOOGLE_APPLICATION_CREDENTIALS']
+        session.run(
+            "gcloud", "auth", "activate-service-account",
+            "--key-file={}".format(key_file)
+        )
+
+        # Set gcloud project
+        project = user_envs['GCLOUD_PROJECT']
+        session.run("gcloud", "config", "set", "project", project)
```
kurtisvg, line 213: I don't think we should be updating gcloud in a test session - this could lead to non-deterministic test results, because now whenever a Cloud Run session is run gcloud could be at a different version for the tests after that. 

If we need a more updated version we should handle that either in the test script or on docker build.

test_config.py
busunkim96: The intent of this particular setting was to allow for the library (e.g., `google-cloud-storage`) to be installed from source `pip install -e ..` instead of from PyPI. It will be `True` in client library repos and `False` in python-docs-samples.

If you'd prefer this happen automagically I could move the detection logic into the noxfile. Searching for a `setup.py` in the repo root is probably sufficient

test_config.py
```diff
@@ -0,0 +1,51 @@
...
+
+    # Set to True if you want to use the Cloud Project configured for each
+    # build.
+    'use_build_specific_project': False,
+
+    # An envvar key for determining the build specific project. Normally you
+    # don't have to modify this.
+    'build_specific_project_env': 'BUILD_SPECIFIC_GCLOUD_PROJECT_ENV',
```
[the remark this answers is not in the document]
busunkim96: +1, would prefer a single value

test_config.py
kurtisvg: Didn't we decide that samples in the client library repos would have 2 different builds: "head vs head" and "published vs published"?

My point was that those two builds are happing to the same tests (so same config), so I think this config can't be in the file (it needs to come from the build, probably via an env var).

busunkim96: Ah I see your point though. I think we can yank this from the config and have the noxfile look for `INSTALL_LIBRARY_FROM_SOURCE` in the environment. 👍 

@tmatsuo, would you mind removing this?

tmatsuo:
Currently py3.7 build is using my personal project and it should fail because I revoked the permissoin.

tmatsuo:
Well, it doesn't work as expected on Kokoro. Locally the test failed when I set BUILD_SPECIFIC_GCLOUD_PROJECT.

I need to think about why.

tmatsuo:
I gave up being an armchair detective. I'm print-debugging now.

test_config.py
```diff
@@ -0,0 +1,51 @@
...
+    'ignored_versions': ["2.7"],
+
+    # Declare optional test sessions you want to opt-in. Currently we
+    # have the following optional test sessions:
+    #     'cloud_run' # Test session for Cloud Run application.
+    'opt_in_sessions': [],
+
+    # Only relevant for the `cloud_run` session. Specify the file
+    # names for your e2e test.
+    'cloud_run_e2e_test_files': ['e2e_test.py'],
```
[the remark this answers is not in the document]
busunkim96: Two ideas, but not particularly attached to either:

* Ask that all end to end tests have 'e2e' in the filename. Glob for files that match the pattern.

  ```py
  from pathlib import Path

  e2e_tests = list(Path('.').glob("*e2e*.py"))
  ```

* Enforce a flat directory structure for all samples projects. (I _think_ this is already the norm?) and use [`norecursedirs`](https://docs.pytest.org/en/latest/reference.html#confval-norecursedirs)

  *pytest.ini*
  ```
  [pytest]

  norecursedirs =  *
  ```

tmatsuo:
Ok, now the py-3.7 buid is correctly failing.

tmatsuo:
@kurtisvg @busunkim96 

This PR is almost ready. `.kokoro/python3.7/common.cfg` is still using my personal project, but it is to make sure the mechanism correctly works. I'll revert it to our test project before merging.

Can you take another look?

tmatsuo:
`detect_test.py::test_async_detect_document` is constantly failing. Maybe I can disable it and file an issue for re-enabling it.

noxfile-template.py
```diff
@@ -113,8 +181,53 @@ def py(session):
...
+    user_envs = get_pytest_env_vars()
+
+    # Only update gcloud on Kokoro.
+    if os.environ.get("KOKORO_JOB_NAME"):
+        # Activate service account
+        key_file = os.environ['GOOGLE_APPLICATION_CREDENTIALS']
+        session.run(
+            "gcloud", "auth", "activate-service-account",
+            "--key-file={}".format(key_file)
+        )
```
kurtisvg, line 215: The auth library should be handling this already - we shouldn't need to do this as a part of the test

noxfile-template.py
```diff
@@ -113,8 +181,53 @@ def py(session):
...
+
+        # Set gcloud project
+        project = user_envs['GCLOUD_PROJECT']
+        session.run("gcloud", "config", "set", "project", project)
```
kurtisvg, line 219: I'd rather the test itself take in the environment variable and throw an error if it doesn't exist. The test can use a flag to set this so it doesn't have to permanently change the gcloud config when it's run. 

Also, I _think_ `GOOGLE_CLOUD_PROJECT` is preferred over `GCLOUD_PROJECT` IIRC

noxfile-template.py
tmatsuo: It's for gcloud command itself, not for the library usage. The current Cloud Run sample author prefers to use gcloud for deploying the app.

noxfile-template.py
tmatsuo: Sure, using a command line flag seems good. I'll remove this.

tmatsuo:
@kurtisvg Stopped setting gcloud project. PTAL

noxfile-template.py
kurtisvg: Apologies, I misunderstood how this was being used. I thought this was being used to set the service account used in the Run instance, but it's actually being used to authenticate the glcoud command to deploy the instance.

In this case, it would be best to handle this in our test script, instead of in the noxfile. The `gcloud auth` makes changes to the local gcloud config, meaning that running this test has side effects and might effect the other tests. If we set this [here](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/2e502b1da42d5de8db1de3e49c6a020e87714c67/.kokoro/tests/run_tests.sh#L45) instead, it's consistent for all of the tests.

tmatsuo: Ok, I can do that.

monitoring/api/v3/alerts-client/noxfile.py
tmatsuo: Now it's using str

test_config.py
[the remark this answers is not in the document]
tmatsuo: I like the 2nd option, I think we can allow users to bring `pytest.ini` and `.flake8` file.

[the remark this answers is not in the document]
tmatsuo: I think users can bring the following pytest.ini:

```
[pytest]

norecursedirs =  *

[flake8]
extend-exclude = myapp1, myapp2
```
to have the desired effect. Thanks @busunkim96 for inspiration!

tmatsuo:
@kurtisvg @busunkim96 I think I addressed all of your concerns. PTAL

noxfile-template.py
```diff
@@ -76,11 +140,25 @@ def lint(session):
...
+    # [flake8]
+    # extend-ignore = I100
+    # extend-exclude = myapp1,myapp2
+    if os.path.isfile("pytest.ini"):
+        options += [
+            "--append-config",
+            "pytest.ini",
+        ]
+    options.append(".")
+    args = FLAKE8_COMMON_ARGS + options
```
kurtisvg, line 161: nit: IMO, I don't think we should let style checks be configurable, but we can discuss on the next sync

tmatsuo: I agree in general, but it's required for the cloud-run case.

Example:
```
my_proj
\
 e2e_test.py
 requirements.txt
 my_app1
 \
  main.py
  main_test.py
  requirements.txt
```

The lint session for the parent my_proj will report that `import main` is in the wrong place because it doesn't recognize the local import files correctly.

So in this case having following `pytest.ini` is the cleanest solution IMO:

```
[pytest]

norecursedirs =  *

[flake8]
extend-exclude = myapp1
```
This files allows to skip subdirectories both for pytest and flake8.

tmatsuo:
The test failed with "google.api_core.exceptions.Aborted: 409 Too many concurrent edits to the project configuration. Please try again.", but this is within expectation. Restarting the build.

PR #1608, opened 2018-08-02
skip automl model create/delete test

sirtorry:
skipping tests for now as they may create multiple models.

PR #3721, opened 2020-05-11
Re-enable AutoML model create tests

### LINKED ISSUE #3706: Re-enable model test
`python-docs-samples/vision/automl/model_test.py`

## Describe the issue

After  [Pull Request 3693](https://github.com/GoogleCloudPlatform/python-docs-samples/pull/3693), there is only one test in this file: `test_model_create_status_delete`, and it is currently skipped. 

This should be re-enabled.

busunkim96:
Closes #3706.

I don't have the context around why these tests were disabled. Does these tests need to be modified in some other way? 

https://github.com/GoogleCloudPlatform/python-docs-samples/pull/1608 disabled these tests.


tmatsuo:
Few things.

1. The tests are using the following code:

```
model_name = "test_" + datetime.datetime.now().strftime("%Y%m%d%H%M%S")
```
This easily conflicts among multiple builds. Can you update this to use uuid?

2. The tests are calceling the operation, but it's a best effort trial of cancellation, so sometimes the models are created before the cancellation happens. I think we should try to delete the models in teardown code.

translate/automl/model_test.py
```diff
@@ -26,29 +26,32 @@
...
-    # create model
+@pytest.fixture
+def model():
     client = automl.AutoMlClient()
-    model_name = "test_" + datetime.datetime.now().strftime("%Y%m%d%H%M%S")
+    model_name = "test_{}".format(uuid.uuid4()).replace("-", "")[:32]
     project_location = client.location_path(project_id, compute_region)
     my_model = {
         "display_name": model_name,
         "dataset_id": "3876092572857648864",
```
busunkim96, line 36: It looks like this dataset doesn't exist. Is there some standard AutoML dataset I can use? @sirtorry @nnegrey

sirtorry:
these samples are outdated. samples and tests can be deleted. translate [doc](https://cloud.google.com/translate/automl/docs/models#automl_translate_create_model-python) and vision [doc](https://cloud.google.com/vision/automl/docs/train) reference other snippets.

tmatsuo:
@sirtorry 
Thanks for the input.

I did a quick search for the region tag, and there are still used in multiple location. Although they're all translated page, can we really delete them now?

sirtorry:
I'm going to submit a request for these samples to be replaced on our localised/non-english documentation. We shouldn't remove the samples until that's been done. In this case, should we re-enable these tests?

busunkim96:
Opened https://github.com/GoogleCloudPlatform/python-docs-samples/issues/3766 to track removing those samples.

I'm alright with closing this and leaving the tests disabled. @tmatsuo Is that alright with you?

tmatsuo:
@busunkim96 Yes, it's alright!

PR #3731, opened 2020-05-12
fix make_iap_request

3a3:
Fixes https://github.com/GoogleCloudPlatform/python-docs-samples/issues/3732


kurtisvg:
@leahecole - FYI

leahecole:
I need to double check it in the context of the tutorial

functions/composer/composer_storage_trigger.py
```diff
@@ -21,7 +21,7 @@
 import google.oauth2.credentials
 import google.oauth2.service_account
 import requests
-
+from google.oauth2 import id_token
```
arithmetic1728, line 24: I think this import line needs to be put right after line 20, and there should be 2 empty lines between `import requests` and `IAM_SCOPE=...`

3a3: Updated

PR #3745, opened 2020-05-12
Add request_method to create example.

slz250:


busunkim96:
@slz250 Thanks for opening a PR! Could you provide some detail on what you're looking to do?

slz250:
> @slz250 Thanks for opening a PR! Could you provide some detail on what you're looking to do?

Hi! I'm on the Cloud Monitoring - Uptime team and will like to update the code samples to include the explicit setting of the `request_method` field.

PR #3751, opened 2020-05-13
chore: some lint fixes

tmatsuo:


tmatsuo:
`detect_test.py::test_async_detect_document` is constantly failing. Maybe I can disable it and file an issue for re-enabling it.

tmatsuo:
Flake reported: https://github.com/GoogleCloudPlatform/python-docs-samples/issues/3754

Restarting the build.

PR #3767, opened 2020-05-15
cleanup: remove stale code block

### LINKED ISSUE #3341: Switch functions tests to Python 3.6+
Currently functions is using a check in the bashfile - instead, we should use the noxfile-templates to control this:

https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/.kokoro/tests/run_tests.sh#L79-87

tmatsuo:
fixes #3341


tmatsuo:
Running py27 periodic:

PR #3772, opened 2020-05-16
[iam] testing: use multiple projects

tmatsuo:
part of #3310


tmatsuo:
py36 project has `Role Administrator`

py37 project has `Security Admin`

My guess is we need `Security Admin`.

tmatsuo:
It seems like you need 3 roles.

tmatsuo:
It seems like `list_service_accounts` errors out with an empty result. Maybe I can just create a dummy service account in each project then it's fine.

leahecole:
removing automerge label until there's a second approval

iam/api-client/noxfile_config.py
```diff
@@ -0,0 +1,42 @@
...
+    # have the following optional test sessions:
+    #     'cloud_run' # Test session for Cloud Run application.
+    'opt_in_sessions': [],
+
+    # An envvar key for determining the project id to use. Change it
+    # to 'BUILD_SPECIFIC_GCLOUD_PROJECT' if you want to opt in using a
+    # build specific Cloud project. You can also use your own string
+    # to use your own Cloud project.
+    # 'gcloud_project_env': 'GCLOUD_PROJECT',
+    'gcloud_project_env': 'BUILD_SPECIFIC_GCLOUD_PROJECT',
```
gguuss, line 37: Just to clarify, this overrides for running locally but doesn't impact CI?

tmatsuo: It affects both.

On Kokoro, it will use one of the python version specific cloud projects (e.g. python-docs-samples-tests-py36).

When you run it locally, you have to set `BUILD_SPECIFIC_GCLOUD_PROJECT` envvar which is little bit cumbersome, but you'll know this by the error message easily, and you can do:

```
export BUILD_SPECIFIC_GCLOUD_PROJECT=${GOOGLE_CLOUD_PROJECT}
```
then it's fine for most cases.

Issue #3442, opened 2020-04-20
Split testing container from testing-infra-docker


kurtisvg:
Currently we use the currently use this [Dockerfile](https://github.com/googleapis/testing-infra-docker/tree/master/python/googleapis/python-multi) for testing. However we really should have our own Dockerfile with our own dependencies here in python-docs-samples. 

Here are the steps that need to be taken to do this:
1. Create a dockerfile based on python:3.6 here in the repo with the dependencies necessary for our tests (maybe in root or `.kokoro`?)
2. Create a Kokoro continuous job that checks if the Dockerfile has changed, and if so pushes a new version of the container
3. Update our tests to use this image instead of the current `python-multi`.

tanjinP:
@kurtisvg I can take this on - looks pretty straightforward.

tmatsuo:
Can we close this now?

PR #3776, opened 2020-05-16
Add testing Dockerfile and Cloud Build configuration

shubha-rajan:
Towards #3442 

I tested the cloud build trigger on my own fork and a separate project and it works - we just need to add a trigger for this repo on the Cloud-Devrel-Kokoro-Resources project.

For now, the Dockerfile is just a copy of python-multi, but we can update it in a separate PR and update the kokoro configs to point to the new image as a final step.


.kokoro/docker/Dockerfile
```diff
@@ -0,0 +1,160 @@
...
+ENV CLOUDSDK_PYTHON python3.6
+RUN wget https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-sdk-$CLOUD_SDK_VERSION-linux-x86_64.tar.gz
+RUN tar xzf google-cloud-sdk-$CLOUD_SDK_VERSION-linux-x86_64.tar.gz
+RUN /google-cloud-sdk/install.sh
+ENV PATH /google-cloud-sdk/bin:$PATH
+
+# Enable redis-server on boot.
+RUN sudo systemctl enable redis-server.service
+
+CMD ["python3.6"]
```
tmatsuo, line 160: Possible to add a newline at the end?

.kokoro/docker/cloudbuild.yaml
```diff
@@ -0,0 +1,22 @@
...
+# limitations under the License.
+
+timeout: 7200s  # 2 hours
+
+steps:
+    - name: 'gcr.io/cloud-builders/docker'
+      args: ['build', '-t', 'gcr.io/$PROJECT_ID/python-samples-testing-docker', '.']
+      dir: '.kokoro/docker'
+images:
+     - 'gcr.io/$PROJECT_ID/python-samples-testing-docker'
```
tmatsuo, line 22: ditto

.kokoro/docker/Dockerfile
```diff
@@ -0,0 +1,160 @@
...
+    build-essential \
+    ca-certificates \
+    curl \
+    git \
+    graphviz \
+    libbz2-dev \
+    libdb5.3-dev \
+    libexpat1-dev \
+    libffi-dev \
+    liblzma-dev \
```
tmatsuo, line 40: Can we add `libmagickwand-dev` here?

Now we install it on the fly:
https://github.com/GoogleCloudPlatform/python-docs-samples/blob/13fb63a93bde73f7394ea9b94f154513f38224ff/.kokoro/tests/run_tests.sh#L58

.kokoro/docker/Dockerfile
shubha-rajan: done

.kokoro/docker/Dockerfile
shubha-rajan: done

tmatsuo:
@shubha-rajan After merging this, can you actually use this file for updating the image? If yes, then once finished, please let us know. We will probably run one of the periodic builds manually to see if the new image works. Then we can remove the unnecessary code in `run_tests.sh`.

tmatsuo:
@shubha-rajan When uploading the docker image, is it possible to have date and/or version in the name or tag? It is great if we have a way to rollback easily.

kurtisvg:
> @shubha-rajan When uploading the docker image, is it possible to have date and/or version in the name or tag? It is great if we have a way to rollback easily.

Commit number would be great. 

Also, taking a second look at this, what do we need to do to add the github actions trigger? Will it rebuild _every_ commit, or can we limit it to only changes in the dockerfile?

shubha-rajan:
@tmatsuo I can add versioning to the 

> > @shubha-rajan When uploading the docker image, is it possible to have date and/or version in the name or tag? It is great if we have a way to rollback easily.
> 
> +1, Commit number would be great.
> 
> Also, taking a second look at this, what do we need to do to add the github actions trigger? Will it rebuild _every_ commit, or can we limit it to only changes in the dockerfile?

I just updated the cloudbuild config to tag images with the shortened commit hash. And you can limit it to only changes to files in specified directories.

shubha-rajan:
For reference, these are the settings I used when setting up the trigger in my test project.
image
image

kurtisvg:
Ok, slick. I think we should merge this as is then and use a different PR to change the image name. 

@tmatsuo Sound good to you?

tmatsuo:
@kurtisvg Yes it sounds good!

shubha-rajan:
@tmatsuo @kurtisvg great! I'll merge this and open another PR to change the image name as soon as the tests pass.

PR #3808, opened 2020-05-18
functions: fix slack + pubsub tests

ace-n:


functions/pubsub/main_test.py
```diff
@@ -24,32 +24,29 @@
...
 
 
-@pytest.mark.skip("broken")
 def test_functions_pubsub_subscribe_should_print_message(capsys):
-    pubsub_message = MagicMock()
-    pubsub_message.data = base64.b64encode(b"Hello, world!")
+    pubsub_message = {"data": base64.b64encode(b"Hello, world!") }
 
-    main.subscribe(pubsub_message)
+    main.subscribe(pubsub_message, None)
```
grant, line 49: I don't think we need an explicit `None`. If we really do, that should be fixed instead.

ace-n:
@tmatsuo yes - I've removed that file. 🙂

functions/pubsub/main_test.py
tmatsuo: It doesn't look like `context` is used in the function. How about just:

```python
def subscribe(event, context=None):
    # ...
```
?

functions/pubsub/main.py
```diff
@@ -62,7 +61,7 @@ def publish(request):
 
 # [START functions_pubsub_subscribe]
 # Triggered from a message on a Cloud Pub/Sub topic.
-def subscribe(event, context):
+def subscribe(event):
```
tmatsuo, line 64: So why did we have `context` argument in the first place? Isn't there a case the functions service is sending the 2nd argument?

ace-n: **Update:** GCF itself sends `context` as a second positional argument, so we _cannot_ ignore it in the method signature.

functions/pubsub/main.py
```diff
@@ -62,7 +61,7 @@ def publish(request):
 
 # [START functions_pubsub_subscribe]
 # Triggered from a message on a Cloud Pub/Sub topic.
-def subscribe(event, context):
+def subscribe(event, _ = None):
```
tmatsuo, line 64: This is sample code, so even if we don't use the `context` (or 2nd positional) argument, I would make it clear that this is a context sent from the functions service.

ace-n: OK - so given that, can we put the `None`-handling logic in the test (instead of the sample) to avoid confusing users?

tmatsuo: What do you mean? Maybe pushing a commit is easier for me to understand?

functions/pubsub/main_test.py
tmatsuo: The function service is sending `context`, but our sample doesn't use it.
We decided to just pass None from our test.

tmatsuo:
@ace-n 
Thanks for the quick fix. I'll merge the master and then merge this PR.

PR #3822, opened 2020-05-19
testing: set and use REPORT_TO_BUILD_COP_BOT

### LINKED ISSUE #3821: Re-consider when to report to build-cop-bot
Currently we send the report in periodic builds.

https://github.com/GoogleCloudPlatform/python-docs-samples/blob/db3f47d05cfb299bd554a487cc573d46c93578ed/.kokoro/tests/run_tests.sh#L109-L112

I think it is better to use more descriptive environment variable, say `REPORT_TO_BUILD_COP_BOT` or something and we can set that environment variable in our job configurations.

This way, we can have a flexibility of only sending build cop report from particular jobs (although I'm not planning to disable in any of our periodic builds), as well as manually run periodic builds without reporting to build-cop-bot (we can set environment variable when manually running the build).

tmatsuo:
fixes #3821

This PR needs an internal Kokoro config change.


.kokoro/tests/run_tests.sh
```diff
@@ -104,9 +104,11 @@ for file in **/requirements.txt; do
...
     EXIT=$?
 
-    # If this is a periodic build, send the test log to the Build Cop Bot.
-    # See https://github.com/googleapis/repo-automation-bots/tree/master/packages/buildcop.
-    if [[ $KOKORO_BUILD_ARTIFACTS_SUBDIR = *"periodic"* ]]; then
+    # If REPORT_TO_BUILD_COP_BOT is set to "True", send the test log
+    # to the Build Cop Bot.
+    # See:
+    # https://github.com/googleapis/repo-automation-bots/tree/master/packages/buildcop.
+    if [[ "${REPORT_TO_BUILD_COP_BOT:-}" == "True" ]]; then
```
kurtisvg, line 111: nit: I _think_ the `:-` is redundant since there are already quotes around the string (but I'm by no means a bash expert.)
```suggestion
    if [[ "${REPORT_TO_BUILD_COP_BOT}" == "True" ]]; then
```

tmatsuo: Now we don't have the `-u` option, but when we add it, this script still works.

Not planning to add `-u` option, but it doesn't do any harms. WDYT?

PR #3833, opened 2020-05-19
Add 'INSTALL_LIBRARY_FROM_SOURCE' to noxfile-template

busunkim96:
This flag will only be used in client library repos.


noxfile-template.py
```diff
@@ -184,6 +188,11 @@ def py(session):
         ))
 
 
+def _is_client_library_repo():
```
tmatsuo, line 191: Is this check really needed?
We can just set `INSTALL_LIBRARY_FROM_SOURCE` to True for specific jobs, right?

This check look alright for now, but it will be problematic when
1) We somehow stopped having `setup.py`
2) We want to use this envvar for other sample project
3) Somehow python-docs-samples started to have setup.py at root

I would just rely on the envvar if possible.

busunkim96: 👍 Removed

PR #3842, opened 2020-05-20
[vision] chore: parameterized FROM, add cloudbuild.yaml

tmatsuo:
also update README with how to build and upload the images.

@genquan9 
Now the images are little bit more maintenable. Potentially we can parameterize the version number too.

### DIFF
--- a/vision/automl/edge_container_predict/Dockerfile
+++ b/vision/automl/edge_container_predict/Dockerfile
@@ -12,8 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-ARG FROM_SUFFIX
-FROM tensorflow/serving:1.14.0${FROM_SUFFIX}
+ARG TF_SERVING_IMAGE_TAG
+FROM tensorflow/serving:${TF_SERVING_IMAGE_TAG}
 
 ENV GCS_READ_CACHE_MAX_STALENESS 300
 ENV GCS_STAT_CACHE_MAX_AGE 300
--- a/vision/automl/edge_container_predict/cloudbuild.yaml
+++ b/vision/automl/edge_container_predict/cloudbuild.yaml
@@ -1,4 +1,4 @@
-# Copyright 2019 Google LLC
+# Copyright 2020 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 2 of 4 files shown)


vision/automl/edge_container_predict/Dockerfile
```diff
@@ -12,7 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM tensorflow/serving:1.14.0
+ARG FROM_SUFFIX
```
genquan9, line 15: FROM_SUFFIX to VERSION ?
And take 1.14.0 or 1.14.0-gpu as versions?

vision/automl/edge_container_predict/cloudbuild.yaml
```diff
@@ -0,0 +1,25 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+timeout: 3600s
+
+steps:
+    - name: 'gcr.io/cloud-builders/docker'
+      args: ['build', '-t', 'gcr.io/$PROJECT_ID/gcloud-container-1.14.0', '.']
```
genquan9, line 19: set VERSION for both cpu and gpu dockers?

vision/automl/edge_container_predict/Dockerfile
tmatsuo: Ah I see.

Other thoughts on this, see:
https://github.com/GoogleCloudPlatform/python-docs-samples/pull/3841

We are using renovate bot which constantly trying to update the versions. Does it make sense to update it to 1.15.0? If so, maybe we should have separate Dockerfiles for cpu and gpu without any ARGs.

WDYT?

tmatsuo: @genquan9 Also, I think it is better to run the test against the newly built images, and if it succeeds, we can push the new image to gcr.io.

tmatsuo:
@genquan9 PTAL

For a record, if we want to update the tensorflow image version, you can do something like the following:

```bash
gawk -i inplace '{ gsub(/1.14.0/, "1.15.0") }; { print }' *
```

tmatsuo:
The error message: "Error response from daemon: client version 1.40 is too new. Maximum supported API version is 1.38"

I'll add an env var to fix it.

PR #3847, opened 2020-05-20
GCF cloudbuild.yaml: clarify "dir" parameter

ace-n:


tmatsuo:
Is this file used for any tests? Probably not.

There is a region tag, so I assume it's used in our docs.
How do we know it's working as expected with this change?

ace-n:
@tmatsuo yes - this is used in our docs.

When I tested this previously (manually), I had to:
1. Create a test Cloud Source repo
1. Link it to a test Cloud Build trigger
1. Push a change to said Cloud Source repo 
1. Confirm the build worked

Historically, we've considered that as "too involved" for a testing workflow. Personally I wouldn't mind changing that (e.g. testing `gcloud` commands and such), but I would like to see that become a {clearly-articulated, team-wide} policy before I update the `functions` samples.

(Maybe it's worth creating an issue about this and discussing further there, as this PR is a fairly minor change?)

PR #3856, opened 2020-05-21
Give option to create a POST check.

slz250:
In the create call demo -- allow uncommenting of POST check code. This is a new feature and will help drive adoption.


tmatsuo:
As @gguuss mentioned, it might be better to have another sample with a test.
If it's in the comment, there's no way to know if it's working.

leahecole:
#4082 was just merged

leahecole:
So I definitely did not intentionally unassign @gguuss - it did that when I commented, and now I can't re-add him. 😬

PR #3857, opened 2020-05-21
[kokoro] testing: use secrets manager

busunkim96:
Turns out Jinja won't let me add `secrets.tar.enc` in the synthtool templates because Jinja can't parse it.

Trying a little experiment with secrets manager.

Note: https://github.com/GoogleCloudPlatform/python-docs-samples/pull/3857/commits/5e87e58782bf7dc0a85b5afa4c9699af53e8143c has some real test runs (because the branch was behind master).


.kokoro/tests/run_tests.sh
```diff
@@ -37,9 +37,15 @@ cd github/python-docs-samples
 # install nox for testing
 pip install -q nox
 
-# Unencrypt and extract secrets
-SECRETS_PASSWORD=$(cat "${KOKORO_GFILE_DIR}/secrets-password.txt")
-./scripts/decrypt-secrets.sh "${SECRETS_PASSWORD}"
+# update gcloud version
+# TODO: move this to the docker image
+gcloud components update
```
busunkim96, line 42: gcloud update in #3859

.kokoro/tests/run_tests.sh
```diff
@@ -37,9 +37,15 @@ cd github/python-docs-samples
...
+
+# Use secrets acessor service account to get secrets
+gcloud auth activate-service-account \
```
tmatsuo, line 45: Does it make sense to just use ${GOOGLE_APPLICATION_CREDENTIALS}? We can give it the right permission on the target project I think.

busunkim96: The scripts are using secrets manager via gcloud, so I think we need to use `gcloud auth activate-service-account`.  I thought only the libraries used `GOOGLE_APPLICATION_CREDENTIALS`?

tmatsuo: We run `activate-service-account` few lines below. It's for using cloudbuild from cloud-run tests.

busunkim96: Ah I see. `testing/service-account.json` doesn't exist until the `scripts/decrypt-secrets.sh` has been executed though.

I could move the python-docs-samples-testing service account into `KOKORO_GFILE_DIR` (or keystore)?

tmatsuo: Oh I see. Then it's fine I guess.

.kokoro/tests/run_tests.sh
tmatsuo: Then we don't need this any more?

busunkim96: Removed

tmatsuo:
`lint` build is using the old image?

PR #3860, opened 2020-05-22
testing: migrate to trampoline_v2

tmatsuo:
This PR changes the build file to `trampoline_v2.sh` for all the jobs.

`trampoline_v2.sh` does 3 things.
1. Prepare the Docker image for the test
2. Run the Docker with appropriate flags to run the test
3. Upload the newly built Docker image
                                                                                                                                                                  
in a way that is somewhat compatible with trampoline_v1.

It also behaves better in many ways.
1. It uses the same uid as the caller, use the same gid as docker group on the host.
2. No permission change required upon cleanup.
3. It's bit more portable so that much easier to run somewhere other than Kokoro including local env.
4. It can optionally build the docker image from the source then use the built image for the test, so that you can have a single PR for changing the Dockerfile and tests.
5. Logging with timestamp. Easier to determine the bottleneck.
6. Only passing down minimum envvars.
7. Somewhat configurable with `.trampolinerc`.

## Python specific info
To run this script, first decrypt our test secret by running `scripts/decrypt-secrets.sh`. This need secret-accessor role on our Cloud Project.
                                                                                                                                                                
Then run the script:
```bash
.kokoro/trampoline_v2.sh
```
You can optionally change these environment variables:
- TRAMPOLINE_IMAGE: The docker image to use.
- TRAMPOLINE_IMAGE_SOURCE: The location of the Dockerfile.
- RUN_TESTS_SESSION: The nox session to run.
- RUN_TESTS_DIRS: colon separated values for which directory you want to run the tests.
- TRAMPOLINE_BUILD_FILE: The script to run in the docker container.

I also added a handy script for running tests locally.

`scripts/run_tests_local.sh` takes only one required argument, and run test sessions for that directory. Example:

```bash
$ pwd
/home/tmatsuo/work/python-docs-samples
$ cd cdn
$ ../scripts/run_tests_local.sh .  # it will runs lint, py-3.6, and py-3.7 
# or
$ ../scripts/run_tests_local.sh . lint py-3.6  # it will only run lint and py-3.6
```

### DIFF
--- a/.kokoro/docker/Dockerfile
+++ b/.kokoro/docker/Dockerfile
@@ -162,8 +162,9 @@ ENV PATH /google-cloud-sdk/bin:$PATH
 RUN sudo systemctl enable redis-server.service
 
 # Create a user and allow sudo
-ARG UID=0
-ARG GID=0
+
+# kbuilder uid on the default Kokoro image
+ARG UID=1000
 ARG USERNAME=kbuilder
 ARG DOCKER_GID=999
 
--- a/.kokoro/docker/Dockerfile
+++ b/.kokoro/docker/Dockerfile
@@ -166,19 +166,12 @@ RUN sudo systemctl enable redis-server.service
 # kbuilder uid on the default Kokoro image
 ARG UID=1000
 ARG USERNAME=kbuilder
-ARG DOCKER_GID=999
 
 # Add a new user to the container image.
-# To allow access to the docker socket on the host, we use the docker
-# group on the host.
+# This is needed for ssh and sudo access.
 
-# First make sure the group exists. Ignore a failure in case
-# there is already a group with that id.
-RUN groupadd -g ${DOCKER_GID} "host-docker" | true
-
-# Add a new user with the caller's uid and the docker goup id on the
-# host.
-RUN useradd -d /h -u ${UID} -g ${DOCKER_GID} ${USERNAME}
+# Add a new user with the caller's uid and the username.
+RUN useradd -d /h -u ${UID} ${USERNAME}
 
 # Allow nopasswd sudo
 RUN echo "${USERNAME} ALL=(ALL) NOPASSWD:ALL" >> /etc/sudoers
--- a/.kokoro/docker/Dockerfile
+++ b/.kokoro/docker/Dockerfile
@@ -167,11 +167,19 @@ ARG GID=0
 ARG USERNAME=kbuilder
 ARG DOCKER_GID=999
 
-# Allow access docker socker in the host.
-RUN groupadd -g ${DOCKER_GID} "host-docker"
-RUN groupadd -g ${GID} "${USERNAME}"
-RUN useradd -d /h -u ${UID} -g ${GID} ${USERNAME}
-RUN adduser "${USERNAME}" "host-docker"
+# Add a new user to the container image.
+# To allow access to the docker socket on the host, we use the docker
+# group on the host.
+
+# First make sure the group exists. Ignore a failure in case
+# there is already a group with that id.
+RUN groupadd -g ${DOCKER_GID} "host-docker" | true
+
+# Add a new user with the caller's uid and the docker goup id on the
+# host.
+RUN useradd -d /h -u ${UID} -g ${DOCKER_GID} ${USERNAME}
+
+# Allow nopasswd sudo
 RUN echo "${USERNAME} ALL=(ALL) NOPASSWD:ALL" >> /etc/sudoers
 
 CMD ["python3.6"]
--- a/.kokoro/docker/Dockerfile
+++ b/.kokoro/docker/Dockerfile
@@ -161,4 +161,17 @@ ENV PATH /google-cloud-sdk/bin:$PATH
 # Enable redis-server on boot.
 RUN sudo systemctl enable redis-server.service
 
+# Create a user and allow sudo
+ARG UID=0
+ARG GID=0
+ARG USERNAME=kbuilder
+ARG DOCKER_GID=999
+
+# Allow access docker socker in the host.
+RUN groupadd -g ${DOCKER_GID} "host-docker"
+RUN groupadd -g ${GID} "${USERNAME}"
+RUN useradd -d /h -u ${UID} -g ${GID} ${USERNAME}
+RUN adduser "${USERNAME}" "host-docker"
+RUN echo "${USERNAME} ALL=(ALL) NOPASSWD:ALL" >> /etc/sudoers
+
 CMD ["python3.6"]
--- a/.kokoro/tests/run_tests.sh
+++ b/.kokoro/tests/run_tests.sh
@@ -175,4 +175,9 @@ for file in **/requirements.txt; do
 done
 cd "$ROOT"
 
+# Remove secrets if we used decrypt-secrets.sh.
+if [[ -f "${KOKORO_GFILE_DIR}/secrets_viewer_service_account.json" ]]; then
+    rm testing/{test-env.sh,client-secrets.json,service-account.json}
+fi
+
 exit "$RTN"
--- a/.kokoro/tests/run_tests.sh
+++ b/.kokoro/tests/run_tests.sh
@@ -101,8 +101,27 @@ for file in **/requirements.txt; do
     file=$(dirname "$file")
     cd "$file"
 
+    # First we look up the environment variable `RUN_TESTS_DIRS`. If
+    # the value is set, we'll iterate through the colon separated
+    # directory list.
+    if [[ -n "${RUN_TESTS_DIRS:-}" ]]; then
+	IFS=":"
+	read -ra run_tests_dirs <<< "${RUN_TESTS_DIRS}"
+	match=0
+	for d in "${run_tests_dirs[@]}"; do
+	    # If the current dir starts with one of the
+	    # RUN_TESTS_DIRS, we should run the tests.
+	    if [[ "${file}" = "${d}"* ]]; then
+		match=1
+		break
+	    fi
+	done
+	IFS=" "
+	if [[ $match -eq 0 ]]; then
+	    continue
+	fi
     # If $DIFF_FROM is set, use it to check for changes in this directory.
-    if [[ -n "${DIFF_FROM:-}" ]] && [[ "${test_all}" == "false" ]]; then
+    elif [[ -n "${DIFF_FROM:-}" ]] && [[ "${test_all}" == "false" ]]; then
         git diff --quiet "$DIFF_FROM" .
         CHANGED=$?
         if [[ "$CHANGED" -eq 0 ]]; then
@@ -145,6 +164,9 @@ for file in **/requirements.txt; do
     else
       echo -e "\n Testing completed.\n"
     fi
+
+    # Remove noxfile.py if it's not tracked by git.
+    git ls-files --error-unmatch noxfile.py > /dev/null 2>&1 | rm noxfile.py
 
 done
 cd "$ROOT"
--- a/.kokoro/tests/run_tests.sh
+++ b/.kokoro/tests/run_tests.sh
@@ -136,13 +136,16 @@ for file in **/requirements.txt; do
 
     # If no local noxfile exists, copy the one from root
     if [[ ! -f "noxfile.py" ]]; then
-      PARENT_DIR=$(cd ../ && pwd)
-      while [[ "$PARENT_DIR" != "$ROOT" && ! -f "$PARENT_DIR/noxfile-template.py" ]];
-      do
-        PARENT_DIR=$(dirname "$PARENT_DIR")
-      done
-      cp "$PARENT_DIR/noxfile-template.py" "./noxfile.py"
-      echo -e "\n Using noxfile-template from parent folder ($PARENT_DIR). \n"
+	PARENT_DIR=$(cd ../ && pwd)
+	while [[ "$PARENT_DIR" != "$ROOT" && ! -f "$PARENT_DIR/noxfile-template.py" ]];
+	do
+            PARENT_DIR=$(dirname "$PARENT_DIR")
+	done
+	cp "$PARENT_DIR/noxfile-template.py" "./noxfile.py"
+	echo -e "\n Using noxfile-template from parent folder ($PARENT_DIR). \n"
+	cleanup_noxfile=1
+    else
+	cleanup_noxfile=0
     fi
 
     # Use nox to execute the tests for the project.
@@ -165,8 +168,10 @@ for file in **/requirements.txt; do
       echo -e "\n Testing completed.\n"
     fi
 
-    # Remove noxfile.py if it's not tracked by git.
-    git ls-files --error-unmatch noxfile.py > /dev/null 2>&1 | rm noxfile.py
+    # Remove noxfile.py if we copied.
+    if [[ $cleanup_noxfile -eq 1 ]]; then
+	rm noxfile.py
+    fi
 
 done
 cd "$ROOT"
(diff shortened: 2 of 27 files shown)


tmatsuo:
Currently most of the tests are passing locally with the new script.
Only failing test is `iap`. I think it's failing due to unrelated reasons. Let's see the periodic builds' result.

tmatsuo:
`iap` tests are passing in periodic builds.
I'm experimenting with running the same tests on Kokoro with trampoline_v2.

tmatsuo:
The `iap` test is passing with `trampoline_v2.sh`:
https://source.cloud.google.com/results/invocations/acd1029b-f6fa-4aa5-940b-cfd936157e5b/targets

Somehow it's not passing locally. I'll skip this test for local runs.

tmatsuo:
I run py-3.7 for all the tests with the following command and they all passed!

```bash
$ TERM=vt100 \
  TRAMPOLINE_BUILD_FILE=.kokoro/tests/run_tests.sh \
  .kokoro/trampoline_v2.sh 2>&1 | tee /tmp/build.log
```

tmatsuo:
Rebased to the master, and experimenting Trampoline V2 with the py-3.7 build.
This should be full build and it'll take long time. I'll examine the invocation logs after the test finishes.

Fingers crossed 🤞

tmatsuo:
Doh. The regex part was wrong.

tmatsuo:
All the presubmit builds passed. This is amazing :)

However, I need to make some more changes.

.kokoro/tests/run_tests.sh
```diff
@@ -92,8 +100,27 @@ for file in **/requirements.txt; do
...
+		break
+	    fi
+	done
+	IFS=" "
+	if [[ $match -eq 0 ]]; then
+	    continue
+	fi
     # If $DIFF_FROM is set, use it to check for changes in this directory.
-    if [[ -n "${DIFF_FROM:-}" ]] && [[ "${test_all}" == "false" ]]; then
+    elif [[ -n "${DIFF_FROM:-}" ]] && [[ "${test_all}" == "false" ]]; then
```
kurtisvg, line 123: Should this be either/or? I think that may be unexpected

.kokoro/tests/run_tests.sh
```diff
@@ -92,8 +100,27 @@ for file in **/requirements.txt; do
     file=$(dirname "$file")
     cd "$file"
 
+    # First we look up the environment variable `RUN_TESTS_DIRS`. If
+    # the value is set, we'll iterate through the colon separated
+    # directory list.
+    if [[ -n "${RUN_TESTS_DIRS:-}" ]]; then
```
kurtisvg, line 106: The other variables are actually set by a flag at the start of this script. Should we do the same for this one (or document?

.kokoro/tests/run_tests.sh
```diff
@@ -92,8 +100,27 @@ for file in **/requirements.txt; do
...
+	    # RUN_TESTS_DIRS, we should run the tests.
+	    if [[ "${file}" = "${d}"* ]]; then
+		match=1
+		break
+	    fi
+	done
+	IFS=" "
+	if [[ $match -eq 0 ]]; then
+	    continue
+	fi
```
kurtisvg, line 121: suggestion: Would this be simpler with a regex?

```bash
if [[ ! "$RUN_TESTS_DIRS" =~ ^(.*:)?("$file"[^:]*)(:.*)?$ ]]; do
    continue
elif
```


If you prefer not to use regexs, how about `tr`? 

```bash
for d in $(echo $RUN_TESTS_DIRS | tr ";" "\n"); do
   // loop stuff
done
```

.kokoro/trampoline_v2.sh
```diff
@@ -0,0 +1,352 @@
...
+# .kokoro/trampoline_v2.sh
+#
+# You can optionally change these environment variables:
+#
+# TRAMPOLINE_IMAGE: The docker image to use.
+# TRAMPOLINE_DOCKERFILE: The location of the Dockerfile.
+# TRAMPOLINE_IMAGE_UPLOAD:
+#     (true|false): Whether to upload the Docker image after the
+#                   successful builds.
+# TRAMPOLINE_BUILD_FILE: The script to run in the docker container.
```
kurtisvg, line 42: Isn't this one required?

.kokoro/trampoline_v2.sh
```diff
@@ -0,0 +1,352 @@
...
+if [[ "${TRAMPOLINE_DOCKERFILE:-none}" != "none" ]]; then
+    # Build the Docker image from the source.
+    context_dir=$(dirname "${TRAMPOLINE_DOCKERFILE}")
+    docker_build_flags=(
+	"-f" "${TRAMPOLINE_DOCKERFILE}"
+	"-t" "${TRAMPOLINE_IMAGE}"
+	"--build-arg" "UID=${user_uid}"
+	"--build-arg" "GID=${user_gid}"
+	"--build-arg" "USERNAME=${user_name}"
+	"--build-arg" "DOCKER_GID=${docker_gid}"
```
kurtisvg, line 234: Why set during build-time? Why can't these just be set at execution time?

.kokoro/trampoline_v2.sh
```diff
@@ -0,0 +1,352 @@
...
+    # Mount the temporary home directory.
+    "--volume" "${tmphome}:/h"
+    "--env" "HOME=/h"
+
+    # Allow docker in docker.
+    "--volume" "/var/run/docker.sock:/var/run/docker.sock"
+
+    # Mount the /tmp so that docker in docker can mount the files
+    # there correctly.
+    "--volume" "/tmp:/tmp"
```
kurtisvg, line 291: Are you sure docker in docker needs `/tmp`? I thought it only needed the socket?

.kokoro/tests/run_tests.sh
```diff
@@ -63,9 +68,12 @@ export GOOGLE_CLIENT_SECRETS=$(pwd)/testing/client-secrets.json
 export DATALABELING_ENDPOINT="test-datalabeling.sandbox.googleapis.com:443"
 
 # Run Cloud SQL proxy (background process exit when script does)
-wget --quiet https://dl.google.com/cloudsql/cloud_sql_proxy.linux.amd64 -O cloud_sql_proxy && chmod +x cloud_sql_proxy
-./cloud_sql_proxy -instances="${MYSQL_INSTANCE}"=tcp:3306 &>> cloud_sql_proxy.log &
-./cloud_sql_proxy -instances="${POSTGRES_INSTANCE}"=tcp:5432 &>> cloud_sql_proxy.log &
+wget --quiet https://dl.google.com/cloudsql/cloud_sql_proxy.linux.amd64 \
+     -O ${HOME}/cloud_sql_proxy && chmod +x ${HOME}/cloud_sql_proxy
```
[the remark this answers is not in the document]
tmatsuo: I can do that easily because we enable nopasswd sudo.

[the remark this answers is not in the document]
tmatsuo: Personally I don't care where to download though

.kokoro/tests/run_tests.sh
tmatsuo: Makes sense, done

.kokoro/tests/run_tests.sh
tmatsuo: Added some documents

.kokoro/tests/run_tests.sh
tmatsuo: I chose `tr`

.kokoro/trampoline_v2.sh
tmatsuo: Done

.kokoro/trampoline_v2.sh
tmatsuo: If you're using docker in docker and try to mount files or directories inside the container, the path must be the same location on the docker host too.

It worked with trampoline V1 because the path was the same (`/tmpfs/src/github/python-docs-samples`).

But now we want to mount the project at `/workspace`, we need a path which is guaranteed to be the same on the docker host.

.kokoro/trampoline_v2.sh
tmatsuo: Sorry for a long explanation.

1. To allow sudo we need a real username
2. One of the tests (compute/oslogin iirc) needs a real username

So we need the real username for those reasons. So anyways we need to create a user and allow sudo at build time. Well, if we create a user, I can also add the user to the docker group.

Then we can pass `--user={$user_uid}:${user_gid}` which is better than `--user={$user_uid}:${docker_gid}`.

The files created by the former will be the same uid:gid as you, but the latter will create a file with different gid.

Usecase: we can use the script to generate the README:

```
scripts/run_tests_local.sh cdn readmegen
```

This will generate the README.rst in cdn directory with the same uid and gid as you.

.kokoro/tests/run_tests.sh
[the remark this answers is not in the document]
tmatsuo: Done

tmatsuo:
@kurtisvg Thanks for the review!

I think I addressed your comments and answered your questions. PTAL

.kokoro/trampoline_v2.sh
tmatsuo: So basically, the direct answer to your question is:

Yes, we can pass `--user={$user_uid}:${docker_gid}` and docker works, but some other tests will fail (compute/oslogin, and potential tests which needs sudo access).

tmatsuo:
Don't worry about the Python 3.8 failure. That's the one I manually started and stopped once I confirmed the build cop bot is working.

.kokoro/trampoline_v2.sh
kurtisvg: Docker already has functionality for setting a [default user/group as part of a Dockerfile](https://docs.docker.com/engine/reference/builder/#user). What do we gain by having trampoline manage this instead of letting the container do it itself? 

I think my concern is that the container is supposed to be the encapsulation of the entire environment, including users/groups. I don't think it makes sense to have trampoline play a part in managing that. If folks want to have a different user/group by default, they should set that in their Dockerfile.

.kokoro/trampoline_v2.sh
kurtisvg: Ah ok, I think I understand. 

I think you can use a named volume and mount it to both containers to avoid exposing the hosts `/tmp` folder. I'm not sure if this is beneficial or not - my instinct is that exposing the entire `/tmp` folder might not be a great idea, but I'm not sure.

.kokoro/trampoline_v2.sh
tmatsuo: The current code makes sure that, docker will run with the same uid:gid as the caller of the trampoline_v2.sh script. This is dynamic. When you run locally, it's going to be your usual uid:gid, when it's run in Kokoro, it's going to be uid:gid of kbuilder user, etc etc.

tmatsuo: So, to summarize:

1. I want to use `--user "${user_uid}:${user_gid}"` for making sure the container will create files with the same uid:gid as the caller.
2. It works for most of our tests, but some tests are failing.
3. compute/oslogin: We need a real user name for this uid -> we need to create the user in the container OS. This is why we passing the build arg.
4. docker in docker: We need to make sure the user in the container have a permission on /var/run/docker.sock. -> This is why we add our user to the group with the same id as the docker group in the host os.
5. sudo: Some tests potentially need root access, so we need to add sudoers entry on the fly. Actually our `run_tests.sh` needs it for storing cloudsql_proxy to /bin.

So there are reasons. If you come up with a solution with the passing tests both for `compute/oslogin` and `vision/automl`, I'm open to such a solution.

You can download this branch and try out the script with the following command:

```sh
$ scripts/decrypt-secrets.sh  # Make sure we have the secrets
$ scripts/run_tests_local.sh compute/oslogin py-3.7
$ scripts/run_tests_local.sh vision/automl py-3.7
# or
$ RUN_TESTS_DIRS=compute/oslogin:vision/automl \
  TRAMPOLINE_BUILD_FILE=.kokoro/tests/run_tests.sh \
  .kokoro/trampoline_v2.sh

```

These commands start from building the Docker image, so you can test end 2 end for your proposed solution.

kurtisvg: Doesn't this also mean that it rebuilds (and potentially pushes a new image) every time a new user runs it?

tmatsuo: Yes, but they are fast because `docker pull` can detect the local layers and Docker build is using `--cache-from`. These command never upload to anywhere by default. You have to provide two envvars to use that option like this:

```sh
TRAMPOLINE_IMAGE=gcr.io/tmatsuo-test/my-python-multi \
  TRAMPOLINE_IMAGE_UPLOAD=true
  scripts/run_tests_local.sh compute/oslogin py-3.7
```

tmatsuo: Having said that, the upload option is not intend to be used by humans.

My intention is to set that flag only in one of the periodic builds. When the tests are all green, it will push the built image. We won't need the Cloud Build hook any more (although it won't harm anything).

tmatsuo: I also hope that in the future, we can merge all the periodic builds into one. If we pass down the multiple nox sessions, it might be possible. The tests will take a very long time, so we may want to have a tool to run tests in parallel. I thought someone is developing such a tool.

tmatsuo:
@kurtisvg Oops. The `vision/automl` test is failing. I need to fix it before merging. Also the script doesn't report failure which is also bad.

tmatsuo:
Now it's correctly failing.

tmatsuo:
@kurtisvg Ok, now I removed `docker_gid` thing from the Dockerfile, but I think we still need to have the user entry and the sudoers config.

I wish I could pass the same uid:gid, but I think we'd better to pass docker gid to allow docker in docker.

PTAL

.kokoro/tests/run_tests.sh
```diff
@@ -63,9 +68,12 @@ export GOOGLE_CLIENT_SECRETS=$(pwd)/testing/client-secrets.json
 export DATALABELING_ENDPOINT="test-datalabeling.sandbox.googleapis.com:443"
 
 # Run Cloud SQL proxy (background process exit when script does)
-wget --quiet https://dl.google.com/cloudsql/cloud_sql_proxy.linux.amd64 -O cloud_sql_proxy && chmod +x cloud_sql_proxy
-./cloud_sql_proxy -instances="${MYSQL_INSTANCE}"=tcp:3306 &>> cloud_sql_proxy.log &
-./cloud_sql_proxy -instances="${POSTGRES_INSTANCE}"=tcp:5432 &>> cloud_sql_proxy.log &
+sudo wget --quiet https://dl.google.com/cloudsql/cloud_sql_proxy.linux.amd64 \
```
kurtisvg, line 71: nit: I was wrong on this, I think it's better if we don't use sudo

.kokoro/docker/Dockerfile
```diff
@@ -161,4 +161,19 @@ ENV PATH /google-cloud-sdk/bin:$PATH
...
+ARG USERNAME=kbuilder
+
+# Add a new user to the container image.
+# This is needed for ssh and sudo access.
+
+# Add a new user with the caller's uid and the username.
+RUN useradd -d /h -u ${UID} ${USERNAME}
+
+# Allow nopasswd sudo
+RUN echo "${USERNAME} ALL=(ALL) NOPASSWD:ALL" >> /etc/sudoers
```
kurtisvg, line 177: nit: I still think it would be better to used a fixed non-root user, but we can revisit later

tmatsuo:
lint and py-2.7 passed. I'll merge this now with the admin power.

PR #3865, opened 2020-05-23
Adding dependency setuptools and grpcio

starsaminf:
Adding dependency setuptools and grpcio


starsaminf:
@googlebot I signed it!

kurtisvg:
@starsaminf Can you provide some motivation on these changes? Neither of these requirements look like they should need to be explicitly listed

busunkim96:
I can confirm that `grpcio` will be pulled in via `google-api-core` (a dependency of all the GAPIC libraries). Similarly a minimum version of `setuptools` is required by `google-auth`.

leahecole:
Alright if we've determined these are brought in somewhere else, I'm going to go ahead and close this.

PR #3877, opened 2020-05-26
testing: various cleanups

tmatsuo:
Cleanups mostly around temp files.

* [iap]: only run iap tests on Kokoro
* [vision/automl]: use temporary directory for temporary files
* [appengine/flexible/scipy]: use temporary directory
* [bigtable/snippets/reads]: update pytest snapshot
* [texttospeech/cloud-client]: added output.mp3 to .gitignore
* [iot/api-client/gcs_file_to_device]: use temporary directory
* [iot/api-client/mqtt_example]: use temporary directory
* [logging/cloud-client]: use uuid and add backoff


iap/iap_test.py
```diff
@@ -36,6 +38,10 @@
 
 @pytest.mark.flaky
 def test_main(capsys):
+    # It only passes on Kokoro now. Skipping in other places.
+    # The envvar `TRAMPOLINE_CI` will be set once #3860 is merged.
+    if os.environ.get('TRAMPOLINE_CI', 'kokoro') != 'kokoro':
+        pytest.skip('Only passing on Kokoro.')
```
kurtisvg, line 44: What? why?

tmatsuo:
@telpirion re: vision/automl, Actually this sample is still used in [one of our cloud site](https://cloud.google.com/vision/automl/docs/containers-gcs-tutorial) (as you know), so in that sense, it's not moved yet.

iap/iap_test.py
tmatsuo: I don't know. Please try locally if you're interested in. I'll file an issue about this.

tmatsuo: Filed https://github.com/GoogleCloudPlatform/python-docs-samples/issues/3878

kurtisvg: I'd prefer not adding this skip if the issue is only permissions related

tmatsuo: Yeah I'll debug this as soon as I get the ownership on the project.

PR #3899, opened 2020-05-28
refactor Cloud SQL MySQL & Postgres samples to demonstrate both TCP and Unix socket connection patterns

SaketramDurbha:


cloud-sql/mysql/sqlalchemy/README.md
```diff
@@ -32,23 +32,78 @@ secure solution such as [Cloud KMS](https://cloud.google.com/kms/) to help keep
...
+```powershell
+Start-Process -filepath "C:\<path to proxy exe>" -ArgumentList "-instances=<project-id>:<region>:<instance-name>=tcp:3306 -credential_file=<CREDENTIALS_JSON_FILE>"
+```
+
+### Launch proxy with Unix Domain Socket
+NOTE: this option is currently only supported on Linux and Mac OS. Windows users should use the
+[Launch proxy with TCP](#launch-proxy-with-tcp) option.
+
+To use a Unix socket, you'll need to create the `/cloudsql` directory and give write access to the
+user running the proxy. Use these commands to create the directory and set permissions:
```
shubha-rajan, line 81: Should we advise users to use the `DB_SOCKET_PATH` environment variable here instead?

cloud-sql/postgres/sqlalchemy/README.md
```diff
@@ -31,25 +31,79 @@ secure solution such as [Cloud KMS](https://cloud.google.com/kms/) to help keep
...
+### Launch proxy with Unix Domain Socket
+NOTE: this option is currently only supported on Linux and Mac OS. Windows users should use the
+[Launch proxy with TCP](#launch-proxy-with-tcp) option.
+
+To use a Unix socket, you'll need to create the `/cloudsql` directory and give write access to the
+user running the proxy. Use these commands to create the directory and set permissions:
 ```bash
 sudo mkdir /cloudsql
 sudo chown -R $USER /cloudsql
 ```
```
shubha-rajan, line 84: Same here, we should probably show them how to create the directory and set `DB_SOCKET_PATH`

cloud-sql/postgres/sqlalchemy/main.py
```diff
@@ -27,34 +27,12 @@
...
-        database=db_name,
-        query={
-            'unix_sock': '/cloudsql/{}/.s.PGSQL.5432'.format(
-                cloud_sql_connection_name)
-        }
-    ),
-    # ... Specify additional properties here.
-    # [START_EXCLUDE]
-
+db_config = {
```
shubha-rajan, line 30: This config dict should probably be moved to right above where we create the connection pool. (same goes for the MySQL sample)
I suggest initializing it inside `init_connection_engine` instead of making it a global variable, and then passing it into the `init_tcp` and `init_unix_socket` functions as a parameter

cloud-sql/mysql/sqlalchemy/README.md
```diff
@@ -32,23 +32,78 @@ secure solution such as [Cloud KMS](https://cloud.google.com/kms/) to help keep
...
+export GOOGLE_APPLICATION_CREDENTIALS=/path/to/service/account/key.json
+export INSTANCE_CONNECTION_NAME='<MY-PROJECT>:<INSTANCE-REGION>:<INSTANCE-NAME>'
+export DB_USER='<DB_USER_NAME>'
+export DB_PASS='<DB_PASSWORD>'
+export DB_NAME='<DB_NAME>'
 ```
+
+Then use this command to launch the proxy in the background:
+```bash
+./cloud_sql_proxy -dir=/cloudsql --instances=$INSTANCE_CONNECTION_NAME --credential_file=$GOOGLE_APPLICATION_CREDENTIALS &
```
shubha-rajan, line 98: The `DB_SOCKET_PATH` env var should probably be used here too, to guarantee that this command will work on all systems, since, depending on the OS, the user might not be able to write to the root directory.

cloud-sql/mysql/sqlalchemy/main.py
```diff
@@ -62,17 +44,87 @@
...
+            drivername="mysql+pymysql",
+            username=db_user,
+            password=db_pass,
+            host=db_host,
+            port=db_port,
+            database=db_name,
+        ),
+        # ... Specify additional properties here.
+        # [START_EXCLUDE]
+        **db_config
```
shubha-rajan, line 90: Question for repo owners: are we still requiring 2.7 compatibility for all tests or is this fine?

shubha-rajan, line 90: Note to repo owners: do we still need to support python 2.7 in all samples or can we use this dictionary unpacking syntax?

busunkim96: Thanks for checking! 3.6+ is what we're now targeting https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/AUTHORING_GUIDE.md#python-version

shubha-rajan: Thanks for the quick answer!

cloud-sql/postgres/sqlalchemy/main.py
```diff
@@ -31,56 +31,104 @@
...
+    # [START cloud_sql_postgres_sqlalchemy_create_tcp]
+    db_scoket_addr = os.environ.get("DB_HOST").split(":")
+
+    # Extract host and port from socket address
+    db_host = db_scoket_addr[0]
+    db_port = int(db_scoket_addr[1])
+
+    return sqlalchemy.create_engine(
+        # Equivalent URL:
+        # postgres+pg8000://<db_user>:<db_pass>@/<db_name>?unix_sock=/cloudsql/<cloud_sql_instance_name>/.s.PGSQL.5432
```
shubha-rajan, line 82: This looks like a Unix Socket connection string. It should be a TCP connection string

cloud-sql/postgres/sqlalchemy/main.py
```diff
@@ -31,56 +31,104 @@
...
+        return init_unix_connection_engine(db_config)
+
+
+def init_tcp_connection_engine(db_config):
+    # [START cloud_sql_postgres_sqlalchemy_create_tcp]
+    db_scoket_addr = os.environ.get("DB_HOST").split(":")
+
+    # Extract host and port from socket address
+    db_host = db_scoket_addr[0]
+    db_port = int(db_scoket_addr[1])
```
shubha-rajan, line 78:
```suggestion
    db_socket_addr = os.environ.get("DB_HOST").split(":")

    # Extract host and port from socket address
    db_host = db_socket_addr[0]
    db_port = int(db_socket_addr[1])
```

cloud-sql/postgres/sqlalchemy/README.md
```diff
@@ -31,25 +31,85 @@ secure solution such as [Cloud KMS](https://cloud.google.com/kms/) to help keep
...
+NOTE: this option is currently only supported on Linux and Mac OS. Windows users should use the
+[Launch proxy with TCP](#launch-proxy-with-tcp) option.
+
+To use a Unix socket, you'll need to create a directory and give write access to the user running
+the proxy. For example:
+
 ```bash
 sudo mkdir /cloudsql
 sudo chown -R $USER /cloudsql
 ```
```
shubha-rajan, line 85: nittiest of nits: the paths in the bash command examples should probably be the same to make it clear that you're referring to the same path
```suggestion
```bash
sudo mkdir /path/to/the/new/directory
sudo chown -R $USER /path/to/the/new/directory
```
```

**Comment by shubha-rajan** [comment]
@SaketramDurbha have you verified that these changes work when deploying to App Engine and Cloud Run?

**Comment by SaketramDurbha** [comment]
Update: working on adding GAE flex deployment instructions.

File: cloud-sql/postgres/sqlalchemy/requirements.txt [hunk 9]
```diff
@@ -2,3 +2,4 @@ Flask==1.1.2
 SQLAlchemy==1.3.17
 pg8000==1.13.2; python_version > '3.0'
 pg8000==1.12.5; python_version < '3.0'
```
dmahugh, line 4: Since the sample uses Python 3+ syntax that doesn't work with 2.7, probably best to just pin to pg8000==1.13.2 to avoid creating the expectation that it will work with Python versions < 3.0.

SaketramDurbha: Done!

PR #3904, opened 2020-05-28
samples: add Pub/Sub dead letter queue samples

anguillanneuf:
Four samples added, similar to the Java ones [here](https://github.com/googleapis/java-pubsub/tree/master/samples/snippets/src/main/java/pubsub).


pubsub/cloud-client/subscriber.py
```diff
@@ -72,16 +72,65 @@ def create_subscription(project_id, topic_name, subscription_name):
...
+def create_subscription_with_dead_letter_topic(
+    project_id, topic_name, subscription_name, dead_letter_topic_name
+):
+    """Create a subscription with dead letter policy."""
+    # [START pubsub_dead_letter_create_subscription]
+    from google.cloud import pubsub_v1
+    from google.cloud.pubsub_v1.types import DeadLetterPolicy
+
+    # TODO project_id = "Your Google Cloud Project ID"
+    # This is an existing topic that the subscription with dead letter policy
```
kurtisvg, line 90: Please put these in something like the following format: 
```
# TODO(developer): This is an existing topic that the subscription with dead letter policy
# project_id = "your-project-id"
```

This way the developer can just uncomment the line and replace the value. Additionally, if any of the variables require a specific format, it's helpful to hint that using the string name.

pubsub/cloud-client/subscriber.py
```diff
@@ -72,16 +72,65 @@ def create_subscription(project_id, topic_name, subscription_name):
...
+            subscription.dead_letter_policy.dead_letter_topic
+        )
+    )
+    print(
+        "After {} delivery attempts.".format(
+            subscription.dead_letter_policy.max_delivery_attempts
+        )
+    )
+
+    subscriber.close()
```
kurtisvg, line 124: What does closing this subscriber do? Should we use a `with subscriber:` statement to ensure it's closed if an error happens? 

It this subscriber being used? It's not clear to me what this sample is doing.

pubsub/cloud-client/subscriber.py
```diff
@@ -171,6 +221,113 @@ def update_subscription(project_id, subscription_name, endpoint):
...
+    dead_letter_topic_path = subscriber.topic_path(
+        project_id, dead_letter_topic_name
+    )
+
+    subscription_before_update = subscriber.get_subscription(subscription_path)
+    print("Before the update: {}".format(subscription_before_update))
+
+    subscription_before_update.HasField("dead_letter_policy")
+
+    assert 'dead_letter_policy' in properties
```
kurtisvg: We shouldn't use `assert` in non-test code

pubsub/cloud-client/subscriber.py
```diff
@@ -171,6 +221,113 @@ def update_subscription(project_id, subscription_name, endpoint):
...
+    topic_path = subscriber.topic_path(project_id, topic_name)
+    subscription_path = subscriber.subscription_path(
+        project_id, subscription_name
+    )
+    dead_letter_topic_path = subscriber.topic_path(
+        project_id, dead_letter_topic_name
+    )
+
+    subscription_before_update = subscriber.get_subscription(subscription_path)
+    print("Before the update: {}".format(subscription_before_update))
```
kurtisvg, line 232: Can you add a comment explaining to the user what is happening here? e.g. "Get the current version of the subscription so we can update it."

pubsub/cloud-client/subscriber.py
```diff
@@ -171,6 +221,113 @@ def update_subscription(project_id, subscription_name, endpoint):
...
+
+    subscription_before_update.HasField("dead_letter_policy")
```
kurtisvg, line 234: What does this line do?

pubsub/cloud-client/subscriber.py
```diff
@@ -461,6 +618,46 @@ def callback(message):
     # [END pubsub_subscriber_error_listener]
 
 
+def receive_messages_with_delivery_attempts(
+    project_id, subscription_name, timeout=None
```
kurtisvg, line 572: Can we provide a default value for this timeout?

pubsub/cloud-client/subscriber.py
```diff
@@ -171,6 +221,113 @@ def update_subscription(project_id, subscription_name, endpoint):
...
+        project_id, subscription_name
+    )
+
+    subscription_before_update = subscriber.get_subscription(subscription_path)
+    print("Before removing the policy: {}".format(subscription_before_update))
+
+    update_mask = {
+        "paths": {"dead_letter_policy": {"dead_letter_topic"}},
+        "paths": {"dead_letter_policy": {"max_delivery_attempts"}},
+    }
```
kurtisvg: Can you add a comment explaining what this does?

pubsub/cloud-client/subscriber.py
```diff
@@ -171,6 +221,113 @@ def update_subscription(project_id, subscription_name, endpoint):
...
+
+    # Construct the subscription with the dead letter policy you expect to have
+    # after the update. Here, values in the required fields (name, topic) help
+    # identify the subscription.
+    subscription = pubsub_v1.types.Subscription(
+        name=subscription_path, topic=topic_path
+    )
```
kurtisvg, line 292: Would it make more sense to modify existing `subscription_before_update`? Otherwise, will updates blow away existing configurations?

pubsub/cloud-client/subscriber.py
```diff
@@ -461,6 +618,46 @@ def callback(message):
...
+    print("Listening for messages on {}..\n".format(subscription_path))
+
+    # Wrap subscriber in a 'with' block to automatically call close() when done.
+    with subscriber:
+        # When `timeout` is not set, result() will block indefinitely,
+        # unless an exception is encountered first.
+        try:
+            streaming_pull_future.result(timeout=timeout)
+        except Exception as e:
```
kurtisvg, line 597: Catching `Exception` here is too broad - can you either narrow this down, or else remove try/catch and just let the exception propagate?

pubsub/cloud-client/subscriber.py
anguillanneuf: Removed.

pubsub/cloud-client/subscriber.py
anguillanneuf: It should default to None, and is expected to stream until the user enters `Ctrl+C` to terminate.

pubsub/cloud-client/subscriber.py
anguillanneuf: Done.

pubsub/cloud-client/subscriber.py
anguillanneuf: Done.

pubsub/cloud-client/subscriber.py
anguillanneuf: Done.

pubsub/cloud-client/subscriber.py
anguillanneuf: Done.

pubsub/cloud-client/subscriber.py
anguillanneuf: This should be the subscription that one expects to get after the update, not before the update. The comment above it describes it.

pubsub/cloud-client/subscriber.py
anguillanneuf: It would be a little misleading to phrase it like that. The print statement is mainly there so people can compare fields. To update a subscription, an expected subscription (not the original subscription) and a field mask are supplied to the update method.

PR #3914, opened 2020-05-29
testing: prototyping a new change detector

tmatsuo:


.kokoro/tests/get_changed_files.py
```diff
@@ -0,0 +1,45 @@
...
+# limitations under the License.
+
+import os
+import sys
+
+import requests
+
+
+_PR_URL_TEMPLATE = (
+    'https://api.github.com/repos/googleapis/google-cloud-python/pulls/'
```
busunkim96, line 24: I think we  eventually started to see GitHub quota issues (too many requests with no associated token). If we start to see that we can add a read-only GH token for this request to use.

tmatsuo: Thanks! Maybe I can do that now

tmatsuo:
I manually ran py-3.7 continuous build against my branch `tmatsuo:pr-only-changed`.

It correctly detect the changed file in the last commit and only ran the test for `cdn` directory.

Looks like it works as intended.

tmatsuo:
Probably we can just use `git diff origin/master...`.

PR #3915, opened 2020-05-29
Update presubmit tests to run full tests if .kokoro/tests or .kokoro/docker have changed. 

kurtisvg:
* Updates "--only-diff-master" to use `...`, which reduces unintentional changes from merge commits
* Runs full test if .kokoro/tests and .kokoro/docker have changed


tmatsuo:
superceded by #3916

PR #3916, opened 2020-05-29
testing: diff against the common commit on the master

tmatsuo:
This change allows us to only run necessary tests for Pull Requests.

### DIFF
--- a/.kokoro/tests/run_tests.sh
+++ b/.kokoro/tests/run_tests.sh
@@ -21,6 +21,10 @@ shopt -s globstar
 
 DIFF_FROM=""
 
+# We need to temporarily turn off -e for detecting changes in the test
+# driver.
+set +e
+
 # `--only-diff-master` will only run tests on project changes on the
 # last common commit from the master branch.
 if [[ $* == *--only-diff-master* ]]; then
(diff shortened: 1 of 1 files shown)


.kokoro/tests/run_tests.sh
```diff
@@ -21,12 +21,28 @@ shopt -s globstar
...
+# driver.
+set +e
+
+# `--only-diff-master` will only run tests on project changes on the
+# last common commit from the master branch.
 if [[ $* == *--only-diff-master* ]]; then
-    DIFF_FROM="origin/master.."
+    git diff --quiet "origin/master..." .kokoro/tests .kokoro/docker \
+	.kokoro/trampoline_v2.sh
+    CHANGED=$?
```
kurtisvg, line 33: nit: I think this would be easier to read (and get away without a comment) if you limit it to the immediate area
```suggestion
    set +e
    git diff --quiet "origin/master..." .kokoro/tests .kokoro/docker \
	.kokoro/trampoline_v2.sh
    CHANGED=$?
    set -e
```

tmatsuo: Done with the next commit

tmatsuo:
The only failures on py-3.7 are dlp and firestore (503). I'm going to merge this.

PR #3922, opened 2020-06-01
Improve Spanner delete_data sample coverage

larkee:
The [Deleting rows in a table](https://cloud.google.com/spanner/docs/modify-mutation-api#deleting_rows_in_a_table) section gives four different delete methods but only two methods were demonstrated.

This PR adds the missing methods.


tmatsuo:
Let's wait LGTM from @mmmarklu

PR #3929, opened 2020-06-02
add example of creating AutoML Tables client with non-default endpoint ('new' sdk)

amygdala:
Torry, this is related to b/146659153. @sirtorry 
I may not be using the right conventions for this sample code— LMK.  (also not sure I'm doing the right things w.r.t. the tests...)

Edit: also, can you share what linter to run?  It looks like you all require 4 spaces, not 2, for one thing.

/assign @sirtorry


dinagraves:
Hi Amy - try running `black` on your files to fix your linting.   ie

`python -m black  .`

tables/automl/endpoint_test.py
```diff
@@ -0,0 +1,36 @@
...
+from google.api_core.client_options import ClientOptions
+
+PROJECT = os.environ["GCLOUD_PROJECT"]
+
+def test_client_creation(capsys):
+    client = automl_tables_set_endpoint.create_client_with_endpoint(
+        PROJECT)
+    client.list_datasets()
+    out, _ = capsys.readouterr()
+    assert "GRPCIterator" in out
```
dinagraves, line 35: Does this pass locally?  list_datasets() doesn't seem like it's writing to stdout.

telpirion:
@sirtorry Correct me if I'm wrong, but I think we want to consolidate all of our AutoML samples into automl/.

tables/automl/endpoint_test.py
amygdala: yeah, thanks— wrapping it in a print stmt makes the assertion pass.  (Not sure this is the cleanest way to test that the client obj is okay, but I didn't want to assume any other history.)

amygdala:
> @sirtorry Correct me if I'm wrong, but I think we want to consolidate all of our AutoML samples into automl/.

I was just following what appeared to be the convention for the location of the 'new sdk' examples— LMK.

tables/automl/endpoint_test.py
```diff
@@ -0,0 +1,27 @@
...
+import os
+
+import automl_tables_set_endpoint
+
+PROJECT = os.environ["GCLOUD_PROJECT"]
+
+
+def test_client_creation(capsys):
+    client = automl_tables_set_endpoint.create_client_with_endpoint(PROJECT)
+    print(client.list_datasets())
```
leahecole, line 25: Would it be possible to have the `create_client_with_endpoint` method print something, and then assert on your print? Then, you would be testing your sample, rather than the actual AutoML underlying function. 

Also, if testing is something you're not 100% comfortable with, do feel free to put some time on my calendar and we can take 20-30 min to have a faster back and forth about what you think is most appropriate for this sample :)

leahecole:
> > @sirtorry Correct me if I'm wrong, but I think we want to consolidate all of our AutoML samples into automl/.
> 
> I was just following what appeared to be the convention for the location of the 'new sdk' examples— LMK.
> Edit: if you have decided to move these examples, there are a number of other files in this dir linked to from the docs too, so I'd suggest pulling this in once it seems okay, then tackling the move separately. (to coordinate the docs updates).

+1 to separate move

telpirion:
I'm fine with moving these samples as part of a separate PR.

tables/automl/endpoint_test.py
telpirion: +1 to @leahecole 's suggestion.

For the set endpoint kind of samples, we typically use a really simple API call to verify that the client was created and operates correctly.

See example [here](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/vision/cloud-client/detect/set_endpoint.py).

I suggest copying the output of the regular Tables [list_datasets sample](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/tables/automl/automl_tables_dataset.py#L59).

busunkim96: +1 to just executing `list_datasets` since the goal is to check that the client was constructed correctly.

PR #3953, opened 2020-06-04
Pub/Sub: replace name with id in samples

anguillanneuf:
Users have pointed out that when we say topic name and subscription name in the samples, we actually mean topic id and subscription id. Updating the samples to reflect this.


anguillanneuf:
An INTERNAL error caused the test to fail. Re-running.

PR #3956, opened 2020-06-04
Add code sample for string replacement based deidentification.

achinkulshrestha:
Adds a code sample corresponding to the replacement based deidentification in the Cloud DLP API. The detected sensitive value is replaced with a specified surrogate.


dlp/deid.py
```diff
@@ -83,8 +83,72 @@ def deidentify_with_mask(
...
+                        "replace_config": {
+                            "new_value": {
+                                "string_value": replacement_str,
+                            }
+                        }
+                    }
+                }
+            ]
+        }
+    }
```
busunkim96, line 131: It's hard for me to wrap my head around this because of how deeply this is nested.

Could we perhaps use two dictionaries?
```py
transformation = {
    "primitive_transformation": {
        "replace_config": {
            "new_value": {
                "string_value": replacement_str,
            }
        }
    }
}

deidentify_config = {
    "info_type_transformations": {
        "transformations": [transformation]
    }
}
```

dlp/deid.py
```diff
@@ -83,8 +83,72 @@ def deidentify_with_mask(
...
+    project,
+    string,
+    info_types,
+    replacement_str=None,
+):
+    """Uses the Data Loss Prevention API to deidentify sensitive data in a
+    string by replacing matched input values with a value you specify.
+    Args:
+        project: The Google Cloud project id to use as a parent resource.
+        string: The string to deidentify (will be treated as text).
```
busunkim96, line 97: `info_types` is missing in docstring

dlp/deid.py
achinkulshrestha: Thanks for the review. I was trying to follow surrounding code for consistency. Do you feel strongly about the nesting.

dlp/deid.py
achinkulshrestha: Again, I was trying to follow surrounding code for consistency, which also didn't have info_types in the docstring. In python, I think it's ok to leave out a parameter from the docstring if their meaning is clear enough.

I can create a separate pull request for improving the entire python style for this repo.

dlp/deid.py
busunkim96: Nope, I didn't notice the existing function. I'm fine with keeping this as is.

dlp/deid.py
busunkim96: I think `info_types` missing was a mistake. See https://github.com/GoogleCloudPlatform/python-docs-samples/blob/4b0d97e95cf5b88387671187deb5d9e95ede7fa7/dlp/inspect_content.py#L40 for comparison. Would you mind updating it for the two functions in this file?

achinkulshrestha: You are right. It should be present for all functions. This pull request only concerns with deid.py. I'll create a separate pull request for docstring changes for the entire repo. Would that be ok or do you suggest making these nits here?

busunkim96: I don't have a strong preference either way, please do whatever is easier for you

dlp/deid.py
```diff
@@ -83,8 +83,72 @@ def deidentify_with_mask(
 
 # [END dlp_deidentify_masking]
 
+# [START dlp_deidentify_replace]
+def deidentify_with_replace(
+    project,
+    string,
```
busunkim96, line 89:
```suggestion
    item,
```
For consistency with the existing function

dlp/deid.py
busunkim96, line 97:
```suggestion
        item: The string to deidentify (will be treated as text).
```

dlp/deid.py
```diff
@@ -83,8 +83,72 @@ def deidentify_with_mask(
...
+
+    # Construct deidentify configuration dictionary
+    deidentify_config = {
+        "info_type_transformations": {
+            "transformations": [
+                {
+                    "primitive_transformation": {
+                        "replace_config": {
+                            "new_value": {
+                                "string_value": replacement_str,
```
busunkim96, line 124: Is the API alright with this being passed as `None`? If not, please change the default in the function signature

dlp/deid.py
achinkulshrestha: Added info_types to the new changes. Will create a separate request for other files.

achinkulshrestha: Done

dlp/deid.py
achinkulshrestha: You are right. The whole file follows the pattern of marking these input vars as None for some reason. Modified. Will change other in a separate pull request.

dlp/deid.py
achinkulshrestha: Done.

dlp/deid.py
kurtisvg: +1 to breaking this out - I think there is significantly more nesting going on here than in the other function that it feel necessary

tmatsuo:
I just bumped the timeout for the failed test. I'll updating the PR.

tmatsuo:
I push "REBUILD" button for py-3.7 build.

PR #3964, opened 2020-06-04
Add text redaction sample using DLP

sethmoo:


busunkim96:
@sethmoo Is this the same as #3956? (Apologies if not, I'm not familiar with the DLP API 😄 )

sethmoo:
> @sethmoo Is this the same as #3956? (Apologies if not, I'm not familiar with the DLP API 😄 )

It's similar, but subtly different. This PR is a sample for https://cloud.google.com/dlp/docs/redacting-sensitive-data.

dlp/deid.py
```diff
@@ -435,6 +435,62 @@ def write_data(data):
 # [END dlp_deidentify_date_shift]
 
 
+# [START dlp_redact_sensitive_data]
+def redact_sensitive_data(project, string, info_types):
```
busunkim96, line 439:
```suggestion
def redact_sensitive_data(project, item, info_types):
```

dlp/deid.py
```diff
@@ -435,6 +435,62 @@ def write_data(data):
...
+    """Uses the Data Loss Prevention API to redact sensitive data in a
+    string by replacing it with the info type.
+    Args:
+        project: The Google Cloud project id to use as a parent resource.
+        item: The string to redact (will be treated as text).
```
busunkim96, line 444: info_type docstring copied from https://github.com/GoogleCloudPlatform/python-docs-samples/blob/88ca408d9575de50cdc37546253ca23ae794bafa/dlp/inspect_content.py#L40-L41
```suggestion
        item: The string to redact (will be treated as text).
        info_types: A list of strings representing info types to look for.
            A full list of info type categories can be fetched from the API.
```

dlp/deid.py
sethmoo: Done

PR #3971, opened 2020-06-05
E2E tests for Cloud Run hello-broken and logging-manual samples

dinagraves:
Tests for hello-broken and logging-manual.  

Original plan was to do all the e2e-tests in one PR, but the logging-manual one got a little more complicated. 

Using the Stackdriver API to check that the message logged properly, as this is the main feature of the sample & tutorial.  I can be convinced to use the python client instead of the HTTP call.


run/logging-manual/e2e_test.py
```diff
@@ -0,0 +1,156 @@
...
+
+    # Test that the logs are writing properly to stackdriver
+    time.sleep(10)  # Slight delay writing to stackdriver
+    response = stackdriver_request(service_name, project, access_token)
+
+    entries = json.loads(response).get("entries")
+    assert entries
+
+
+def stackdriver_request(service_name, project, access_token):
```
grayside, line 128: 1. Why not use client library?
2. I notice no retries, it's been found necessary in Node.js and Golang to avoid flakiness:
  * https://github.com/GoogleCloudPlatform/nodejs-docs-samples/blob/master/run/logging-manual/test/system.test.js#L20
  * https://github.com/GoogleCloudPlatform/golang-samples/blob/master/logging/simplelog/simplelog_test.go#L63

run/hello-broken/noxfile_config.py
```diff
@@ -0,0 +1,39 @@
...
+    # You can opt out from the test for specific Python versions.
+
+    # We only run the cloud run tests in py38 session.
+    'ignored_versions': ["2.7", "3.6", "3.7"],
+
+    # An envvar key for determining the project id to use. Change it
+    # to 'BUILD_SPECIFIC_GCLOUD_PROJECT' if you want to opt in using a
+    # build specific Cloud project. You can also use your own string
+    # to use your own Cloud project.
+    'gcloud_project_env': 'GCLOUD_PROJECT',
```
grayside, line 33: `GCLOUD_PROJECT` is deprecated in GAE at least, in favor of `GOOGLE_CLOUD_PROJECT` which I see used elsewhere in this PR. Should this be standardized?

busunkim96: @GoogleCloudPlatform/python-samples-owners We should probably switch to GOOGLE_CLOUD_PROJECT in this repo.  Opened #4019

dinagraves: Updated

run/logging-manual/e2e_test.py
dinagraves: Updated to add retries.

Updated to the client library.  The reason why I didn't use it as first is because  doing the direct HTTP calls is generally easier. The documentation for the client libraries is often not great.  By contrast, I can use the API explorer to figure out the correct API call and then copy the HTTP request from the UI.

PR #3973, opened 2020-06-05
[jobs] testing: use multiple projects

### LINKED ISSUE #3969: jobs failed in py 3.6 for a google-auth PR
build_id: 	[0ca1c2e6-6b63-4568-9f4d-9d6917eb3679](https://source.cloud.google.com/results/invocations/0ca1c2e6-6b63-4568-9f4d-9d6917eb3679/targets)
test failures: email_alert_search_sample_test.test_email_alert_search_sample

tmatsuo:
fixes #3969


tmatsuo:
It's concerning to see the same quota error even with using multiple projects.
I'll dig a bit deeper.

tmatsuo:
I actually check the quota value and found that the quota is `6` for `quota metric that handles periodic task requests (searchForAlert) per minute`. This is very low.

I'm going to let the test wait more between retries.

tmatsuo:
A different test is failing for different reason.

tmatsuo:
@leahecole 

During the process of opting in to use multiple projects, usually I need to do some setup tasks. For example, in this case, I needed to enable the API, also needed to opt in to data logging (although I don't know what exactly it is), add a role.

Whenever I go through this process, I try to document the required process as much as possible. For example, in this case, I added some field in the README.rst.in. Then I needed to re-generate the README.

Does it make sense?

PR #3975, opened 2020-06-05
Change string format to f-strings

sdlin:
f-strings is the preferred way to format strings in Python 3

### DIFF
--- a/firestore/cloud-client/snippets.py
+++ b/firestore/cloud-client/snippets.py
@@ -61,7 +61,7 @@ def quickstart_get_collection():
     docs = users_ref.stream()
 
     for doc in docs:
-        print(u'{} => {}'.format(doc.id, doc.to_dict()))
+        print(f'{doc.id} => {doc.to_dict()}')
     # [END quickstart_get_collection]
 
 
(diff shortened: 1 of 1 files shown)


firestore/cloud-client/snippets.py
```diff
@@ -61,7 +61,7 @@ def quickstart_get_collection():
     docs = users_ref.stream()
 
     for doc in docs:
-        print(u'{} => {}'.format(doc.id, doc.to_dict()))
+        print(f'{doc.id} => {doc.to_dict()}')
```
tmatsuo, line 64: Change looks good.
Just a thought. Does it make sense to use [`pprint`](https://docs.python.org/3/library/pprint.html)?

sdlin: Just tried, it's not very pretty unfortunately.
```
>>> pprint(doc)
<google.cloud.firestore_v1.document.DocumentSnapshot object at 0x10dfa1780>
```

tmatsuo: Thanks, LGTM

PR #3991, opened 2020-06-05
Add custom infoType snippets to DLP samples

sethmoo:


dlp/custom_infotype.py
```diff
@@ -0,0 +1,83 @@
...
+    """
+
+    # Import the client library.
+    import google.cloud.dlp
+
+    # Instantiate a client.
+    dlp = google.cloud.dlp_v2.DlpServiceClient()
+
+    # Construct a list of infoTypes for DLP to locate in `content_string`
+    info_types_to_locate = [{"name": "PERSON_NAME"}, {"name": "EMAIL_ADDRESS"}]
```
kurtisvg, line 45: nit: Can you add a link to the docs where infoTypes are defined so users can find these magic strings?

sethmoo: Sure thing!

PR #3993, opened 2020-06-06
feat(tables): update samples for making prediction with exported model

helinwang:
Exported models runs inside a docker image in localhost. This commit
adds code sample for making prediction with exported model with
Python. It will be used in
https://cloud.google.com/automl-tables/docs/model-export#automl-tables-example-drest

### DIFF
--- a/tables/automl/automl_tables_predict.py
+++ b/tables/automl/automl_tables_predict.py
@@ -112,6 +112,31 @@ def batch_predict(
     # [END automl_tables_batch_predict]
 
 
+def exported_model_predict():
+    """Make a prediction for the exported model."""
+    # [START automl_tables_exported_model_predict]
+    import requests
+
+    response = requests.post(
+        "http://localhost:8080/predict",
+        json={
+            "instances": [
+                {
+                    "categorical_col": "mouse",
+                    "num_array_col": [1, 2, 3],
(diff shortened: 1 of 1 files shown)


telpirion:
It looks like this sample is missing tests. Would you please add them?

sirtorry:
this is going to be part of a [tutorial](https://cloud.google.com/automl-tables/docs/model-export) where the reader will be asked to stand up a local server before running this snippet. how should we test it? should we test it?

helinwang:
Maybe we can test the correctness of the request JSON example.

busunkim96:
Passing along testing suggestions from @tmatsuo and @kurtisvg in chat:

kurtisvg 
> My first attempt would be just to mock the response to the request being sent to the local server
>
> My second suggestion (and perhaps a little more e2e, but also potentially more flaky) would be to create a pytest fixture > that runs the server... I think we have some other tests that already use docker in docker but I'm not sure.

tmatsuo 
> docker in docker should work with a caveat. If you mount files when running the image, the path should be under /tmp. For example, this docker run -v /somewhere/mydir:/mydir hello doesn't work but docker run -v /tmp/mydir:/mydir hello should work

busunkim96:
Closing since this PR has gone stale. These samples have been moved to the python-automl repository.

PR #4017, opened 2020-06-08
fix: changes positional to named pararameters in Video samples

telpirion:
Changes calls to `VideoClient.annotate_video()` so that GCS URIs are provided as named parameters.

Example:
```
operation = video_client.annotate_video(path, features=features)
```
Becomes:
```
operation = video_client.annotate_video(input_uri=path, features=features)
```


telpirion:
Thank you!

PR #4020, opened 2020-06-09
Rename DLP code samples from 'redact' to 'replace'

jlmwise:
In the DLP API, redaction and replacement are two separate, named concepts. Code samples recently added by #3964 were named 'redact' but are actually examples of replacement. This change renames those samples for clarity.


jlmwise:
Thanks - the author of the first pull request hasn't added the region tags to the documentation yet.

jlmwise:
Can a reviewer merge this CL?

PR #4051, opened 2020-06-10
Add DLP sample code for reid w/ fpe using surrogate type and unwrapped security key

lxhfirenking:
Add DLP code sample for reid w/ fpe using surrogate type and unwrapped security key, meant for https://cloud.google.com/dlp/docs/pseudonymization#re-identification_in_free_text_code_example
Add corresponding unit test.

Note: The added function deviates in terms of style from other functions in the same file. This is necessary because as a documentation snippet it has to be more specific to fit in the particular narrative of the documentation linked above. It is not integrated as part of the commandline tool in deid.py for the same reason.

Please automerge upon approval

### DIFF
--- a/dlp/deid.py
+++ b/dlp/deid.py
@@ -336,6 +336,86 @@ def reidentify_with_fpe(
 
 # [END dlp_reidentify_fpe]
 
+# [START dlp_reidentify_free_text_with_fpe_using_surrogate]
+def reidentify_free_text_with_fpe_using_surrogate(
+    project,
+    input_str,
+    alphabet="NUMERIC",
+    surrogate_type="PHONE_TOKEN",
+    unwrapped_key="YWJjZGVmZ2hpamtsbW5vcA==",
+):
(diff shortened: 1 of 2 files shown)


dlp/deid.py
```diff
@@ -336,6 +336,86 @@ def reidentify_with_fpe(
...
+    parent = dlp.project_path(project)
+
+    # The unwrapped key is base64-encoded, but the library expects a binary
+    # string, so decode it here.
+    import base64
+
+    unwrapped_key = base64.b64decode(unwrapped_key)
+
+    # Construct Deidentify Config
+    reidentify_config = {
```
busunkim96, line 378: Would you mind splitting this up into two dictionaries? It should make it a little easier for folks to understand and edit.

e.g.

```py
transformation = {"primitive_transformation": ...}

info_type_transformations = {"transformations": transformation}
```

lxhfirenking: Refactored in new commit.

PR #4065, opened 2020-06-11
[bigtable] update reads/filters region tags

billyjacobson:


billyjacobson:
@kurtisvg I'm not sure why this keeps failing lint. I run it locally and it passes and then every time a new lint error appears in kokoro

PR #4066, opened 2020-06-11
Create PULL_REQUEST_TEMPLATE.md

kurtisvg:


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,14 @@
...
+Fixes #<ISSUE-NUMBER>
+
+Note: It's a good idea to open an issue first for discussion.
+
+## Checklist
+- [ ] I have followed [Sample Guidelines from AUTHORING_GUIDE.MD](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/AUTHORING_GUIDE.md)
+- [ ] README is updated (if needed)
+- [ ] **Tests** pass:   `nox -s py-3.6` (see [Test Enviroment Setup](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/AUTHORING_GUIDE.md#test-environment-setup))
+- [ ] **Lint** pass:   `nox -s lint` (see [Test Enviroment Setup](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/AUTHORING_GUIDE.md#test-environment-setup))
+- [ ] These samples need a new **API enabled** to pass (let us know which ones)
```
tmatsuo, line 12: FYI, this can be done by adding `required_api_url` if you're using `README.rst.in`.

Does it make sense to say `README has all the required steps including enabling the API, and setting environment variables` or sth?

kurtisvg: This is meant to mean that the samples need a new API enabled for our testing projects. I would what you've described to fall under "[ ] README is updated (if needed)", which is a few steps up.

tmatsuo: "README is updated if needed" is not clear enough for readers to recognize what should be in it.

kurtisvg: @tmatsuo - Fair enough, I've updated all three lines to clarify

PR #4067, opened 2020-06-11
Ndb migration sample

engelke:


engelke:
index.yaml not being used by tests, but works when manually deployed.

PR #4068, opened 2020-06-11
[jobs] testing: use multiple projects

### LINKED ISSUE #4061: [jobs] use multiple projects
We started to run tests in parallel, and enabled py-3.8 presubmit build.
Then we started to see some resource conflict errors in jobs tests.

We'd better use multiple projects for jobs tests.

tmatsuo:
fixes #4061


jobs/v4beta1/job_search_list_tenants_test.py
```diff
@@ -13,13 +13,30 @@
...
 
 import os
+import uuid
+
+from google.cloud import talent
+import pytest
 
 import job_search_list_tenants
 
 PROJECT_ID = os.environ["GOOGLE_CLOUD_PROJECT"]
```
munkhuushmgl, line 23: Since this env var is commented out in the noxfile_config.py, Does it get assigned from somewhere else?

tmatsuo: It's not commenting out :)

Long story, but we're copying [`noxfile-template.py`](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/noxfile-template.py) at the project root directory to `noxfile.py`, and use it for driving the test.

noxfile.py will [override the values](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/70677d25b9bcd9b490af9150172475a089804a03/noxfile-template.py#L74) set by the build config files (e.g. [py-3.6 config](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/70677d25b9bcd9b490af9150172475a089804a03/.kokoro/python3.6/common.cfg#L50))

munkhuushmgl: I see

munkhuushmgl:
LGMT!

munkhuushmgl:
@gguuss would you please take a look this? Looks like my Approval is not legit enough :')

PR #4069, opened 2020-06-11
Add a simplified inspect string example to DLP code samples

jlmwise:
Add a basic DLP code sample that just inspects with an infotype. This sample will be linked to the documentation at http://cloud/dlp/docs/concepts-infotypes#how_to_use_infotypes.

Please merge this PR after approving.

### DIFF
--- a/dlp/inspect_content.py
+++ b/dlp/inspect_content.py
@@ -66,11 +66,7 @@ def inspect_string_basic(
     # Print out the results.
     if response.result.findings:
         for finding in response.result.findings:
-            try:
-                if finding.quote:
-                    print("Quote: {}".format(finding.quote))
-            except AttributeError:
-                pass
+            print("Quote: {}".format(finding.quote))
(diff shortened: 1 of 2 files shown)


dlp/inspect_content.py
```diff
@@ -22,6 +22,64 @@
...
+    # Call the API.
+    response = dlp.inspect_content(parent, inspect_config, item)
+
+    # Print out the results.
+    if response.result.findings:
+        for finding in response.result.findings:
+            try:
+                if finding.quote:
+                    print("Quote: {}".format(finding.quote))
+            except AttributeError:
```
gguuss, line 72: When does the AttributeError occur?  We generally try to avoid `try ... catch` logic like this in sample code, maybe use `finding.get('quote', None)`.

jlmwise: It was copied from another example. But actually, in this example every finding should have a quote, so it can be removed.

PR #4074, opened 2020-06-12
pin rsa to 4.0

### LINKED ISSUE #4072: [py-2.7] testing: pin rsa-4.0
[Build](https://source.cloud.google.com/results/invocations/87900b4d-f290-4a67-a16c-58822fd225b3/targets)

We started to have deterministic failures on py-2.7 builds.

I think we need to pin to rsa-4.0 for some projects.

tmatsuo:
fixes #4072


busunkim96:
We just released a new version of google-auth that should handle this (install an older version of RSA for Python 2 ) https://github.com/googleapis/google-auth-library-python/pull/528

tmatsuo:
@busunkim96 Thanks. So maybe we can just upgrade google-auth.

PR #4080, opened 2020-06-12
[asset] Add sample code for two new RPCs.

yuyifan-google:


asset/cloud-client/quickstart_searchallresources_test.py
```diff
@@ -0,0 +1,62 @@
...
+
+@pytest.fixture(scope='module')
+def asset_dataset(bigquery_client):
+    dataset = bigquery_client.create_dataset(DATASET)
+
+    yield DATASET
+
+    try:
+        bigquery_client.delete_dataset(dataset)
+    except Exception as e:
```
busunkim96, line 44: We prefer not to swallow all exceptions - could you modify this to look for the `NotFound` exception instead?

https://github.com/GoogleCloudPlatform/python-docs-samples/blob/ee5be6df3f8e43bd7798bc56a39b7c5f273ca90f/dlp/triggers_test.py#L55-L60

yuyifan-google:
Do you know why the kokoro tests are not running?

busunkim96:
@yuyifan-google They only run automatically for the members of the `GoogleCloudPlatform` org. I just kicked off the tests, sorry to keep you waiting.

asset/cloud-client/quickstart_searchalliampolicies.py
```diff
@@ -0,0 +1,48 @@
...
+    # [START asset_quickstart_search_all_iam_policies]
+    from google.cloud import asset_v1
+
+    client = asset_v1.AssetServiceClient()
+    response = client.search_all_iam_policies(
+        scope, query=query, page_size=page_size)
+    for page in response.pages:
+        for policy in page:
+            print(policy)
+        break
```
busunkim96, line 31: Did you mean to leave the `break` here?

yuyifan-google: Yes, for this sample code I just want to print the first page, as there might be too many results.

asset/cloud-client/quickstart_searchallresources_test.py
```diff
@@ -0,0 +1,62 @@
...
+        print('Failed to delete dataset {}'.format(DATASET))
+        raise e
+
+
+def test_search_all_resources(asset_dataset, capsys):
+    scope = "projects/{}".format(PROJECT)
+    query = "name:{}".format(DATASET)
+
+    @backoff.on_exception(
+        backoff.expo, (AssertionError, InvalidArgument), max_time=30
```
kurtisvg, line 54: Why are we added a backoff for this sample? Can we add a comment with context on why this is?

yuyifan-google: It takes some time for the created dataset to be searchable. I realized sleep() may be more intuitive in this case than using backoff to retry the search.

tmatsuo: Sorry, but why? I recommend backoff over sleep.

yuyifan-google: Ok I'll switch back to backoff. I don't code python a lot so may I ask why backoff is preferred?

kurtisvg: Using backoff is tmatsuo's personal preference, and it's not one I agree with here. I believe `backoff.on_exception` has max_tries default to `None`, which means this will never fail and instead will continue to retry until the whole build fails out.

asset/cloud-client/quickstart_searchallresources_test.py
```diff
@@ -0,0 +1,63 @@
...
+    except NotFound as e:
+        print('Failed to delete dataset {}'.format(DATASET))
+        raise e
+
+
+def test_search_all_resources(asset_dataset, capsys):
+    scope = "projects/{}".format(PROJECT)
+    query = "name:{}".format(DATASET)
+
+    # The test will fail before the dataset creation event goes to CAI.
```
kurtisvg, line 53: This comment doesn't really provide enough context for me. Can we make it something like "The asset_dataset is not immediately searchable, so some time may be needed before the snippet will pass".

asset/cloud-client/quickstart_searchallresources_test.py
```diff
@@ -0,0 +1,63 @@
...
+    @backoff.on_exception(
+        backoff.expo, (AssertionError), max_time=30
+    )
```
kurtisvg, line 56: Is there a way to tell if this dataset is searchable as part of the fixture? I would much prefer a check at the fixture level to determine if the resource is searchable before yielding it. If that can only be accomplished by searching it, it's fine to leave this as is (but please make sure to include either `max_time` or `max_tries` so it doesn't timeout forever).

yuyifan-google: I don't think it's possible unless we do sleep() for a few seconds.

asset/cloud-client/quickstart_searchallresources_test.py
yuyifan-google: I have `max_time` set to 30. It take no more than a few seconds to propagate, so maybe reduce it to 10s?

kurtisvg: I'm fine with 30 - the difference seems negligible.

PR #4123, opened 2020-06-18
testing: update the python versions in the Dockerfile

### LINKED ISSUE #4126: [flake] Build failures due to python build failure
[Example build](https://source.cloud.google.com/results/invocations/b4676752-ef00-41af-a7c6-3999ecaf7697/targets)

tmatsuo:
also enable the Docker image upload in python 3.8 periodic build.

fixes #4126

### DIFF
--- a/.kokoro/docker/Dockerfile
+++ b/.kokoro/docker/Dockerfile
@@ -92,7 +92,7 @@ RUN curl https://packages.microsoft.com/keys/microsoft.asc | apt-key add - \
   && rm -f /var/cache/apt/archives/*.deb
 
 # Install the desired versions of Python.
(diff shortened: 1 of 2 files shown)


tmatsuo:
Doh. I overlooked there is python 2.7.18

tmatsuo:
The current logic for installing Python seems flaky. I think it fails to import gpg keys.
Also `gpg --verify ...` returns 0 even if there's no public key.

The tests in py-3.7 builds are all skipped because there's no python-3.7
Some tests in py-2.7 builds failed because there's non python.h for python2.7.

I'll modify the python installation to fail when key verification fails.

tmatsuo:
py-2.7 build failed with several errors with "bad zip files". I'll restart it.

tmatsuo:
The builds are all green, merging

PR #4128, opened 2020-06-19
testing: use default service account

tmatsuo:
We should be able to use Compute Engine default service account on Kokoro builds.

### DIFF
--- a/.kokoro/tests/run_tests.sh
+++ b/.kokoro/tests/run_tests.sh
@@ -55,13 +55,20 @@ fi
 # quota related errors. I think we can avoid this by changing the
 # order of tests to execute (e.g. reverse order for py-3.8
 # build). Currently there's no easy way to do that with btlr, so we
-# temporarily wait few minutes to avoid quota issue for py-3.8
-# presubmit build.
-if [[ "${RUN_TESTS_SESSION}" == "py-3.8" ]] \
-       && [[ "${KOKORO_JOB_NAME}" == *presubmit ]]; then
-    echo -n "Detected py-3.8 presubmit build,"
-    echo "Wait few minutes to avoid quota issues."
-    sleep 5m
+# temporarily wait few minutes to avoid quota issue for some
+# presubmit builds.
+if [[ "${KOKORO_JOB_NAME}" == *presubmit ]] \
+       && [[ -z "${DIFF_FROM:-}" ]]; then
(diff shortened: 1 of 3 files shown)


.kokoro/tests/run_tests.sh
```diff
@@ -55,13 +55,20 @@ fi
...
-    echo "Wait few minutes to avoid quota issues."
-    sleep 5m
+# temporarily wait few minutes to avoid quota issue for some
+# presubmit builds.
+if [[ "${KOKORO_JOB_NAME}" == *presubmit ]] \
+       && [[ -z "${DIFF_FROM:-}" ]]; then
+    if [[ "${RUN_TESTS_SESSION}" == "py-3.7" ]]; then
+	echo -n "Detected py-3.7 presubmit full build,"
+	echo "Wait 5 minutes to avoid quota issues."
+	sleep 5m
```
kurtisvg, line 65: Should these sleeps perhaps be configurable via an env var rather than session specifc?

tmatsuo: Well, technically it's possible, but I think it's a temporary workaround until btlr supports list of directories. Once btlr supports a list of directories, this effort will be moot.

tmatsuo:
The lint periodic build in a new VM pool successfully used the default service account:

https://source.cloud.google.com/results/invocations/f5191a17-52b2-470a-964b-1f0d9b624b52/targets

It worked for gcr.io download as well as secret access.

tmatsuo:
The iap tests are failing. My guess is the test picks up the new default service account attached to the VM. I'll try to add the new service account to the IAP access list.

tmatsuo:
I added the new service account to the IAP access list. I'll re-run the 3.6 build first.

tmatsuo:
Yes my guess is correct.

https://github.com/googleapis/google-auth-library-python/blob/06d7f97adaebb3b34ce6a159c23061dd2554e8ac/google/oauth2/id_token.py#L168

> This function acquires ID token from the environment in the following order:
> 
>   1. If the application is running in Compute Engine, App Engine or Cloud Run,
>       then the ID token are obtained from the metadata server.
>   2. If the environment variable ``GOOGLE_APPLICATION_CREDENTIALS`` is set
>        to the path of a valid service account JSON file, then ID token is
>        acquired using this service account credentials.
>   3. If metadata server doesn't exist and no valid service account credentials
>       are found, :class:`~google.auth.exceptions.DefaultCredentialsError` will
>       be raised.

PR #4136, opened 2020-06-20
fixed requirements (specified the version of grpcio and setuptools)

thesugar:
I fixed `codelabs/flex_and_vision/requirements.txt` (specified the version of grpcio and setuptools).

This code sample is used in https://google.qwiklabs.com/focuses/3339?parent=catalog.
If you don't specify the versions of `grpcio` and `setuptools` (= current configuration), build fails and you can't deploy to App Engine as below. So you can't complete the lab.

スクリーンショット 2020-06-20 12 23 58

After this change, you can deploy the app and get to finish the lab.

### DIFF
--- a/codelabs/flex_and_vision/requirements.txt
+++ b/codelabs/flex_and_vision/requirements.txt
@@ -4,3 +4,5 @@ gunicorn==19.10.0; python_version < '3.0'
 google-cloud-vision==1.0.0
 google-cloud-storage==1.29.0
 google-cloud-datastore==1.12.0
+grpcio==1.27.2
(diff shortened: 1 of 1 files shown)


busunkim96:
This is actually because of an old version of pip in the default Python image for App Engine Flex. I believe that problem was resolved last week. Can you try re-deplying @thesugar?

thesugar:
@tmatsuo Thank you for prompt check and reply!

@busunkim96 Thank you for your confirmation!
I got it. I'm glad to hear that this problem has been resolved.
So feel free to forget about this and discard my PR.

PR #4142, opened 2020-06-23
Update automl_tables_predict.py with batch_predict_bq sample

### LINKED ISSUE #4141: Doc: sample for AutoML batch_predict using BigQuery contains code for GCS case
The sample of executing batch prediction in "Using BigQuery tables" section for Python contains code using GCS not BQ. Please see:
https://cloud.google.com/automl-tables/docs/predict-batch#automl-tables-example-python

I'm suggesting adding a specific sample for using BigQuery.

evil-shrike:
Added a new method `batch_predict_bq` demonstrating running batch_prediction using BigQuery.
Added notes in comments about asynchronicity for `batch_predict` method.

The region `automl_tables_batch_predict_bq` will be used on cloud.google.com (currently both sections for GCS and BigQuery use the same sample code which is incorrect).

Fixes #4141


telpirion:
Please add unit tests and be sure to read the [AUTHORING_GUIDE](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/AUTHORING_GUIDE.md).

evil-shrike:
Hi Eric, I'll try to add some tests (it's a bit complicated setup), 
but FYTI currently there's no tests for batch_predict at all as I can see - https://github.com/GoogleCloudPlatform/python-docs-samples/blob/patch-1/tables/automl/predict_test.py (only for `predict`). 
Shall we move `batch_predict` samples into a separate file with its own tests?

telpirion:
Yes, if you want to break out the batch_predict samples and add tests, that would be awesome.

sirtorry:
I'll take care of the tests.

PR #4201, opened 2020-06-29
chore: changes boilerplate to LLC

telpirion:
Small change: switches to "Google, LLC" in boilerplate.

No region tags are affected.


tmatsuo:
@telpirion @leahecole 

Actually, you should not change this.

tmatsuo:
Ah it depends on when the file is added and whether Google was Inc or LLC at that time.

So go ahead if you're sure Google was already an LLC when the file is added :)

telpirion:
No worries ... I was using this as a demo of the PR process for a new team member.

It looks like Google became an LLC in 2015/2016.  I'll defer to you folks as to whether we should change this comment (or not). 

I'll find another example for this CL.

tmatsuo:
@telpirion Then this looks good.

PR #4223, opened 2020-07-01
feat(speech): Add Speech-to-Text On-Prem sample

### LINKED ISSUE #4222: Add Speech-to-Text On-Prem sample
## Description

Add `transcribe_onprem.py` to demonstrate Speech-to-Text On-Prem usage. The sample will be manually tested against an internally hosted service but **does not have automated tests** as this is a on-prem feature.

vinnysenthil:
## Description

Fixes #4222 

Adding sample to demonstrate speech transcription for on-prem customers. This sample does not currently contain tests due to being on-prem.

## Checklist
- [ ] **Tests** pass:   `nox -s py-3.6` (see [Test Enviroment Setup](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/AUTHORING_GUIDE.md#test-environment-setup))
- [x] **Lint** pass:   `nox -s lint` (see [Test Enviroment Setup](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/AUTHORING_GUIDE.md#test-environment-setup))


speech/cloud-client/transcribe_onprem.py
```diff
@@ -0,0 +1,91 @@
...
+    }
+    with io.open(local_file_path, "rb") as f:
+        content = f.read()
+    audio = {"content": content}
+
+    response = client.recognize(config, audio)
+    for result in response.results:
+        # First alternative is the most probable result
+        alternative = result.alternatives[0]
+        print(u"Transcript: {}".format(alternative.transcript))
```
busunkim96, line 68:
```suggestion
        print(f"Transcript: {alternative.transcript}")
```

nit: New samples only need to be compatible with 3.6+, so please use f-strings where it makes sense

speech/cloud-client/transcribe_onprem.py
```diff
@@ -0,0 +1,91 @@
...
+
+    # Set the API endpoint to direct requests to
+    client_options = {"api_endpoint": api_endpoint}
+
+    # Create a gRPC channel to your server
+    channel = grpc.insecure_channel(target=api_endpoint)
+
+    client = speech_v1p1beta1.SpeechClient(
+        client_options=client_options, channel=channel
+    )
```
busunkim96, line 44: There's no need to pass the `api_endpoint` through client options on the client constructor. Passing a channel is considered "advanced" mode so you're expected to set the endpoint on the channel itself.

speech/cloud-client/transcribe_onprem.py
vinnysenthil: Fixed, will opt for f-strings in the future

speech/cloud-client/transcribe_onprem.py
vinnysenthil: Thanks for pointing this out, removed `client_options` and just set the endpoint once in the channel

leahecole:
I understand that this sample will look a little different test-wise than others because of the nature of it being on-prem - I'm hesitant to have absolutely no testing though. My two questions related to that are:

1. How would we recommend a customer test this if they were running on prem? Without thinking too hard about it, my guess is it would be a system test with something on prem or something with mocks (or maybe something else - I really didn't go too deep in my thought process here) - and if that's the case, would it be valuable to include that here?
With an on-prem sample like this, how can you as the codeowner and we as repo owners know that this code is out of date? Normally we rely on tests for breaking changes, and I'm worried that with nothing we won't be able to figure out quickly when something is broken.

leahecole:
Just adding the "do not merge" label while we figure out an appropriate test strategy

PR #4230, opened 2020-07-02
feat(spanner): add sample for create instance

larkee:


spanner/cloud-client/snippets_test.py
```diff
@@ -22,19 +21,27 @@
...
+    return f'test-instance-{uuid.uuid4().hex[:10]}'
+
+
 def unique_database_id():
     """ Creates a unique id for the database. """
     return f'test-db-{uuid.uuid4().hex[:10]}'
 
 
-INSTANCE_ID = os.environ['SPANNER_INSTANCE']
+INSTANCE_ID = unique_instance_id()
```
hengfengli, line 34: Would it have a case that I want to use an existing instance for other tests (not `create_instance` test)?

hengfengli: In other words, is there a way to keep `SPANNER_INSTANCE` but also be able to run `create_instance` test?

PR #4248, opened 2020-07-06
Error checking on billing sample code for Slack

markmirch:
## Description
Some additional error checking for the Slack code, referenced on https://cloud.google.com/billing/docs/how-to/notify

## Checklist
- [x] **Tests** pass:   `nox -s py-3.6` (see [Test Enviroment Setup](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/AUTHORING_GUIDE.md#test-environment-setup))
- [x] **Lint** pass:   `nox -s lint` (see [Test Enviroment Setup](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/AUTHORING_GUIDE.md#test-environment-setup))


functions/billing/main.py
```diff
@@ -42,26 +44,36 @@
...
+
+    try:
+        slack_client.api_call(
+            'chat.postMessage',
+            json={
+                'channel': CHANNEL,
+                'text'   : budget_notification_text
+            }
+        )
+    except Exception:
```
leahecole, line 75: Can you specify an exception to catch? I worry that using the general `Exception` is too broad and may lead to poor DevX if things go wrong

markmirch: I think technically it would be SlackApiError, but I don't know the full details of the scope of Slack's API and the types of errors that may be thrown.

functions/billing/main.py
```diff
@@ -42,26 +44,40 @@
...
 
 def notify_slack(data, context):
     pubsub_message = data
 
-    notification_attrs = json.dumps(pubsub_message['attributes'])
-    notification_data = base64.b64decode(data['data']).decode('utf-8')
-    budget_notification_text = f'{notification_attrs}, {notification_data}'
+    # Default to blank strings to make testing the function easier
+    # See https://cloud.google.com/billing/docs/how-to/budgets-programmatic-notifications#notification_format
+    notification_attr = ''
```
ace-n, line 59: If this is _purely_ for testing, is there a way we can do without it?

(AFAIK we generally try to avoid any mention of test code in our samples, unless the test code itself has some associated educational/explanatory value.)

markmirch: Gotcha, let me rework that area.

functions/billing/main.py
markmirch: I'll specify SlackAPIError

functions/billing/main.py
```diff
@@ -42,26 +45,42 @@
...
+        notification_data = base64.b64decode(data['data']).decode('utf-8')
+    except KeyError:
+        notification_data = "No data passed in"
+
+    # This is just a quick dump of the budget data (or an empty string)
+    # You can modify and format the message to meet your needs
+    budget_notification_text = f'{notification_attr}, {notification_data}'
+
+    try:
+        slack_client.api_call(
```
jmgreger, line 75: Slack documentation presents a slightly different way of using the client here: https://github.com/slackapi/python-slackclient#sending-a-message-to-slack
The .api_call() function failed when passing in a channel parameter during testing.

```suggestion
        slack_client.chat_postMessage(
                'channel': CHANNEL,
                'text'   : budget_notification_text
                )
        
```

markmirch: Yeah I didn't update that part but it seemed to be working fine passing in a channel name (with #) or channel ID?

markmirch: That is, per the documentation here: https://python-slackclient.readthedocs.io/en/latest/basic_usage.html#calling-any-api-methods

PR #4257, opened 2020-07-07
cleanup(jobs): remove unnecessary code

tmatsuo:
## Description

Just removing unnecessary code.


leahecole:
I'm merging this and opening a bug with GitHub because it seems like the second review didn't persist

PR #4298, opened 2020-07-14
Ask contributors to put library samples in library repository

busunkim96:


tmatsuo:
I think it's totally fine to ignore `out-of-date` restriction for docs only changes :)

PR #4311, opened 2020-07-16
Remove no longer used environment variable

### LINKED ISSUE #4268: Confusion about use of FIREBASE_PROJECT_ID in firenotes sample

[appengine/standard/firebase/firenotes/backend/app.yaml](../blob/master/appengine/standard/firebase/firenotes/backend/app.yaml)

## Describe the issue

In [this tutorial](https://cloud.google.com/appengine/docs/standard/python/authenticating-users-firebase-appengine) a `FIREBASE_PROJECT_ID` environment variable is specified in the `app.yaml`, but it doesn't appear to be used anywhere.  I imagine it should be passed as the `audience` argument to [`verify_firebase_token`](https://google-auth.readthedocs.io/en/latest/reference/google.oauth2.id_token.html#google.oauth2.id_token.verify_firebase_token).  Since this wasn't done in the example code I thought that environment variable might be recognized by the library as providing a default for this argument, but as far as I can tell that isn't the case.

It would be nice to clarify this example.

engelke:
## Description

Fixes #4268

### DIFF
--- a/appengine/standard/firebase/firenotes/backend/app.yaml
+++ b/appengine/standard/firebase/firenotes/backend/app.yaml
@@ -8,6 +8,4 @@ handlers:
   script: main.app
 
 env_variables:
-  # Replace with your Firebase project ID.
-  FIREBASE_PROJECT_ID: '<PROJECT_ID>'
   GAE_USE_SOCKETS_HTTPLIB : 'true'


engelke:
Resolves issue https://github.com/GoogleCloudPlatform/python-docs-samples/issues/4268

embray:
Is `GOOGLE_CLOUD_PROJECT` a standard environment variable that is expected to be set?

engelke:
> Is `GOOGLE_CLOUD_PROJECT` a standard environment variable that is expected to be set?

In general, yes. The tests mostly expect it to be set, and platforms like App Engine set it automatically.

PR #4305, opened 2020-07-16
functions: consolidate redundant GCS samples

ace-n:
**Do not merge** until `functions_helloworld_storage_generic` has been removed from the docs.

Python fix for b/135560881

PR #4324, opened 2020-07-17
functions: add missing integration test samples

ace-n:
(The `storage` integration test will be added in #4305)


engelke:
Fix imports per lint. Otherwise LGTM.

PR #4335, opened 2020-07-20
testing: allow missing secrets

tmatsuo:
This allows users to use `scripts/run_tests_local.sh` without having
secret files.

A typical use case is to run readmegen or lint locally without preparing
their own secrets.

@rebeccawshaw FYI


tmatsuo:
I'm pretty sure it's safe to merge, using admin power.

PR #4359, opened 2020-07-22
fix: expand bigquery/cloud-client optional dependencies

shollyman:
This updates the bigquery client version and widens optional dependencies to include storage API integration.


shollyman:
Closing this as #174 is now merged, which changes how we do this.

PR #4369, opened 2020-07-24
functions: Use Slack client lib for request validation

ace-n:
**N.B:** this adds an additional package (`slackclient`) to do signature verification. (The code readability/maintainability benefits likely outweigh the drawbacks of adding an additional package.)


ace-n:
@grant this tutorial was written quite awhile ago - likely before [GSuite Chat[(https://gsuite.google.com/products/chat/) was released. I'm _potentially_ open to replacing it with one based around GSuite Chat.

@engelke I agree with your concern RE making this similar to other language samples. (Even if those languages have their own samples, this Python-specific implementation may come off as arbitrary.) Which would you (and @grant) say is more important here - consistency, or simpler code? (Personally I'm on the fence - I see value in both.)

ace-n:
Ping on this. 🙂

PR #4391, opened 2020-07-28
Delete Region tag run_events_gcs_server_setup

grant:
Delete Region tag run_events_gcs_server_setup. Not needed.


grant:
@engelke 1 region tag was renamed. 2 region tags were deleted.

`[START/END run_events_gcs_handler]` is there. The PR should be correct.

PR #4402, opened 2020-07-31
[ABANDONED] draft of DRIFT region tag tracking

ace-n:


.kokoro/tests/run_single_test.sh
```diff
@@ -41,6 +41,19 @@ fi
 nox -s "$RUN_TESTS_SESSION"
 EXIT=$?
 
+# Inject region tag data into the test log
+REGION_TAG_PARSER_DIR="$PARENT_DIR/region-tag-parser"
```
tmatsuo, line 45: Can you use a temporary directory? Also it's great if you make sure the temporary directory will be deleted after the test.

ace-n:
FYI - this is a draft PR. (No review necessary until I actually get this
working. 🙂)

.kokoro/tests/run_single_test.sh
tmatsuo: Update: maybe you can just install them in `/opt` or somewhere then it will only exists in the container.

`/tmp` is shared between the container and the host for a reason, so it's probably better to use somewhere else.

tmatsuo:
FYI, you can run the whole test locally. Here is an example for running lint full build:

```sh
$ cd python-docs-samples
$ TRAMPOLINE_BUILD_FILE=.kokoro/tests/run_tests.sh RUN_TESTS_SESSION=lint .kokoro/trampoline_v2.sh
```

For python sessions, you probably want to have secrets locally:
```sh
$ cd python-docs-samples
$ scripts/decrypt-secrets.sh # You need permission of course
$ TRAMPOLINE_BUILD_FILE=.kokoro/tests/run_tests.sh RUN_TESTS_SESSION=py-3.7 .kokoro/trampoline_v2.sh
```

If you want to use your own Cloud projects:
```sh
$ cd python-docs-samples
$ cp testing/test-env.tmpl.sh testing/test-env.sh  # and edit it
$ cp /some/where/your-service-account-json.key testing/service-account.json
$ TRAMPOLINE_BUILD_FILE=.kokoro/tests/run_tests.sh RUN_TESTS_SESSION=py-3.7 .kokoro/trampoline_v2.sh
```

This way, it will run the tests in an identical way as on Kokoro builds, so you can develop it locally without pushing any changes.

ace-n:
@tmatsuo thanks! I _think_ this needs to be run on Kokoro itself though, as part of what I'm testing is the ability to _export_ these test results (which I think happens outside the immediate build context).

ace-n:
Closing in favor of #4506

PR #4454, opened 2020-08-07
feat(firestore): async snippets

rafilong:
Add snippets for new [async Firestore interface](https://github.com/googleapis/python-firestore/tree/v2-staging). Snippets are ported from `firestore/cloud-client` to `firestore/cloud-async-client`.


rafilong:
Force push to overwrite history and make review easier. 6fb670d752c7719a46d45fc6455339689e5a7761 directly copies sync snippets to async directory, and 495ae796ff395a5b517ab9ccd47d40b5f705131f converts to async. Would recommend reviewing the latter for `snippets*.py`.

rafilong:
Question: are annotations scoped to the repo, project, or file? I assumed the latter, so snippets tags are shared between sync and async.

crwilcox:
We don't have a pattern yet but I feel like appending async to the region tag makes sense as it would group with the other sample. Still haven't settled on how to display these though. For instance, should they both show under python or do we want a new Python (async) tab similar to how Java has the Android version.

Either way we need to settle on a new region tag. How are the Android ones separated?

BenWhitehead:
It looks like in Android the java/kotlin separation is handled by the sample registry indirection (though I'm not familiar with how that actually works).

I'm okay with having an `_async` suffix if that's how we want to do it. Though, I'm not sure of the best way to do this. I know lots of snippets are managed lots of different ways.

rafilong:
@BenWhitehead @crwilcox would adding an `_async` suffix as a temporary stopgap so we can merge this in reasonable? It should be a relatively quick fix in the future to implement a better solution, and a suffix seems fairly reasonable.

BenWhitehead:
Adding the `_async` suffix SGTM

rafilong:
@BenWhitehead @crwilcox suffix added!

firestore/cloud-async-client/snippets_test.py
```diff
@@ -0,0 +1,308 @@
+# Copyright 2017, Google, Inc.
```
crwilcox, line 1: I think there may be enough change to rev this?

firestore/cloud-async-client/snippets.py
```diff
@@ -0,0 +1,774 @@
...
+    # [START structure_doc_ref_alternate_async]
+    a_lovelace_ref = db.document("users/alovelace")
+    # [END structure_doc_ref_alternate_async]
+
+    return a_lovelace_ref
+
+
+async def structure_subcollection_ref():
+    db = firestore.AsyncClient()
+    # [START structure_subcollection_ref_async]
```
crwilcox, line 290: this sample seems to be identical to sync? Do you have an idea of how common this is?

firestore/cloud-async-client/snippets_test.py
busunkim96: +1, I think there are enough differences to bump this to 2020

rafilong: Resolved!

firestore/cloud-async-client/snippets.py
rafilong: Not terribly common. A couple snippets just create documents/collections locally, of which there are no async calls. I would guess less than 10, maybe around 5. Should we remove these? I felt as though leaving them to illustrate how local creation is the same between sync / async was useful.

PR #4484, opened 2020-08-12
chore: fix comments on Dataflow example

sofisl:


composer/workflows/dataflowtemplateoperator_tutorial.py
```diff
@@ -15,10 +15,10 @@
 # [START composer_dataflow_dag]
 
 
-"""Example Airflow DAG that creates a Cloud Dataproc cluster, runs the Hadoop
-wordcount example, and deletes the cluster.
+"""Example Airflow DAG that creates a Cloud Dataflow workflow which takes a
+text file and adds the rows to a BigQuery table.
```
tmatsuo, line 19: Do we need `the`?
I mean `adds rows` instead of `adds the rows`?

leahecole: I think it's fine because it's referring to the text file

PR #4506, opened 2020-08-15
feat: Add DRIFT region tag parser to CI

ace-n:
Cleaned-up version of #4402 

@jmdobry FYI


tmatsuo:
python-3.7 build has a failure, and the new script is throwing an error:

```
Command pytest --junitxml=sponge_log.xml failed with exit code -6
Session py-3.7 failed.
cp: cannot stat '/workspace/speech/cloud-client/sponge_log.xml': No such file or directory
cat: /workspace/speech/cloud-client/drift_tmp.xml: No such file or directory
  File "/region-tag-parser/wizard-py/cli.py", line 200, in <module>
    inject_snippet_mapping(args.root_dir, sys.stdin.readlines())
  File "/region-tag-parser/wizard-py/cli.py", line 89, in inject_snippet_mapping
    xunit_tree = etree.fromstring("".join(stdin_lines))
  File "/usr/local/lib/python3.8/xml/etree/ElementTree.py", line 1321, in XML
    return parser.close()
xml.etree.ElementTree.ParseError: no element found: line 1, column 0
Testing failed: Nox returned a non-zero exit code.
```

Seems like the script needs to be little bit more robust.

Also, since you have changes in the Dockerfile, all the CI builds are full build.
I strongly recommend that you develop the script further locally.

You can emulate the Kokoro full build with the following command:

```
$ cd python-docs-samples
$ TRAMPOLINE_BUILD_FILE=.kokoro/tests/run_tests.sh .kokoro/trampoline_v2.sh
```

Full builds consume resources and interfere other PRs, so please be considerable.

.kokoro/docker/Dockerfile
```diff
@@ -174,4 +174,11 @@ RUN useradd -d /h -u ${UID} ${USERNAME}
 # Allow nopasswd sudo
 RUN echo "${USERNAME} ALL=(ALL) NOPASSWD:ALL" >> /etc/sudoers
 
+# Setup DRIFT region tag parser
+ENV REGION_TAG_PARSER_DIR /region-tag-parser
+ENV PARSER_PATH /region-tag-parser/wizard-py/cli.py
```
tmatsuo, line 179: Can you do `ENV PARSER_PATH ${REGION_TAG_PARSER_DIR}/wizard-py/cli.py`?

.kokoro/docker/Dockerfile
```diff
@@ -174,4 +174,11 @@ RUN useradd -d /h -u ${UID} ${USERNAME}
...
+
+RUN git clone https://github.com/GoogleCloudPlatform/repo-automation-playground "$REGION_TAG_PARSER_DIR" --single-branch
```
tmatsuo, line 181: It will always fetch it from head in the docker build.

This will prevent you from "releasing" the tool at will.

We always build the docker image, but we're utilizing previously built image as the cache. `docker build` command can not detect new changes in `repo-automation-playground` repo, so `docker build` will just use the cache layer and we won't get the new changes.

Maybe you want to have commit hash at least?

.kokoro/tests/run_single_test.sh
```diff
@@ -41,6 +41,13 @@ fi
 nox -s "$RUN_TESTS_SESSION"
 EXIT=$?
 
+# Inject region tag data into the test log
+XUNIT_PATH="$PWD/sponge_log.xml"
+XUNIT_TMP_PATH="$PWD/drift_tmp.xml"
```
tmatsuo, line 46: Can you use temp directory for XUNIT_TMP_PATH?

.kokoro/tests/run_single_test.sh
```diff
@@ -41,6 +41,13 @@ fi
...
+
+cp $XUNIT_PATH $XUNIT_TMP_PATH
```
tmatsuo, line 48: Maybe you want to guard it with checking if the file exists.

.kokoro/tests/run_single_test.sh
```diff
@@ -41,6 +41,13 @@ fi
...
+cat "$XUNIT_TMP_PATH" | python3.8 "$PARSER_PATH" inject-snippet-mapping "$PWD" > "$XUNIT_PATH"
```
tmatsuo, line 49: ditto

.kokoro/docker/Dockerfile
ace-n: I thought about doing that myself. 🙂 

Done.

.kokoro/tests/run_single_test.sh
ace-n: Is there a situation where `pytest` could fail to output an XUnit file?

.kokoro/tests/run_single_test.sh
ace-n: Done for `XUNIT_TMP_PATH`, though I _think_ `XUNIT_PATH` needs to stay in `$PWD`.

.kokoro/tests/run_single_test.sh
ace-n: See above comment.

.kokoro/tests/run_single_test.sh
ace-n: Never mind, added a guard here (for "defensive design" if nothing else).

.kokoro/docker/Dockerfile
ace-n: Good point - fixed.

.kokoro/docker/Dockerfile
```diff
@@ -174,4 +174,14 @@ RUN useradd -d /h -u ${UID} ${USERNAME}
...
+# Setup DRIFT region tag parser
+ENV REGION_TAG_PARSER_DIR /region-tag-parser
+ENV PARSER_PATH ${REGION_TAG_PARSER_DIR}/wizard-py/cli.py
+
+RUN git clone https://github.com/GoogleCloudPlatform/repo-automation-playground "$REGION_TAG_PARSER_DIR" --single-branch
+
+# Add commit hash so Docker rebuild images whenever region tag parser changes
+RUN cd ${REGION_TAG_PARSER_DIR} && git checkout 3a1a1c51ff92264d36489d8e9ceb1a9f2f760f6b
+RUN chmod +x $PARSER_PATH
+
```
tmatsuo, line 186: This certainly works.

However, if we change the commit hash, all the builds will be full builds.

Also, do you need this feature for all of the builds?

For example, is it ok to only run this in py-3.7 periodic build?

Sorry for being back and forth, but I'd like to avoid unnecessary full builds if possible.

.kokoro/tests/run_single_test.sh
```diff
@@ -41,6 +41,17 @@ fi
...
+XUNIT_PATH="$PWD/sponge_log.xml"
+XUNIT_TMP_PATH="tmp/drift_tmp.xml"
+
+if [ -f "$XUNIT_PATH" ]; then
+    cp $XUNIT_PATH $XUNIT_TMP_PATH
+    cat "$XUNIT_TMP_PATH" | python3.8 "$PARSER_PATH" inject-snippet-mapping "$PWD" > "$XUNIT_PATH"
+else
+    echo "No XUnit output file found!"
+fi
+
```
tmatsuo, line 54: The same question.

If it is reasonable to run this only on a single Python version periodic builds, can we introduce an environment variable for the switch?

Then can we only fetch the parser in `run_tests.sh`?

.kokoro/tests/run_tests.sh
```diff
@@ -83,6 +83,9 @@ export PATH="${HOME}/.local/bin:${PATH}"
 # install nox for testing
 pip install --user -q nox
 
+# install PyYaml (used by the DRIFT region tag parsing system)
+pip install --user -q pyyaml
+
```
tmatsuo, line 88: If we introduce an envvar, can we only install pyyaml if the envvar is set to true? Also I think it is ok to install the parser alongside of pyyaml.

Or even, how about to have requirements.txt in the parser repo and run `pip install --user -r ${PARSER_DIR}/requirements.txt` or sth?

tmatsuo:
I run py37 periodic build against your branch:
https://source.cloud.google.com/results/invocations/34c925d1-5261-466a-826b-6e8bc00c30ad/targets

I think we have little more to fix.

.kokoro/tests/run_single_test.sh
```diff
@@ -41,6 +41,18 @@ fi
 nox -s "$RUN_TESTS_SESSION"
 EXIT=$?
 
+# Inject region tag data into the test log
+if [ "${INJECT_REGION_TAGS:-}" == "true" ]]; then
```
tmatsuo, line 45: I think it should be:

```
if [[ "${INJECT_REGION_TAGS:-}" == "true" ]]; then
```
(one leading bracket is missing)


Also can you add info log?
```
echo "Injecting snippet mapping data"
```

.kokoro/tests/run_single_test.sh
```diff
@@ -41,6 +41,18 @@ fi
...
+    XUNIT_PATH="$PWD/sponge_log.xml"
+    XUNIT_TMP_PATH="tmp/drift_tmp.xml"
+    if [ -f "$XUNIT_PATH" ]; then
+        cp $XUNIT_PATH $XUNIT_TMP_PATH
+        cat "$XUNIT_TMP_PATH" | python3.7 "$PARSER_PATH" inject-snippet-mapping "$PWD" > "$XUNIT_PATH"
```
tmatsuo, line 50: If piped command `python3.7 "$PARSER_PATH" ...` fails, it will empty the `$XUNIT_PATH`.

Can you use a temporary file and move it after it succeeds?

.kokoro/tests/run_tests.sh
```diff
@@ -154,6 +154,20 @@ testing/btlr "${btlr_args[@]}"
 RTN=$?
 cd "$ROOT"
 
+# Setup DRIFT region tag parser
+# (only run on *some* builds)
+if [ "${INJECT_REGION_TAGS:-}" == "true" ]]; then
+    # install PyYaml (used by the DRIFT region tag parsing system)
```
tmatsuo, line 160: It's not obvious from the log, so can you add info log?

```
    echo "Downloading region tag parser"
```

or something?

Also, do you prefer always installing it from HEAD?

If you install with a SHA1 hash or release tag, can you also log the hash or the release tag?

ace-n: 1) Fixed missing bracket
2) Added debug log statements
3) Yes - installing from `HEAD` is the intent here (esp. since we don't need to use a commit hash to bust Docker caches!)

.kokoro/tests/run_single_test.sh
ace-n: Done.

.kokoro/tests/run_single_test.sh
ace-n: Done.

ace-n:
Reran the build [here](https://g3c.corp.google.com/results/invocations/e6ea889e-9ead-4601-862e-3276593cf5ee/targets)

.kokoro/tests/run_single_test.sh
```diff
@@ -41,6 +41,21 @@ fi
 nox -s "$RUN_TESTS_SESSION"
 EXIT=$?
 
+# Inject region tag data into the test log
+if [[ "${INJECT_REGION_TAGS:-}" == "true" ]]; then
+    echo "=== Injecting region tags into XUnit output ==="
+    XUNIT_PATH="$PWD/sponge_log.xml"
+    XUNIT_TMP_PATH="tmp/drift_tmp.xml"
```
tmatsuo, line 48: I don't think there is `tmp` directory.
How about:
```
XUNIT_TMP_PATH=$(mktemp ${HOME}/xunit-tmp.xxxxx)
```
 or something? The trampoline script will clean up ${HOME} upon exit, so you don't have to clean up on your own.

.kokoro/tests/run_single_test.sh
```diff
@@ -41,6 +41,21 @@ fi
...
+    if [[ -f "$XUNIT_PATH" ]]; then
+        echo "Processing XUnit output file: $XUNIT_PATH"
+        cat "$XUNIT_PATH" | python3.7 "$PARSER_PATH" inject-snippet-mapping "$PWD" > "$XUNIT_TMP_PATH"
+        mv $XUNIT_TMP_PATH $XUNIT_PATH
```
tmatsuo, line 52: I think you have to check the return value, maybe

```
if [[ $? -eq 0 ]]; then
    mv $XUNIT_TMP_PATH $XUNIT_PATH
fi
```
?

.kokoro/tests/run_tests.sh
tmatsuo: Ok, this raises another question. How is the parser script tested?

.kokoro/tests/run_tests.sh
ace-n: I got this working by installing `pyyaml` in a specific directory and adding that to the `PYTHONPATH`. (I think there are some `virtualenv` issues here that prevent the usual "`pip install` and go" approach.)

.kokoro/tests/run_tests.sh
ace-n: The _Python_ script itself is fetched [from this repo](https://github.com/googlecloudplatform/repo-automation-playground/tree/master/wizard-py); there are tests in that folder.

.kokoro/tests/run_single_test.sh
ace-n: AFAIK `/tmp` is a pretty standard directory on Linux.

(I fixed the missing `/`)

.kokoro/tests/run_single_test.sh
ace-n: Not necessary if you expect the script to consistently succeed. 😉 

(I added a guard here anyways, as it makes failures easier to debug.)

.kokoro/tests/run_tests.sh
tmatsuo: Do you have CI builds setup?

.kokoro/tests/run_single_test.sh
tmatsuo: It will work, but here is a reason I prefer ${HOME} over `/tmp`.

Currently Trampoline mount host's `/tmp` under `/tmp` for an unrelated reason, so if you create files under `/tmp`, the files will be remain in the host's `/tmp` directory.

Probably we should change this, but this is how it works.

Whereas, files in ${HOME} will be automatically clean up by Trampoline.

.kokoro/tests/run_single_test.sh
tmatsuo: Well, it's safe to assume that any process can fail for random reasons.

.kokoro/tests/run_single_test.sh
```diff
@@ -41,6 +41,33 @@ fi
...
+    export PARSER_PATH="${REGION_TAG_PARSER_DIR}/wizard-py/cli.py"
+    export PIP_PATH=/tmp/pyyaml
+    export PYTHONPATH="$PYTHONPATH:$PIP_PATH"
+
+    XUNIT_PATH="$PWD/sponge_log.xml"
+    XUNIT_TMP_PATH="/tmp/drift_tmp.xml"
+
+    if [[ -f "$XUNIT_PATH" ]]; then
+        echo "=== Injecting region tags into XUnit output ==="
+        pip install pyyaml -q -t $PIP_PATH
```
tmatsuo, line 56: Why do you need `$PIP_PATH`?

.kokoro/tests/run_single_test.sh
ace-n: Gotcha. I'll change this to `$HOME`, and keep the _parsing script itself_ in `tmp` (since that can be reused between invocations).

.kokoro/tests/run_single_test.sh
ace-n: The script wasn't able to find this library without it being installed in a specific directory, that is then added to the `$PYTHONPATH`.

(There might be cleaner workarounds, but this was the easiest one to implement. And given that these are largely-ephemeral CI machines, I didn't think "leaving some stuff behind" like this would be all that bad.)

tmatsuo: Hm. Can we run `pip install --user pyyaml` in `run_tests.sh`?

If we install it in `run_single_test.sh`, it will be installed over 200 times.

.kokoro/tests/run_tests.sh
ace-n: Not yet - though we plan to add them in the coming months.

(If you'd prefer to do that immediately after merging this PR though, that can be done.)

.kokoro/tests/run_single_test.sh
ace-n: Done.

ace-n:
New periodic build [here](https://g3c.corp.google.com/results/invocations/14e7aceb-e0e8-421f-95a2-f390c8722c19)

ace-n:
New periodic build [here](https://source.cloud.google.com/results/invocations/14e7aceb-e0e8-421f-95a2-f390c8722c19).

ace-n:
New periodic build [`here`](https://source.cloud.google.com/results/invocations/e978843c-e9d2-4f3d-a1ee-be8a777f9625/targets).

.kokoro/tests/run_single_test.sh
```diff
@@ -41,6 +41,32 @@ fi
...
 
+# Inject region tag data into the test log
+if [[ "${INJECT_REGION_TAGS:-}" == "true" ]]; then
+    export REGION_TAG_PARSER_DIR="/tmp/region-tag-parser"
+    export PARSER_PATH="${REGION_TAG_PARSER_DIR}/wizard-py/cli.py"
+    export PIP_PATH=/tmp/pyyaml
+    export PYTHONPATH="$PYTHONPATH:$PIP_PATH"
+
+    export XUNIT_PATH="$PWD/sponge_log.xml"
+    export XUNIT_TMP_PATH="${HOME}/drift_tmp_$(uuidgen).xml"
```
tmatsuo, line 52: Why do you prefer `uuidgen` over `mktemp`? I prefer less dependency.

.kokoro/tests/run_single_test.sh
```diff
@@ -41,6 +41,32 @@ fi
...
+
+    if [[ -f "$XUNIT_PATH" ]]; then
+        echo "=== Injecting region tags into XUnit output ==="
+        echo "Processing XUnit output file: $XUNIT_PATH (saving output to $XUNIT_TMP_PATH)"
+
+        cat "$XUNIT_PATH" | python3.7 "$PARSER_PATH" inject-snippet-mapping --output_file "$XUNIT_TMP_PATH" "$PWD"
+        if [[ $? -eq 0 ]]; then
+            mv $XUNIT_TMP_PATH $XUNIT_PATH
```
tmatsuo, line 60: Are you 100% sure that there is a file when the script exit with 0?
If not, please check the file existance.

Also, does it make sense to log something here as well?

.kokoro/tests/run_tests.sh
```diff
@@ -121,6 +121,27 @@ set +e
...
 
+# Setup DRIFT region tag injector
+# (only run on *some* builds)
+if [[ "${INJECT_REGION_TAGS:-}" == "true" ]]; then
+    echo "=== Setting up DRIFT region tag injector ==="
+    # install PyYaml (used by the DRIFT region tag parsing system)
+
+    export REGION_TAG_PARSER_DIR="/tmp/region-tag-parser"
+    export PARSER_PATH="${REGION_TAG_PARSER_DIR}/wizard-py/cli.py"
+    export PIP_PATH=/tmp/pyyaml
```
tmatsuo, line 132: Why did you need PIP_PATH again?

Did you try:

```
pip install --user pyyaml -q 
```
?

ace-n: IIRC, I did (and it didn't work).

.kokoro/tests/run_single_test.sh
ace-n: If the file at `$XUNIT_TMP_PATH` doesn't exist, then the `mv` command should fail.

IMO, logging something here would be overkill. 🙂

.kokoro/tests/run_tests.sh
tmatsuo: What's the error message?

.kokoro/tests/run_single_test.sh
tmatsuo: So if `mv` fails, the test is marked as failed even if just the injection here failed?
I don't like it.

.kokoro/tests/run_tests.sh
ace-n: If memory serves, it couldn't find the package (so probably something like `cannot find module 'yaml'`)

ace-n:
> So if mv fails, the test is marked as failed even if just the injection here failed?
I don't like it.

Maybe the cleaner answer here (instead of including a bunch of checks) is to use `set +e`?

tmatsuo:
@ace-n 
Question: is there any reason you don't want to check the file existence?

.kokoro/tests/run_tests.sh
tmatsuo: I think it should work.
Can you try this locally and paste the error log?

ace-n:
> @ace-n
> Question: is there any reason you don't want to check the file existence?

I'm not against it, but explicitly checking for every edge case makes this more complex (and in my view, harder to maintain). Personally, I think the `set e` approach is a (syntactically) cleaner solution.

.kokoro/tests/run_single_test.sh
tmatsuo: Well, actually, there's no `-e` flag set in this file. So it may keep running.
I still like the explicit check, but I think I can live with the current code.

tmatsuo:
@ace-n 
Actually this file does not `-e` flag set, so it will keep running even if the mv fails.

However, I think being explicit and verbose logging will help us debug things if something wrong happens. I still recommend my approach.

.kokoro/tests/run_tests.sh
ace-n: If I use that line and remove the `PYTHONPATH` modification:

```
Processing XUnit output file: /workspace/functions/env_vars/sponge_log.xml (saving output to /tmp/tmp.I0Hx5IhTZh)
  File "/tmp/region-tag-parser/wizard-py/cli.py", line 17, in <module>
    import analyze
  File "/tmp/region-tag-parser/wizard-py/analyze.py", line 19, in <module>
    import yaml_utils
  File "/tmp/region-tag-parser/wizard-py/yaml_utils.py", line 19, in <module>
    import yaml
ModuleNotFoundError: No module named 'yaml'
```

.kokoro/tests/run_single_test.sh
ace-n: Done - removed `uuidgen` in favor of `mktemp`.

ace-n:
@tmatsuo I went ahead and added a file existence check.

I also added `set -e` to trap any errors we might not think to explicitly check for.

.kokoro/tests/run_tests.sh
busunkim96: Hmm maybe the package is being installed to a pip associated with a different Python version. Could you try something like `python3.7 -m pip install ...`?

ace-n:
FYI @tmatsuo build failures are unrelated to this PR.

.kokoro/tests/run_single_test.sh
```diff
@@ -41,6 +41,34 @@ fi
 nox -s "$RUN_TESTS_SESSION"
 EXIT=$?
 
+# Inject region tag data into the test log
+set +e  # Don't fail the entire test if this step fails
+if [[ "${INJECT_REGION_TAGS:-}" == "true" ]]; then
+    export REGION_TAG_PARSER_DIR="/tmp/region-tag-parser"
+    export PARSER_PATH="${REGION_TAG_PARSER_DIR}/wizard-py/cli.py"
+    export PIP_PATH=/tmp/pyyaml
```
tmatsuo, line 49: Is `PIP_PATH` used?

tmatsuo:
Test failures are known ones, let's merge this.

PR #4533, opened 2020-08-20
chore: first pass of assign_issues_by

tmatsuo:
We're planning to start using blunderbuss's `assign_issues_by` feature, and here is the first pass of the config file changes.

### DIFF
--- a/.github/blunderbuss.yml
+++ b/.github/blunderbuss.yml
@@ -1,20 +1,20 @@
 assign_issues_by:
-  - labels:
-      - 'api: appengine'
-      to:
-      - engelke
-  - labels:
-      - 'api: automl'
-      to:
-      - telpirion
-  - labels:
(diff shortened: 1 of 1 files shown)


tmatsuo:
@tbpg FYI we're trying `assign_issues_by`

tmatsuo:
@leahecole It should be possible. I added torry

tbpg:
Yes, you can add as many people as you want for a particular set of labels. See https://github.com/googleapis/repo-automation-bots/tree/master/packages/blunderbuss.

Yes, you should be able to add teams. It doesn't assign the issue to someone _in the team_, though. See https://github.com/googleapis/repo-automation-bots/issues/284.

tmatsuo:
Not hearing from @billyjacobson , but let's merge this.

PR #4539, opened 2020-08-20
chore: Add kokoro-lite label

ace-n:
See b/165845522


ace-n:
FYI - this _does_ work, but the Docker portion of the build takes ~10 minutes.

tmatsuo:
@ace-n Not sure, but if we do so, we should generalize it a bit more. Feel free to file an issue on https://github.com/googlecloudplatform/docker-ci-helper

PR #4546, opened 2020-08-24
Fix modify_policy_add_role method

melaniedejong:
## Description

The current modify_policy_add_role method will always add a new binding, regardless of whether a binding for the role already exists. This fixes that issue by modifying the way that the method checks if the role exists in the array of policy bindings.

## Checklist

   For tests and linting: My local configuration is a bit messed up, which makes it difficult to run tests. I'll troubleshoot if this modificaiton causes any tests to fail.

- [X] Please **merge** this PR for me once it is approved.


iam/api-client/quickstartv2.py
```diff
@@ -63,8 +63,12 @@ def modify_policy_add_role(crm_service, project_id, role, member):
...
     policy = get_policy(crm_service, project_id)
 
-    if role in policy["bindings"]:
-        binding = next(b for b in policy["bindings"] if b["role"] == role)
+    binding = None
+    for b in policy["bindings"]:
+        if b["role"] == role:
+            binding = b
+            break
+    if binding != None:
```
busunkim96, line 71: Please use `is not` instead of `!=` for comparisons with None. See https://realpython.com/python-is-identity-vs-equality/#comparing-the-python-comparison-operators

```suggestion
    if binding is not None:
```

melaniedejong: Done.

PR #4594, opened 2020-08-28
testing: btlr update

### LINKED ISSUE #4103: testing: avoid timeout with btlr
@kurtisvg 

[Build](https://source.cloud.google.com/results/invocations/30295de3-f92a-44a5-9d5d-d5aff4ad2bda/targets)

This build timed out.

I think this happened because btlr unluckily run slow tests in a single worker.

Imagine there are 3 slow tests which takes one hour 5 minutes and btlr unfortunately schedule those 3 tests in a particular worker. This worker can not finish those 3 tests in 3 hours.

What came to mind is to manually pick some slow tests, put them in the beginning of the list.

```
tests_dirs=("datalabeling", "dlp")

# Then add the rest.
tests_dirs+=("proj1", "proj2")

testing/btlr run --max-concurrency 10 "${tests_dirs[@]}" -- "{$test_prog}"
```

This way, those slow tests are executed in parallel in the first batch.

What do you think?

### LINKED ISSUE #4104: testing: consider how to preserve logs for btlr workers even if the CI job timed out
[Example build](https://source.cloud.google.com/results/invocations/30295de3-f92a-44a5-9d5d-d5aff4ad2bda)

This build timed out. We lost majority of the invocation logs.
Questions:

Is it worth spending our time to preserve logs even the CI job times out.
(Actually all the xml test result files are uploaded, so maybe we can live with the current situation)
If yes, how to achieve this?

I think Kokoro uploads all the `build.log` in the artifact dir, but I'm not sure.

kurtisvg:
## Description

Updates btlr with a version that finishes printing logs upon sigterm signals, and adds a per-command timeout to prevent long jobs from hogging a thread for too long. 

Fixes #4104 
Fixes #4103


.kokoro/tests/run_tests.sh
```diff
@@ -146,6 +146,7 @@ test_prog="${PROJECT_ROOT}/.kokoro/tests/run_single_test.sh"
 
 btlr_args=(
     "run"
+    "--max-cmd-duration=30m"
```
tmatsuo, line 149: What happens when it timeouts? I presume the test will fail. Am I right?

kurtisvg: Yes (specifically, it sends a sigterm to the command and which will end the python process and give a non-zero exit code). 

The error reported will be "signal: killed" in the summary statement for the cmd.

PR #4643, opened 2020-09-08
feat(spanner): add samples for NUMERIC type

larkee:
Adds samples showing how to use the NUMERIC type in Cloud Spanner.


spanner/cloud-client/snippets.py
```diff
@@ -1397,6 +1449,34 @@ def query_data_with_string(instance_id, database_id):
     # [END spanner_query_with_string_parameter]
 
 
+def query_data_with_numeric_parameter(instance_id, database_id):
+    """Queries sample data using SQL with a NUMERIC parameter. """
+    # [START spanner_query_with_numeric_parameter]
+    # instance_id = "your-spanner-instance"
```
hengfengli, line 1455: nits: should we add these comments for the above samples?

hengfengli (approved):
LGTM.

thiagotnunes (approved):
LGTM

spanner/cloud-client/snippets.py
larkee: I was matching similar samples. The difference seems to be that the region here begins inside the the function so the comments show that they are defined, whereas the other sample region tags after around the function which means the variables are defined as arguments.

We can consider making the region tags more consistent but that's out of scope for this PR

PR #4648, opened 2020-09-09
chore(functions/firebase): add guard against infinite loops

ace-n:
N.B: this doesn't actually fail on Cloud Functions (since Firebase doesn't emit `update` events for documents that don't change) - but it _does_ rely on undocumented (and thus, not necessarily guaranteed!) behavior.

### DIFF
--- a/functions/firebase/main.py
+++ b/functions/firebase/main.py
@@ -106,7 +106,7 @@ def make_upper_case(data, context):
         })
     else:
         # Value is already upper-case
-        # Don't perform a(nother) write to avoid infinite loops
+        # Don't perform a second write (which can trigger an infinite loop)
         print('Value is already upper-case.')
 # [END functions_firebase_reactive]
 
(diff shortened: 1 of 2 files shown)


functions/firebase/main.py
```diff
@@ -98,11 +98,16 @@ def make_upper_case(data, context):
...
-        u'original': new_value
-    })
+    if cur_value != new_value:
+        print(f'Replacing value: {cur_value} --> {new_value}')
+        affected_doc.set({
+            u'original': new_value
+        })
+    else:
+        # Value is already upper-case
+        # Don't perform a(nother) write to avoid infinite loops
```
grant, line 109: - spell"another" as one word
- There is only one infinite loop per value write. Say loop singular or pluralize values and loops.

grant (approved):

functions/firebase/main.py
ace-n: Reworded - PTAL?

tmatsuo (approved):

grant (approved):

PR #4665, opened 2020-09-10
remove iot manager samples

arithmetic1728:
samples as moved to python-iot library.
docs reference updated: cl/330961368


busunkim96 (approved):

tmatsuo (approved):
There are also some samples in other directory.

busunkim96:
@tmatsuo This seems to be the only directory that uses `google-cloud-iot`. The others seem to be using `google-api-python-client.`

PR #4716, opened 2020-09-21
IAM: update quickstart comments

anguillanneuf:
1. Project IDs usually use dashes.
1. IAM members should not be limited to one form. 

Reference: https://cloud.google.com/sdk/gcloud/reference/iam/service-accounts/add-iam-policy-binding#--member


busunkim96 (approved):

iam/api-client/quickstartv2.py
```diff
@@ -117,8 +117,9 @@ def set_policy(crm_service, project_id, policy):
 if __name__ == '__main__':
 
     # TODO: replace with your project ID
-    project_id = "your_project_id"
-    # TODO: Replace with the ID of your member in the form 'user:member@example.com.
-    member = "your_member"
+    project_id = "your-project-id"
+    # TODO: Replace with the ID of your member in the form
+    # 'user|group|serviceAccount:email' or 'domain:domain'
```
kurtisvg, line 122: Can we add a link to the IAM reference documentation for this format?

kurtisvg (approved):

iam/api-client/quickstartv2.py
anguillanneuf: The link has some good examples of what `member` should look like.

kurtisvg (approved):

melaniedejong:
Since this is a quickstart and we want to be as directive as possible, I would prefer not to give the user multiple options for the member type. We're assuming at this point that they will be testing out IAM by adding a user.

The goal of the quickstart is just to give users a first-touch experience with IAM. If we need to instruct users about how to modify policy, including modifying policy for different member types, we should use the [Granting, changing, and revoking access page](https://cloud.google.com/iam/docs/granting-changing-revoking-access#iam-get-policy-python), and/or the [access sample](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/iam/api-client/access.py).

PR #4758, opened 2020-09-24
cleanup: replace sanity check with confidence check

tmatsuo:
Part of the inclusive fixit


leahecole (approved):
This is great - thank you for making this much needed change!!

tmatsuo:
@leahecole Thanks, admin merging

Issue #4362, opened 2020-07-23
[tests] Firestore: add asserts into snippets tests


IlyaFaer:
It was noticed that most of the [snippet tests](https://github.com/q-logic/python-docs-samples/blob/f5f3744438e468b4f71cd4524e9f2f1cf7811826/firestore/cloud-client/snippets_test.py) in Firestore doesn't do any asserts.
As per [comments](https://github.com/GoogleCloudPlatform/python-docs-samples/pull/4196#discussion_r451797229) it should be fixed.

/cc @tmatsuo, @busunkim96

dandhlee:
Friendly ping @IlyaFaer, could you work on adding this soon? Thanks!

PR #4763, opened 2020-09-28
test(firestore): add asserts into samples tests

IlyaFaer:
Towards #4362

Opening a preliminary PR to check if we're all okay with such a samples/tests style. I'm mimicking the [Storage fashion](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/2c15a82cc7a245513ccfb84419ba19e8dc165a1e/storage/cloud-client/storage_add_bucket_default_owner.py#L23-L33), where they send all the ids and data with sample function args. Within sample itself they write comments to describe, which values variables has.

If there are no counter-arguments on this, I'll do the same for all the other tests.

### DIFF
--- a/firestore/cloud-client/snippets.py
+++ b/firestore/cloud-client/snippets.py
@@ -29,15 +29,15 @@ def quickstart_new_instance():
     return db
 
 
-def quickstart_add_data_one():
+def quickstart_add_data_one(collection_id, document_id, data):
     db = firestore.Client()
     # [START quickstart_add_data_one]
-    doc_ref = db.collection(u'users').document(u'alovelace')
-    doc_ref.set({
-        u'first': u'Ada',
-        u'last': u'Lovelace',
-        u'born': 1815
-    })
+    # collection_id = "users"
+    # document_id = "alovelace"
+    # data = {"first": "Ada", "last": "Lovelace", "born": 1815}
+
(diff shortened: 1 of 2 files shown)


firestore/cloud-client/snippets.py
```diff
@@ -29,15 +29,15 @@ def quickstart_new_instance():
...
+def quickstart_add_data_one(collection_id, document_id, data):
     db = firestore.Client()
     # [START quickstart_add_data_one]
-    doc_ref = db.collection(u'users').document(u'alovelace')
-    doc_ref.set({
-        u'first': u'Ada',
-        u'last': u'Lovelace',
-        u'born': 1815
-    })
+    # collection_id = "users"
```
tmatsuo, line 35: The problem with this approach is that these comments are fundamentally not tested.

Here is an idea. Can you have a docstring with these value assignment examples, then in the test, extract those value assignments and test with those values?

IlyaFaer: I've tried to extract docstrings with `inspect` and run them with `eval/exec`. It doesn't look like a good variant, as because of scopes there is no way to write something universal - to parse all the snippets docstrings with one helper. And it actually looks ugly. So, for now I don't see how we can do it.

busunkim96:
Given how long this has been open I am going to close it for now. Please re-open it the change is still desired. Thanks!

PR #4785, opened 2020-10-01
Data cleaning region tags

leahecole:
## Description
Adds region tags for the data cleaning section, removes the gender column from the dataframe. Link to doc in staging available if you ping me :) 

## Checklist


leahecole:
fixing lint now ugh 🙈

busunkim96 (approved):

data-science-onramp/data-cleaning/clean.py
```diff
@@ -129,26 +137,33 @@ def compute_end_udf(duration, start, end):
...
     spark = SparkSession.builder.appName("data_cleaning").getOrCreate()
 
-    # Load data into dataframe if table exists
+    # Load data into temporary dataframe if table exists, drop unused column and load into new dataframe
+    # (using a temporary dataframe is an alternative to calling df.drop(column_name).collect() and is used
+    # to avoid OOM issues)
     try:
-        df = spark.read.format("bigquery").option("table", TABLE).load()
+        temp_df = spark.read.format("bigquery").option("table", TABLE).load()
+        df = temp_df.drop("gender")
```
bradmiro, line 155: Should this be more explicitly explained in the code why we are removing gender? Either way, I would move it outside of this try-block. It's also ok to do `df = df.<operation>` and not use the `temp_df` var name as this ends up happening in multiple other places in the script.

data-science-onramp/data-cleaning/clean_test.py
```diff
@@ -109,7 +106,7 @@ def setup_and_teardown_cluster():
         project_id=PROJECT_ID,
         region=CLUSTER_REGION,
         cluster_name=DATAPROC_CLUSTER,
-        timeout=300
+        timeout=300,
```
bradmiro, line 109: This doesn't hurt but you shouldn't need to do this. Were there timeout errors deleting the cluster?

bradmiro (changes requested):

data-science-onramp/data-cleaning/clean.py
leahecole: Np for moving it outside of try block and making that df = df.operation change. Reasoning for removal of the column is in the **Putting it together** section of the doc I sent you - do you think it also needs to be mentioned here?

data-science-onramp/data-cleaning/clean_test.py
leahecole: leftover from Tushar - can remove if need be

leahecole:
ugh, ofc these were passing locally. I'll figure out why they don't pass here!

data-science-onramp/data-cleaning/clean.py
bradmiro: I think that's fair, maybe it should have its own region tag then?

data-science-onramp/data-cleaning/clean_test.py
bradmiro: Removing is fine unless it somehow breaks

data-science-onramp/data-cleaning/clean.py
leahecole: Added region tag, but have to wait a few min before re-staging

data-science-onramp/data-cleaning/clean_test.py
leahecole: Will remove next, just testing locally now

leahecole: local tests passed, just pushed removal

data-science-onramp/data-cleaning/clean.py
```diff
@@ -147,17 +147,21 @@ def compute_end_udf(duration, start, end):
...
     except Py4JJavaError as e:
         raise Exception(f"Error reading {TABLE}") from e
 
-    # [END datascienceonramp_sparksession]
-    # [START datascienceonramp_sparksingleudfs]
+# [END datascienceonramp_sparksession]
+
+# [START datascienceonramp_removecolumn]
+    # remove unused column
+    # not using .collect() notation to avoid OOM
```
bradmiro, line 160: No need to mention `.collect()` here.

leahecole: done

bradmiro (approved):

PR #4797, opened 2020-10-05
Functions: add comment

### LINKED ISSUE #4671: PubSub future result is an indicator of publish success

https://github.com/GoogleCloudPlatform/python-docs-samples/blob/04a7a7f7ecdc2cfe343486e54ea15b9fcf7a9f89/functions/pubsub/main.py#L50-L57

## Describe the issue

It is my understanding that the `result()` method of the future will return the message ID on success. The `publisher.publish()` method may raise an exception, but only in certain ways (too large, or publisher was stopped). In this code block, by merely calling `result()` then returning a successful message, Cloud Functions will ack the message even if `result()` is `None` or `False`.

On a related note, I do something like this and I indeed have seen messages fail to publish.

```py
        if future.result():
            message.ack()
        else
            message.nack()
```

anguillanneuf:
Fixes #4671 

Users get confused by when a push message that triggers a function is ack'ed, despite [docs](https://cloud.google.com/functions/docs/calling/pubsub).

> Cloud Functions acks the message internally upon successful function invocation.

This comment helps clear this confusion. Even if the function subsequently fails to publish (any message, not limited to the push message that triggered the function), the original push message that triggered the function is ack'ed. 

Also see internal cl/335474263.


busunkim96 (approved):

grant:
Thanks for the PR.

I don't believe this suggested change will address the users's issue. Every function is invoked in every function sample, so we don't need to state that we invoked the function. If we made this change, we would also need to change all of our samples like Java and Node which use the same log statement.

Rather, I think a change to the docs would clarify and address the non-language-specific issue. Or a different comment than the changed one in this PR.

grant (changes requested):
Thanks for the PR.

I don't believe this suggested change will address the users's issue. Every function is invoked in every function sample, so we don't need to state that we invoked the function. If we made this change, we would also need to change all of our samples like Java and Node which use the same log statement.

Rather, I think a change to the docs would clarify and address the non-language-specific issue. Or a different comment than the changed one in this PR.

anguillanneuf:
That sounds good to me. I can close this PR after the internal CL is submitted. Thanks!

engelke (approved):

functions/pubsub/main.py
```diff
@@ -35,7 +35,7 @@ def publish(request):
     if not topic_name or not message:
         return ('Missing "topic" and/or "message" parameter.', 400)
 
-    print(f'Publishing message to topic {topic_name}')
+    print(f'Function is invoked. Next, publish message to topic {topic_name}')
```
ace-n, line 38: Nit:
> Function **was** invoked. **Now publishing** message to topic ...

leahecole:
@anguillanneuf I'm going to close this because the internal CL looks like it's submitted - please reopen if needed.

PR #4820, opened 2020-10-06
samples: refactored samples to v2

munkhuushmgl:
Fixes the problem with location_path

#4663


iot/api-client/manager/manager.py
```diff
@@ -93,6 +94,12 @@ def create_rs256_device(
 ):
     """Create a new device with the given id, using RS256 for
     authentication."""
+    # [START iot_create_rsa_device]
```
busunkim96, line 97: Please remove the region tags from this file. This file is only being used for test setup. The version displayed to users is in python-iot.

munkhuushmgl: ok

engelke (approved):
LGTM

munkhuushmgl:
@busunkim96 friendly ping

busunkim96 (approved):

PR #4823, opened 2020-10-07
chore(functions/pubsub): use consistent param names

ace-n:


functions/pubsub/main.py
```diff
@@ -22,7 +22,7 @@
 
 # Instantiates a Pub/Sub client
 publisher = pubsub_v1.PublisherClient()
-PROJECT_ID = os.getenv('GCP_PROJECT')
+PROJECT_ID = os.getenv('GOOGLE_CLOUD_PROJECT')
```
busunkim96, line 25: Do these point to the same thing in the Functions environment? https://cloud.google.com/functions/docs/env-var#nodejs_6_nodejs_8_python_37_and_go_111

grant (approved):

functions/pubsub/main.py
ace-n: Good catch - @grayside I assume this is still the right env var to use?

grayside: `GOOGLE_CLOUD_PROJECT` is the generally correct variable to use and present in some of the older runtimes. It must be user-supplied on deployment for newer runtimes, the current practice is to use the metadata server to get the current projectID.

grayside (approved):

busunkim96 (approved):

PR #4835, opened 2020-10-09
Adding e2e test for pubsub sample

dinagraves:
Adding end to end tests for the pub/sub sample.  B/c pub/sub doesn't return a message from the service, it uses the logging API to check that the service handled the message properly.


grant (approved):

run/pubsub/e2e_test.py
```diff
@@ -0,0 +1,151 @@
...
+    filters = (
+        f"timestamp>=\"{filter_date.isoformat('T')}\" "
+        "resource.type=cloud_run_revision "
+        f"AND resource.labels.service_name={CLOUD_RUN_SERVICE} "
+    )
+
+    # Retry a maximum number of 10 times to find results in stackdriver
+    found = False
+    for x in range(10):
+        iterator = client.list_log_entries(resource_names, filter_=filters)
```
averikitsch, line 143: Maybe add a sleep in between tries

averikitsch (approved):
Thanks @dinagraves !

run/pubsub/e2e_test.py
dinagraves: Added!  Also realized I was only breaking from the inner loop so added an extra break statement for the outer loop.

PR #4852, opened 2020-10-14
Adding e2e_test for Cloud Run image processing sample

dinagraves:
- Uploads image to one storage bucket
- Pub/Sub Notification to Cloud Run service
- Uploads blurred image to another storage bucket


grant (approved):

run/image-processing/e2e_test.py
```diff
@@ -0,0 +1,162 @@
...
+    ).stdout.strip()
+
+    yield service_url.decode()
+
+
+@pytest.fixture()
+def pubsub_topic(service_url):
+    # Create pub/sub topic
+    topic = f"image_proc_{SUFFIX}"
+    subprocess.run(
```
averikitsch, line 90: This is all really great, but in the future would it be better to this to be set up as a part of cloud build?

averikitsch (approved):

run/image-processing/e2e_test.py
dinagraves: Good question -- I didn't go down that road b/c the Cloud Build service agent didn't have pub/sub permissions in this project, but obviously we could update the permissions.

PR #4884, opened 2020-10-21
Fix appengine standard endpoints region tags

busunkim96:
These samples are not currently published to devsite (okay to merge immediately).

"endpoint" was added to the beginning of the region tags so they are attributed to the correct project. For duplicate region tags I added additional words to disambiguate them from each other.


engelke (approved):
LGTM

leahecole (approved):

PR #4934, opened 2020-11-02
Added tests for Cloud SQL samples

craiglabenz:
Adds tests for Postgres, MySQL, and SQL Server sample apps in the `/cloud-sql` directory.

## Checklist
- [x] These samples need a new/updated **env vars** in testing projects set to pass (let us know which ones)


cloud-sql/postgres/sqlalchemy/main.py
```diff
@@ -151,7 +174,13 @@ def create_tables():
 
 @app.route('/', methods=['GET'])
 def index():
+    context = get_index_context()
+    print(context)
```
shubha-rajan, line 178: should this print statement be here?

cloud-sql/sql-server/sqlalchemy/connection_test.py
```diff
@@ -0,0 +1,77 @@
...
+        pool = main.init_tcp_connection_engine()
+    except pyodbc.OperationalError as e:
+        logger.warning(
+            'Could not connect to the production database. '
+            'If running tests locally, is the cloud_sql_proxy currently running?'
+        )
+        raise e
+
+    with pool.connect() as conn:
+        conn.execute("SELECT GETDATE()")
```
shubha-rajan, line 56: Should this setup function create a table here like it does in the other tests?

cloud-sql/postgres/sqlalchemy/main.py
craiglabenz: Sure shouldn't!

cloud-sql/sql-server/sqlalchemy/connection_test.py
craiglabenz: It certainly could. My thinking was that the GET test confirms the table is created and read from. Since the SQL Server process involves all those heavy ORM objects to create a table (as opposed to a simple `CREATE TABLE` string), this can do its job by successfully executing any given query.

.kokoro/tests/run_tests.sh
```diff
@@ -108,9 +108,9 @@ export DATALABELING_ENDPOINT="test-datalabeling.sandbox.googleapis.com:443"
 # Run Cloud SQL proxy (background process exit when script does)
 wget --quiet https://dl.google.com/cloudsql/cloud_sql_proxy.linux.amd64 \
      -O ${HOME}/cloud_sql_proxy && chmod +x ${HOME}/cloud_sql_proxy
-${HOME}/cloud_sql_proxy -instances="${MYSQL_INSTANCE}"=tcp:3306 &>> \
+${HOME}/cloud_sql_proxy -instances="${MYSQL_INSTANCE}"=tcp:3306,"${MYSQL_INSTANCE}" -dir "${HOME}" &>> \
        ${HOME}/cloud_sql_proxy.log &
-${HOME}/cloud_sql_proxy -instances="${POSTGRES_INSTANCE}"=tcp:5432 &>> \
+${HOME}/cloud_sql_proxy -instances="${POSTGRES_INSTANCE}"=tcp:5432,"${POSTGRES_INSTANCE}"  -dir "${HOME}" &>> \
        ${HOME}/cloud_sql_proxy-postgres.log &
```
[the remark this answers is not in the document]
craiglabenz: Yes, I've since added that!

craiglabenz:
> Would it be better to just include the `**db_config` instead of this comment?

@kurtisvg Interesting idea. Now that the config samples show where `db_config` comes from, that probably would be sufficient for this spot as well.

I'll make that change.

cloud-sql/mysql/sqlalchemy/connection_test.py
```diff
@@ -0,0 +1,97 @@
...
+    except pymysql.err.OperationalError as e:
+        logger.warning(
+            'Could not connect to the production database. '
+            'If running tests locally, is the cloud_sql_proxy currently running?'
+        )
+        raise e
+
+    with pool.connect() as conn:
+        conn.execute(
+            "CREATE TABLE IF NOT EXISTS test_table "
```
kurtisvg, line 70: Should we add a UUID to this test? This test will run an few times simultaneously for different versions of Python

cloud-sql/mysql/sqlalchemy/main.py
```diff
@@ -132,11 +146,14 @@ def init_unix_connection_engine(db_config):
...
 # managing a pool of connections to your database
-db = init_connection_engine()
+# global db
+db = None
 
 
 @app.before_first_request
 def create_tables():
+    global db
+    db = db or init_connection_engine()
```
kurtisvg, line 156: Can we add a comment explaining why we want might already have an existing db?

cloud-sql/postgres/sqlalchemy/main.py
```diff
@@ -26,7 +26,16 @@
 
 
 def init_connection_engine():
+    # [START cloud_sql_postgres_sqlalchemy_limit]
```
kurtisvg, line 29: Can we pull the test changes and the region tag changes into separate PRs?

cloud-sql/postgres/sqlalchemy/main.py
```diff
@@ -109,11 +124,12 @@ def init_unix_connection_engine(db_config):
     cloud_sql_connection_name = os.environ["CLOUD_SQL_CONNECTION_NAME"]
 
     pool = sqlalchemy.create_engine(
+
         # Equivalent URL:
         # postgres+pg8000://<db_user>:<db_pass>@/<db_name>
         #                         ?unix_sock=<socket_path>/<cloud_sql_instance_name>/.s.PGSQL.5432
         sqlalchemy.engine.url.URL(
-            drivername="postgres+pg8000",
+            drivername="postgresql+pg8000",
```
kurtisvg, line 132: are we sure this is right? if so, we need to update the URL in the comments as well

kurtisvg (changes requested):

cloud-sql/postgres/sqlalchemy/main.py
craiglabenz: SQLAlchemy was spitting out lots of warnings about the driver name `"postgres"` being deprecated.

cloud-sql/mysql/sqlalchemy/connection_test.py
craiglabenz: ✅

shubha-rajan (approved):

grant (changes requested):
Changes to Cloud Run samples seem out of scope for the desired Cloud SQL sample changes.

I believe @averikitsch has more context/requested changes for the Cloud Run region tags.

craiglabenz:
Closed in favor of cleaner #4977

PR #4944, opened 2020-11-05
feat(samples): added new region tag

HemangChothani:


busunkim96 (approved):
CC @crwilcox 

It looks like there's some duplication of region tags (one example `firestore_setup_dataset_pt1`). Is this intentional? I believe region tags are expected to be globally unique.

crwilcox:
@busunkim96 that isn't expected, also, due to potential lagging behind as we publish documentation including these snippets, previous region tags should be left 

From notes: "Update all samples to be ‘dual homed’, being surrounded by both the old region tag and a new region tag that will be consistent across languages."

Once these are all changed, I intent to update our inclusions for these in docs.

leahecole (approved):

HemangChothani:
CC @crwilcox 
I have updated all samples into ‘dual homed'.

HemangChothani:
@crwilcox @leahecole @leahecole Can anyone help to run the kokoro, so this PR can merge. I am unable to add kokoro.

PR #4971, opened 2020-11-12
public access prevention samples & tests

shaffeeullah:
Samples for the upcoming Public Access Prevention feature. Pending the merge of client library (https://github.com/googleapis/python-storage/pull/304) and release of backend feature.


storage/cloud-client/storage_get_public_access_prevention.py
```diff
@@ -0,0 +1,49 @@
...
+
+import sys
+
+# [START storage_get_public_access_prevention]
+from google.cloud import storage
+from google.cloud.storage.bucket import PUBLIC_ACCESS_PREVENTION_UNSPECIFIED
+
+
+def get_public_access_prevention(bucket_name):
+    """Get public access prevention bucket"""
```
tritone, line 25: `Gets the public access prevention setting for a bucket.` (these should all be full sentences with periods I think)

storage/cloud-client/storage_get_public_access_prevention.py
```diff
@@ -0,0 +1,49 @@
...
+    # bucket_name = "my-bucket"
+
+    storage_client = storage.Client()
+    bucket = storage_client.get_bucket(bucket_name)
+    iam_configuration = bucket.iam_configuration
+
+    if iam_configuration.public_access_prevention is PUBLIC_ACCESS_PREVENTION_UNSPECIFIED:
```
tritone, line 32: As in the node sample, I think this conditional is a bit confusing. I don't think a simple if/else makes sense unless the value is a bool, which it isn't. Maybe just have the value printed and include a comment that possible values are `enforced` and `unspecified`?

storage/cloud-client/storage_unspecify_public_access_prevention.py
```diff
@@ -0,0 +1,42 @@
...
+
+import sys
+
+# [START storage_unspecify_public_access_prevention]
+from google.cloud import storage
+from google.cloud.storage.bucket import PUBLIC_ACCESS_PREVENTION_UNSPECIFIED
+
+
+def unspecify_public_access_prevention(bucket_name):
+    """Unspecify public access prevention for a bucket"""
```
tritone, line 25: I think it's not that useful to have a docstring that just restates the function name basically. More relevant here is that this removes public access prevention for the bucket (so we will instead inherit the org-level policy). I guess we should work with Karen to clarify what should go in the docstring and what should go in the text above the sample block, though.

storage/cloud-client/public_access_prevention_test.py
```diff
@@ -0,0 +1,79 @@
...
+from google.cloud import storage
+import pytest
+
+import storage_enforce_public_access_prevention
+import storage_get_public_access_prevention
+import storage_unspecify_public_access_prevention
+
+
+@pytest.fixture()
+def bucket():
```
tritone, line 28: Is this something that could/should be shared across sample tests that use buckets?

storage/cloud-client/public_access_prevention_test.py
```diff
@@ -0,0 +1,79 @@
...
+    bucket.create()
+    yield bucket
+    time.sleep(3)
+    bucket.delete(force=True)
+    # Set the value back.
+    os.environ['GOOGLE_CLOUD_PROJECT'] = original_value
+
+
+def test_get_public_access_prevention(bucket, capsys):
+    short_name = storage_get_public_access_prevention
```
tritone, line 47: What is the reason for declaring this as a variable?

tritone (commented):
Generally looks good. Let's chat on Monday about how to assign out snippets for the remaining languages.

storage/cloud-client/storage_unspecify_public_access_prevention.py
shaffeeullah: Good point, I will follow up with Karen on this.

storage/cloud-client/public_access_prevention_test.py
shaffeeullah: I based these tests off of the [UBLA tests](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/f0a5cbec18011f956c8d5b795837f1ad7b9d7fbf/storage/cloud-client/uniform_bucket_level_access_test.py#L58) and this is how those are written

storage/cloud-client/public_access_prevention_test.py
shaffeeullah: It could be shared. I'm not sure it should be. I based these tests off of the UBLA tests since these are similar features and the UBLA tests do not share the bucket.

storage/cloud-client/storage_unspecify_public_access_prevention.py
karenarialin: +1, something along the lines of "Sets the public access prevention status to unspecified, so that the bucket inherits its setting from its parent project."

storage/cloud-client/public_access_prevention_test.py
tritone: Oh I didn't mean that the bucket should be shared, but rather the fixture code could be. Does that make sense?

shaffeeullah: Ah I understand now. I will look into it!

storage/cloud-client/public_access_prevention_test.py
frankyn: supa long lines, was my rationale for using this lols...

storage/cloud-client/storage_set_public_access_prevention_enforced.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+def set_public_access_prevention_enforced(bucket_name):
+    """Enforce public access prevention for a bucket."""
+    # bucket_name = "my-bucket"
+
+    storage_client = storage.Client()
+    bucket = storage_client.get_bucket(bucket_name)
+
+    bucket.iam_configuration.public_access_prevention = PUBLIC_ACCESS_PREVENTION_ENFORCED
```
frankyn, line 31: I would recommend switching to string literals instead of constants because they're not used actively in our samples. 

I may have missed samples that are now using string constants though. I would still stay away from it unless it's a strict requirement.

shaffeeullah: Other samples in this library (such as `change_default_storage_class`) are using string constants. Why do you suggest staying away from it? I believe this is a safer implementation. Someone using this constant from the library will not have to change their code if we change the value of this constant down the line.

frankyn: I'm making a change to address the usage of constants in that sample here: https://github.com/GoogleCloudPlatform/python-docs-samples/pull/5329

The constants.py isn't intentional about how a developer is going to use them. I agree that this is something we should do (rely on constants provided by the library), but we need to audit the existing constants experience.

For example, some are located [constants.py](https://github.com/googleapis/python-storage/blob/master/google/cloud/storage/constants.py) and others are located here: [iam.py](https://github.com/googleapis/python-storage/blob/master/google/cloud/storage/iam.py#L22-L41). For PAP, they were added [here](https://github.com/googleapis/python-storage/blob/f9498b982d5011971ca7809ab279cf738bcc07f1/google/cloud/storage/bucket.py#L89).

I agree this makes sense for supportability reasons as you stated, but we should be intentional / guided about where these constants exist which in turn will make it easier for developers to find them. Otherwise, it's a maze.

I might be too cautious as well, but samples have used string literals for the most part.

storage/cloud-client/public_access_prevention_test.py
andrewsg: Nothing wrong with this, but it's also ok to just type it in long and let the autoformatter deal with it.

storage/cloud-client/public_access_prevention_test.py
```diff
@@ -0,0 +1,53 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import storage_get_public_access_prevention
+import storage_set_public_access_prevention_enforced
+import storage_set_public_access_prevention_unspecified
+
+
+def test_get_public_access_prevention(bucket, capsys):
+    short_name = storage_get_public_access_prevention
```
andrewsg, line 21: I would prefer to either use the long name and let the autoformatter deal with it, or if that's really unacceptably long, to do the original import differently ("import storage_get_public_access_prevention as storage_get_pap" or something). Not a critical change since it's only in the test - up to you.

andrewsg (approved):
LGTM, with one optional style suggestion.

storage/cloud-client/public_access_prevention_test.py
shaffeeullah: @andrewsg , I wrote these tests to match the [UBLA tests](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/f0a5cbec18011f956c8d5b795837f1ad7b9d7fbf/storage/cloud-client/uniform_bucket_level_access_test.py#L58) since they're similar features. Those are written with `short_name`. I don't mind changing it, but changing it will create an inconsistency. Is that okay?

storage/cloud-client/conftest.py
```diff
@@ -0,0 +1,25 @@
+import os
```
dinagraves, line 1: Please add license

dinagraves (changes requested):
Hi! Can you re-request review when the tests pass? LMK if you need help.

shaffeeullah:
@andrewsg ,

We would like to merge these samples next week after the client library has been released. Does that work for you?

storage/cloud-client/public_access_prevention_test.py
andrewsg: Sorry I didn't respond here earlier. Sure, fine to keep it for consistency. Thanks!

andrewsg:
@shaffeeullah Yep sounds great. Let me know if you need anything from me to make that happen.

shaffeeullah:
Public access prevention rollout has been delayed due to a bug surfaced during Googler preview. I will keep this PR updated as I learn new release timeline details.

shaffeeullah:
@dinagraves The client library implementation has been released https://github.com/googleapis/python-storage/pull/485 , but it seems these tests are still failing. Do you know how I can have them run against the latest version of the library?

storage/cloud-client/storage_get_bucket_metadata.py
```diff
@@ -39,6 +39,7 @@ def bucket_metadata(bucket_name):
     )
     print("Default KMS Key Name: {}".format(bucket.default_kms_key_name))
```
dinagraves, line 40: Please use f-strings.

storage/cloud-client/storage_get_public_access_prevention.py
```diff
@@ -0,0 +1,41 @@
...
+    # The ID of your GCS bucket
+    # bucket_name = "my-bucket"
+
+    storage_client = storage.Client()
+    bucket = storage_client.get_bucket(bucket_name)
+    iam_configuration = bucket.iam_configuration
+
+    print(
+        "Public access prevention is {} for {}."
+        .format(iam_configuration.public_access_prevention, bucket.name)
```
dinagraves, line 34: f string

dinagraves (changes requested):

storage/cloud-client/storage_get_bucket_metadata.py
tritone: These samples still need to work with Py 2.7 right? I thought that ruled out fstrings.

dinagraves: AFAIK, App Engine Standard is the only product that is still required to work with 2.7.    These samples do not run tests on 2.7.

storage/cloud-client/public_access_prevention_test.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+def test_set_public_access_prevention_enforced(bucket, capsys):
+    short_name = storage_set_public_access_prevention_enforced
+    short_name.set_public_access_prevention_enforced(
+        bucket.name
+    )
+    out, _ = capsys.readouterr()
+    assert (
+        "Public access prevention is set to enforced for {}.".format(bucket.name)
```
dinagraves, line 39:
```suggestion
        f"Public access prevention is set to enforced for {bucket.name}."
```

storage/cloud-client/public_access_prevention_test.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+def test_get_public_access_prevention(bucket, capsys):
+    short_name = storage_get_public_access_prevention
+    short_name.get_public_access_prevention(
+        bucket.name
+    )
+    out, _ = capsys.readouterr()
+    assert (
+        "Public access prevention is unspecified for {}.".format(bucket.name)
```
dinagraves, line 27:
```suggestion
        f"Public access prevention is unspecified for {bucket.name}."
```

storage/cloud-client/conftest.py
```diff
@@ -0,0 +1,41 @@
...
+@pytest.fixture(scope="function")
+def bucket():
+    """Yields a bucket that is deleted after the test completes."""
+    # The new projects enforces uniform bucket level access, so
+    # we need to use the old main project for now.
+    original_value = os.environ['GOOGLE_CLOUD_PROJECT']
+    os.environ['GOOGLE_CLOUD_PROJECT'] = os.environ['MAIN_GOOGLE_CLOUD_PROJECT']
+    bucket = None
+    while bucket is None or bucket.exists():
+        bucket_name = "uniform-bucket-level-access-{}".format(uuid.uuid4().hex)
```
dinagraves, line 34:
```suggestion
        bucket_name = f"uniform-bucket-level-access-{uuid.uuid4().hex}"
```

storage/cloud-client/public_access_prevention_test.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+def test_set_public_access_prevention_unspecified(bucket, capsys):
+    short_name = storage_set_public_access_prevention_unspecified
+    short_name.set_public_access_prevention_unspecified(
+        bucket.name
+    )
+    out, _ = capsys.readouterr()
+    assert (
+        "Public access prevention is 'unspecified' for {}.".format(bucket.name)
```
dinagraves, line 51:
```suggestion
        f"Public access prevention is 'unspecified' for {bucket.name}."
```

dinagraves (approved):

PR #5052, opened 2020-12-08
fix: add a comment to draw attention to using get_blob, not blob

crwilcox:
There have been reports that perhaps `blob` and `get_blob` and there differences can be confusing to users. This attempts to bring in some context to popular samples to help clarify.


crwilcox:
@frankyn @sdspikes I have added a bit of documentation to two samples. This may be a bit much but thought we can hone in a bit.

busunkim96 (approved):

frankyn (approved):

PR #5221, opened 2021-01-11
Add feature engineering notebook

leahecole:
CI/CD was attempted [here](https://github.com/GoogleCloudPlatform/ai-platform-samples/pull/234), but that process isn't 100% fleshed out yet so let me explain the failures:

- Lint - this is to be expected. There are three failures - two are about the block comments in the first and last cells letting the user know they can ignore them. They are formatted as my review (bmiro@) asked. The third failure is due to a redefinition of an import. This is also expected and okay - the first instance of the import is in a block that users are not expected to run. 
- notebook run - Torry and I spent a really long time trying to get this to work properly but ran into failures with the underlying infrastructure. What the notebook run script does is tests that the notebook runs from top to bottom. I was able to not only run the notebook from top to bottom, but was also able to execute the [test script](https://github.com/GoogleCloudPlatform/ai-platform-samples/blob/master/.kokoro/tests/run_notebooks_only_diff.sh) locally with slight modifications (removing the part that parses the git tree looking for differences and changing the directory it points to in 173, no changes to the actual notebook running parts). Torry, my notebook expert, felt okay with this

@bradmiro has reviewed this extensively in a branch and can give approval here as well, and we can wait for @sirtorry to approve when back in office (1/22) if you'd like


leahecole:
Code written not by me was written by interns who had signed CLAs over the summer

bradmiro (approved):

engelke (approved):
I recommend adding a requirements.txt file to make it easier to run this locally.

LGTM

leahecole:
@engelke this is a good point! I actually just double checked with Gonzalo while Torry is OOO about what is preferred for AI Platforms and notebooks. Given that this is to be run directly on AI Platform Notebooks as part of a tutorial (ping me for staged docs) the preference is to have installs inline with the cells (instead of a requirements.txt). For other parts of AI Platform though, the preference is to have requirements.txt, and I'll ensure those are there for those tutorials.

leahecole:
As previously stated CLA was signed by interns

PR #5266, opened 2021-01-20
feat(AST parser): use v2 parser architecture

ace-n:
This migrates the `python-docs-samples` repo off the Python-only `v1` AST parser architecture and onto the more language-agnostic `v2` (or "polyglot parser") architecture.

### DIFF
--- a/.kokoro/tests/run_single_test.sh
+++ b/.kokoro/tests/run_single_test.sh
@@ -50,28 +50,36 @@ if [[ "${INJECT_REGION_TAGS:-}" == "true" ]]; then
     export XUNIT_PATH="$PWD/sponge_log.xml"
     export XUNIT_TMP_PATH="$(mktemp)"
 
+    # We use `python3.8` because it is the version pip3 installs to.
+    export PYTHON_CMD="python3.8"
+
     if [[ -f "$XUNIT_PATH" ]]; then
         echo "=== Injecting region tags into XUnit output ==="
         echo "Processing XUnit output file: $XUNIT_PATH (saving output to $XUNIT_TMP_PATH)"
 
+        TARGET_DIR=$PWD
+
         # First, we generate a Python-specific "polyglot_snippet_data.json" file
-    	#   We use `python3` because it will work even if we remove old
-    	#   python versions from the docker image.
-        echo "Calling python3 ${POLYGLOT_PARSER_PATH} ${PWD}"
-        python3 "$POLYGLOT_PARSER_PATH" "$PWD"
+        # (cd'ing into script's working dir is necessary for imports to work)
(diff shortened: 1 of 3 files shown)


.kokoro/tests/run_single_test.sh
```diff
@@ -50,20 +50,36 @@ if [[ "${INJECT_REGION_TAGS:-}" == "true" ]]; then
     export XUNIT_PATH="$PWD/sponge_log.xml"
     export XUNIT_TMP_PATH="$(mktemp)"
 
+    # We use `python3.8` because it is the version pip3 installs to.
+    export PYTHON_CMD="python3.8"
```
busunkim96, line 54: Is it okay to assume this @tmatsuo @kurtisvg?

ace-n: N.B: If not, we'll have to find some other way to have `pip` target a different version.

(We could manually add to `PYTHONPATH`, but that feels kludgey IMO.)

.kokoro/tests/run_tests.sh
```diff
@@ -130,18 +130,22 @@ ROOT=$(pwd)
 if [[ "${INJECT_REGION_TAGS:-}" == "true" ]]; then
     echo "=== Setting up DRIFT region tag injector ==="
     # install PyYaml (used by the DRIFT region tag parsing system)
-    echo "--- Installing PyYaml ---"
-    python3 -m pip install --user pyyaml
+    echo "--- Installing pip packages ---"
+    pip3 --version
+    pip3 install --user pyyaml frozendict recordclass
```
tmatsuo, line 135: Isn't it possible to use `python3 -m pip install --user` here and `python3` in `run_single_test.sh`?

.kokoro/tests/run_single_test.sh
tmatsuo: If possible, I'd like to just use major version e.g. `python3`.

.kokoro/tests/run_tests.sh
ace-n: Not without some kludgery - I get a `ModuleNotFound` error for `frozendict` when using `python3`. Furthermore, I get `pip 20.3.3 from /usr/local/lib/python3.8/site-packages/pip (python 3.8)` when I run `pip3 --version`.

If I run `python3 -m pip install --user frozendict` (in the container's `bash` prompt, not the script itself) I get `No module named pip`.

.kokoro/tests/run_single_test.sh
tmatsuo: I will add some commits to use `python3` instead of using the specific version.

tmatsuo (approved):

ace-n:
Python 3.7 failures look unrelated. @busunkim96 can you please approve and merge? (Thanks! 🙂)

busunkim96 (approved):

PR #5278, opened 2021-01-21
(storage) add sample for check latest transfer operation in transfer …

JesseLovelace:
Adds a sample for check latest transfer operation in transfer service


storage/transfer_service/sts_snippets_test.py
```diff
@@ -0,0 +1,37 @@
...
+    storagetransfer = googleapiclient.discovery.build("storagetransfer", "v1")
+    result = storagetransfer.transferJobs().create(body=transfer_job).execute()
+
+    job_name = result.get("name")
+    check_latest_transfer_operation.check_latest_transfer_operation(
+        project_id, job_name
+    )
+    out, _ = capsys.readouterr()
+
+    assert job_name in out
```
frankyn, line 37: Can you assert `The latest operation for job` is included as well because job_name is both in the success case and failure case above.

frankyn (changes requested):
Few nits, otherwise LGTM

storage/transfer_service/sts_snippets_test.py
```diff
@@ -0,0 +1,37 @@
+import os
```
frankyn, line 1: Missing copyright header*

storage/transfer_service/noxfile.py
```diff
@@ -0,0 +1,245 @@
+# Copyright 2019 Google LLC
```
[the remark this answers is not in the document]
JesseLovelace: This is what the authoring guide says to do:

>python-docs-samples: As a temporary workaround, each project currently uses first noxfile-template.py found in a parent folder above the current sample. In order to simulate this locally, you need to copy + rename the parent noxfile-template.py as noxfile.py in the folder of the project (containing the requirements.txt for the file).

>cd python-docs-samples
>cp noxfile-template.py PATH/TO/YOUR/PROJECT/noxfile.py
>cd PATH/TO/YOUR/PROJECT/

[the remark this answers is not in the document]
frankyn: Thanks for clarifying, this is new!

frankyn (approved):
LGTM, Pending Kokoro

tmatsuo:
The test failed with the following error:

```
googleapiclient.errors.HttpError: <HttpError 400 when requesting https://storagetransfer.googleapis.com/v1/transferJobs?alt=json returned "Failed to obtain the location of the Google Cloud Storage (GCS) bucket python-docs-samples-tests-storagetransfer-source due to insufficient permissions.  Please verify that the necessary permissions have been granted.". Details: "Failed to obtain the location of the Google Cloud Storage (GCS) bucket python-docs-samples-tests-storagetransfer-source due to insufficient permissions.  Please verify that the necessary permissions have been granted.">
```

JesseLovelace:
Turns out the STS samples never had any tests, and their testing environment was never set up for this account. I'm working on setting it up now.

leahecole:
@JesseLovelace if you need any help wrt testing environment stuff, let me (or any of the owners) know! There's an office hours from 10-12 pacific on 3/12, or, if you ping any of us we can always set up time outside of that, especially if pacific time is not great for you.

Regarding the exchange about the noxfile - this repo is somewhat unusual. You actually don't need to add a `noxfile` unless yours is different than the `noxfile-template.py`, and if that's the case, you add a `noxfile_config` (here's an [example in the cloud run directory where they enforce type hints and also change the versions they test](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/run/django/noxfile_config.py)) . In kokoro, that template is copied to your directory behind the scenes. I will be the first to admit it's not the greatest DevX for DREs, but it means that we have a single source of truth which is less painful for long term maintenance.  

if you think that the phrasing/explanation could be improved in the Authoring guide please lmk or feel free to make a PR for it (you too, @frankyn !)

frankyn (approved):

storage/transfer_service/sts_snippets_test.py
tmatsuo: @JesseLovelace Do you want to address this comment?

tmatsuo (approved):
LGTM. Can you also address the comment from @frankyn ?

storage/transfer_service/sts_snippets_test.py
JesseLovelace: Frank and I talked about this offline, it's fine how it is, avoids a flaky test

PR #5287, opened 2021-01-26
chore(functions/OCR): fix using reserved var name

ace-n:
`GCP_PROJECT` -> `GCP_PROJECT_ID`

~**DO NOT MERGE** until the docs have been updated accordingly. (cl/354393805)~


engelke:
I'm not clear on the reason for this change. Both the test environment and the instructions at https://cloud.google.com/functions/docs/tutorials/ocr#functions-prepare-environment-python set the GCP_PROJECT environment variable, not GCP_PROJECT_ID.

grant:
There's also many uses in functions/ for GCP_PROJECT. I'd expect any PR that changes that to do so throughout this repo.

grant (changes requested):
Requesting more of an explanation and answer to above questions.

ace-n:
@engelke we're planning on updating the docs accordingly.

@grant the reasoning here is that `GCP_PROJECT` is a reserved variable for some older runtimes (e.g. Python 3.7, see b/157737766). Right now, the OCR tutorial doc tells them to specify their project ID in their `gcloud functions deploy` command. Naturally, that will fail if they use `GCP_PROJECT` because it's a reserved variable (on certain older runtimes).

IMO the cleanest solution is to remove the ambiguity entirely and use an environment variable name that is **not** reserved by _any_ runtime.

RE changing the other mentions: I was hoping this would be a one-off PR, but good catch nonetheless - so I'm happy to make this change if you're on board.

grant:
> change `functions`-wide if you're on board.

Yes, that is cleaner and what I expected.

ace-n:
Closing in favor of adding a note to the docs.

**Googlers:** see b/157737766 for more info.

PR #5292, opened 2021-01-26
fix: use pytds driver in SQL Server sample

shubha-rajan:
## Description

Update sample to use pytds, a pure-python driver instead of pyodbc, so that it can work on Cloud Functions

## Checklist


kurtisvg (approved):

cloud-sql/sql-server/sqlalchemy/requirements.txt
```diff
@@ -1,4 +1,5 @@
 Flask==1.1.2
 SQLAlchemy==1.3.22
-pyodbc==4.0.30
+python-tds==1.10.0
```
engelke, line 3: connection_test is still using pyodbc.

engelke (approved):
LGTM once tests pass.

shubha-rajan:
The tests are passing locally while running the proxy, but not on Kokoro. Getting this error.
```
sqlalchemy.exc.DBAPIError: (pytds.tds_base.Error) Client does not have encryption enabled but it is required by server, enable encryption and try connecting again
(Background on this error at: http://sqlalche.me/e/13/dbapi)
```
Any ideas why this error would show up here but not locally?

kurtisvg:
> The tests are passing locally while running the proxy, but not on Kokoro. Getting this error.
> 
> ```
> sqlalchemy.exc.DBAPIError: (pytds.tds_base.Error) Client does not have encryption enabled but it is required by server, enable encryption and try connecting again
> (Background on this error at: http://sqlalche.me/e/13/dbapi)
> ```
> 
> I'm in the process of debugging it. Any ideas why this error would show up here but not locally?

Are we using the same instance? Is it possible there is a "required SSL" bit that was flipped for overground?

shubha-rajan:
> Are we using the same instance? Is it possible there is a "required SSL" bit that was flipped for overground?
Different instance, but both are under the Google org so overground rules should be the same. Running the application works locally too, so I don't think the tests passing is a fluke. 

Also, shouldn't using the proxy mean the connection to the server is SSL-encrypted? Is the issue that it expects encryption while connecting to the proxy server on the client side too?

PR #5305, opened 2021-01-29
fix(run): Clarify the source of GOOGLE_CLOUD_PROJECT var

grayside:
Responding to developer feedback that it is confusing to come from the docs to the main.py and see `GOOGLE_CLOUD_PROJECT` environment variable. This adds the missing context. For more discussion of this feedback, see b/178792002 (internal).

This applies to both Cloud Run and Cloud Functions, let's see what the bot thinks about my tagging both on the PR.

### DIFF
--- a/run/logging-manual/main.py
+++ b/run/logging-manual/main.py
@@ -23,6 +23,9 @@ app = Flask(__name__)
 
 @app.route("/", methods=["GET"])
 def index():
+    # This is set as a custom environment variable on deployment.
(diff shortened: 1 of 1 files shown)


engelke (approved):
LGTM

leahecole (approved):

averikitsch (approved):
This note works but the Java and Node.js samples include the call to the metadata server even if not shown in the sample.

PR #5315, opened 2021-01-31
feat: python version of idp-sql tutorial

glasnt:
## Checklist
- [x] These samples need a new **API enabled** in testing projects to pass (let us know which ones)
- [x] These samples need a new/updated **env vars** in testing projects set to pass (let us know which ones)


run/idp-sql/main.py
```diff
@@ -0,0 +1,304 @@
...
+# is safe to initialize your database connection pool when your script starts
+# -- there is no need to wait for the first request.
+db = None
+
+
+@app.before_first_request
+def create_tables() -> None:
+    global db
+    db = init_connection_engine()
+    # Create tables (if they don't already exist)s
```
averikitsch, line 169: Can we update the votes table to pet_votes. From the node version, we found that there can be an issue between this sample and cloud sql sharing the same table name but not schema. 
```suggestion
    # Create pet_votes table if it doesn't already exist
```

averikitsch (commented):
A few thoughts:
1) Is there an idiomatic way to split main.py into separate files to reduce complexity?
2) A priority is to add a sig term handler to show of that functionality (depending on this postgres library, the db connections can be closed when sigterm is caught)
3) A lesser priority would be to implement better logging. In order for stackdriver to parse the logs correctly they need to be json formatted with a severity property to set for the log level to be parsed.

run/idp-sql/database.py
```diff
@@ -0,0 +1,197 @@
...
+
+    return {
+        "dogs_count": dogs_count,
+        "recent_votes": votes,
+        "cats_count": cats_count,
+    }
+
+
+def save_vote(team: str, uid: str, time_cast: datetime.datetime) -> None:
+
```
averikitsch, line 177: nit: newline

run/idp-sql/Dockerfile
```diff
@@ -6,6 +6,9 @@
 # https://hub.docker.com/_/python
 FROM python:3.9
 
+# Send stdout/stderr out, do not buffer.
+ENV PYTHONUNBUFFERED 1
```
averikitsch, line 10: I'm wondering if this is the best method for flushing logs

run/idp-sql/Dockerfile
```diff
@@ -23,4 +26,5 @@ COPY . ./
 # webserver, with one worker process and 8 threads.
 # For environments with multiple CPU cores, increase the number of workers
 # to be equal to the cores available.
-CMD exec gunicorn --bind :$PORT --workers 1 --threads 8 main:app
\ No newline at end of file
+# Using ENTRYPOINT to ensure SIGTERM is correctly handled.
+ENTRYPOINT ["gunicorn", "--bind", ":8080", "--workers", "1", "--threads", "8", "main:app"]
```
averikitsch, line 30: `CMD exec gunicorn --bind :$PORT --workers 1 --threads 8 --timeout 0 main:app` should work here. We prefer using CMD over entrypoint because it's easily over-writable. Adding the `exec` should ensure SIGTERM is handled.

run/idp-sql/middleware.py
```diff
@@ -0,0 +1,84 @@
...
+
+    return decorated_function
+
+
+# [END cloudrun_python_user_auth_jwt]
+
+# adapted from https://github.com/ymotongpoo/cloud-logging-configurations/blob/master/python/structlog/main.py
+
+
+def field_name_modifier(
```
averikitsch, line 54: Can you add a comment that this is used for formatting the logger

run/idp-sql/requirements.txt
```diff
@@ -3,4 +3,6 @@ SQLAlchemy==1.3.22
 pg8000==1.16.5 # Do not upgrade >=1.6.6 until https://github.com/tlocke/pg8000/commit/3a2e7439ae3613367ec231218d7e0f541466d1e5#commitcomment-43174891 is resolved
 gunicorn==20.0.4
 firebase-admin==4.5.1
-google-cloud-secret-manager==2.1.0
\ No newline at end of file
+google-cloud-secret-manager==2.1.0
+#python-json-logger==2.0.1
```
averikitsch, line 7: Is this needed?

run/idp-sql/main.py
```diff
@@ -0,0 +1,106 @@
...
+    )
+
+
+# https://cloud.google.com/blog/topics/developers-practitioners/graceful-shutdowns-cloud-run-deep-dive
+def shutdown_handler(signal: int, frame: FrameType) -> None:
+    logger.info("Signal received, safely shutting down.")
+    database.shutdown()
+    middleware.logging_flush()
+    print("Exiting process.", flush=True)
+    sys.exit(0)
```
averikitsch, line 97: I've heard that this might not be the right way to exit, is there a way to shutdown the flask server?

run/idp-sql/Dockerfile
glasnt: "Force the stdout and stderr streams to be unbuffered." https://docs.python.org/3/using/cmdline.html#envvar-PYTHONUNBUFFERED

You can also use `import sys; sys.stdout.flush()` to manually force a flush. 

For this example, should I be including multiple settings to ensure logs are flushed, connections are closed, etc, even though they may not have any effect?

run/idp-sql/main.py
glasnt: In this case we are using gunicorn to run flask. Gunicorn handles signals to configure graceful timeouts on TERM ([gunicorn docs](https://docs.gunicorn.org/en/stable/signals.html)) However, we default our gunicorn invocations with `--timeout 0` ([gunicorn docs](https://docs.gunicorn.org/en/stable/settings.html?highlight=timeout#timeout)) ([stack overflow](https://stackoverflow.com/questions/60537977/critical-worker-timeout-in-logs-when-running-hello-cloud-run-with-python-f)). There is docs for flask regarding `werkzeug.server.shutdown`, but I believe that is only useful when having request-based shutdown, and I don't believe with gunicorn that werkzerg would be running anyway.

run/idp-sql/Dockerfile
averikitsch: > For this example, should I be including multiple settings to ensure logs are flushed, connections are closed, etc, even though they may not have any effect?

No, just the main way. A lot of my questions are figuring out what is idiomatic to Python versus what I've seen elsewhere. I assume `PYTHONUNBUFFERED` is the more idiomatic/common way?

run/idp-sql/main.py
averikitsch: sgtm. Thanks for explaining.

run/idp-sql/Dockerfile
glasnt: I believe it is, as it's the method we've used to ensure timely logs getting received by Cloud Logging when working with Cloud Functions.

averikitsch (approved):

tmatsuo (approved):
It's quite a masterpiece.

averikitsch:
@bradmiro @leahecole @grant codeowners please review

grant (approved):

glasnt:
The `retry.sh` script was updated to ensure that failed parts of a step fails the entire step. 

The db connections aren't being cleaned up properly, according to the error: 

```
ERROR: (gcloud.sql.databases.delete) HTTPError 400: Invalid request: failed to delete 
database "idp-sql-database-1c7f6db56c". Detail: pq: database 
"idp-sql-database-1c7f6db56c" is being accessed by other users. (Please use psql 
client to delete database that is not owned by "cloudsqlsuperuser")
```

Adding `close_all_sessions()` didn't work.

The settings from `init_connection_engine` are taken from [another sample](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/3ca7807ec1f97cd693dbe1ee4c00461e74289f4f/cloud-sql/postgres/sqlalchemy/main.py#L28), but I'm not sure if these settings are causing issues (I'm not an sqlalchemy developer). The same sort of cleanup operation (cleaning up databases) doesn't fail in the django version because I think database connections are managed by django itself, so it's not erroring. 

I'd appreciate any assistance.

averikitsch:
The service should already be deleted by the time that the  database is being deleted, so I am not sure what could still be connected. 
@kvg @shubha-rajan can you advise?

> The `retry.sh` script was updated to ensure that failed parts of a step fails the entire step.
> 
> The db connections aren't being cleaned up properly, according to the error:
> 
> ```
> ERROR: (gcloud.sql.databases.delete) HTTPError 400: Invalid request: failed to delete 
> database "idp-sql-database-1c7f6db56c". Detail: pq: database 
> "idp-sql-database-1c7f6db56c" is being accessed by other users. (Please use psql 
> client to delete database that is not owned by "cloudsqlsuperuser")
> ```
> 
> Adding `close_all_sessions()` didn't work.
> 
> The settings from `init_connection_engine` are taken from [another sample](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/3ca7807ec1f97cd693dbe1ee4c00461e74289f4f/cloud-sql/postgres/sqlalchemy/main.py#L28), but I'm not sure if these settings are causing issues (I'm not an sqlalchemy developer). The same sort of cleanup operation (cleaning up databases) doesn't fail in the django version because I think database connections are managed by django itself, so it's not erroring.
> 
> I'd appreciate any assistance.

run/idp-sql/database.py
```diff
@@ -0,0 +1,201 @@
...
+
+def shutdown() -> None:
+    # Each connection was released on execution, so just formally
+    # dispose of the db connection if it's been instantiated
+    if db:
+        db.dispose()
+        logger.info("Database connection disposed.")
+    # Additionally, close all sessions.
+    close_all_sessions()
```
kurtisvg, line 200: I'd suggest flipping these - [from the SQLAlchemy docs](https://docs.sqlalchemy.org/en/13/core/connections.html?highlight=dispose#sqlalchemy.engine.Engine.dispose):

> This has the effect of fully closing all currently checked in database connections. Connections that are still checked out will not be closed, however they will no longer be associated with this Engine, so when they are closed individually, eventually the Pool which they are associated with will be garbage collected and they will be closed out fully, if not already closed on checkin.

run/idp-sql/test/e2e_test_setup.yaml
```diff
@@ -0,0 +1,87 @@
...
+
+steps:
+
+- id: 'Create a dedicated database'
+  name: 'gcr.io/cloud-builders/gcloud'
+  entrypoint: '/bin/bash'
+  args:
+    - '-c'
+    - |
+      ./test/retry.sh "gcloud sql databases create ${_DB_NAME} \
```
kurtisvg, line 23: So ideally, we really should be reusing infrastructure as opposed to recreating / tearing down each time (see [External Resources](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/AUTHORING_GUIDE.md#external-resources)).

run/idp-sql/test/e2e_test_cleanup.yaml
```diff
@@ -0,0 +1,41 @@
...
+        "gcloud run services delete ${_SERVICE} --region ${_REGION} --platform ${_PLATFORM} --quiet"
+
+      ./test/retry.sh "gcloud secrets describe ${_SERVICE}-secrets" \
+        "gcloud secrets delete ${_SERVICE}-secrets --quiet --project $PROJECT_ID"
+
+      ./test/retry.sh "gcloud container images describe gcr.io/${PROJECT_ID}/${_SERVICE}:${_VERSION}" \
+        "gcloud container images delete gcr.io/${PROJECT_ID}/${_SERVICE}:${_VERSION} --quiet"
+
+      ./test/retry.sh "gcloud sql databases describe ${_DB_NAME} --instance ${_DB_INSTANCE} --project $PROJECT_ID" \
+        "gcloud sql databases delete ${_DB_NAME} --instance ${_DB_INSTANCE} --quiet --project $PROJECT_ID"
```
kurtisvg, line 33: You may want to consider using fixtures to create / teardown the temporary resources. There is a good example of this in the [run/logging-manual](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/run/logging-manual/e2e_test.py) sample.

run/idp-sql/test/e2e_test_setup.yaml
kurtisvg: Realized now this is this database (and not the instance) but it looks like in the [Cloud SQL samples we reuse the database and just create/drop tables](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/d384d10314f89dba48b68f2874da7e1861a5f9f7/cloud-sql/mysql/sqlalchemy/connection_test.py#L81)

run/idp-sql/database.py
kurtisvg: To clarify: 
```
  # Additionally, close all sessions to return connections to the pool
  close_all_sessions()
  # Close the database to clean up any connections
 if db:
        db.dispose()
        logger.info("Database connection disposed.")
```

run/idp-sql/Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+# Copy local code to the container image.
+ENV APP_HOME /app
+WORKDIR $APP_HOME
+COPY . ./
+
+# Run the web service on container startup. Here we use the gunicorn
+# webserver, with one worker process and 8 threads.
+# For environments with multiple CPU cores, increase the number of workers
+# to be equal to the cores available.
+CMD exec gunicorn --bind :$PORT --workers 1 --threads 8 --timeout 0 main:app
```
kurtisvg, line 29: Not sure if `--timeout 0` is hindering things either -- looks like that disables work timeouts?

glasnt:
Current version passes tests, but removes `--timeout 0` from gunicorn and uses NullPool. I'm not certain if these are settings that are suitable for this type of application, if the interface is usable, or if there's other considerations; just because the tests pass now doesn't mean these are good configuration changes for this sample.

run/idp-sql/test/e2e_test_setup.yaml
glasnt: See comment on current state of PR https://github.com/GoogleCloudPlatform/python-docs-samples/pull/5315#issuecomment-783866308

Using this method may make the tests pass, but it will also mean there's potential namespace collision, etc, because in this example the database is created in the client code. 

The nodejs version of this doesn't create/delete databases either https://github.com/GoogleCloudPlatform/nodejs-docs-samples/blob/master/run/idp-sql/test/e2e_test_cleanup.yaml

glasnt:
As discussed with @averikitsch, this now uses NullPool when testing only to work around service delete issues; will re-address when upstream changes happen around those. I'm comfortable with this PR being merged as it is.

(I have the ability to hit the [squash and merge] button with the approvals received, but I'm not sure if I should be merging my own PRs)

averikitsch:
> As discussed with @averikitsch, this now uses NullPool when testing only to work around service delete issues; will re-address when upstream changes happen around those. I'm comfortable with this PR being merged as it is.
> 
> (I have the ability to hit the [squash and merge] button with the approvals received, but I'm not sure if I should be merging my own PRs)

LGTM. Feel free to merge

Issue #5307, opened 2021-01-29
Flaky test in run/django


tmatsuo:
[an example build](https://source.cloud.google.com/results/invocations/7cc562e3-5d16-4c55-9cc4-6838d5375c11/targets)

tmatsuo:
It fails in teardown. I think the relevant message is `ERROR: (gcloud.sql.databases.delete) HTTPError 409: Operation failed because another operation was already in progress.`

tmatsuo:
Here is my guess what's happening.

1. `gcloud sql databases delete` command send an underlining api request
2. the request errored (probably timeout), but the server got the request and processing
3. `gcloud` automatically retry the request
4. the server returns 409 because it is processing the previous request
5. `gcloud` reports the error

I think it's worth reporting to upstream. 

FYI @dinagraves

leahecole:
Added @glasnt also who is the original author of this sample

glasnt:
I think this sample would benefit from moving to the "setup/cleanup/retry.sh" method of setting up and taking down testing infra ([borrowed from node.js](https://github.com/GoogleCloudPlatform/nodejs-docs-samples/tree/master/run/idp-sql/test), being used in an [upcoming sample](https://github.com/GoogleCloudPlatform/python-docs-samples/pull/5315
)

glasnt:
#5340 is now merged, should fix this issue

leahecole:
I am seeing that [some storage buckets aren't being deleted](https://screenshot.googleplex.com/4opkmbmqQrWfAsF) - could that be an oversight in this refactor?

glasnt:
Good catch, @leahecole; I'll clean that up and submit a PR fix.

PR #5340, opened 2021-02-07
fix(django): retry-based e2e testing

glasnt:
## Description

Addresses #5307 #5252 

Uses retry for setup/cleanup of infra, should help with flaky.

Some infra commands outside of retry.sh because permissions (would require cloud sql admin)

## Checklist


run/django/test/retry.sh
```diff
@@ -0,0 +1,72 @@
...
+
+# Usage: try "cmd1" "cmd2"
+# If first cmd executes successfully then execute second cmd
+runIfSuccessful() {
+  echo "running: $1"
+  $($1 > /dev/null)
+  if [ $? -eq 0 ]; then
+    echo "running: $2"
+    $($2 > /dev/null)
+  else return 1
```
glasnt, line 44: @averikitsch Updated line from nodejs-docs-samples version: If running in a two-arg state, if the first command failed, this failure wouldn't be captured by the until loop

run/django/test/retry.sh
```diff
@@ -0,0 +1,72 @@
...
+else
+  cmd="runIfSuccessful \"$arg1\" \"$arg2\""
+fi
+
+until eval $cmd
+do
+    if ((attempt_num==max_attempts))
+    then
+        echo "Attempt $attempt_num / $max_attempts failed! No more retries left!"
+        exit 1
```
glasnt, line 67: @averikitsch Update from the nodejs-docs-sample version (opinionated): should the retry script fail entirely, it should return an error state, meaning the cloud build step should fail, thus the job fails, and the test should fail. 
This should allow for setup failures to fail early, rather than steps failing later due to the earlier failure

run/django/test/retry.sh
averikitsch: My original assumption here is if the first command fails, we don't want to retry it since it's most likely a "describe" command. I would assume that "describe" commands wouldn't fail or be flaky like delete or deploy commands. The goals is aimed at capturing the failure of the second command.

run/django/test/retry.sh
averikitsch: I'm not sure I follow. The `exit 1` should cause the script and then the build step to fail.

run/django/test/retry.sh
glasnt: I was using the script in an "create/describe" format, and without this change 409 errors were falling through. Adding this change would extend it's usage. The script comments describe usage for "action" or "exists/action", so I was using it incorrectly. I will address.

run/django/test/retry.sh
glasnt: Even with this change, the step will only fail if the last exit status is 1, so in theory if you use multiple retry.sh in one step, there is a chance it can fail and continue, depending on the order and success of commands.

averikitsch: Good catch. Would this have to be handled in the cloud build to exit sooner?

glasnt: In the example I borrowed from, the cleanup was all done in one step. If and only if the last command failed in the step would the step fail.

averikitsch (approved):

busunkim96 (approved):

run/django/test/retry.sh
glasnt: FYI I backed out these changes and updated my use of the script to correct the issues I was experiencing.

PR #5389, opened 2021-02-19
add holidays dataset and readme about it

leahecole:
The Kaggle dataset we originally used for this tutorial is no longer available, so we now are including it with the code for users to download.

### DIFF
--- a/data-science-onramp/data-ingestion/README.md
+++ b/data-science-onramp/data-ingestion/README.md
@@ -1,0 +1,7 @@
+# Data Ingestion
+This is part of a larger tutorial found on cloud.google.com
+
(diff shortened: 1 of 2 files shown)


data-science-onramp/data-ingestion/README.md
```diff
@@ -0,0 +1,7 @@
+# Data Ingestion
+This is part of a larger tutorial found on cloud.google.com
```
bradmiro, line 2: Super annoying and tiny nit to add a period to the end of this but otherwise lgtm

bradmiro (approved):

busunkim96 (approved):

PR #5536, opened 2021-03-17
docs: added type hints and comments to Datastore samples

craiglabenz:
Improved clarity around Datastore samples for beginners


datastore/cloud-client/tasks.py
```diff
@@ -25,47 +26,60 @@ def create_client(project_id):
 # [END datastore_build_service]
 
 
+import google.cloud.datastore  # noqa
```
craiglabenz, line 29: Adding this import to allow for full type-hints

craiglabenz:
This PR is down to a linting warning that is purely an artifact of how we're dividing up code so they appear nicely in rendered snippets on devsite. I think it is good to go, otherwise.

datastore/cloud-client/snippets.py
```diff
@@ -110,10 +229,13 @@ def properties(client):
...
 
 def array_value(client):
-    key = client.key("Task")
     # [START datastore_array_value]
+    key = client.key("Task")
     task = datastore.Entity(key)
     task.update({"tags": ["fun", "programming"], "collaborators": ["alice", "bob"]})
+
+    # To access array values:
+    task["tags"]  # ["fun", "programming"]
```
crwilcox, line 238: this ticks out as a bit strange. Do we have a read array sample? if not including here may be fine. What do other langs do? Should they also be updated?

craiglabenz: Resolving this with additional tickets, per our conversation.

datastore/cloud-client/snippets.py
```diff
@@ -16,8 +16,128 @@
 import datetime
 from pprint import pprint
 
-from google.cloud import datastore
 import google.cloud.exceptions
+from google.cloud import datastore
```
engelke, line 20: You could add "# noqa: I100" to the end of this line (and perhaps the line before it) to suppress the lint message and declare that the order is intentional, not an error.

datastore/cloud-client/snippets.py
```diff
@@ -66,6 +186,9 @@ def basic_entity(client):
             "description": "Learn Cloud Datastore",
         }
     )
+    # Access fields
+    task["category"]  # "Personal"
+    task["priority"]  # 4
```
engelke, line 191: I find this confusing. I get showing how to access a field, but statements that are just a value aren't something I'd expect. You might want to consider doing something with these values, such as logging them, to show not only how to reference them but also how to use the result.

engelke (approved):
LGTM, though I recommend suppressing the import order lint issue with # noqa.

crwilcox (approved):

PR #5547, opened 2021-03-19
Pub/Sub: enable parallel writes to GCS in Pub/Sub Dataflow example

### LINKED ISSUE #5441: Limited Parallelism due to fixed key within GroupWindowsIntoBatches

https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/pubsub/streaming-analytics/PubSubToGCS.py

No changes. 

## Describe the issue

GroupWindowsIntoBatches PTransform uses a fixed "None" key, which limits parallelism on the pipeline to 1. This means under high load, especially with I/O to an external system, we are unable to scale out horizontally. Having a fixed key also means unless range assignment changes, all data across windows is sent to a single worker. 

Perhaps having a rotating set of keys, for each invocation of the PTransform would provide the single key/batch functionality with some additional parallelism - aka parallelizing batch writes.

anguillanneuf:
Fixes #5441 

Used this opportunity to update comments and clean up this example too.

Tried [`Splittable DoFn`](https://beam.apache.org/documentation/programming-guide/#splittable-dofns) but the documentation on parallel writes is very slim, and the Python support lags behind Java.


pubsub/streaming-analytics/PubSubToGCS.py
```diff
@@ -14,119 +14,126 @@
...
-    time and outputs a list of dictionaries, where each contains one message
-    and its publish timestamp.
+class GroupMessagesByFixedWindows(PTransform):
+    """A composite transform that groups Pub/Sub messages based on publish time
+    and outputs a list of tuples, each containing a message and its publish time.
     """
 
-    def __init__(self, window_size):
-        # Convert minutes into seconds.
+    def __init__(self, window_size, num_shards):
```
davidcavazos, line 31: Since `num_shards` is more of an optimization parameter, how about we have a default value even if we don't use it? So if anyone wants to copy-paste this class, it would be a little easier to use. Elasticsearch uses 5 as a default, maybe we can use that as well or whatever you prefer. I would just avoid having 1 as a default so that we "optimize by default" and to showcase this as a feature.

pubsub/streaming-analytics/PubSubToGCS.py
```diff
@@ -14,119 +14,126 @@
...
-            # for this. If the windowed elements do not fit into memory,
-            # please consider using `beam.util.BatchElements`.
-            # https://beam.apache.org/releases/pydoc/current/apache_beam.transforms.util.html#apache_beam.transforms.util.BatchElements
-            | "Add Dummy Key" >> beam.Map(lambda elem: (None, elem))
-            | "Groupby" >> beam.GroupByKey()
-            | "Abandon Dummy Key" >> beam.MapTuple(lambda _, val: val)
+            # Bind window info to each element using element timestamp (or publish time).
+            | "Window into fixed intervals"
+            >> WindowInto(FixedWindows(self.window_size))
+            | "Add key to windowed elements" >> ParDo(AddKey(self.num_shards))
```
davidcavazos, line 42: I would actually add the timestamp in one transform, and add the random key in another transform. It's usually clearer to have each transform do only one simple function. Dataflow fuses all the element-wise transforms together into a single function anyways, so there are no performance concerns from this.

```
  | 'Add timestamps' >> beam.ParDo(AddTimestamps())
  | 'Random shard keys' >> beam.WithKeys(lambda _: random.randint(0, num_shards - 1))
```

It would also be a good idea to explain at some point _why_ we are doing sharding, what shards are, and how to choose a good number of shards. Maybe the README would be a good place for this?

pubsub/streaming-analytics/PubSubToGCS.py
```diff
@@ -14,119 +14,126 @@
...
-    def process(self, batch, window=beam.DoFn.WindowParam):
-        """Write one batch per file to a Google Cloud Storage bucket. """
+    def process(self, batch, window=DoFn.WindowParam):
+        """Write messages in a batch to Google Cloud Storage."""
 
         ts_format = "%H:%M"
         window_start = window.start.to_utc_datetime().strftime(ts_format)
         window_end = window.end.to_utc_datetime().strftime(ts_format)
-        filename = "-".join([self.output_path, window_start, window_end])
+        shard_id = str(batch[0])
```
davidcavazos, line 79: I would rename some things and do this:

```py
shard_id, batch = key_value
```

pubsub/streaming-analytics/PubSubToGCS.py
```diff
@@ -14,119 +14,126 @@
...
-                f.write("{}\n".format(json.dumps(element)).encode("utf-8"))
+        with io.gcsio.GcsIO().open(filename=filename, mode="w") as f:
+            for message in batch[1]:
+                f.write(f"{message[0]},{message[1]}\n".encode("utf-8"))
 
 
-def run(input_topic, output_path, window_size=1.0, pipeline_args=None):
-    # `save_main_session` is set to true because some DoFn's rely on
-    # globally imported modules.
+def run(input_topic, output_path, window_size=1.0, num_shards=1, pipeline_args=None):
```
davidcavazos, line 87: Lets match the `num_shards` default value to whatever you choose :)

pubsub/streaming-analytics/PubSubToGCS.py
```diff
@@ -14,119 +14,126 @@
...
+    with Pipeline(options=pipeline_options) as pipeline:
         (
             pipeline
-            | "Read PubSub Messages"
-            >> beam.io.ReadFromPubSub(topic=input_topic)
-            | "Window into" >> GroupWindowsIntoBatches(window_size)
-            | "Write to GCS" >> beam.ParDo(WriteBatchesToGCS(output_path))
+            # Because `timestamp_attribute` is unspecified in `ReadFromPubSub`, Beam
+            # binds the publish time returned by the Pub/Sub server for each message
+            # to the element. Element timestamp is accessible via `DoFn.TimestampParam`.
```
davidcavazos, line 98: Lets link to the API reference:

https://beam.apache.org/releases/pydoc/current/apache_beam.io.gcp.pubsub.html#apache_beam.io.gcp.pubsub.ReadFromPubSub

pubsub/streaming-analytics/PubSubToGCS.py
```diff
@@ -14,119 +14,126 @@
...
     )
     parser.add_argument(
         "--output_path",
-        help="GCS Path of the output file including filename prefix.",
+        help="Path of the output GCS file including the prefix.",
+    )
+    parser.add_argument(
+        "--num_shards",
+        type=int,
+        default=1,
```
davidcavazos, line 127: Again, lets match this default value :)

pubsub/streaming-analytics/PubSubToGCS.py
```diff
@@ -14,119 +14,126 @@
...
+        help="Number of shards to use when writing windowed elements to GCS.",
     )
     known_args, pipeline_args = parser.parse_known_args()
 
     run(
         known_args.input_topic,
```
davidcavazos, line 133: Nit: these are starting to grow, it might be a good idea to start doing named arguments so that the order doesn't matter, otherwise it might be error-prone introducing more arguments.

```py
run(
  input_topic=known_args.input_topic,
  output_path=known_args.output_path,
  ...
)
```

pubsub/streaming-analytics/PubSubToGCS.py
```diff
@@ -14,119 +14,126 @@
...
             pipeline
-            | "Read PubSub Messages"
-            >> beam.io.ReadFromPubSub(topic=input_topic)
-            | "Window into" >> GroupWindowsIntoBatches(window_size)
-            | "Write to GCS" >> beam.ParDo(WriteBatchesToGCS(output_path))
+            # Because `timestamp_attribute` is unspecified in `ReadFromPubSub`, Beam
+            # binds the publish time returned by the Pub/Sub server for each message
+            # to the element. Element timestamp is accessible via `DoFn.TimestampParam`.
+            | "Read from Pub/Sub" >> io.ReadFromPubSub(topic=input_topic)
+            | "Window into" >> GroupMessagesByFixedWindows(window_size, num_shards)
```
davidcavazos, line 100: Totally optional, but it might be easier to read if you inlined the `GroupMessagesByFixedWindows` transform here. That way you can see the whole pipeline from beginning to end without having to scroll up and down to see what's happening.

I would recommend having a custom PTransform like this if you were using it in multiple places so you don't duplicate code. Or if you add unit tests for your transforms, but since we're already doing an e2e test, I would go for readability.

davidcavazos (commented):
It's looking great, thank you!

pubsub/streaming-analytics/README.md
```diff
@@ -142,25 +145,27 @@ The following example will run a streaming pipeline. It will read messages from
...
   --region=us-central1 \
-  --input_topic=projects/$PROJECT_NAME/topics/$TOPIC_NAME \
-  --output_path=gs://$BUCKET_NAME/samples/output \
+  --input_topic=projects/$PROJECT_ID/topics/$TOPIC_ID \
+  --output_path=gs://$BUCKET_ID/samples/output \
   --runner=DataflowRunner \
-  --window_size=2 \
-  --temp_location=gs://$BUCKET_NAME/temp
+  --window_size=1 \
+  --num_shards=2 \
```
davidcavazos, line 159: I don't have a strong opinion about this, but it might be a good idea to _not_ include `--num_shards` here since it slightly complicates how we run the sample and it's a technical word not everyone knows.

We could have like an info box saying that you might see more than one file per window, and that's because the elements in the windows might be processed in parallel, and that you can control how many files you want by specifying the number of shards, and that each shard corresponds to one file. This might be a good place to explain what shards are (like the number of files per window), why you would use them, what it defaults to, and how to choose a number (it would be a function of the number of messages per second, the window size, and the approximate number of messages you want per file).

pubsub/streaming-analytics/PubSubToGCS.py
anguillanneuf: Good point. Setting default to 5.

pubsub/streaming-analytics/PubSubToGCS.py
anguillanneuf: Moved this out of this DoFn and mentioned sharding in README.

pubsub/streaming-analytics/PubSubToGCS.py
anguillanneuf: I started doing this, then I realized that it would diverge from the Java sample and a screenshot we show on c.g.c. I will keep this in mind when doing the new Pub/Sub Lite to GCS quickstart.

pubsub/streaming-analytics/PubSubToGCS.py
anguillanneuf: Ack.

pubsub/streaming-analytics/README.md
anguillanneuf: Commented this one out in the README.

davidcavazos (approved):
Thanks for the explanation, I have some minor comments, but overall LGTM!

pubsub/streaming-analytics/README.md
```diff
@@ -134,33 +137,43 @@ The following instructions will help you prepare your development environment.
 
 * [PubSubToGCS.py](PubSubToGCS.py)
 
-The following example will run a streaming pipeline. It will read messages from a Pub/Sub topic, then window them into fixed-sized intervals, and write one file per window into a GCS location.
+The following example will run a streaming pipeline. The pipelien does the following:
+1. Reads messages from a Pub/Sub topic.
+1. Group messages by fixed windows and break the grouped messages into batches.
+  1. Adds window info to each element/message.
+  1. Adds timestamp to each element/message.
+  1. Adds a random shard ID as key to each windowed element. This step helps shard the elements in the same window into small batches, so that multiple workers can help write the batched elements to Cloud Storage later.
```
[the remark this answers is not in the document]
anguillanneuf: I like the suggestion! Applied.

pubsub/streaming-analytics/README.md
```diff
@@ -134,33 +137,43 @@ The following instructions will help you prepare your development environment.
...
+  1. Groups the windowed elements by key.
```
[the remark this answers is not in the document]
anguillanneuf: Done.

pubsub/streaming-analytics/PubSubToGCS.py
```diff
@@ -14,119 +14,122 @@
...
-        filename = "-".join([self.output_path, window_start, window_end])
+        shard_id, batch = key_value
+        filename = "-".join([self.output_path, window_start, window_end, str(shard_id)])
 
-        with beam.io.gcp.gcsio.GcsIO().open(filename=filename, mode="w") as f:
-            for element in batch:
-                f.write("{}\n".format(json.dumps(element)).encode("utf-8"))
+        with io.gcsio.GcsIO().open(filename=filename, mode="w") as f:
+            for message in batch:
+                f.write(f"{message[0]},{message[1]}\n".encode("utf-8"))
```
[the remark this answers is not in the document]
anguillanneuf: Done.

davidcavazos:
Looks great! Thanks for these improvements, LGTM!

hongalex (approved):

busunkim96 (approved):

PR #5555, opened 2021-03-22
dataflow: allow custom Python version in Dockerfile

davidcavazos:
## Description

One of the highest friction points in the current GPU sample is that it depends on the Python version that comes pre-installed with the `tensorflow/tensorflow:2.4.0-gpu` image. They currently have Python 3.6 which will be [deprecated by the end of this year](https://endoflife.date/python).

Since TensorFlow 2.4 now uses the newer CUDA 11 and cuDNN 8, there's less configuration needed, so we can now base directly off the `nvidia/cuda` images. They don't come with any Python pre-installed, so we can easily install whichever version we want with Miniconda, and then `pip install` tensorflow and the rest of the dependencies. Installing tensorflow 2.4 in the nvidia images now works out of the box without further configurations.

This way builds faster, is simpler, more flexible, allows any supported Python version, and results in smaller images.

R: @tvalentyn - can you please help us review this whenever you have a chance?

## Checklist

### DIFF
--- a/dataflow/gpu-workers/Dockerfile
+++ b/dataflow/gpu-workers/Dockerfile
@@ -39,4 +39,5 @@ RUN apt-get update \
     && pip install --no-cache-dir -r requirements.txt \
     && conda clean -y --all --force-pkgs-dirs
 
+# Set the entrypoint to Apache Beam SDK worker launcher.
 ENTRYPOINT [ "/opt/apache/beam/boot" ]
--- a/dataflow/gpu-workers/Dockerfile
+++ b/dataflow/gpu-workers/Dockerfile
@@ -12,22 +12,29 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM tensorflow/tensorflow:2.4.0-gpu
+FROM nvidia/cuda:11.0.3-cudnn8-runtime-ubuntu20.04
+
+ARG python_version=3.8
 
 WORKDIR /root
 
-# Installing the requirements here makes the workers start faster since they
-# don't need to install the requirements at runtime.
-#   ℹ️ Make sure your requirements.txt includes `apache-beam[gcp]`.
+COPY --from=apache/beam_python3.8_sdk:2.28.0 /opt/apache/beam /opt/apache/beam
 COPY requirements.txt .
 
-# Don't install `tensorflow` since the base image already comes with it.
-RUN egrep -v '^tensorflow([=<>]|$)' requirements.txt > /tmp/requirements.txt \
-    && pip install --no-cache-dir -U pip \
-    && pip install --no-cache-dir -r /tmp/requirements.txt
+ENV PATH=$PATH:/opt/conda/bin
 
-# Copy the Apache Beam required files from the Beam Python SDK image.
-COPY --from=apache/beam_python3.6_sdk:2.28.0 /opt/apache/beam /opt/apache/beam
+RUN apt-get update \
+    && apt-get install -y wget \
+    && rm -rf /var/lib/apt/lists/* \
+    # The nvidia image doesn't come with Python pre-installed.
+    # We use Miniconda to install the Python version of our choice.
+    && wget -q https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh \
+    && sh Miniconda3-latest-Linux-x86_64.sh -b -p /opt/conda \
+    && rm Miniconda3-latest-Linux-x86_64.sh \
+    # Create a new Python environment and install our requirements.
+    # We don't need to update $PATH since /usr/local is already in $PATH.
+    && conda create -y -p /usr/local python=$python_version pip \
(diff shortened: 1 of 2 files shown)


dataflow/gpu-workers/Dockerfile
```diff
@@ -12,22 +12,32 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM tensorflow/tensorflow:2.4.0-gpu
+# Make sure the CUDA and cuDNN versions are compatible with your TensorFlow version.
```
tvalentyn, line 15: I recommend you check a user journey for user who has old version of Tensorflow, say, 2.1.0, and see if they can figure the configuration given the instructions hear, and whether it works.
I noticed that we often focus in instructions/examples on 'latest' versions, but customers are often several releases back, and there are quirks that they need to do because they are on the old version. So for old versions, we either need to document quirks, or mention to customers: this approach won't work unless you use version x or higher.

dataflow/gpu-workers/Dockerfile
```diff
@@ -12,22 +12,32 @@
...
+    && wget -q https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh \
+    && sh Miniconda3-latest-Linux-x86_64.sh -b -p /opt/conda \
+    && rm Miniconda3-latest-Linux-x86_64.sh \
+    # Create a new Python environment and install our requirements.
+    # We don't need to update $PATH since /usr/local is already in $PATH.
+    && conda create -y -p /usr/local python=$python_version pip \
+    && pip install --no-cache-dir -r requirements.txt \
+    && conda clean -y --all --force-pkgs-dirs
 
 # Set the entrypoint to Apache Beam SDK worker launcher.
```
tvalentyn, line 42: I think we need an override for pip variable or a symlink for pip (unless it's Beam 2.30.0 and higher).

tvalentyn: Smth like:
```
# Beam SDK worker looks for pip at a different location by default. Can be omitted with 2.30.0 and higher versions of Beam.
RUN ln -s $(which pip) /usr/local/bin/pip
```

tvalentyn (approved):

dataflow/gpu-workers/Dockerfile
davidcavazos: I tried updating the TensorFlow version to 2.1.0, which would require CUDA 10.1 and cuDNN 7. I manually updated the base image to `nvcr.io/nvidia/cuda:10.1-cudnn7-runtime-ubuntu18.04`, and had to downgrade to Python 3.7 to install the older TensorFlow version. After that, it still worked out of the box.

I think it's straightforward enough to not have to include explicit instructions, the Dockerfile already has comments that the TensorFlow, CUDA and cuDNN versions have to match a compatible configuration, and have a link to the TensorFlow page with the compatible configurations. I added a link to the Nvidia container registry so they can find the image they need.

tvalentyn:
cc: @rezarokni

dataflow/gpu-workers/e2e_test.py
```diff
@@ -24,7 +25,7 @@
 SUFFIX = uuid.uuid4().hex[0:6]
 PROJECT = os.environ["GOOGLE_CLOUD_PROJECT"]
 BUCKET_NAME = f"dataflow-gpu-test-{SUFFIX}"
-IMAGE_NAME = f"gcr.io/{PROJECT}/dataflow/gpu-workers/test-{SUFFIX}:latest"
+IMAGE_TAG = f"dataflow/gpu-workers/test-{SUFFIX}:latest"
```
tvalentyn, line 28: hm, this actually sounds like image name to me.

dataflow/gpu-workers/Dockerfile
davidcavazos: Thanks, added

davidcavazos:
FYI: @rosetn -- once this merges, we'll have to update [this section](https://cloud.google.com/dataflow/docs/samples/satellite-images-gpus#preparing_your_working_environment) in the docs. We no longer _require_ Python 3.6, so we can go back to using the usual "just create a virtualenv" :)

I would also think we should update the [custom GPU containers](https://cloud.google.com/dataflow/docs/guides/using-gpus#configuring_your_container_image) page as well. I would go ahead and just recommend basing directly from `nvidia/cuda` and installing your Python version of choice like in the sample. I would probably get rid of the section that installs the GPU drivers manually in favor of using the prebuilt nvidia images. But all is not urgent at all and could be done after launch.

davidcavazos:
Reza mentioned this looks good to him via chat. So I think we have the necessary approvals.

@dinagraves and @engelke, if you're happy with the tests, this should be ready to merge :)

engelke (approved):
LGTM

engelke (approved):

leahecole (approved):

PR #5604, opened 2021-03-30
feat: add spritesheet samples and tests

irataxy:
## Description

Fixes #181782257

## Checklist
- [X] **Tests** pass:   `nox -s py-3.6` (see [Test Environment Setup](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/AUTHORING_GUIDE.md#test-environment-setup))
- [X] **Lint** pass:   `nox -s lint` (see [Test Environment Setup](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/AUTHORING_GUIDE.md#test-environment-setup))
- [X] Please **merge** this PR for me once it is approved.


media/transcoder/create_job_with_periodic_images_spritesheet.py
```diff
@@ -0,0 +1,127 @@
...
+
+import argparse
+
+from google.cloud.video import transcoder_v1beta1
+from google.cloud.video.transcoder_v1beta1.services.transcoder_service import (
+    TranscoderServiceClient,
+)
+from google.protobuf import duration_pb2 as duration
+
+# [START transcoder_create_job_with_periodic_images_spritesheet]
```
kurtisvg, line 31: Please include the imports in the region tag - [See "Region Tags" in the Samples Format Guide](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/AUTHORING_GUIDE.md#region-tags)

media/transcoder/create_job_with_periodic_images_spritesheet.py
```diff
@@ -0,0 +1,127 @@
...
+        elementary_streams=[
+            transcoder_v1beta1.types.ElementaryStream(
+                key="video-stream0",
+                video_stream=transcoder_v1beta1.types.VideoStream(
+                    codec="h264",
+                    height_pixels=360,
+                    width_pixels=640,
+                    bitrate_bps=550000,
+                    frame_rate=60,
+                ),
```
kurtisvg, line 61: Presumably, these are all specific to the URI for the video used for testing? Can you add some comments to indicate that the user should be updating these values to match the video submitted, and if there are any constraints on what they might be? 

There's also quite a bit of nesting here, which [we generally try to avoid having too much of](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/AUTHORING_GUIDE.md#arrange-act-assert). It would be good to pull maybe the top level components out so it context can be added?

media/transcoder/job_test.py
```diff
@@ -242,6 +257,96 @@ def test_create_job_with_animated_overlay(capsys, test_bucket):
...
+    str_slice = out.split("/")
+    job_id = str_slice[len(str_slice) - 1].rstrip("\n")
+    job_name = f"projects/{project_number}/locations/{location}/jobs/{job_id}"
+    assert job_name in out
+
+    get_job.get_job(project_id, location, job_id)
+    out, _ = capsys.readouterr()
+    assert job_name in out
+
+    time.sleep(30)
```
kurtisvg, line 280: Please add a comment as to why this sleep is necessary

media/transcoder/job_test.py
```diff
@@ -242,6 +257,96 @@ def test_create_job_with_animated_overlay(capsys, test_bucket):
...
+
+    _get_job_state(capsys, job_id)
```
kurtisvg, line 282: This name feels deceptive - it's not getting the state, it's waiting until it succeeded? Can you add a comment at least to clarify how it's being used here.

media/transcoder/job_test.py
```diff
@@ -252,3 +357,11 @@ def _get_job_state(capsys, job_id):
 
     out, _ = capsys.readouterr()
     assert job_succeeded_state in out
+
+
+def _get_file_in_bucket(capsys, directory_and_filename):
+    storage_client = storage.Client()
```
kurtisvg, line 363: We shouldn't be creating a new client like this - please just reuse the same client. You can create a fixture for it and scope to the module to avoid having to recreate it several times.

media/transcoder/job_test.py
```diff
@@ -252,3 +357,11 @@ def _get_job_state(capsys, job_id):
 
     out, _ = capsys.readouterr()
     assert job_succeeded_state in out
+
+
+def _get_file_in_bucket(capsys, directory_and_filename):
```
kurtisvg, line 362: This name is deceptive too - it should be more like `_assert_file_in_bucket`.

media/transcoder/job_test.py
```diff
@@ -242,6 +257,96 @@ def test_create_job_with_animated_overlay(capsys, test_bucket):
...
+    assert job_name_prefix in out
+
+    str_slice = out.split("/")
+    job_id = str_slice[len(str_slice) - 1].rstrip("\n")
+    job_name = f"projects/{project_number}/locations/{location}/jobs/{job_id}"
+    assert job_name in out
+
+    get_job.get_job(project_id, location, job_id)
+    out, _ = capsys.readouterr()
+    assert job_name in out
```
kurtisvg, line 278: Does asserting the job_name actually men it was successful? Should we be asserting some kind of "succeeded" message in the response instead?

kurtisvg (changes requested):

kurtisvg:
Also, PTAL at the lint check.

media/transcoder/create_job_with_periodic_images_spritesheet.py
irataxy: Done. I have a b/182777706 to fix the rest of the samples.

media/transcoder/job_test.py
irataxy: Refactored to use the test bucket

media/transcoder/job_test.py
irataxy: Changed name to reflect actual function

media/transcoder/job_test.py
irataxy: Added comment to explain this

media/transcoder/job_test.py
irataxy: Changed name to _assert_file_in_bucket

media/transcoder/create_job_with_periodic_images_spritesheet.py
irataxy: These attributes are for the output video. I added some comments to clarify, but these particular samples focus on the spritesheet functionality.

This nesting is somewhat unavoidable. The job config is lengthy. I added further comments to clarify the functions of each section, but not sure pulling out the top-level components will help and may just add unnecessary length.

kurtisvg (approved):

PR #5714, opened 2021-04-19
test: adds sample and test for firestore bundles

craiglabenz:


firestore/cloud-client/snippets.py
```diff
@@ -1,4 +1,5 @@
-# Copyright 2017, Google, Inc.
+# Copyright 2019 Google, LLC.
```
busunkim96, line 1: Please preserve the original copyright year

firestore/cloud-client/snippets.py
```diff
@@ -1091,3 +1092,31 @@ def list_document_subcollections():
             print(f'{doc.id} => {doc.to_dict()}')
     # [END firestore_data_get_sub_collections]
     # [END fs_list_document_subcollections]
+
+
+def _setup_bundle():
+    from google.cloud import firestore
+    db = firestore.Client()
+    db.collection("stories").document("news-item").set({"title": "Wow!"})
+
```
busunkim96, line 1101: If this setup is only required for the test, could it be moved to the test file? 

Alternatively if the sample is hard to use without this additional setup I'd prefer it's inside the region tags.

firestore/cloud-client/snippets.py
craiglabenz: Haha, I was focusing on the `Inc.` vs `LLC` and didn't even notice it changed to 2019 😆

firestore/cloud-client/snippets.py
craiglabenz: Thanks @busunkim96!

The sample's construction is supposed to closely follow [the other language's Bundle samples](https://firebase.google.com/docs/firestore/bundles#building_data_bundles_on_the_server) in existing documentation, and those samples operate on imaginary records we only care about actually existing for the tests. I'm happy to refactor this if we can think of an optimization, but I don't think the sample record's creation is worth including in the region tag.

craiglabenz: And to further clarify, I put this in its own function to resolve awkwardness around duplicate imports and linting.

busunkim96: Makes sense! Thank you for the explanation. :)

busunkim96 (approved):

engelke (approved):
LGTM

PR #5760, opened 2021-04-28
Add retry sample to STS

JesseLovelace:
The example [retry code ](https://cloud.google.com/storage-transfer/docs/create-client#python_1)on the STS docs page is both incorrect and in-line, this adds a proper sample that simply executes a request using retries to replace it


frankyn (approved):
Nice catch Jesse! This LGTM

storage/transfer_service/get_transfer_job_with_retries.py
```diff
@@ -0,0 +1,39 @@
...
+
+    transferJob = (
+        storagetransfer.transferJobs()
+        .get(projectId=project_id, jobName=job_name)
+        .execute(num_retries=retries)
+    )
+    print(
+        "Fetched transfer job: "
+        + transferJob.get("name")
+        + " using {} retries".format(retries)
```
leahecole, line 37: nit - can you switch this to f-strings?

storage/transfer_service/sts_snippets_test.py
```diff
@@ -50,3 +51,39 @@ def test_latest_transfer_operation(capsys):
...
+
+    retries = 3
+
+    get_transfer_job_with_retries.get_transfer_job_with_retries(
+        project_id, job_name, retries
+    )
+    out, _ = capsys.readouterr()
+    # This sample isn't really meant to do anything, just check that it ran without any issues
+    # when we populated num_retries
+    assert str(retries) in out
```
leahecole, line 89:
```suggestion
    assert f"using {str(retries)} retries" in out
```

leahecole (changes requested):
Just some silly little nits and an update to the retry just to make sure it doesn't pass on another hypothetical random 3

storage/transfer_service/get_transfer_job_with_retries.py
JesseLovelace: I believe f strings aren't compatible with python 2.7 which we do have kokoro tests for, am I mistaken?

leahecole (approved):

PR #5837, opened 2021-05-14
chore: add snippet-bot's checks to the protection rule

tmatsuo:
These checks are from snippet-bot. For product prefix, some legacy samples might have a trouble. You can ignore the prefix requirements with `snippet-bot:no-prefix-req` label.


dandhlee (approved):
LGTM. Thank you! We'll just have to remember that this label is a thing.

sofisl (approved):

PR #5874, opened 2021-05-21
chore(deps): update Renovate to ignore patches

### LINKED ISSUE #5873: RenovateBot: disable patch updates
Currently RenovateBot has been pushing a lot of patch updates. This creates a huge burden on maintaining the dependencies when tests take more than 20 minutes to run and there's significantly high number of traffic.

The initial problem is that RenovateBot by default does not distinguish between patch and minor updates. Thus, since currently we have `major` and `minor` updates enabled: it enables RenovateBot to pick up on patches as well.

As desired we should fix this behavior and only update on major/minor releases.

dandhlee:
## Description

The problem with RenovateBot currently is that it actually treats `minor` and `patch` releases to be the same. See [separateMinorPatch](https://docs.renovatebot.com/configuration-options/#separateminorpatch) docs for more detail.

To mitigate this, we first include a rule to tell RenovateBot to separate and recognize between `minor` and `patch` releases. Then, we create a second rule to disable `patch` updates. This should now successfully let RenovateBot to only update on `major` and `minor` patches.

Fixes #5873 

## Checklist

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -17,6 +17,10 @@
       "packagePatterns": ["pytest"],
       "updateTypes": ["patch"],
       "enabled": false
+    },
+    {
+      "packagePatterns": ["pytest"],
+      "updateTypes": ["minor", "major"]
     }
   ]
 }
--- a/renovate.json
+++ b/renovate.json
@@ -11,7 +11,12 @@
   "packageRules": [
     {
       "packagePatterns": ["pytest"],
-      "updateTypes": ["minor", "major"]
+      "separateMinorPatch": true
+    },
+    {
+      "packagePatterns": ["pytest"],
+      "updateTypes": ["patch"],
+      "enabled": false
     }
   ]
 }


dandhlee:
Adding back the original snippet for specifying `major` and `minor` updates back for good faith :S

I'm not sure if specifying them will be required, but this should now be explicit to say

```
separate minor and patch releases
ignore patch releases
update on major and minor releases
```

rather than assuming that the third line (update on major and minor) is achieved by the second line.

busunkim96 (approved):

renovate.json
```diff
@@ -9,6 +9,15 @@
...
+      "separateMinorPatch": true
+    },
+    {
+      "packagePatterns": ["pytest"],
+      "updateTypes": ["patch"],
+      "enabled": false
+    },
     {
       "packagePatterns": ["pytest"],
       "updateTypes": ["minor", "major"]
```
leahecole, line 23: DO we need to also remove this section? Or do these rules work with the existing pattern?

leahecole: nvm, i see the comment

leahecole (approved):
YAY

dandhlee:
Hoping this works! I'll follow up with `boto3` since that one gets a daily update.

PR #5893, opened 2021-05-26
docs(firestore): clean-up region tagging

Strykrol:
1. Removes any dual-homed region tags where correct _firestore_ prefixed tags have proliferated in docs
2. Adds correct product prefix for any tags that aren't on devsite tenants (e.g. safe to update)
3. Dual-wrapped some tags to match similar other language snippets.

## Checklist
- [ x] I have followed [Sample Guidelines from AUTHORING_GUIDE.MD](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/AUTHORING_GUIDE.md)
- [x ] Please **merge** this PR for me once it is approved.


Strykrol:
Committed updates for 2/4 region tags that were in violation (on devsite). The remaining two found in Quickstart docs are false positives, awaiting updates from cl/376017137.

crwilcox (approved):

tmatsuo:
@Strykrol I think the checks on `quickstart_add_data_one` and `quickstart_add_data_two` are valid.

For example, this page:
https://cloud.google.com/firestore/docs/quickstart-servers#add_data
has a link to
https://github.com/GoogleCloudPlatform/python-docs-samples/blob/ca852daec6c5e3d31180ffbd6f1e3a4096e875f4/firestore/cloud-client/snippets.py#L38-L45

But I may be wrong. Why do you think it's a false positive?

The link has commit hash, so it might be ok, but the fact that the master branch doesn't have the same file is still confusing.

tmatsuo:
For a record, the html element has the following property:
```
data-region-tag="quickstart_add_data_one"
```
So I assume the devsite build will break if we remove this region tag now.

Strykrol:
@tmatsuo let's talk offline - I'm not seeing what you're seeing, but the source file for that page doesn't _appear to me_ to have these region tags anymore.

Strykrol:
Looks like the violations resolved after we manually re-published last week.

firestore/cloud-async-client/snippets.py
```diff
@@ -17,66 +17,57 @@
```
dandhlee, line 17: I think if you remove the comma after the year if should be fine, or `Copyright 2020 Google LLC` should do.

dandhlee (changes requested):
LGTM, minor license header issue. Please see comment below

firestore/cloud-async-client/snippets.py
Strykrol: Done

dandhlee:
One more minor lint issue then we should be good:
```
nox > flake8 --show-source --builtin=gettext --max-complexity=20 --import-order-style=google --exclude=.nox,.cache,env,lib,generated_pb2,*_pb2.py,*_pb2_grpc.py --ignore=E121,E123,E126,E203,E226,E24,E266,E501,E704,W503,W504,I202 --max-line-length=88 --application-import-names distributed_counters,noxfile,distributed_counters_test,.nox,snippets_test,snippets .
./snippets.py:106:1: E302 expected 2 blank lines, found 1
class City(object):
^
./snippets.py:166:1: E302 expected 2 blank lines, found 1
def add_example_data():
^
```

Strykrol:
Done

dandhlee (approved):

dandhlee:
Thank you! Merged as crwilcox@ is part of `cloud-native-db-dpes`

PR #6055, opened 2021-06-01
ignore composer constraints file

leahecole:
This file is pulled directly from Airflow so we shouldn't be changing it


busunkim96 (approved):

dandhlee (approved):
LGTM. Thank you!

PR #6096, opened 2021-06-07
dataflow: new minimal gpu samples

davidcavazos:
## Description

- Adds new minimal examples to use GPU with tensorflow and pytorch.
- Simplifies the landsat sample with newer patterns.
- Fixes some resource leaks due to uncaught errors in some fixture's teardown code.

## Checklist


kurtisvg:
@davidcavazos Would you mind pinging me for review once you have the tests passing?

dataflow/gpu-workers/pytorch-minimal/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Running the Dataflow job with GPUs
+
+We use Cloud Build to run the Dataflow job.
+We launch the job using the worker image to make sure the job launches
+with the same Python version as the workers.
+
+```sh
+export REGION="us-central1"
+export WORKER_ZONE="us-central1-f"
```
tvalentyn, line 30: We no longer need to  specify the zone, zone adviser will pick the best zone for the user.

dataflow/gpu-workers/pytorch-minimal/README.md
```diff
@@ -0,0 +1,45 @@
...
+export GPU_TYPE="nvidia-tesla-t4"
+
+gcloud beta builds submit \
+    --config run.yaml \
+    --substitutions _REGION=$REGION,_WORKER_ZONE=$WORKER_ZONE,_GPU_TYPE=$GPU_TYPE \
+    --no-source
+```
+
+> ℹ️ Make sure the GPU type you choose is available in the worker zone for the job.
```
tvalentyn, line 39: Accordingly, GPUs should be available in some zones in this region.

dataflow/gpu-workers/tensorflow-landsat/main.py
```diff
@@ -276,39 +276,42 @@ def run(
...
-    options = PipelineOptions(beam_args, save_main_session=True)
-    with beam.Pipeline(options=options) as pipeline:
-        # Optionally, validate that the workers are using GPUs.
-        gpu_check = (
-            pipeline
-            | beam.Create([None])
-            | "Check GPU availability" >> beam.Map(check_gpus, gpus_optional)
+    beam_options = PipelineOptions(beam_args, save_main_session=True)
+
+    # We currently cannot use the `with` statement to run without waiting.
```
tvalentyn, line 281: I think users may not understand this comment, let's remove it. Re `with` behavior, that may be WAI - for example when I access files in a `with` statement, resources are cleaned up after we exit from the context, so it makes sense to me that the `with` statement waits for pipeline completion.

dataflow/gpu-workers/tensorflow-landsat/run.yaml
```diff
@@ -0,0 +1,54 @@
...
+#   https://cloud.google.com/build/docs/build-config
+#
+# To learn more about Cloud Build variable substitutions:
+#   https://cloud.google.com/build/docs/configuring-builds/substitute-variable-values#using_user-defined_substitutions
+# -----------------------------------------------------------------------------
+
+substitutions:
+  _GCS_PATH: please set --substitutions _GCS_PATH=gs://my-bucket/samples/dataflow/landsat
+  _REGION: us-central1
+  _WORKER_ZONE: us-central1-f
```
tvalentyn, line 31: Same here (no need to hardcode a zone)

dataflow/gpu-workers/tensorflow-landsat/main.py
```diff
@@ -276,39 +276,42 @@ def run(
...
+    #   https://issues.apache.org/jira/browse/BEAM-12455
+    pipeline = beam.Pipeline(options=beam_options)
+
+    # Convert Landsat 8 scenes into images.
+    # ℹ️ We pass `gpu_check` as an unused side input to force that step in
+    # the pipeline to wait for the check before continuing.
```
tvalentyn, line 287: How abount:

We pass `gpu_check` as an unused side input to make a blocking one-time check
in the pipeline to verify GPUs are available.

dataflow/gpu-workers/pytorch-minimal/main.py
```diff
@@ -0,0 +1,61 @@
...
+def run(input_text: str, beam_args: Optional[List[str]] = None) -> None:
+    beam_options = PipelineOptions(beam_args, save_main_session=True)
+
+    # We currently cannot use the `with` statement to run without waiting.
+    #   https://issues.apache.org/jira/browse/BEAM-12455
+    pipeline = beam.Pipeline(options=beam_options)
+    (
+        pipeline
+        | "Create data" >> beam.Create([input_text])
+        | "Check GPU availability" >> beam.Map(check_gpus)
```
tvalentyn, line 44: Maybe add a comment: See landsat example to make this a one-time check instead of a per-element check.

tvalentyn: Or perhaps just make it a parallel check if you think it's not too complex - one of CE's suggested that would be helpful - leaving this up to you..

dataflow/gpu-workers/pytorch-minimal/main.py
```diff
@@ -0,0 +1,61 @@
...
+        | "My transform" >> beam.Map(logging.info)
```
tvalentyn, line 45: We could add a comment that users can sub this with a code that actually uses GPU here or perhaps make it a separate function + add a comment.

dataflow/gpu-workers/pytorch-minimal/README.md
davidcavazos: Awesome, that's great to know :)

dataflow/gpu-workers/pytorch-minimal/README.md
davidcavazos: Thanks, removed this note.

dataflow/gpu-workers/tensorflow-landsat/main.py
davidcavazos: Alright, I'm removing this comment for now.

dataflow/gpu-workers/tensorflow-landsat/run.yaml
davidcavazos: Removed

dataflow/gpu-workers/tensorflow-landsat/main.py
davidcavazos: Removed this and used a more compact pattern.

dataflow/gpu-workers/pytorch-minimal/main.py
davidcavazos: I used a more compact and copy-paste friendly pattern.

dataflow/gpu-workers/tensorflow-minimal/README.md
```diff
@@ -0,0 +1,41 @@
+# PyTorch GPU minimal pipeline
```
tvalentyn, line 1: Tensorflow

davidcavazos: Good catch, fixed!

dataflow/gpu-workers/pytorch-minimal/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+WORKDIR /pipeline
+
+# Copy the Apache Beam worker files and the pipeline source files.
+COPY --from=apache/beam_python3.8_sdk:2.29.0 /opt/apache/beam /opt/apache/beam
+COPY requirements.txt .
+COPY *.py ./
+
+# Install the pipeline requirements and check that there are no conflicts.
+RUN pip install --no-cache-dir --upgrade pip \
+    && pip install --no-cache-dir -r requirements.txt \
```
tvalentyn, line 26: We could make a comment somewhere that since users already install requirements.txt in the dockerfile, they should not pass --requirements_file when they launch the pipeline.

dataflow/gpu-workers/conftest.py
```diff
@@ -0,0 +1,503 @@
...
+    @staticmethod
+    def underscore_name(name: str) -> str:
+        return UNDERSCORE_NAME_RE.sub("_", Utils.hyphen_name(name))
+
+    @staticmethod
+    def storage_bucket(name: str) -> str:
+        from google.cloud import storage
+
+        storage_client = storage.Client()
+        bucket = storage_client.create_bucket(Utils.hyphen_name(name))
```
tvalentyn, line 62: Do you use some random suffixes here to make sure two tests don't step on each other?

dataflow/gpu-workers/conftest.py
```diff
@@ -0,0 +1,503 @@
...
+
+        logging.info(f"Created storage_bucket: {bucket.name}")
+        yield bucket.name
+
+        # Print all the objects in the bucket before deleting for debugging.
```
tvalentyn, line 67: Does the cleanup code  after `yield`  get executed if we only create one bucket?

dataflow/gpu-workers/pytorch-minimal/e2e_test.py
```diff
@@ -0,0 +1,60 @@
...
+@pytest.fixture(scope="session")
+def build_image(utils: Utils) -> str:
+    yield from utils.cloud_build_submit(
+        image_name=NAME,
+        config="build.yaml",
+        substitutions={"_IMAGE": f"{NAME}:{utils.uuid}"},
+    )
+
+
+@pytest.fixture(scope="session")
```
tvalentyn, line 42: Does this mean you only run one dataflow job per session? It doesn't seem to make sense as the test suite should run as many jobs as requested, on the other hand it's fine to create only one bucket for test session if you'd like to reuse the same bucket.

dataflow/gpu-workers/pytorch-minimal/main.py
```diff
@@ -0,0 +1,63 @@
...
+    )
+    pipeline.run()
+
+
+if __name__ == "__main__":
+    logging.getLogger().setLevel(logging.INFO)
+
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "--input-text",
```
tvalentyn, line 57: no strong opinion  but this might be additional complication to the example. You could as well just create a pcollection with this value.

dataflow/gpu-workers/tensorflow-minimal/run.yaml
```diff
@@ -0,0 +1,55 @@
...
+  entrypoint: python
+  args:
+  - /pipeline/main.py
+  - --runner=DataflowRunner
+  - --project=$PROJECT_ID
+  - --region=$_REGION
+  - --job_name=$_JOB_NAME
+  - --temp_location=$_TEMP_LOCATION
+  - --sdk_container_image=gcr.io/$PROJECT_ID/$_IMAGE
+  - --experiment=worker_accelerator=type:$_GPU_TYPE;count:$_GPU_COUNT;install-nvidia-driver
```
tvalentyn, line 47: since this is using tensorflow, let's add:
--experiment=no_use_multiple_sdk_containers

tvalentyn (approved):

dataflow/flex-templates/streaming_beam/requirements.txt
```diff
@@ -1 +1 @@
-apache-beam[gcp]==2.29.0
+apache-beam[gcp]==2.30.0
```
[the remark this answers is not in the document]
davidcavazos: From a user perspective, Beam is a requirement, so it should be here. If Beam is already installed, reinstalling it should be a no-op.

dataflow/gpu-workers/pytorch-minimal/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+FROM pytorch/pytorch:1.8.1-cuda11.1-cudnn8-runtime
+
+WORKDIR /pipeline
+
+# Copy the Apache Beam worker files and the pipeline source files.
+COPY --from=apache/beam_python3.8_sdk:2.29.0 /opt/apache/beam /opt/apache/beam
```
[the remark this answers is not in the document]
davidcavazos: Thanks, updated in all. Thankfully we also have a bot to automatically update Docker image versions.

dataflow/gpu-workers/pytorch-minimal/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+COPY requirements.txt .
+COPY *.py ./
+
+# Install the pipeline requirements and check that there are no conflicts.
+# Since the image already has all the dependencies installed,
+# there's no need to run with the --requirements_file option.
+RUN pip install --no-cache-dir --upgrade pip \
```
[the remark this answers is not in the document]
davidcavazos: It's always a good idea to upgrade pip. And yes, the pytorch image came with an older version of pip.

If it breaks with a newer version of pip, that sounds like we should file an issue to Beam to fix the dependency version constraints.

dataflow/gpu-workers/pytorch-minimal/e2e_test.py
davidcavazos: It's one job for this specific test suite. It doesn't really make any difference since there is only one test that uses this.

In case we wanted to check different things about the job, it would make sense to only launch one job and just poll different properties from it, although in this case we're only waiting until it finishes successfully.

dataflow/gpu-workers/pytorch-minimal/main.py
davidcavazos: Yeah, my train of thought was that most pipelines need one or more input options, so it would be easier to modify by having that as a pattern. I also don't have a strong opinion, but I think it's common enough that it would still be useful.

dataflow/gpu-workers/tensorflow-minimal/run.yaml
davidcavazos: Thanks, added to the landsat example as well.

dataflow/flex-templates/streaming_beam/requirements.txt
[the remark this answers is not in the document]
davidcavazos: I'll split these changes to another PR to unblock this one.

dataflow/gpu-workers/conftest.py
davidcavazos: Yes, we use a unique identifier for each test so each test has its own unique resources.

dataflow/gpu-workers/conftest.py
davidcavazos: Yes, fixtures do cleanup after the yield.

dataflow/gpu-workers/pytorch-minimal/e2e_test.py
davidcavazos: Also, if we sent multiple jobs, they would have the same job name so subsequent jobs would fail. For these tests, having exactly one job is what we want. It could be different in other test suites, but each one defines how they run their jobs.

dataflow/gpu-workers/tensorflow-minimal/run.yaml
tvalentyn: landsat uses a 1 core vm, so it's not necessary. you could remove the custom VM machine type, but as far as I remember, you needed a highmem machine type so you had to set some non-standard machine type anyway.

tvalentyn: it wouldn't hurt, but users might get confused thinking that both measures are required.

tvalentyn: If the default machine type works for you, you could just use the flag and omit the machine type.

davidcavazos:
Hi @kurtisvg, all tests are passing, this should be ready for review.

dataflow/gpu-workers/tensorflow-landsat/README.md
```diff
@@ -0,0 +1,39 @@
+# Processing Landsat satellite images with GPUs
```
tvalentyn, line 1: Can we decouple changes to Landsat example into a separate change, or make a new folder like 'landsat_v2'?

We currently refer to Landsat example in https://cloud.google.com/dataflow/docs/samples/satellite-images-gpus , and I think these instructions will no longer work once this is merged, since Python version is no longer a parameter in this dockerfile.

Alternatively, we should merge this simultaneously with an update to https://cloud.google.com/dataflow/docs/samples/satellite-images-gpus.

tvalentyn: The second option may be easier, but we would need to make and test the website changes first.

dataflow/gpu-workers/requirements.txt
```diff
@@ -0,0 +1,4 @@
+Pillow==8.2.0
+apache-beam[gcp]==2.29.0
```
tvalentyn: thanks for restoring the example. nit: 2.30.0 here as well

davidcavazos: Good catch! Thanks

davidcavazos:
Restored the old landsat sample to avoid breaking the current docs, and moved the new sample into a new directory because otherwise there were conflicts in the directory pytest ran tests from.

dataflow/gpu-examples/pytorch-minimal/README.md
```diff
@@ -0,0 +1,39 @@
...
+We use Cloud Build to run the [Dataflow](https://cloud.google.com/dataflow) job.
+
+> ℹ️ We launch the job using the worker image to make sure the job launches
+> with the same Python version as the workers and all the dependencies installed.
+
+```sh
+export REGION="us-central1"
+export GPU_TYPE="nvidia-tesla-t4"
+
+gcloud beta builds submit \
```
leahecole, line 30: If it's not necessary to use beta here, let's use regular `gcloud builds submit`

dataflow/gpu-examples/tensorflow-landsat/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+> ℹ️ We launch the job using the worker image to make sure the job launches
+> with the same Python version as the workers and all the dependencies installed.
+
+```sh
+export OUTPUT_PATH="gs://$BUCKET/samples/dataflow/landsat/output-images/"
+export REGION="us-central1"
+export GPU_TYPE="nvidia-tesla-t4"
+
+gcloud beta builds submit \
```
leahecole, line 35: same comment about using `beta`

dataflow/gpu-examples/tensorflow-landsat/main.py
```diff
@@ -0,0 +1,334 @@
...
+from PIL import Image
+import rasterio
+import tensorflow as tf
+
+DEFAULT_RGB_BAND_NAMES = ["B4", "B3", "B2"]
+DEFAULT_MIN_BAND_VALUE = 0.0
+DEFAULT_MAX_BAND_VALUE = 12000.0
+DEFAULT_GAMMA = 0.5
+
+DEFAULT_SCENES = [
```
leahecole, line 69: Thank you for commenting what the locations are!! So cool. 🌍 🌎 🌏

dataflow/gpu-examples/tensorflow-minimal/README.md
```diff
@@ -0,0 +1,39 @@
...
+We use Cloud Build to run the [Dataflow](https://cloud.google.com/dataflow) job.
+
+> ℹ️ We launch the job using the worker image to make sure the job launches
+> with the same Python version as the workers and all the dependencies installed.
+
+```sh
+export REGION="us-central1"
+export GPU_TYPE="nvidia-tesla-t4"
+
+gcloud beta builds submit \
```
leahecole, line 30: same comment about beta

leahecole (changes requested):
Mostly little nits - I also wanted to check that it was intentional to commit the .gcloudignore and .dockerignore files

dataflow/gpu-examples/pytorch-minimal/README.md
davidcavazos: I was using beta because only beta supports streaming the logs, so without beta it looked like a black box process that hanged indefinitely. It definitely works without the beta. There is a console message suggesting to use beta if you want logs, so I guess we can leave that to the user.

davidcavazos:
The .gcloudgnore and .dockerignore were part of the sample before, mostly to avoid uploading the Python virtualenv and other files to Cloud Build which were several GiB. However, now we're doing everything in Cloud Build, so we don't even need Python installed locally. I removed them because now they're irrelevant.

We'll remove the older landsat example under gpu-workers once we update the docs.

davidcavazos:
@leahecole @tvalentyn, tests are passing, this should be ready to merge if there are no more comments.

dataflow/gpu-examples/conftest.py
```diff
@@ -0,0 +1,508 @@
...
+                time.sleep(sleep_sec)
+
+        # Start a subprocess in the background to do the publishing.
+        logging.info(f"Starting publisher on {topic_path}")
+        p = mp.Process(target=_infinite_publish_job)
+        p.start()
+
+        yield p.is_alive()
+
+        # For cleanup, terminate the background process.
```
leahecole, line 185: Is there ever a circumstance where this process wouldn't be stopped?

dataflow/gpu-examples/conftest.py
```diff
@@ -0,0 +1,508 @@
...
+    @staticmethod
+    def dataflow_jobs_get(
+        job_id: Optional[str] = None,
+        job_name: Optional[str] = None,
+        project: str = PROJECT,
+        list_page_size: int = 30,
+    ) -> Optional[Dict[str, Any]]:
+        from googleapiclient.discovery import build
+
+        dataflow = build("dataflow", "v1b3")
```
leahecole, line 296: If we update past v1b3 will this build command fail? (just making sure it's future proofed so that when dependencies get updated we don't break samples)

dataflow/gpu-examples/conftest.py
```diff
@@ -0,0 +1,508 @@
...
+    @staticmethod
+    def dataflow_jobs_wait(
+        job_id: Optional[str] = None,
+        job_name: Optional[str] = None,
+        project: str = PROJECT,
+        region: str = REGION,
+        until_status: str = "JOB_STATE_DONE",
+        timeout_sec: str = 30 * 60,
+        poll_interval_sec: int = 60,
+        list_page_size: int = 100,
```
leahecole, line 334: Non blocking nit - what do you think about these also being constants like the other `RETRY_MAX_TIME`? Might help make them more visible for long term maintenance.

dataflow/gpu-examples/conftest.py
```diff
@@ -0,0 +1,508 @@
...
+        ] + [
+            f"--parameters={name}={value}"
+            for name, value in {
+                **parameters,
+            }.items()
+        ]
+        logging.info(f"{cmd}")
+
+        try:
+            # The `capture_output` option was added in Python 3.7, so we must
```
leahecole, line 481: Okay @davidcavazos in general I'm a fan of your extremely helpful comments but this one is my favorite and I think the most helpful in this file so far 😄

dataflow/gpu-examples/pytorch-minimal/e2e_test.py
```diff
@@ -0,0 +1,60 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+try:
+    # `conftest` cannot be imported when running in `nox`, but we still
+    # try to import it for the autocomplete when writing the tests.
```
leahecole, line 19: clever!

dataflow/gpu-examples/pytorch-minimal/noxfile_config.py
```diff
@@ -0,0 +1,41 @@
...
+
+# You can copy this file into your directory, then it will be imported from
+# the noxfile.py.
+
+# The source of truth:
+# https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/noxfile_config.py
+
+TEST_CONFIG_OVERRIDE = {
+    # You can opt out from the test for specific Python versions.
+    # > ℹ️ We're opting out of all Python versions except 3.8.
```
leahecole, line 25: 🎉

dataflow/gpu-examples/pytorch-minimal/run.yaml
```diff
@@ -0,0 +1,55 @@
...
+#   https://cloud.google.com/build/docs/build-config
+#
+# To learn more about Cloud Build variable substitutions:
+#   https://cloud.google.com/build/docs/configuring-builds/substitute-variable-values#using_user-defined_substitutions
+# -----------------------------------------------------------------------------
+
+substitutions:
+  _IMAGE: samples/dataflow/pytorch-gpu:latest
+  _JOB_NAME: ''
+  _TEMP_LOCATION: ''
```
leahecole, line 31: Does it matter that these substitutions aren't defined? (and/or do they need to be defined in our test project cloud build?)

leahecole (changes requested):
David this is so so so close and also it's a really fantastic PR. When you're back, let's sit down for like 15 min and just knock out my questions. (I sent you an email about this too in case your inbox is gross from being OOO)

dataflow/gpu-examples/conftest.py
davidcavazos: I made some experiments:

* Joining with timeout 0 actually doesn't do anything since it's an infinite loop so it never finishes.
* Terminate always stops the subprocess, but my experiments are linear, and pytest might be doing some nasty concurrency
* If I don't terminate the process, the whole thing hangs because the main process waits for the subprocess to finish, which might explain the intermittent hangs I was getting (if for some reason the fixture cleanup doesn't execute)
* Apparently, we can set up [`daemon`](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.Process.daemon) to True and the main process no longer waits for the subprocess to finish, so it terminates it automatically even if I don't explicitly terminate it :)

I deleted the `join`, kept the `terminate`, but also made the subprocess into a daemon, and added a large explanation on it.

dataflow/gpu-examples/conftest.py
davidcavazos: I don't think it would break, but I don't think it receives a ton of maintenance either. I personally prefer the gcloud commands for these things, but I know the general rule is to avoid them if there's a client library.

dataflow/gpu-examples/conftest.py
davidcavazos: Sure, sounds good.

On another note, the `RETRY_MAX_TIME` constant is actually not used anymore so I deleted it as well :)

dataflow/gpu-examples/pytorch-minimal/run.yaml
davidcavazos: If they're undefined, Dataflow will find suitable defaults. The README doesn't include to use these, but for the tests we actually define these so we can keep track of the resources.

davidcavazos:
@leahecole the new Beam version depends on `google-cloud-profiler` which requires a C++ compiler and Python build tools in the container image. Also the newer PyTorch image is quite larger so I had to adjust the timeout for building the image.

All tests are passing again.

leahecole (approved):

leahecole:
Used admin merge just because second approver didn't have correct permissions, but they are a Dataflow expert.

PR #6250, opened 2021-06-29
ppai/image-classification: fix dependency conflicts

davidcavazos:
## Description

In pip 21 with Beam 2.30 and AI Platform 1.1.1, there is a dependency version conflict causing pip to hang for a really long time. Pinning `google-cloud-bigquery==2.18.0`, which is required by both libraries, seems to resolve this conflict.

The notebook also now needs to restart the runtime due to dependency updates, so I'm also updating that.

## Checklist

### DIFF
--- a/people-and-planet-ai/image-classification/requirements.txt
+++ b/people-and-planet-ai/image-classification/requirements.txt
@@ -1,3 +1,4 @@
 Pillow==8.2.0
-apache-beam[gcp]==2.29.0
+apache-beam[gcp]==2.30.0
 google-cloud-aiplatform==1.1.1
+google-cloud-bigquery==2.18.0 # Indirect dependency, but there is a version conflict that causes pip to hang unless we constraint this.
--- a/people-and-planet-ai/image-classification/README.ipynb
+++ b/people-and-planet-ai/image-classification/README.ipynb
@@ -164,6 +164,69 @@
     {
       "cell_type": "markdown",
       "metadata": {
+        "id": "5S5_eLKE-M2o"
+      },
+      "source": [
+        "### ⛏️ Preparing the project environment\n",
+        "\n",
+        "Click run ▶️ for the following cells to download and install the necessary libraries and resources for this solution.\n",
+        "\n",
+        "> 💡 You can _optionally_ view the entire\n",
+        "> [code in GitHub](https://github.com/GoogleCloudPlatform/python-docs-samples/tree/master/people-and-planet-ai/image-classification)."
+      ]
+    },
+    {
+      "cell_type": "code",
+      "metadata": {
+        "id": "zdiQJibg9iEn"
+      },
+      "source": [
+        "# We need libffi-dev to launch the Dataflow pipeline.\n",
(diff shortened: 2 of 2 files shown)


people-and-planet-ai/image-classification/requirements.txt
```diff
@@ -1,3 +1,4 @@
 Pillow==8.2.0
-apache-beam[gcp]==2.29.0
+apache-beam[gcp]==2.30.0
 google-cloud-aiplatform==1.1.1
+google-cloud-bigquery==2.18.0 # Indirect dependency, but there is a version conflict that causes pip to hang unless we constraint this.
```
leahecole, line 4: should this possibly go into a constraints.txt instead?

davidcavazos: There isn't a way to pass a constraints.txt file to Beam, so the Dataflow launch process would hang when staging the dependencies.

busunkim96:
@davidcavazos Is there a bug where I can read more about the dependency conflict? I'd like to make sure we haven't pinned something in an incompatible manner on our side.

davidcavazos:
I just discovered this issue yesterday. I'm also not sure if it has to do with colab's pre-installed dependencies.

If you install `apache-beam[gcp]` in isolation, or `google-cloud-aiplatform` in isolation, they both work on their own. But if you try to install them together, then pip backtracks on every possible version combination. The culprit was `google-cloud-bigquery`, but I haven't found a root cause.

people-and-planet-ai/image-classification/requirements.txt
leahecole: Makes sense

leahecole (approved):

leahecole:
@busunkim96 I changed you to be reviewer 2 just in case - it's fine from a data analytics standpoint but if you come up with something else, feel free to suggest changes! A reminder also that if you need to pin to pip < 21 you can set a specific pip version in the noxfile config (Airflow requires this, so if that's the case, I feel your pain!)

busunkim96 (approved):
This seems like the right thing to to get the sample to work.  I think it's worth poking around a bit more to figure out what's happening with pip's dependency resolution.

davidcavazos:
Update: It has to do with Colab's pre-installed dependencies.

* Beam has [`google-cloud-bigquery>=1.6.0,<3`](https://github.com/apache/beam/blob/5fffad689e3fa7b6f788f3037a02fb215c1864ec/sdks/python/setup.py#L192)
* AI Platform has [`google-cloud-bigquery >= 1.15.0, < 3.0.0dev`](https://github.com/googleapis/python-aiplatform/blob/b95e0406566879e8f71cefda72b41dc6fe4e578f/setup.py#L65)

Colab has `google-cloud-bigquery==1.21.0` preinstalled, so when we install the Beam and AI Platform, that version still falls into a valid range and tries to keep it, but some other indirect dependency must be incompatible with it. Most likely another Colab dependency.

I think it's safe to update.

PR #6271, opened 2021-07-07
Ai platform -> vertex ai

leahecole:
branding update

### DIFF
--- a/data-science-onramp/feature-engineering/feature_engineering.ipynb
+++ b/data-science-onramp/feature-engineering/feature_engineering.ipynb
@@ -95,13 +95,13 @@
    "source": [
     "# Feature engineering with pandas and scikit-learn\n",
     "\n",
-    "This notebook demonstrates how to use AI Platform notebooks to perform feature engineering on a dataset using Pandas.\n",
+    "This notebook demonstrates how to use Notebooks to perform feature engineering on a dataset using Pandas.\n",
     "\n",
     "For each dataset, you will load the data into a Pandas DataFrame, clean and transform the columns into a usable format, and then restructure the data into feature and target data columns.\n",
     "\n",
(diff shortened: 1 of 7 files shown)


leahecole:
having an internal discussion about what exactly this region tag prefix is supposed to be

leahecole:
Okay! This should be ready for review. The region tag prefix for Vertex Ai products is going to be `aiplatform` to match with the underlying googleapis service name.

busunkim96 (approved):

dandhlee (commented):
Should we leave a note to clear the confusion between AI platform and Vertex AI? In case folks get confused with the rebranding.

On the other hand, the region tag is staying as aiplatform (which I followed the discussion and it makes sense) but we're moving stuff to vertex-ai directories, would this be temporary and the region tags eventually change to `vertexai` prefixes?

bradmiro (approved):

leahecole:
> Should we leave a note to clear the confusion between AI platform and Vertex AI? In case folks get confused with the rebranding.
> 
> On the other hand, the region tag is staying as aiplatform (which I followed the discussion and it makes sense) but we're moving stuff to vertex-ai directories, would this be temporary and the region tags eventually change to `vertexai` prefixes?

I think to solve this what I'll eventually do is add a link in the README to the actual published tutorial. Right now, the only place where `aiplatform` appears in the code is the region tags, and the region tags themselves won't show up in the tutorial. The region tag will not be changing as long as the underlying `aiplatform.googleapis.com` servicename doesn't change

leahecole:
I can talk about this whole service name confusion on a GVC at some point if you want

dandhlee (approved):
Sounds good! I think I understand the gist of what's going on, but a reminder in our bi-weekly meeting would be good, so we're all up to date C:

PR #6280, opened 2021-07-09
docs: clarified reference to project parameter

craiglabenz:
Region tags `firestore_setup_client_create` and `firestore_setup_client_create_with_project_id` are being consolidated. The latter was never a thing in the Python samples, but here we keep what Python samples that do exist in line with the target state for all 7 languages (1 snippet that explains the optional parameter).


firestore/cloud-async-client/snippets.py
```diff
@@ -20,8 +20,9 @@ async def quickstart_new_instance():
     # [START firestore_setup_client_create_async]
     from google.cloud import firestore
 
-    # Project ID is determined by the GCLOUD_PROJECT environment variable
-    db = firestore.AsyncClient()
+    # The `project` parameter is optional and if not supplied will
+    # fall back to your gcloud project configuration.
```
crwilcox, line 24: I may be too concerned, but the way we infer projects is broader than gcloud. https://github.com/googleapis/python-cloud-core/blob/master/google/cloud/client.py#L237

```
    # If you don't specify credentials when constructing the client, the
    # client library will look for credentials in the environment.
```

Is how we describe it in auth, https://cloud.google.com/docs/authentication/getting-started#verifying_authentication. Even that isn't very specific to what is happening though, maybe intentionally.

craiglabenz: I share this concern - went back and forth on it with @kolea2 last week on the phrasing. Maybe the three of us can put our heads together on one succinct phrasing to rule them all.

kolea2: I think you still need to update this comment to match the other?

craiglabenz: Oh my goodness - this Python a/sync split will be the end of me one day.

leahecole (approved):

craiglabenz:
Like Sisyphus, I appear doomed to merge master into this branch for the rest of time.

leahecole:
> Like Sisyphus, I appear doomed to merge master into this branch for the rest of time.

Gotta get two approvals in this repo! @crwilcox or @kolea2 should be your number 2

kolea2 (approved):
lgtm!

kolea2:
@googlebot

PR #6439, opened 2021-07-19
feat: Fixing the fetch_gpg_keys.sh script

### LINKED ISSUE #6440: ha.pool.sks-keyservers.net no longer resolveable
Since this repo uses similar kokoro setup to the one [python-compute](https://github.com/googleapis/python-compute) uses, I figured it will face similar problems as described in https://github.com/googleapis/python-compute/issues/79

I am working on a fix for this, so we don't use the unavailable key server.

m-strzelczyk:
## Description

Fixes #6440

The `fetch_gpg_keys.sh` script doesn't work anymore since `ha.pool.sks-keyservers.net` is no longer resolvable, as can be seen on their [website](https://sks-keyservers.net/status/):

> This service is deprecated. This means it is no longer maintained, and new HKPS certificates will not be issued. Service reliability should not be expected.
> 
> Update 2021-06-21: Due to even more GDPR takedown requests, the DNS records for the pool will no longer be provided at all.

Here I try to provide an update to the key acquisition, based on what I found on https://www.python.org/downloads/


parthea:
@m-strzelczyk , I'm going to close this PR as this change doesn't appear to be required in this repo. Please feel free to re-open it though.

PR #6456, opened 2021-07-22
chore(CODEOWNERS): Update functions owners

ace-n:

PR #6457, opened 2021-07-22
chore: Update CODEOWNERS

grant:
Removes @grant from sample owners. (I'm on the SWE team now)


grant:
Using https://github.com/GoogleCloudPlatform/python-docs-samples/pull/6456.

PR #6534, opened 2021-08-06
Update example to apps/v1

SimonBiggs:
## Description

I'm not certain if this is correct, but when I was following the following documentation:

https://cloud.google.com/endpoints/docs/grpc/get-started-kubernetes-engine#deploying_the_sample_api_and_esp_to_the_cluster

I was getting the following error message:

```
no matches for kind "Deployment" in version "extensions/v1beta1"
```

Which lead me to the following GitHub issue:

https://github.com/coreos/etcd-operator/issues/2126#issuecomment-561100503

So, here's the fix that made the example work for me again.


kurtisvg (approved):

kurtisvg:
@daniel-sanche PTAL

parthea (approved):

dandhlee (approved):
I've made similar changes for others that needed change to `apps/v1`. Should be good.

SimonBiggs:
Thanks @dandhlee, @parthea and @kurtisvg :slightly_smiling_face:.

Cheers,
Simon

PR #6545, opened 2021-08-10
more codeowner/blunderbuss updates

leahecole:
I forgot to add @loferris to blunderbuss along with #6540 and I removed Torry's GH Handle from things as he left Google last Friday


dandhlee (approved):
I wonder if we should come up with a tool that syncs and/or checks if CODEOWNERS and Blunderbuss are in sync with each other 🐈

leahecole:
oh we def should - it probably aligns with longer term work of syncing google groups and github groups too

busunkim96 (approved):

PR #6581, opened 2021-08-16
docs: require timezone aware datetimes

busunkim96:
Follow up to discussion in https://github.com/googleapis/python-tasks/issues/147#issuecomment-895454979.

If this is an OK blanket recommendation to make, I will follow up with a PR to adjust all the calls to `datetime.datetime` in this repository.

For supporting documentation, see warnings for these two methods:
- https://docs.python.org/3/library/datetime.html#datetime.datetime.utcfromtimestamp
- https://docs.python.org/3/library/datetime.html#datetime.datetime.utcnow


dandhlee (approved):

leahecole (approved):

leahecole:
Now I'm curious about what kinds of mayhem not doing this can cause 😁

tswast:
There's a DATETIME data type in BigQuery that's specifically not timezone-aware, but I still think this makes sense as general advice when not dealing with a data type where it doesn't make sense.

https://stackoverflow.com/a/47724366/101923

AUTHORING_GUIDE.md
```diff
@@ -288,6 +288,20 @@ When documenting primitive types, be sure to note if they have a particular set
 of constraints. For example, `A base64-encoded string` or `Must be between 0
 and 10`.
 
+### `datetime.datetime` Objects
+
+Always create timezone aware datetime objects. For libraries that use protobuf,
+omitting the timezone may lead to unexpected behavior when the datetime
+is converted to a protobuf tiemstamp.
```
tswast, line 295:
```suggestion
is converted to a protobuf timestamp.
```

PR #6582, opened 2021-08-17
docs: add guidelines around imports and initializing GAPIC objects

busunkim96:
This proposes two changes to the style guidelines:

1. Imports

    Ask folks to always use the `from google.cloud import bar` style. I think this is already [the most common style](https://github.com/GoogleCloudPlatform/python-docs-samples/search?p=5&q=%22from+google.cloud%22&type=code) but we don't explicitly spell it out anywhere. 

2. GAPIC library object construction.

    proto-plus provides several different ways to create objects, and our samples currently use a mix. I've seen a steady stream of questions around this on library repos (example: https://github.com/googleapis/python-tasks/issues/148) since the microgenerator rolled out.  I think we also need better guidance at the library ref doc (or maybe in the "language specific, API agnostic" docs that are being planned), but asking sample authors to use one of the "preferred" styles is another way to nudge folks towards the happy path.


AUTHORING_GUIDE.md
```diff
@@ -171,6 +171,84 @@ comments and docstrings only as needed to further clarify the code’s intent.
...
+
+client = texttospeech_v1.TextToSpeechClient()
+
+audio_config = texttospeech.AudioConfig(
+        audio_encoding=texttospeech.AudioEncoding.MP3
+)
+```
+
+
+### Creating Request Objects for GAPICs
```
busunkim96, line 196: I had trouble wrapping my head around where in the tree these sections belong.  Please let me know if you have suggestions on a better layout!

AUTHORING_GUIDE.md
```diff
@@ -171,6 +171,84 @@ comments and docstrings only as needed to further clarify the code’s intent.
...
+
+GAPIC libraries are generated from [protos](https://github.com/googleapis/googleapis)
+that define the API surface via a [generator](https://github.com/googleapis/gapic-generator-python)).
```
dandhlee, line 199: nit: extra closing bracket.

AUTHORING_GUIDE.md
dandhlee: I think the current layout is good for `Importing Google Cloud Libraries`. 

For `Creating Request Objects for GAPICs` section, maybe it makes more sense to put it between `Functions and Classes` and `README File` but since GAPIC instantiated objects aren't necessarily about functions/classes I think it's fine to put it where it is now.

dandhlee (approved):
Sorry for the late review! LGTM, one minor nit.

leahecole (approved):
LGTM once Dan's changes are incorporated 😄

AUTHORING_GUIDE.md
leahecole, line 199:
```suggestion
that define the API surface via a [generator](https://github.com/googleapis/gapic-generator-python).
```

leahecole (approved):

PR #6643, opened 2021-08-31
Simplify trigger dag example

leahecole:
## Description
part of b/198306911

### DIFF
--- a/composer/workflows/trigger_response_dag.py
+++ b/composer/workflows/trigger_response_dag.py
@@ -18,28 +18,18 @@
 import datetime
 
 import airflow
-from airflow.operators import bash_operator
+from airflow.operators.bash_operator import BashOperator
 
 
-default_args = {
-    'owner': 'Composer Example',
-    'depends_on_past': False,
-    'email': [''],
-    'email_on_failure': False,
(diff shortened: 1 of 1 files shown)


tswast (approved):

composer/workflows/trigger_response_dag.py
```diff
@@ -18,28 +18,17 @@
...
-    'email_on_retry': False,
-    'retries': 1,
-    'retry_delay': datetime.timedelta(minutes=5),
-    'start_date': datetime.datetime(2017, 1, 1),
-}
-
 with airflow.DAG(
         'composer_sample_trigger_response_dag',
-        default_args=default_args,
+        start_date=datetime.datetime(2021, 1, 1),
```
busunkim96, line 26: Please add a timezone if it makes sense in this context. 😄

busunkim96 (approved):

composer/workflows/trigger_response_dag.py
leahecole: It does not matter for this use case because the DAG is triggered by GCF, but good lookin out! This date update just makes this look up to date :)

PR #6648, opened 2021-09-01
chore(functions/logging): exclude trace code from sample

ace-n:
Fixes b/183751712


run/logging-manual/main.py
```diff
@@ -30,12 +30,14 @@ def index():
...
     # [START cloudrun_manual_logging]
     # [START run_manual_logging]
+    # [START functions_manual_logging]
     # Uncomment and populate this variable in your code:
     # PROJECT = 'The project ID of your Cloud Run service';
 
     # Build structured log messages as an object.
     global_log_fields = {}
-
+
```
dandhlee, line 39: lint is complaining about this extra newline.

dandhlee (dismissed):
Minor lint issue, otherwise LGTM.

ace-n:
@grayside PTAL?

dinagraves:
This is causing this test to fail: https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/run/logging-manual/main_test.py#L38

dinagraves:
When I added a print statement to see what was in `locals()`, it returns `{'PROJECT': 'XXXXXXX', 'global_log_fields': {}}`.

dinagraves:
FYI - 3.9 is the only python version that is actually running the samples, which is why it's the only one failing.

ace-n:
@dinagraves thanks - let's try `globals()` instead and see if that works. 🙂

grayside (commented):
I like the approach of skipping the trace header if it's not available to the code. I don't know enough Python to say if this is a code smell. If I understand correctly, we're trying to infer if the code is running in an HTTP serving context by whether the request object is imported from flask?

Maybe there's a standard approach? https://stackoverflow.com/questions/40284800/check-if-flask-request-context-is-available/40286162

ace-n:
> I like the approach of skipping the trace header if it's not available to the code. I don't know enough Python to say if this is a code smell. If I understand correctly, we're trying to infer if the code is running in an HTTP serving context by whether the request object is imported from flask?

Yes - or at least a context where the Flask `request` value is available.

> Maybe there's a standard approach? https://stackoverflow.com/questions/40284800/check-if-flask-request-context-is-available/40286162

Will this (`has_request_context`) work even if `Flask` isn't imported? (I _suspect_ it won't, as `has_request_context` _itself_ won't be defined.)

dinagraves:
@ace-n  Could you write a test for this?

run/logging-manual/main.py
```diff
@@ -12,9 +12,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# [START cloudrun_manual_logging]
+# [START run_manual_logging]
```
dandhlee, line 16: `run_manual_logging` doesn't seem to actually be used, is this region tag needed?

dandhlee: snippetbot also is complaining that the product prefix should be `cloudrun`

dinagraves: Confirming the prefix should be `cloudrun` --when in doubt you an check the product in DevRel Services

run/logging-manual/main.py
```diff
@@ -37,13 +42,17 @@ def index():
...
-    trace_header = request.headers.get("X-Cloud-Trace-Context")
-
-    if trace_header and PROJECT:
-        trace = trace_header.split("/")
-        global_log_fields[
-            "logging.googleapis.com/trace"
-        ] = f"projects/{PROJECT}/traces/{trace[0]}"
+    # This is only relevant in HTTP-based contexts, and is ignored elsewhere.
+    # (In particular, non-HTTP-based Cloud Functions.)
+    request_is_defined = 'request' in globals() or 'request' in locals()
```
dinagraves, line 47: Can you run `black` on this?

dinagraves (approved):
Conditional approval:

- Please run `black` to make the formatting more consistent
- If `run_manual_logging` isn't used in the docs, please remove it

run/logging-manual/main.py
ace-n: Removed that one, thanks y'all!

run/logging-manual/main.py
ace-n: Done.

dandhlee (approved):
LGTM.

PR #6684, opened 2021-09-12
Functions v2 storage sample

engelke:
New sample for functions v2 sample triggered by storage cloudevent.


functions/v2/storage/main.py
```diff
@@ -0,0 +1,37 @@
...
+    event_id = cloudevent["id"]
+    event_type = cloudevent["type"]
+
+    bucket = data["bucket"]
+    name = data["name"]
+    metageneration = data["metageneration"]
+    timeCreated = data["timeCreated"]
+    updated = data["updated"]
+
+    print('Event ID: {}'.format(event_id))
```
ace-n, line 30: Given that GCF only supports Python 3, would it be better to use `f`-strings here?

engelke: I have been slow to move to f-strings due to older versions, but you're right. f-strings would be better. I'll do a quick update.

engelke: Done.

ace-n (approved):

busunkim96 (approved):

PR #6699, opened 2021-09-16
Two character update to Airflow DAG

leahecole:
I was going through [this tutorial](https://cloud.google.com/composer/docs/how-to/using/using-kubernetes-pod-operator#secret-config) earlier and found that the environment variable set here throws TemplateNotFound errors. It's because of a stupid quirk of Airflow and adding a space after the filename fixes it. 🙄 (found in [common pitfalls](https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=62694614))


dandhlee (approved):
Interesting that it needs a space at the end 🧐

busunkim96 (approved):

leahecole:
> Interesting that it needs a space at the end 🧐

I think the word you're looking for is "obnoxious" 😁

PR #6764, opened 2021-09-29
fix: fail nox session if python version is missing

busunkim96:
By default nox skips the session if the interpreter is missing.
This asks it to error instead.


leahecole (approved):

dandhlee (approved):
LGTM!

PR #6777, opened 2021-09-30
Update Cloud SQL sample apps with SSL example

jsimonweb:


loferris:
code LGTM so far, but we should figure out what's going on with the Python 3.10 build.

PR #7015, opened 2021-11-02
Update Cloud SQL sample apps with Kubernetes Engine yaml files.

jsimonweb:


cloud-sql/mysql/sqlalchemy/deployment.yaml
```diff
@@ -0,0 +1,66 @@
...
+        # Replace $YOUR_PROJECT_ID with your project ID.
+        image: $LOCATION-docker.pkg.dev/$YOUR_PROJECT_ID/gke-cloud-sql-repo/gke-sql:latest
+        # This app listens on port 8080 for web traffic by default.
+        ports:
+        - containerPort: 8080
+        env:
+        - name: PORT
+          value: "8080"
+        - name: DB_HOST
+          value: "0.0.0.0"
```
kurtisvg, line 28: Should this be localhost? or a filler value? 
```suggestion
          value: "0.0.0.0"

```

kurtisvg (changes requested):

cloud-sql/mysql/sqlalchemy/deployment.yaml
jsimonweb: Updated to be 
value: "127.0.0.1"

kurtisvg (approved):

crwilcox (approved):

PR #7076, opened 2021-11-09
Exempt dataflow from 3.10

leahecole:
@davidcavazos this was an educated guess based on which examples skip 3.9. Please edit or close as needed


busunkim96 (approved):

davidcavazos (approved):
Thanks! Yes, Beam just now has plans to begin supporting 3.9 sometime in the (hopefully) near future, but 3.10 is not under the radar yet.

PR #7101, opened 2021-11-12
Add Cloud Run sample

averikitsch:
## Description

I will create a bug to add E2E tests

## Checklist


run/jobs/main.py
```diff
@@ -0,0 +1,64 @@
...
+# Retrieve Job-defined env vars
+TASK_NUM = os.getenv('TASK_NUM') or 0
+ATTEMPT_NUM = os.getenv('ATTEMPT_NUM') or 0
+# Retrieve User-defined env vars
+SLEEP_MS = os.getenv('SLEEP_MS')
+FAIL_RATE = os.getenv('FAIL_RATE')
+
+
+# Define main script
+def main(sleep_ms=None, fail_rate=None):
```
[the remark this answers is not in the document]
averikitsch: I've set it up this way to make testing easier of the main function, so instead of setting env vars for the test I can just pass them in as `main.main(SLEEP_MS, RATE)`. Therefore I have to check for Nones anyways. Is there something that is more idiomatic?

engelke (approved):
LGTM

run/jobs/main.py
```diff
@@ -0,0 +1,64 @@
...
+    print(f"Starting Task #{TASK_NUM}, Attempt #{ATTEMPT_NUM}...")
+    # Simulate work
+    if (sleep_ms):
+        # Wait for a specific amount of time
+        time.sleep(float(sleep_ms)/1000)  # Convert to seconds
+
+    # Simulate errors
+    if (fail_rate):
```
dinagraves, line 38:
```suggestion
    if fail_rate:
```

run/jobs/main.py
```diff
@@ -0,0 +1,64 @@
...
+# Retrieve User-defined env vars
+SLEEP_MS = os.getenv('SLEEP_MS')
+FAIL_RATE = os.getenv('FAIL_RATE')
+
+
+# Define main script
+def main(sleep_ms=None, fail_rate=None):
+    print(f"Starting Task #{TASK_NUM}, Attempt #{ATTEMPT_NUM}...")
+    # Simulate work
+    if (sleep_ms):
```
dinagraves, line 33:
```suggestion
    if sleep_ms:
```

run/jobs/main.py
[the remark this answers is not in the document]
dinagraves: I was thinking that if fail_rate defaults to 0, then the logic in the random_failure function could be simplified.

run/jobs/main.py
```diff
@@ -0,0 +1,64 @@
...
+            random_failure(float(fail_rate))
+        except Exception:
+            raise Exception(f"Task #{TASK_NUM}, Attempt #{ATTEMPT_NUM} failed.")
+
+    print(f"Completed Task #{TASK_NUM}.")
+
+
+# Throw an error based on fail rate
+def random_failure(rate):
+    if (rate is None or rate < 0 or rate > 1):
```
dinagraves, line 49: I'm not sure you need to check for `None` here, even as written.  This function is only ever called after `if fail_rate`  -- so it should never be `None`.   

That being said, defaulting to 0 will better ensure that it is definitely never `None.`

dinagraves: Also, please remove the parentheses!  https://google.github.io/styleguide/pyguide.html#33-parentheses

run/jobs/main.py
```diff
@@ -0,0 +1,61 @@
...
+# limitations under the License.
+
+import json
+import os
+import random
+import sys
+import time
+
+# Retrieve Job-defined env vars
+TASK_NUM = os.getenv('TASK_NUM') or 0
```
dinagraves, line 22:
```suggestion
TASK_NUM = os.getenv('TASK_NUM', 0)
```

run/jobs/main.py
```diff
@@ -0,0 +1,61 @@
...
+ATTEMPT_NUM = os.getenv('ATTEMPT_NUM') or 0
```
dinagraves, line 23:
```suggestion
ATTEMPT_NUM = os.getenv('ATTEMPT_NUM', 0)
```

run/jobs/main.py
```diff
@@ -0,0 +1,61 @@
...
+# Retrieve User-defined env vars
+SLEEP_MS = os.getenv('SLEEP_MS') or 0
```
dinagraves, line 25:
```suggestion
SLEEP_MS = os.getenv('SLEEP_MS', 0) 
```

run/jobs/main.py
```diff
@@ -0,0 +1,61 @@
...
+FAIL_RATE = os.getenv('FAIL_RATE') or 0
```
dinagraves, line 26:
```suggestion
FAIL_RATE = os.getenv('FAIL_RATE', 0)
```

run/jobs/main.py
```diff
@@ -0,0 +1,61 @@
...
+    except Exception:
+        raise Exception(f"Task #{TASK_NUM}, Attempt #{ATTEMPT_NUM} failed.")
+
+    print(f"Completed Task #{TASK_NUM}.")
+
+
+# Throw an error based on fail rate
+def random_failure(rate):
+    if rate < 0 or rate > 1:
+        print(f"Invalid FAIL_RATE env var value: {rate}. Must be a float between 0 and 1 inclusive.")
```
dinagraves, line 47: This isn't raising an exception!

dinagraves (commented):
I hope you don't mind - I updated your test file.  The `try/exept` format doesn't test for the scenario when a failure *should* happen.  In this case, I noticed that you expected a failure to happen when the fail_rate wasn't valid, but instead of raising an exception, it printed a message.

Also added a suggestion on a better use of `getenv`.   

Could you run `black` on both files to ensure formatting consistency?

dinagraves:
Just noticed that tests were not actually running in here, so I added the `noxfile_config`

run/jobs/main_test.py
```diff
@@ -0,0 +1,59 @@
...
+
+def test_bad_env_vars():
+    with pytest.raises(Exception, match=r".*Invalid.*"):
+        main.main(fail_rate="2")
+
+
+def test_run_script():
+    output = (
+        subprocess.run(
+            ["python", "main.py"],
```
dinagraves, line 41: Not sure if you want to add a note here about python aliases?  If `python` is pointing at `2.*`, then it will fail because our samples only support python3.   Locally, I have to change this to `python3` to make it work.

run/jobs/main.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+# Define main script
+def main(sleep_ms=0, fail_rate=0):
+    print(f"Starting Task #{TASK_NUM}, Attempt #{ATTEMPT_NUM}...")
+    # Simulate work by waiting for a specific amount of time
+    time.sleep(float(sleep_ms)/1000)  # Convert to seconds
+
+    # Simulate errors
+    try:
```
dinagraves, line 36: I would remove this `try/except` block, because otherwise you'll have nested Exceptions.  E.g. if the `random_failure < rate` exception is raised, then that will trigger raising the 2nd exception with `Task ...`.  

It would be better to call this just like
`random_failure(...)`

And then have all the exception messaging and logic happen in the `random_failure` function. 

Example:
```
def main(sleep_ms=0, fail_rate=0):
    print(f"Starting Task #{TASK_NUM}, Attempt #{ATTEMPT_NUM}...")
    # Simulate work by waiting for a specific amount of time
    time.sleep(float(sleep_ms)/1000)  # Convert to seconds

    # Simulate errors
    random_failure(float(fail_rate))

    print(f"Completed Task #{TASK_NUM}.")


# Throw an error based on fail rate
def random_failure(rate):
    if rate < 0 or rate > 1:
        raise Exception(f"Invalid FAIL_RATE env var value: {rate}. Must be a float between 0 and 1 inclusive.")

    random_failure = random.random()
    if random_failure < rate:
        raise Exception(f"Task #{TASK_NUM}, Attempt #{ATTEMPT_NUM} failed.")
```

run/jobs/main.py
averikitsch: I've decided that I don't want this throw an exception so it won't retry

dinagraves (approved):

PR #7140, opened 2021-11-19
Using check_call to expose underlying error messages

dinagraves:
Changed `subprocess.run` to `subprocess.check_call` whenever we're running a command and not saving the stdout.  This bubbles up the error message.  

We had some flaky tests failing on subprocess.run (eg [hello-broken](https://source.cloud.google.com/results/invocations/31e5fdd2-bcc3-4012-a153-1fd81271dba6/targets/github%2Fpython-docs-samples%2Frun%2Fhello-broken/tests)), so this should help us debug it in the future.

PR #7149, opened 2021-11-22
test: use `check_output` for memorystore e2e test

busunkim96:
Use `subprocess.check_output`, which is equivalent to:

```py
run(..., check=True, stdout=PIPE).stdout
```

This ensures we can see the error from the subprocess in the log. This is helpful for debugging issues like https://github.com/GoogleCloudPlatform/python-docs-samples/issues/7104.

https://docs.python.org/3/library/subprocess.html#subprocess.check_output

For this test file:

```py
import subprocess
import pytest

@pytest.fixture
def auth_token():
    auth_token = subprocess.check_output(
        ["gcloud", "auth", "print-idetity-token"],
    ).strip().decode()
    yield auth_token


def test_dummy(auth_token):
    assert auth_token is not None
```

The test results look like this:

```
=============================================== test session starts ================================================
platform linux -- Python 3.9.7, pytest-6.2.5, py-1.11.0, pluggy-1.0.0
rootdir: /usr/local/google/home/busunkim
collected 1 item                                                                                                   

test-header.py E                                                                                             [100%]

====================================================== ERRORS ======================================================
___________________________________________ ERROR at setup of test_dummy ___________________________________________

    @pytest.fixture
    def auth_token():
>       auth_token = subprocess.check_output(
            ["gcloud", "auth", "print-idetity-token"],
        ).strip().decode()

test-header.py:6: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/usr/lib/python3.9/subprocess.py:424: in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

input = None, capture_output = False, timeout = None, check = True
popenargs = (['gcloud', 'auth', 'print-idetity-token'],), kwargs = {'stdout': -1}
process = <Popen: returncode: 2 args: ['gcloud', 'auth', 'print-idetity-token']>, stdout = b'', stderr = None
retcode = 2

    def run(*popenargs,
            input=None, capture_output=False, timeout=None, check=False, **kwargs):
        """Run command with arguments and return a CompletedProcess instance.
    
        The returned instance will have attributes args, returncode, stdout and
        stderr. By default, stdout and stderr are not captured, and those attributes
        will be None. Pass stdout=PIPE and/or stderr=PIPE in order to capture them.
    
        If check is True and the exit code was non-zero, it raises a
        CalledProcessError. The CalledProcessError object will have the return code
        in the returncode attribute, and output & stderr attributes if those streams
        were captured.
    
        If timeout is given, and the process takes too long, a TimeoutExpired
        exception will be raised.
    
        There is an optional argument "input", allowing you to
        pass bytes or a string to the subprocess's stdin.  If you use this argument
        you may not also use the Popen constructor's "stdin" argument, as
        it will be used internally.
    
        By default, all communication is in bytes, and therefore any "input" should
        be bytes, and the stdout and stderr will be bytes. If in text mode, any
        "input" should be a string, and stdout and stderr will be strings decoded
        according to locale encoding, or by "encoding" if set. Text mode is
        triggered by setting any of text, encoding, errors or universal_newlines.
    
        The other arguments are the same as for the Popen constructor.
        """
        if input is not None:
            if kwargs.get('stdin') is not None:
                raise ValueError('stdin and input arguments may not both be used.')
            kwargs['stdin'] = PIPE
    
        if capture_output:
            if kwargs.get('stdout') is not None or kwargs.get('stderr') is not None:
                raise ValueError('stdout and stderr arguments may not be used '
                                 'with capture_output.')
            kwargs['stdout'] = PIPE
            kwargs['stderr'] = PIPE
    
        with Popen(*popenargs, **kwargs) as process:
            try:
                stdout, stderr = process.communicate(input, timeout=timeout)
            except TimeoutExpired as exc:
                process.kill()
                if _mswindows:
                    # Windows accumulates the output in a single blocking
                    # read() call run on child threads, with the timeout
                    # being done in a join() on those threads.  communicate()
                    # _after_ kill() is required to collect that and add it
                    # to the exception.
                    exc.stdout, exc.stderr = process.communicate()
                else:
                    # POSIX _communicate already populated the output so
                    # far into the TimeoutExpired exception.
                    process.wait()
                raise
            except:  # Including KeyboardInterrupt, communicate handled that.
                process.kill()
                # We don't call process.wait() as .__exit__ does that for us.
                raise
            retcode = process.poll()
            if check and retcode:
>               raise CalledProcessError(retcode, process.args,
                                         output=stdout, stderr=stderr)
E               subprocess.CalledProcessError: Command '['gcloud', 'auth', 'print-idetity-token']' returned non-zero exit status 2.

/usr/lib/python3.9/subprocess.py:528: CalledProcessError
---------------------------------------------- Captured stderr setup -----------------------------------------------
ERROR: (gcloud.auth) Invalid choice: 'print-idetity-token'.
Maybe you meant:
  gcloud auth activate-service-account
  gcloud auth configure-docker
  gcloud auth list
  gcloud auth login
  gcloud auth print-access-token
  gcloud auth print-identity-token
  gcloud auth revoke

To search the help text of gcloud commands, run:
  gcloud help -- SEARCH_TERMS
============================================= short test summary info ==============================================
ERROR test-header.py::test_dummy - subprocess.CalledProcessError: Command '['gcloud', 'auth', 'print-idetity-toke...
================================================= 1 error in 1.01s =================================================
```

### DIFF
--- a/memorystore/redis/cloud_run_deployment/e2e_test.py
+++ b/memorystore/redis/cloud_run_deployment/e2e_test.py
@@ -35,7 +35,7 @@ MEMORYSTORE_REDIS_NAME = "static-test-instance"
 @pytest.fixture
 def redis_host():
     # Get the Redis instance's IP
-    redis_host = subprocess.run(
+    redis_host = subprocess.check_output(
         [
             "gcloud",
             "redis",
@@ -47,9 +47,7 @@ def redis_host():
             "--project",
             PROJECT,
         ],
-        stdout=subprocess.PIPE,
-        check=True
-    ).stdout.strip().decode()
+    ).strip().decode()
     yield redis_host
 
     # no deletion needs to happen, this is a "get" of a static instance
@@ -59,14 +57,14 @@ def redis_host():
 def container_image():
     # Build container image for Cloud Run deployment
     image_name = f"gcr.io/{PROJECT}/test-visit-count-{SUFFIX}"
-    subprocess.run(
+    subprocess.check_output(
         [
             "cp",
             "cloud_run_deployment/Dockerfile",
             ".",
-        ], check=True
+        ],
     )
-    subprocess.run(
+    subprocess.check_output(
         [
             "gcloud",
             "builds",
@@ -75,14 +73,14 @@ def container_image():
             image_name,
             "--project",
             PROJECT,
-        ], check=True
+        ],
     )
     yield image_name
 
-    subprocess.run(["rm", "Dockerfile"], check=True)
+    subprocess.check_output(["rm", "Dockerfile"])
 
     # Delete container image
-    subprocess.run(
+    subprocess.check_output(
         [
             "gcloud",
             "container",
@@ -92,7 +90,7 @@ def container_image():
             "--quiet",
             "--project",
             PROJECT,
-        ], check=True
+        ],
     )
 
 
@@ -100,7 +98,7 @@ def container_image():
 def deployed_service(container_image, redis_host):
     # Deploy image to Cloud Run
     service_name = f"test-visit-count-{SUFFIX}"
-    subprocess.run(
+    subprocess.check_output(
         [
             "gcloud",
             "run",
@@ -117,12 +115,12 @@ def deployed_service(container_image, redis_host):
             f"REDISHOST={redis_host},REDISPORT=6379",
             "--project",
             PROJECT,
-        ], check=True
+        ]
     )
     yield service_name
 
     # Delete Cloud Run service
-    subprocess.run(
+    subprocess.check_output(
         [
             "gcloud",
             "run",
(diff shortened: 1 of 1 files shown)


busunkim96:
I forgot DIna already opened #7140, please ignore this one!

PR #7157, opened 2021-11-23
feat(functions/v2): Add generic audit logging sample

dinagraves:
Following canonical sample: https://github.com/GoogleCloudPlatform/nodejs-docs-samples/blob/main/functions/v2/index.js#L65


functions/v2/audit_log/main.py
```diff
@@ -0,0 +1,38 @@
...
+    # See https://github.com/cloudevents/spec/blob/v1.0.1/spec.md#subject
+    print(f"Subject: {cloudevent['subject']}")
+
+    # Print out details from the `protoPayload`
+    # This field encapsulates a Cloud Audit Logging entry
+    # See https://cloud.google.com/logging/docs/audit#audit_log_entry_structure
+
+    payload = cloudevent.data.get("protoPayload")
+    if payload:
+        print(f"API method: {payload['methodName']}")
```
dinagraves, line 33: Arguably, these should all be `.get` so it doesn't break if the key is missing.

engelke (approved):
LGTM, though I agree that using get instead of square brackets is probably better practice.

muncus (approved):
I agree on the `get` suggestion. This is a nice, concise sample!

PR #7159, opened 2021-11-24
Fix flaky tests by running slow deletion as a background task

dinagraves:
There have been many flaky tests failing on `gcloud run services delete`.   The underlying error seems to show that the deletion call is slow, but it succeeds eventually.  I am proposing running this asynchronously and trusting that it will delete.


ace-n (commented):
This _seems_ OK to me, but I'll defer to @averikitsch or @grayside.

leahecole:
If this is a best practice for cloud run products or for samples that use gcloud within, should we add this to the authoring guide?

parthea (approved):

grayside:
We usually adjust timeouts and add retries per gcloud command in an effort to guarantee the process completes successfully before the next test run. Cloud Run automatically persists unspecified configuration to new deployments, so if a deletion fails, any configuration bits that weren't explicitly set for a new test will be inherited and potentially disrupt service behaviors.

This is not a problem if the Cloud Run service is named per build.

If deletions do tend to fail, there's also a risk of reaching the 1000 service and 1000 revision quota limits, which would lead to test failures.

leahecole (approved):

PR #7204, opened 2021-12-07
dataflow/gpu: add dataflow prime gpu hints

davidcavazos:
## Description

Use Dataflow Prime to give hints for GPU usage.

R: @tvalentyn
R: @rezarokni

## Checklist
- [ ] **Tests** pass:   `nox -s py-3.6` (see [Test Environment Setup](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/main/AUTHORING_GUIDE.md#test-environment-setup))


dataflow/gpu-examples/tensorflow-landsat/Dockerfile
```diff
@@ -16,7 +16,7 @@
 #   https://www.tensorflow.org/install/source#gpu
 # For a list of all the nvidia images:
 #   https://ngc.nvidia.com/catalog/containers/nvidia:cuda/tags
-FROM nvcr.io/nvidia/cuda:11.4.2-cudnn8-runtime-ubuntu20.04
+FROM nvcr.io/nvidia/cuda:11.2.2-cudnn8-runtime-ubuntu20.04
```
davidcavazos, line 19: As per https://www.tensorflow.org/install/source#gpu for TensorFlow 2.7 we **must** use CUDA 11.2, not the latest. Otherwise it fails.

dataflow/gpu-examples/tensorflow-landsat/Dockerfile
```diff
@@ -37,5 +37,11 @@ RUN apt-get update \
     && pip check
 
 # Set the entrypoint to Apache Beam SDK worker launcher.
-COPY --from=apache/beam_python3.8_sdk:2.34.0 /opt/apache/beam /opt/apache/beam
-ENTRYPOINT [ "/opt/apache/beam/boot" ]
+# COPY --from=apache/beam_python3.8_sdk:2.34.0 /opt/apache/beam /opt/apache/beam
+# ENTRYPOINT [ "/opt/apache/beam/boot" ]
+
+# Workaround for b/199755022
```
davidcavazos, line 43: We might want to wait until this fix is published before we merge, probably by Beam 2.35 (?)

davidcavazos: This works with Beam 2.35 RC3

dataflow/gpu-examples/tensorflow-landsat/main.py
```diff
@@ -290,45 +297,54 @@ def run(
...
-        "--gamma", type=float, default=DEFAULT_GAMMA, help="Gamma correction factor."
-    )
+        "--gamma",
+        type=float,
+        default=DEFAULT_GAMMA,
+        help="Gamma correction factor.")
+    parser.add_argument(
+        "--gpu-type",
+        default=DEFAULT_GPU_TYPE,
+        help="GPU type hint to use.")
```
tvalentyn, line 332: technically, we have one resource hint (--accelerator), that captures both type and count. you could still customize type and count separately if you prefer, but perhaps don't call it a 'hint' to avoid confusion.

tvalentyn (commented):
Main comment: looks like we replace the non-prime example with an example that uses Dataflow Prime.

It may be better to have both examples. In which case, the non-prime example can be a part of GPU tutorial, and Prime example could be something that additionally demonstrates some benefits of Dataflow Prime and a part of Prime tutorial.

dataflow/gpu-examples/tensorflow-landsat/Dockerfile
```diff
@@ -37,5 +37,5 @@ RUN apt-get update \
     && pip check
 
 # Set the entrypoint to Apache Beam SDK worker launcher.
-COPY --from=apache/beam_python3.8_sdk:2.34.0 /opt/apache/beam /opt/apache/beam
+COPY --from=apache/beam_python3.8_sdk:2.35.0_rc3 /opt/apache/beam /opt/apache/beam
```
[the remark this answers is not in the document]
davidcavazos: Good point! I'll add a note in the README

dataflow/gpu-examples/tensorflow-landsat/main.py
davidcavazos: I would prefer to capture them separately to avoid user errors as much as possible. That way the accelerator hint format can stay as an implementation detail. I'll remove the word "hint".

dataflow/gpu-examples/tensorflow-landsat/main.py
```diff
@@ -249,33 +255,34 @@ def run(
...
             lambda x, unused_side_input: x,
             unused_side_input=beam.pvalue.AsSingleton(
-                pipeline | beam.Create([None]) | beam.Map(check_gpus)
+                pipeline
+                | beam.Create([None])
+                | beam.Map(check_gpus).with_resource_hints(accelerator=gpu_hint)
             ),
         )
         | "Get RGB band paths" >> beam.Map(get_band_paths, rgb_band_names)
+        | "Reshuffle" >> beam.Reshuffle()
```
rezarokni, line 278: Explain with comments why we are using Reshuffle, can link the break fusion page in Dataflow docs

dataflow/gpu-examples/tensorflow-landsat/main.py
```diff
@@ -66,6 +66,9 @@
 DEFAULT_MAX_BAND_VALUE = 12000.0
 DEFAULT_GAMMA = 0.5
 
+DEFAULT_GPU_TYPE = "nvidia-tesla-t4"
```
rezarokni, line 69: Link to docs which show different types.

dataflow/gpu-examples/tensorflow-landsat/run.yaml
```diff
@@ -40,16 +40,16 @@ steps:
...
   - /pipeline/main.py
   - --output-path-prefix=$_OUTPUT_PATH
+  - --gpu-type=$_GPU_TYPE
+  - --gpu-count=$_GPU_COUNT
   - --runner=DataflowRunner
   - --project=$PROJECT_ID
   - --region=$_REGION
-  - --job_name=$_JOB_NAME
   - --temp_location=$_TEMP_LOCATION
-  - --worker_machine_type=custom-1-13312-ext
```
rezarokni, line 48: As this needed a 13 Gig to 1 mem ratio before, maybe worth adding a memory min limit to the hint as well

davidcavazos: Yes, I actually added a `--resource_hints=min_ram=32GB`. I tried 16GB but I was still hitting out of memory errors. Something changed where it tried to load more images per worker I guess.

davidcavazos:
@tvalentyn I also thought of having two separate examples, but it meant too much duplicated code. That would be maintaining two almost identical samples. Maybe what we can do is having a config file to run with Prime and a config file to run without it. With that in mind, we would have to pass the accelerator flag directly as you suggested.

But that would involve having the accelerator flag passed to the entire pipeline instead of having the per-transform hints. Or maybe we can have the per-transform hints, but pass None if it's in non-Prime. What do you think?

dataflow/gpu-examples/tensorflow-landsat/run.yaml
tvalentyn: I may be running 2 cores on the VM? I think we may need to set --experiments=no_use_multiple_sdk_containers

tvalentyn:
> @tvalentyn I also thought of having two separate examples, but it meant too much duplicated code. That would be maintaining two almost identical samples. Maybe what we can do is having a config file to run with Prime and a config file to run without it.

We currently have two GPU offerings: GPUs without prime and GPUs with prime, it's different use-cases in Dataflow. Prime is in Preview, while Dataflow GPUs are GA. I think right now we cannot recommend prime as a drop in replacement for non-prime, so it may be good to show users how to use both.

We could decide that a non-prime GPU offering does not need a tutorial and the documentation is sufficient. then we can move GPU tutorial to use Prime and have only one example. 

Alternatively, we can have two tutorials. they may not necessarily be the same. Ideally, the GPU prime tutorial would focus on prime advantages, for example demonstrate that having multiple worker pools results in cost savings, while running the same job on a single pool is more expensive. 

would leave that to DPE folks to decide.

dataflow/gpu-examples/tensorflow-landsat/run.yaml
davidcavazos: It's already using it in line 53

davidcavazos:
@tvalentyn I like the "We could decide that a non-prime GPU offering does not need a tutorial and the documentation is sufficient. then we can move GPU tutorial to use Prime and have only one example" alternative.

We also have the `tensorflow-minimal` and `pytorch-minimal` samples, which are non-prime. So how about we use the Landsat tutorial as the prime example, and simply link to the minimal examples for the non-prime examples for now.

We still need to update the Landsat tutorial in the docs for the prime setup and instructions.

How would you feel about this?

tvalentyn:
Once prime is GA that is completely fine; while prime is preview, but GPU is GA , it's still possible, but we may need extra wording around this in the tutorial

davidcavazos:
@pcoet what are your thoughts?

dataflow/gpu-examples/tensorflow-landsat/main.py
davidcavazos: Thanks, adding a comment and linking to https://cloud.google.com/dataflow/docs/guides/deploying-a-pipeline#preventing-fusion

dataflow/gpu-examples/tensorflow-landsat/main.py
davidcavazos: Thanks, I added a comment and a link to https://cloud.google.com/dataflow/docs/resources/locations#gpu_availability

I also added a similar note in the README when running the pipeline and passing the GPU type.

pcoet:
@davidcavazos I don't have a lot of context here, but my initial thought is that we should keep both the Prime and non-Prime examples for now. That seems like the best solution for users.

dataflow/gpu-examples/tensorflow-landsat/Dockerfile
leahecole: Can you please add this as an inline comment so that other maintainers and I don't upgrade it by accident?

davidcavazos:
Alright, sounds good to me. I'll create a copy and we'll take it from there.

davidcavazos:
The sample was moved to `tensorflow-landsat-prime` and left the original `tensorflow-landsat` sample untouched. After GA we might want to revisit if we want to adapt the tutorial in the docs, point to the prime sample, and remove the old sample.

davidcavazos:
Tests are passing, but we should still wait until 2.35 stable is published before merging.

rezarokni:
+1 on two copies, but fine if you use same code with branches.

davidcavazos:
Hi @leahecole, Beam 2.35 stable is out, tests are passing with it. I think this should be ready to merge.

kurtisvg (approved):

dataflow/gpu-examples/tensorflow-landsat/Dockerfile
kurtisvg: Should we consider pinning this in renovate?

leahecole (approved):

PR #7218, opened 2021-12-09
feat: Prevent overriding the value from config

irvifa:
## Description

Fixes #<ISSUE-NUMBER>Previously the DEBUG env var will always be overridden into True, while
we already have a way to read it from env var. By setting the default
value as True and not overridding the value later we can retainthe DEBUG
value provided from the users.

## Checklist
- [x] **Tests** pass:   `nox -s py-3.6` (see [Test Environment Setup](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/main/AUTHORING_GUIDE.md#test-environment-setup))


glasnt (approved):

run/django/mysite/settings.py
```diff
@@ -63,7 +63,7 @@
 
 # SECURITY WARNING: don't run with debug turned on in production!
 # Change this to "False" when you are ready for production
```
dandhlee, line 65: Could you move this warning up above to where it's set to `True` as well?

dandhlee (changes requested):
LGTM, one minor comment.

run/django/mysite/settings.py
irvifa: Addressed your comment @dandhlee

dandhlee (approved):
LGTM.

PR #7222, opened 2021-12-09
exempt some samples from python 3.10 tests

leahecole:
We are subscribed to and tracking two issues that are blocking these samples from being 3.10 compatible. I'll remove the exemption when they're resolved. 
https://github.com/google/tink/issues/553
https://github.com/GoogleCloudPlatform/cloud-debug-python/issues/33


appengine/standard_python3/cloud_debugger/noxfile_config.py
```diff
@@ -24,7 +24,7 @@
     # You can opt out from the test for specific Python versions.
     # google-python-cloud-debugger doesn't support Python 3.9.
     # https://github.com/GoogleCloudPlatform/cloud-debug-python/issues/22
```
dandhlee, line 26: nit: this isn't an issue anymore, could you remove it?

dandhlee (approved):
LGTM! One minor nit below..

appengine/standard_python3/cloud_debugger/noxfile_config.py
leahecole, line 26:
```suggestion
```

parthea (approved):

PR #7291, opened 2021-12-21
Added sample for geospatial classification

nikitamaia:
## Description

Geospatial classification sample for the People + Planet AI series.


people-and-planet-ai/geospatial-classification/trainer/task.py
```diff
@@ -0,0 +1,132 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import tensorflow as tf
+import argparse
+
```
davidcavazos, line 17: This would be a good place to have a high level overview of what the sample does. For people who bump into this code, it might be nice to have a short high level description of what we're trying to achieve and how the data looks like.

Maybe something like [this](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/d8d4951911029b4776bd8d34ad5a8832fe21787b/dataflow/gpu-examples/tensorflow-landsat/main.py#L17-L48)?

davidcavazos (commented):
Hi Nikita! Thanks for taking the time to collaborate with this sample, we really appreciate it!

This is looking great! I made some comments, mostly on style and nitpicking.

> Some background: I learned C and C++ in college and I grew writing imperative code, but the more that I've been delving into purely functional languages (like Elm and Haskell), I've been learning really cool ways of doing the same things but without side effects like mutating variables. I've noticed it's usually shorter, more concise, easier to read, and less error prone, but it does require a slightly different way of doing things. I think in summary, avoiding to mutate variables, avoiding for-loops and while-loops, and using list comprehensions or functions instead. I've reinforced this with Beam / Dataflow, Earth Engine, and even TensorFlow Datasets, where everything is done via `map`, `filter`, and some kind of `reduce`, and it makes it easier to parallelize / distribute.

people-and-planet-ai/geospatial-classification/trainer/task.py
```diff
@@ -0,0 +1,132 @@
...
+    inputs = tf.keras.Input(shape=(None, None, 13))
+    x = normalizer(inputs)
+    x = tf.keras.layers.Conv2D(filters=32, kernel_size=33, activation="relu")(x)
+    outputs = tf.keras.layers.Dense(1, activation="sigmoid")(x)
+
+    model = tf.keras.Model(inputs=inputs, outputs=outputs)
+
+    model.compile(
+        optimizer=tf.keras.optimizers.Adam(0.0001),
```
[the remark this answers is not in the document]
nikitamaia: learning rate doesn't make a ton of difference in this example, but when doing ml experimentation it tends to be one of the most impactful hyperparameters to tweak. For that reason I generally prefer to avoid the string.

people-and-planet-ai/geospatial-classification/requirements-test.txt
```diff
@@ -0,0 +1,4 @@
+earthengine-api==0.1.292
```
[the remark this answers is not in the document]
nikitamaia: The requirements file is only for the serving app which doesn't need all of this. Should I have a separate requirements file for the whole project?

people-and-planet-ai/geospatial-classification/e2e_test.py
```diff
@@ -0,0 +1,376 @@
...
+
+def get_neighboring_patch(feature):
+    """Gets image pixel values for patch."""
+
+    # filter ImageCollection at start/end dates.
+    image = (
+        ee.ImageCollection("COPERNICUS/S2")
+        .filterDate(feature.get("start"), feature.get("end"))
+        .select(BANDS)
+        .median()
```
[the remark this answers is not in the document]
nikitamaia: Does it make a difference since the start and end dates are only 1 day apart?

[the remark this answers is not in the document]
davidcavazos: I guess not, but it would also be consistent with what is done for prediction. Maybe we could even use the same function?

people-and-planet-ai/geospatial-classification/requirements-test.txt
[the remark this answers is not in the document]
davidcavazos: Mmm that's a good question. I think the requirements.txt file should contain anything required to run the sample locally. And then we could have another requirements for the serving app. It doesn't matter if there's duplication since we have a bot that auto-updates requirements.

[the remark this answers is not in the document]
nikitamaia: Ok, sounds good! Will do that

people-and-planet-ai/geospatial-classification/e2e_test.py
```diff
@@ -0,0 +1,376 @@
...
+
+    logging.info(f"bucket_name: {bucket_name}")
+    yield bucket_name
+
+    bucket.delete(force=True)
+
+
+@pytest.fixture(scope="session")
+def test_data(bucket_name):
+    label_fc = create_label_fc("labeled_geospatial_data.csv")
```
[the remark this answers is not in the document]
nikitamaia: Have you tried to run this export? I ran this snippet but some reason the export keeps failing with the message `Collection.randomColumn: Encoded object too large`

people-and-planet-ai/geospatial-classification/e2e_test.py
```diff
@@ -0,0 +1,376 @@
...
+    )
+
+    validation_task.start()
+
+    train_status = None
+    val_status = None
+
+    logging.info("Waiting for data export to complete.")
+    for _ in range(0, TIMEOUT_SEC, POLL_INTERVAL_SEC):
+        train_status = ee.data.getOperation(training_task.name)["metadata"]["state"]
```
davidcavazos, line 126: While trying this, apparently we can fetch the status like this as well:

```
train_status = training_task.status()['state']
```

nikitamaia: Did you try this with pytest? I think I had this earlier and it resulted in a bunch of warnings when I ran the tests that said `getOperation` was the non deprecated way to do this.

people-and-planet-ai/geospatial-classification/requirements.txt
```diff
@@ -0,0 +1,7 @@
+numpy==1.21.4
```
leahecole, line 1:
```suggestion
numpy==1.21.4; python_version > "3.6"
numpy==1.19.5; python_version <= "3.6"
```

people-and-planet-ai/geospatial-classification/trainer/task.py
```diff
@@ -0,0 +1,132 @@
...
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--bucket", required=True, type=str, help="GCS Bucket")
+    args = parser.parse_args()
+    return args
+
+
+def parse_tfrecord(example_proto, features_dict):
+    """Parses a single tf.train.Example."""
+
+    return tf.io.parse_single_example(example_proto, features_dict)
```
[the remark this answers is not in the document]
nikitamaia: Keeping it this way since this is generally how it's shown in the TF examples.

people-and-planet-ai/geospatial-classification/requirements-test.txt
```diff
@@ -0,0 +1,5 @@
...
```
davidcavazos, line 1: Any requirement that is already on the requirements.txt file is not necessary here. This is like an "additional" requirements for testing. Nox will install requirements.txt first, and then install requirements-test.txt on top of it. So I think we can keep only `pytest` and `pytest-xdist` :)

people-and-planet-ai/geospatial-classification/README.ipynb
```diff
@@ -0,0 +1,1135 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 557,
+   "id": "1395bd18-e6eb-47cf-b0ca-7d7d04cce8f0",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import requests"
```
davidcavazos, line 839: Can we merge this into the cell below?

people-and-planet-ai/geospatial-classification/README.ipynb
```diff
@@ -0,0 +1,1135 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 559,
+   "id": "a7f6b8c2-cab3-4030-8528-9afe644a800d",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import json"
```
davidcavazos, line 894: Same here, can we merge this into the cell below?

people-and-planet-ai/geospatial-classification/README.ipynb
```diff
@@ -0,0 +1,1135 @@
...
+    "\n",
+    "  </button>\n",
+    "\n",
+    "1. In the project list, select the project that you want to delete, and then click **Delete**.\n",
+    "\n",
+    "1. In the dialog, type the project ID, and then click **Shut down** to delete the project."
+   ]
+  },
+  {
+   "cell_type": "code",
```
davidcavazos, line 1100: Lets remove this empty cell.

davidcavazos (commented):
This is looking great! I left some minor comments on the notebook. In general, can we keep the cell outputs of the code cells? That way people can skim through the notebook and see the outputs without having to actually run it. Although we should remove any cell that includes personal information like a project ID or a bucket name.

people-and-planet-ai/geospatial-classification/e2e_test.py
```diff
@@ -0,0 +1,342 @@
...
+LABEL = "is_powered_on"
+
+IMAGE_COLLECTION = "COPERNICUS/S2"
+SCALE = 10
+
+TRAIN_VALIDATION_SPLIT = 0.7
+
+PATCH_SIZE = 16
+
+credentials, project = google.auth.default()
```
davidcavazos, line 77: Nit: `project` is no longer needed, can we do `credentials, _ = google.auth.default()`?

brijk7:
I've tried running through this notebook as well. Nice work!
One quick comment: please add pip package "folium" to requirements.txt (needed for the last display step in the notebook).

davidcavazos:
Also, can we add a README.md with a short description like the other samples? We'll add links to the video and blog posts once they're published. Can we also add an entry into the README at the `people-and-planet-ai` parent directory?

brijk7:
Nice work on this sample, Nikita! I ran through the notebook and had one comment: as a user who's new to satellite imagery and Earth Engine data, it's somewhat unclear how the time dimension is used in training the model and for prediction. After looking through the EE ImageCollection doc, I gathered that the input feature (both in training and in prediction) is created as a composite "mosaic" of images (filtered to the time range in question). So effectively the time dimension is not used in the model at all, is that right? Clarifying this in the doc explicitly would be helpful.

davidcavazos:
The time is used to know which image to extract from Earth Engine since the same power plant tower can be on or off at different times, but the model itself is not using the time for the prediction.

brijk7:
Thanks, it would be great to clarify this explicitly. In the model building section, we use a time window of 1 day for compositing the satellite image corresponding to the label. In the "Get Predictions" section, we extract the satellite image for the whole month of October 2021 and use it as the input. That assumes that the power plant was in a (mostly) static state during that period of time, correct?

people-and-planet-ai/geospatial-classification/serving_app/noxfile_config.py
```diff
@@ -0,0 +1,46 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Default TEST_CONFIG_OVERRIDE for python repos.
+
+# You can copy this file into your directory, then it will be imported from
+# the noxfile.py.
+
```
leahecole, line 19: This file should not be needed now because of #7342

davidcavazos:
> Thanks, it would be great to clarify this explicitly. In the model building section, we use a time window of 1 day for compositing the satellite image corresponding to the label. In the "Get Predictions" section, we extract the satellite image for the whole month of October 2021 and use it as the input. That assumes that the power plant was in a (mostly) static state during that period of time, I guess.

That's a good point, we should clarify that further. The `mosaic` function in Earth Engine basically gives you the first image similar to `first`, but it will stitch images together at the seams. So it really doesn't matter if it's one day or a year, it'll grab the first image of that time range. For training we know the exact date of the Sentinel 2 image, so we only add a day to make sure we include it. For user supplied images, we don't know the exact date so we grab the entire month, but still only the first image is relevant and the others are discarded. Sentinel 2 takes pictures every 10 days, so we just need to make sure we have at least one image in the time range. It complicates a little on the edges of images since we need to stitch them together to get one cohesive image. Each of the 4 images in an edge/corner will have different timestamps, so we're using one month to make sure we always get something.

nikitamaia:
> > Thanks, it would be great to clarify this explicitly. In the model building section, we use a time window of 1 day for compositing the satellite image corresponding to the label. In the "Get Predictions" section, we extract the satellite image for the whole month of October 2021 and use it as the input. That assumes that the power plant was in a (mostly) static state during that period of time, I guess.
> 
> That's a good point, we should clarify that further. The `mosaic` function in Earth Engine basically gives you the first image similar to `first`, but it will stitch images together at the seams. So it really doesn't matter if it's one day or a year, it'll grab the first image of that time range. For training we know the exact date of the Sentinel 2 image, so we only add a day to make sure we include it. For user supplied images, we don't know the exact date so we grab the entire month, but still only the first image is relevant and the others are discarded. Sentinel 2 takes pictures every 10 days, so we just need to make sure we have at least one image in the time range. It complicates a little on the edges of images since we need to stitch them together to get one cohesive image. Each of the 4 images in an edge/corner will have different timestamps, so we're using one month to make sure we always get something.

Yep thanks for pointing this out! I will update the sample with more detail on this.

people-and-planet-ai/geospatial-classification/trainer/task.py
```diff
@@ -0,0 +1,132 @@
...
+    feature_ds = training_dataset.map(lambda x, y: x)
+    normalizer = tf.keras.layers.experimental.preprocessing.Normalization()
+    normalizer.adapt(feature_ds)
+
+    inputs = tf.keras.Input(shape=(None, None, 13))
+    x = normalizer(inputs)
+    x = tf.keras.layers.Conv2D(filters=32, kernel_size=33, activation="relu")(x)
+    outputs = tf.keras.layers.Dense(1, activation="sigmoid")(x)
+
+    model = tf.keras.Model(inputs=inputs, outputs=outputs)
```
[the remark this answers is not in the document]
nikitamaia: Keeping functional API since it's the idiomatic way to do things in TF keras.

davidcavazos (approved):
LGTM, I'll be OOO next week, but this is looking great!

people-and-planet-ai/geospatial-classification/README.md
```diff
@@ -0,0 +1,18 @@
+# 🗺 Coal Plant Predictions -- _geospatial-classification_
```
davidcavazos, line 1: Nit: can we use 🏭 instead? We'd like to have unique icons for each sample :)

people-and-planet-ai/geospatial-classification/README.md
```diff
@@ -0,0 +1,18 @@
...
+
+[![Open in Colab](https://colab.research.google.com/assets/colab-badge.svg)](https://colab.research.google.com/github/GoogleCloudPlatform/python-docs-samples/blob/main/people-and-planet-ai/geospatial-classification/README.ipynb)
+
+> [Watch the video in YouTube<br> ![thumbnail](http://img.youtube.com/vi/LnEhSVEJUuY/0.jpg)](https://youtu.be/LnEhSVEJUuY)
```
davidcavazos, line 5: Let's remove the video until this episode launches.

nikitamaia:
> LGTM, I'll be OOO next week, but this is looking great!

Thanks David! Just resolved your last few comments :)

davidcavazos:
All tests are passing, I think this should be ready to merge. Thanks @nikitamaia for contributing in this lovely sample!

people-and-planet-ai/geospatial-classification/noxfile_config.py
```diff
@@ -0,0 +1,47 @@
...
+
+TEST_CONFIG_OVERRIDE = {
+    # You can opt out from the test for specific Python versions.
+    # > ℹ️ We're opting out of all Python versions except 3.8.
+    # > The Python version used is defined by the Dockerfile, so it's redundant
+    # > to run multiple tests since they would all be running the same Dockerfile.
+    "ignored_versions": ["2.7", "3.6", "3.7", "3.9", "3.10"],
+    # Old samples are opted out of enforcing Python type hints
+    # All new samples should feature them
+    # "enforce_type_hints": True,
```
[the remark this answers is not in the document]
leahecole: We do ask that new samples use type hints as per the authoring guide https://github.com/GoogleCloudPlatform/python-docs-samples/blob/main/AUTHORING_GUIDE.md#documenting-types - please switch that to true and add them in

leahecole (changes requested):
Please add type hints or, if for some reason an underlying product doesn't support that, make a note in the noxfile config and set the "enforce" boolean to false

people-and-planet-ai/geospatial-classification/README.ipynb
```diff
@@ -0,0 +1,1440 @@
...
+     "output_type": "display_data"
+    }
+   ],
+   "source": [
+    "import folium\n",
+    "import folium.plugins as folium_plugins\n",
+    "import branca.colormap as cm\n",
+    "\n",
+    "colormap = cm.LinearColormap(colors=['lightblue', 'red'], index=[0,1], vmin=0, vmax=1)\n",
+    "map = folium.Map([43.45, -7.87], zoom_start=14, tiles='CartoDB positron')\n",
```
davidcavazos, line 1339: Can we visualize satellite images for the tiles? I found the ESRI has freely available ones that we can use.

```
map = folium.Map(
    location=[43.44, -7.86],
    zoom_start=16,
    tiles='https://server.arcgisonline.com/ArcGIS/rest/services/World_Imagery/MapServer/tile/{z}/{y}/{x}',
    attr = 'ESRI'
)
```

I also slightly modified the coordinates and zoom level to have the power plant more centered.

davidcavazos (approved):
I made some minor comments while running through the sample. Other than that, LGTM

people-and-planet-ai/geospatial-classification/README.ipynb
```diff
@@ -0,0 +1,1440 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "bb8c6ede",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "!pip install -r -requirements.txt"
```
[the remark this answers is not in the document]
nikitamaia: thanks for catching this! Don't think I've used a constraints.txt file before. Just added!

people-and-planet-ai/geospatial-classification/README.ipynb
```diff
@@ -0,0 +1,1459 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "3b86fba8",
+   "metadata": {
+    "id": "3b86fba8"
+   },
+   "outputs": [],
+   "source": [
+    "%cd ~/python-docs-samples/people-and-planet-ai/geospatial-classification"
```
davidcavazos, line 265: Nit: sorry for the nitpicking, but can we move this to the end of the next cell? That way whenever we authenticate we make sure we're in the right directory

leahecole (approved):

PR #7314, opened 2022-01-05
add note about using finally

leahecole:
@tswast reminded me about `finally` the other day and I thought I'd add a note about it in the authoring guide


tswast (approved):

dandhlee (commented):
It is a large document but I think without the examples it can be easily glanced over. Could you include the example snippets in the document directly rather than referencing it through the link? I think this is a great example that should be stood out rather than have the chance of this slipping through the reader when they skim through the authoring guide.

dandhlee (approved):
LGTM!

PR #7342, opened 2022-01-10
Noxfile check for tests

leahecole:
This adds a check to the noxfile to see if there are tests in the directory. To test, use the noxfile-template in any directory with tests. Then try it on a directory like [this one](https://github.com/nikitamaia/python-docs-samples/tree/geospatial-sandbox/people-and-planet-ai/geospatial-classification/serving_app) which is found in PR #7291. This does assume that all tests are of the format `*_test.py` - if we want, I can also include a check for `test_*.py`, because I know that's a common Python pattern even though it's not the one we use. 

Corresponding PR in synthtool https://github.com/googleapis/synthtool/pull/1321

cc @davidcavazos - this will hopefully remove the need for your workaround

### DIFF
--- a/noxfile-template.py
+++ b/noxfile-template.py
@@ -39,17 +39,15 @@ import nox
 
 TEST_CONFIG = {
     # You can opt out from the test for specific Python versions.
-    'ignored_versions': ["2.7"],
-
+    "ignored_versions": ["2.7"],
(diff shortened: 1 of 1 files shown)


davidcavazos:
Thank you!

davidcavazos (approved):
LGTM

busunkim96 (approved):

PR #7393, opened 2022-01-25
docs(iam-samples): extract json_key_file from privateKeyData

Sita04:
Fixes googleapis/google-cloud-python#7824 

 Extract json_key_file from privateKeyData field in service account

## Checklist
- [ ] **Tests** pass:   `nox -s py-3.6` (see [Test Environment Setup](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/main/AUTHORING_GUIDE.md#test-environment-setup))


busunkim96 (approved):

dandhlee (changes requested):
Lint failure:
```
./service_account_keys.py:56:5: F841 local variable 'json_key_file' is assigned to but never used
    json_key_file = base64.b64decode(key['privateKeyData']).decode('utf-8')
    ^
./service_account_keys.py:56:21: F821 undefined name 'base64'
    json_key_file = base64.b64decode(key['privateKeyData']).decode('utf-8')
```

Sita04:
Lint error:
./service_account_keys.py:57:5: F841 local variable 'json_key_file' is assigned to but never used
    json_key_file = base64.b64decode(key['privateKeyData']).decode('utf-8')
    
Justification:
The json_key_file contains sensitive data and we don't want to log it. We want the customers to make an informed decision on how to handle the data.

hegemonic (approved):
LGTM

iam/api-client/service_account_keys.py
```diff
@@ -50,7 +51,11 @@ def create_key(service_account_email):
         name='projects/-/serviceAccounts/' + service_account_email, body={}
         ).execute()
 
-    print('Created key: ' + key['name'])
+    # The privateKeyData field contains the base64-encoded service account key
+    # in JSON format. Save this key to a secure location. You cannot download it
```
leahecole, line 55: To avoid the lint error, can we instead comment out line 57 and then suggest "Uncomment the next line and save this key to a secure location"

Sita04: If we comment the line 57, then the variable in line 50 will not be used which would throw a lint error :(
Is there any other way around this?

dandhlee: Could you try:
```suggestion
    print(f'Created json key: {json_key_file}')
```
or something similar to print that the key has been made (or other properties of the json key file that's more appropriate), that way we know that the key was created for sure?

hegemonic: IAM tech writer here. We shouldn't print the entire key file; it contains a private key, which is sensitive data.

dandhlee: That sounds fair. I'll leave it to @Sita04 to figure out what would be the most appropriate way to use that key without revealing sensitive data.

Sita04: I commented out the `json_key_file` and instead used the key status param `key['disabled']` to validate if the key was created.
@hegemonic @dandhlee PTAL

iam/api-client/service_account_keys.py
```diff
@@ -23,6 +23,7 @@
 
 import argparse
 # [START iam_create_key]
+# import base64
```
dandhlee, line 26: Could you add a comment on uncommenting this when running the code, or move it down further where the rest of the base64 items with the `TODO` are commented out? It can be easily glanced over and users could easily get confused on why it's commented out.

Sita04: That's a great suggestion! Makes sense. 
Thank you Dan.

dandhlee (approved):
LGTM!

PR #7405, opened 2022-01-28
feat(pubsublite): attributes field in sink example

anguillanneuf:
Because this bug was fixed in the Pub/Sub Lite Spark connector https://github.com/googleapis/java-pubsublite-spark/issues/261 with a new release,  we are now able to write to Pub/Sub Lite with an `attributes` field. 

Updating the sample to show how to do that.


pubsublite/spark-connector/spark_streaming_to_pubsublite_example.py
```diff
@@ -35,13 +36,32 @@ def spark_streaming_to_pubsublite(
     # |-- value: long (nullable = true)
     sdf = spark.readStream.format("rate").option("rowsPerSecond", 1).load()
 
+    # Transform the dataframe to match the required data fields and data types:
+    # https://github.com/googleapis/java-pubsublite-spark#data-schema
```
busunkim96, line 40: Is this the canonical link for this schema?

busunkim96 (approved):

pubsublite/spark-connector/spark_streaming_to_pubsublite_example.py
```diff
@@ -35,13 +36,32 @@ def spark_streaming_to_pubsublite(
...
+            ),
+        )
+        .drop("value")
     )
 
+    # After the transformation, the schema of the dataframe should look like:
+    # |-- key: binary (nullable = false)
+    # |-- data: binary (nullable = true)
+    # |-- event_timestamp: timestamp (nullable = true)
+    # |-- attributes: map (nullable = false)
```
jiangmichaellll, line 61: our guide and everywhere actually specifies attributes is nullable.

jiangmichaellll: but our code doesn't really check since they are compatible, I think it's fine.

jiangmichaellll (approved):

pubsublite/spark-connector/spark_streaming_to_pubsublite_example.py
anguillanneuf: Yes. It's the only place it exists for now.

pubsublite/spark-connector/spark_streaming_to_pubsublite_example.py
anguillanneuf: @jiangmichaellll  Sounds like we need to update our guide? When you say "everywhere", which places do you refer to?

hongalex (approved):

PR #7412, opened 2022-01-31
Add django CSRF setting

glasnt:
## Description

This updates the tutorials to support Django 4.0, specifically in relation to CSRF changes. 

## Checklist
- [x] README is updated to include [all relevant information](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/main/AUTHORING_GUIDE.md#readme-file)
- [x] **Tests** pass:   `nox -s py-3.6` (see [Test Environment Setup](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/main/AUTHORING_GUIDE.md#test-environment-setup))


ace-n (approved):
Python LGTM. It's been years since I worked with Django though. 🙂

glasnt:
For reference, https://docs.djangoproject.com/en/4.0/ref/csrf/#how-it-works has 4.0 specific changes (this is explained in the upcoming cl)

leahecole (approved):

appengine/standard_python3/django/mysite/settings.py
```diff
@@ -59,10 +60,22 @@
...
+# SECURITY WARNING: It's recommended that you use this when
+# running in production. The URL will be known once you first deploy
+# to App Engine. This code takes the URL and converts it to both these settings formats.
+APPENGINE_URL = env("APPENGINE_URL", default=None)
+if APPENGINE_URL:
+    # Ensure the HTTPS is in the URL before it's used.
+    APPENGINE_URL = urlparse(APPENGINE_URL, "https").geturl()
+
+    ALLOWED_HOSTS = [APPENGINE_URL]
+    CSRF_TRUSTED_ORIGINS = [urlparse(APPENGINE_URL).netloc]
```
lesteral, line 74: @glasnt - could this line could simply be?
`CSRF_TRUSTED_ORIGINS = [APPENGINE_URL]`
 as per Django 4.0 updates to this (CSRF_TRUSTED_ORIGINS) setting (vs. [The values in older versions must only include the hostname (possibly with a leading dot) and not the scheme or an asterisk.](https://docs.djangoproject.com/en/4.0/ref/settings/#csrf-trusted-origins:~:text=The%20values%20in%20older%20versions%20must%20only%20include%20the%20hostname%20,%20and%20not%20the%20scheme%20or%20an%20asterisk)

appengine/standard_python3/django/mysite/settings.py
```diff
@@ -59,10 +60,22 @@
...
-ALLOWED_HOSTS = ["*"]
+
+# [START gaestd_py_django_csrf]
+# SECURITY WARNING: It's recommended that you use this when
+# running in production. The URL will be known once you first deploy
+# to App Engine. This code takes the URL and converts it to both these settings formats.
+APPENGINE_URL = env("APPENGINE_URL", default=None)
+if APPENGINE_URL:
+    # Ensure the HTTPS is in the URL before it's used.
+    APPENGINE_URL = urlparse(APPENGINE_URL, "https").geturl()
```
lesteral, line 71: @glasnt - this line seems, to me, only marginally effective at "Ensur[ing] the HTTPS is in the URL before it's used", since:
(1) urlparse() uses the "scheme" argument (here: "https") only if a scheme isn't already present in the "urlstring" arg
(ref. [The scheme argument gives the default addressing scheme, to be used only if the URL does not specify one](https://docs.python.org/3/library/urllib.parse.html#:~:text=The%20scheme%20argument%20gives%20the%20default%20addressing%20scheme%2C%20to%20be%20used%20only%20if%20the%20URL%20does%20not%20specify%20one))
(2) if there's no "scheme://" prefix in the urlstring argument, urlparse() won't insert the "//" anyway, as per [urlparse recognizes a netloc only if it is properly introduced by ‘//’](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlparse:~:text=urlparse%20recognizes%20a%20netloc%20only%20if%20it%20is%20properly%20introduced%20by%20%E2%80%98//%E2%80%99)
So, if a user has supplied (e.g., in "app.yaml") "APPENGINE_URL: PROJECT_ID.uc.r.appspot.com", instead of "APPENGINE_URL: https://PROJECT_ID.uc.r.appspot.com", then you'll end up w/ what seems to me a weird/unintended result (3 slashes):
`$ python -c 'from urllib.parse import urlparse; print( urlparse( "PROJECT_ID.uc.r.appspot.com", "https").geturl())'`
`https:///PROJECT_ID.uc.r.appspot.com`
And, of course, the documentation example does already specify use of "https" (ref. [APPENGINE_URL: https://PROJECT_ID.uc.r.appspot.com](https://cloud.google.com/python/django/appengine#:~:text=...%0Aenv_variables%3A-,APPENGINE_URL%3A%20https%3A//PROJECT_ID.uc.r.appspot.com,-Upload%20your%20configuration))
My suggestion is to remove the line (& associated comment) entirely.

lesteral:
@glasnt - in addition to fix #7516 proposed by @sakimyto, please consider the comments at specific lines here

Thank you for the helpful GCP example, BTW.

appengine/standard_python3/django/mysite/settings.py
glasnt: You're right here, I got the order wrong in this sample for which value needed the scheme and which didn't. Good catch!

appengine/standard_python3/django/mysite/settings.py
glasnt: > "urlparse recognizes a netloc only if it is properly introduced by ‘//’"

TIL! Not sure how this slipped in my testing.

Issue #4505, opened 2020-08-15
billing: clean up sample added in #4494


ace-n:
The following items need to be cleaned up in the `billing` directory once #4494 is merged:

1) Update the code to use GAPIC libraries.
2) Add tests similar to [the ones the Node.js sample uses](https://github.com/GoogleCloudPlatform/nodejs-docs-samples/tree/master/functions/billing/test).

meredithslota:
I'm reclassifying this as a `bug` to address the lack of tests.

busunkim96:
I've rewritten this to use the `google-cloud-billing` library in #7432, but the tests still need work.

@ace-n The Node.js samples look like they use a live project, is that correct? There were some warnings about removing billing accounts in the docs, but if the Node CI project is okay the python ones will probably be okay as well.

ace-n:
@busunkim96 you're correct; that project ID is `cdpe-functions-billing-test`.

NimishC:
Thank you, @busunkim96 !

busunkim96:
Closed by #7432.

PR #7432, opened 2022-02-05
Add tests for billing sample

busunkim96:
For #4505. 

Use `google-cloud-billing` to make calls to the Billing API.


busunkim96 (commented):
@ace-n Could you give `jenkins-and-travis@python-docs-samples-tests.iam.gserviceaccount.com` the same permissions as the Node account to  `cdpe-functions-billing-test`?

@parthea This is ready for a first pass review (the tests need to have permissions sorted out).

busunkim96:
This test now runs OK for me with my own user credentials, but fails with the service account (403 permission denied). I'm having trouble figuring out which credentials are needed (or if they can be attached to a service account).

billing/noxfile_config.py
```diff
@@ -0,0 +1,44 @@
...
+    # Only run the tests once so tests do not interfere with each other.
+    "ignored_versions": ["2.7", "3.6", "3.7", "3.8", "3.10"],
+    # Old samples are opted out of enforcing Python type hints
+    # All new samples should feature them
+    "enforce_type_hints": False,
+    # An envvar key for determining the project id to use. Change it
+    # to 'BUILD_SPECIFIC_GCLOUD_PROJECT' if you want to opt in using a
+    # build specific Cloud project. You can also use your own string
+    # to use your own Cloud project.
+    "gcloud_project_env": "FUNCTIONS_BILLING_PROJECT",
```
parthea, line 35: Can you help me locate where `FUNCTIONS_BILLING_PROJECT` is defined?  I see `BUILD_SPECIFIC_GCLOUD_PROJECT` [here](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/main/.kokoro/python3.9/common.cfg#L50) so I was expecting to see `FUNCTIONS_BILLING_PROJECT` in the same location.

billing/test_main.py
```diff
@@ -0,0 +1,85 @@
...
+    request = billing.GetProjectBillingInfoRequest(name=project_name)
+    project_billing_info = cloud_billing_client.get_project_billing_info(request)
+
+    yield PROJECT_ID
+
+    # Re-enable billing in teardown
+    request = billing.UpdateProjectBillingInfoRequest(
+        name=project_name,
+        project_billing_info=billing.ProjectBillingInfo(
+
```
parthea, line 43:
```suggestion
```

billing/test_main.py
```diff
@@ -0,0 +1,85 @@
...
+from main import stop_billing
+
+
+PROJECT_ID = google.auth.default()[1]
+
+
+@pytest.fixture
+def project() -> str:
+    # This test assumes billing is enabled on the project
```
parthea, line 30: Please add a comment to clarify the reason for not enabling billing at the start of the test

parthea (approved):
LGTM, add minor observations

billing/noxfile_config.py
busunkim96: I added it as a new environment variable to test-env.sh (still needs to be added to test-env.tmpl.sh). There's more explanation here: https://github.com/GoogleCloudPlatform/python-docs-samples/blob/main/AUTHORING_GUIDE.md#test-environment-setup

billing/test_main.py
busunkim96: Fair point, if I can figure out if a service account can actually modify the billing account, I'll add the billing account as another secret and always enable bliling at the start.

busunkim96:
@parthea I've made significant changes, could you please re-review? 

@NimishC or @tod-hilton please also review from the Billing API side.

I originally wrote the tests to disable billing on the live project, but ended up using mocks instead. The service account has enough permissions to _disable billing_, but not to _enable billing_. I'm not sure it's possible to give the service account permissions on the billing account (nor am I sure if it's a good idea) so I switched to mocks.

https://cloud.google.com/billing/docs/how-to/modify-project

Permissions required to enable billing:

> To link a project to a Cloud Billing account, you need`resourcemanager.projects.createBillingAssignment` on the project AND `billing.resourceAssociations.create` on the Cloud Billing account.

Permissions required to disable billing:

> To unlink a project from a Cloud Billing account, you need `resourcemanager.projects.deleteBillingAssignment` on the project OR `billing.resourceAssociations.delete` on the Cloud Billing account.

tod-hilton (approved):

parthea (approved):

PR #7559, opened 2022-03-08
fix(serverless): clean up ID token example

ace-n:
Partial fix for b/220462035


auth/service-to-service/auth.py
```diff
@@ -16,33 +16,42 @@
...
-    ID token obtained from the google-auth client library.
+    by authenticating with the ID token obtained from the google-auth client library
+    using the specified audience value.
     """
 
+    # [END functions_bearer_token]
+    """
+    # Cloud Run uses your service's hostname as the `audience` value
+    audience = 'https://my-cloud-run-service.run.app/'
+    """
```
busunkim96, line 38: If folks are expected to uncomment and use `audience`, I think individually commented lines better match the style used in other Python samples.

```suggestion
    # Cloud Run uses your service's hostname as the `audience` value
    # audience = 'https://my-cloud-run-service.run.app/'
```

busunkim96 (approved):

arithmetic1728 (approved):

PR #7602, opened 2022-03-15
refactor(storagetransfer): Use public `transport` rather than `_transport`

danielbankhead:
Minor, but makes the sample a bit more 'official' looking

Ref: https://github.com/googleapis/python-storage-transfer/issues/59#issuecomment-1034342559


dandhlee (approved):
LGTM. That was an interesting read 🤓

cojenco (approved):
Thanks Dan!

PR #7751, opened 2022-04-04
Fix: Change variable name for consistency

engelke:
b/224052838

Listener report showed confusion with variable name, which isn't used in any other languages but Python and isn't consistent with comments.


dandhlee (approved):
LGTM.

I still see references to `service_url` in the test file: https://github.com/GoogleCloudPlatform/python-docs-samples/blob/main/auth/service-to-service/auth_test.py; although it isn't user-facing, should that be changed as well?

m-strzelczyk (approved):

PR #7801, opened 2022-04-18
fix(appengine/static-service): fix non-extension'ed file handling

ace-n:
This was tested locally, but **not on App Engine**.

(It's a simple enough change, so I figured it didn't require a full deployment **if** we know the underlying sample works.)


engelke (approved):

parthea (approved):

nicain (approved):
LGTM

PR #7889, opened 2022-05-12
feat(functions/slack): add GCF v2 decorators

ace-n:
✅  I've **confirmed** this works on GCF itself.


functions/slack/main.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2018, Google, LLC.
+# Copyright 2022 Google LLC.
```
dandhlee, line 1:
```suggestion
# Copyright 2018 Google LLC.
```
Please do not update the years! It should reflect when the file was first introduced.

functions/slack/main.py
```diff
@@ -89,6 +90,7 @@ def make_search_request(query):
 
 
 # [START functions_slack_search]
+@functions_framework.http
```
dandhlee, line 93: Sorry I'm not too informed about this decorator, is this for good practice? If this is a recommended approach, and this is going to show up on the Sample Browser as well, I think it might be good to be able to point users towards some sort of documentation that explains why using this might be beneficial!

dandhlee (changes requested):

functions/slack/main.py
iennae: I don't think it was documented; This PR https://github.com/GoogleCloudPlatform/functions-framework-nodejs/pull/347 is the most detailed bit I could find. Thoughts on whether we should point to it?

dandhlee (approved):
LGTM.

iennae (approved):

PR #8025, opened 2022-05-31
feat(cloudsql/sql-server): update SQL Server to v2 samples

jackwotherspoon:
Updating SQL Server connectivity samples to v2.


cloud-sql/sql-server/sqlalchemy/app.flexible.yaml
```diff
@@ -1,4 +1,4 @@
-# Copyright 2020 Google LLC
+# Copyright 2018 Google LLC
```
enocom, line 1: Should this be changing?

enocom (approved):

PR #8090, opened 2022-06-14
[bug] fix discontinued message string

zhichaog-mineral:
Original codes are two statements in two lines. They should be concatenated to one message string.

### DIFF
--- a/run/jobs/main.py
+++ b/run/jobs/main.py
@@ -61,8 +61,8 @@ if __name__ == "__main__":
     try:
         main(SLEEP_MS, FAIL_RATE)
     except Exception as err:
-        message = f"Task #{TASK_INDEX}, "
-        + f"Attempt #{TASK_ATTEMPT} failed: {str(err)}"
+        message = f"Task #{TASK_INDEX}, " \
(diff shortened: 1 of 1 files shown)


ace-n:
@zhichaog-mineral thanks for the PR!

The sample as-is passes {tests, style checks}, but maybe there's something I'm missing here. 🙂 @loferris can you weigh in?

dandhlee:
I think the portion that is being edited here is actually not tested, hence the test passes and lint unfortunately... does not picks it up? Lint not picking it up is weird.

In terms of Python syntax it should not work and this PR should fix it, however without a test this issue wouldn't have been caught unless someone ran into an issue and ran into the syntax that's not tested.

@loferris wdyt?

ace-n:
@dandhlee maybe it's worth reporting [here](https://github.com/PyCQA/pycodestyle/issues)?

dandhlee:
Our linter runs `flake` and I've tried running flake and still wasn't able to catch the issue 🙃

averikitsch (approved):

dandhlee (approved):
Approving for unblocking folks who'd run into undesired issues. I'll file an issue to seek coverage for this bit.

PR #8121, opened 2022-06-24
Kaiyang intern 2022

kaiyang-code:
## Description

@leahecole 

## Checklist
- [x] **Tests** pass:   `nox -s py-3.8` (see [Test Environment Setup](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/main/AUTHORING_GUIDE.md#test-environment-setup))


composer/workflows/s3togcsoperator_tutorial_test.py
```diff
@@ -0,0 +1,27 @@
...
+def test_dag_import():
+    """Test that the DAG file can be successfully imported.
+    This tests that the DAG can be parsed, but does not run it in an Airflow
+    environment. This is a recommended confidence check by the official Airflow
+    docs: https://airflow.incubator.apache.org/tutorial.html#testing
+    """
+
+    from . import s3togcsoperator_tutorial as module
+
+    internal_unit_testing.assert_has_valid_dag(module)
```
dandhlee, line 27:
```suggestion
    import s3togcsoperator_tutorial

    internal_unit_testing.assert_has_valid_dag(s3togcsoperator_tutorial)
```

`import x as y` should only be used if the short form `y` can still capture the original but in the short form, for example `import tensorflow as tf`. Renaming is not a good style.

composer/workflows/s3togcsoperator_tutorial_test.py
```diff
@@ -0,0 +1,27 @@
...
+
+import internal_unit_testing
+
+
+def test_dag_import():
+    """Test that the DAG file can be successfully imported.
+    This tests that the DAG can be parsed, but does not run it in an Airflow
+    environment. This is a recommended confidence check by the official Airflow
+    docs: https://airflow.incubator.apache.org/tutorial.html#testing
+    """
```
dandhlee, line 23:
```suggestion
    """Test that the DAG file can be successfully imported.
    
    This tests that the DAG can be parsed, but does not run it in an Airflow
    environment. This is a recommended confidence check by the official Airflow
    docs: https://airflow.incubator.apache.org/tutorial.html#testing
    """
```

The first line of the docstring should capture a one line summary, followed by a newline, then more descriptions as needed.

composer/workflows/s3togcsoperator_tutorial.py
```diff
@@ -0,0 +1,149 @@
...
+        bucket=S3_BUCKET_NAME,
+        gcp_conn_id="google_cloud_default",
+        aws_conn_id="aws_s3_connection",
+        dest_gcs=f"gs://{BUCKET_NAME}",
+    )
+
+    create_batch = dataproc.DataprocCreateBatchOperator(
+        task_id="create_batch",
+        project_id=PROJECT_NAME,
+        region="{{ var.value.gce_region }}",
```
dandhlee, line 92: We're using a lot of mix of constants, predefined values and dynamic values. A mix of constants (i.e. `task_id="create_batch"`) and predefined values (i.e. `project_id=PROJECT_NAME`) seem good, we should stick to this style being consistent and also define values like `region="{{ var.value.gce_region }}"`.

composer/workflows/s3togcsoperator_tutorial.py
```diff
@@ -0,0 +1,149 @@
...
+            {"name": "Date", "type": "DATE"},
+            {"name": "Holiday", "type": "STRING"},
+        ],
+        skip_leading_rows=1,
+        write_disposition="WRITE_TRUNCATE"
+    )
+
+    with TaskGroup("join_bq_datasets") as bq_join_group:
+
+        for year in range(1997, 2022):
```
dandhlee, line 113: Any reason as to why `1997` or `2022` is special? Naming these as constants and/or comments should help clear this:

```
YEAR_TO_START = 1997
CURRENT_YEAR = datetime.datetime.now().year

...

# Run the dataset from the predefined years to current year.
for year in range(YEAR_TO_START, CURRENT_YEAR):
```

composer/workflows/s3togcsoperator_tutorial.py
```diff
@@ -0,0 +1,149 @@
...
+            # BigQuery configs
+            BQ_DATASET_NAME = f"bigquery-public-data.ghcn_d.ghcnd_{str(year)}"
+            BQ_DESTINATION_TABLE_NAME = "holidays_weather_joined"
+            # Specifically query a Chicago weather station
+            WEATHER_HOLIDAYS_JOIN_QUERY = f"""
+            SELECT Holidays.Date, Holiday, id, element, value
+            FROM `{PROJECT_NAME}.holiday_weather.holidays` AS Holidays
+            JOIN (SELECT id, date, element, value FROM {BQ_DATASET_NAME} AS Table WHERE Table.element="TMAX" AND Table.id="USW00094846") AS Weather
```
dandhlee, line 121: beware of line length, please break this line over multiple lines for readability.

composer/workflows/s3togcsoperator_tutorial.py
```diff
@@ -0,0 +1,149 @@
...
+    )
+
+    with TaskGroup("join_bq_datasets") as bq_join_group:
+
+        for year in range(1997, 2022):
+            # BigQuery configs
+            BQ_DATASET_NAME = f"bigquery-public-data.ghcn_d.ghcnd_{str(year)}"
+            BQ_DESTINATION_TABLE_NAME = "holidays_weather_joined"
+            # Specifically query a Chicago weather station
+            WEATHER_HOLIDAYS_JOIN_QUERY = f"""
```
dandhlee, line 118: nit: `f-strings` are okay for multi lines, but using `string.format()` sometimes makes it clearer to see what variables are getting replaced in a multi-line text.

composer/workflows/s3togcsoperator_tutorial.py
```diff
@@ -0,0 +1,149 @@
...
+            {"name": "Holiday", "type": "STRING"},
+        ],
+        skip_leading_rows=1,
+        write_disposition="WRITE_TRUNCATE"
+    )
+
+    with TaskGroup("join_bq_datasets") as bq_join_group:
+
+        for year in range(1997, 2022):
+            # BigQuery configs
```
dandhlee, line 114: nit: preserve comments to explain why the code behaves in a certain way or to capture unique information that is not obvious at first glance, rather than re-iterating what's shown in the code.

composer/workflows/s3togcsoperator_tutorial.py
```diff
@@ -0,0 +1,149 @@
...
+            ON Holidays.Date = Weather.Date;
+            """
+
+            # for demo purposes we are using WRITE_APPEND
+            # but if you run the DAG repeatedly it will continue to append
+            # Your use case may be different, see the Job docs
+            # https://cloud.google.com/bigquery/docs/reference/rest/v2/Job
+            # for alternative values for the writeDisposition
+            # or consider using partitioned tables
+            # https://cloud.google.com/bigquery/docs/partitioned-tables
```
dandhlee, line 131:
```suggestion
            # For demo purposes we are using WRITE_APPEND
            # but if you run the DAG repeatedly it will continue to append.
            # Your use case may be different, see the Job docs
            # https://cloud.google.com/bigquery/docs/reference/rest/v2/Job
            # for alternative values for the writeDisposition
            # or consider using partitioned tables
            # https://cloud.google.com/bigquery/docs/partitioned-tables
```

composer/workflows/s3togcsoperator_tutorial.py
```diff
@@ -0,0 +1,149 @@
...
+                            "datasetId": BQ_DESTINATION_DATASET_NAME,
+                            "tableId": BQ_DESTINATION_TABLE_NAME,
+                        },
+                        "writeDisposition": "WRITE_APPEND",
+                    }
+                },
+                location="US",
+            )
+
+        s3_to_gcs_op >> load_external_dataset >> bq_join_group >> create_batch
```
dandhlee, line 149: Could you add a comment to highlight what the user might be able to expect? Will it print a result, update your DAG and be able to verify somewhere or...? Apologies if this is obvious as I'm not a Composer expert 😅

dandhlee (changes requested):
Reviewing for Python style

composer/workflows/s3togcsoperator_tutorial.py
leahecole: These are the only dates found in one of the datasets which is why we chose them - they will stay fixed instead of the current year. This is also identical to [this line](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/f2b85fb2e5fabcdfc46c2e7fd407db364f7acdde/composer/2022_airflow_summit/data_analytics_dag.py#L104) - my original intent was to not change anything that was identical to code that had already been approved @dandhlee, would you want to see this constant change in both files?

composer/workflows/s3togcsoperator_tutorial.py
leahecole: Do you still want to see a comment if this will have an accompanying tutorial on c.g.c?

composer/workflows/s3togcsoperator_tutorial.py
dandhlee: Hmm I can see the benefit of running this for fixed year. Let's stick with the proposed set years.

composer/workflows/s3togcsoperator_tutorial.py
leahecole: This is pretty common in Airflow to have predefined values and dynamic values. Dynamic values (which are populated with the jinja templates) can be shared across DAGs. I tend to use constants with dynamic values that are repeated over and over within a single Dag file (like the project_id one). Can you clarify what you mean about defining the region? If it's specifying a region to be prescriptive, I disagree, because this Airflow variable is used in multiple DAG files in the tutorial, so it's best Airflow practice to not have it be static here. If you mean something else though lmk! I am open to it being turned into a constant just to make it look nicer (and if Airflow considers it valid) if that is an okay compromise.

composer/workflows/s3togcsoperator_tutorial.py
leahecole: This matches with another DAGfile that was already approved - I'd ideally like to stay with f-strings because that's what we've been enforcing consistently since we removed the Python 2 requirement in our samples and imho it's the most "modern" approach

composer/workflows/s3togcsoperator_tutorial.py
leahecole: It will have a region tag, we're just not at that stage of the program yet.

composer/workflows/s3togcsoperator_tutorial_test.py
kaiyang-code: `nox -s py-3.8` won't pass unless we use ' from . import s3togcsoperator_tutorial' instead of `import s3togcsoperator_tutorial`

composer/workflows/s3togcsoperator_tutorial.py
kaiyang-code: Done!

composer/workflows/s3togcsoperator_tutorial.py
kaiyang-code: Marked as resolved according to @leahecole . Feel feel to change back if further action needed.

composer/workflows/s3togcsoperator_tutorial.py
kaiyang-code: Done!

composer/workflows/s3togcsoperator_tutorial_test.py
kaiyang-code: Done!

composer/workflows/s3togcsoperator_tutorial.py
kaiyang-code: Done!

composer/workflows/s3togcsoperator_tutorial.py
dandhlee: Ah I meant initializing a constant for `{{ var.value.gce_region }}`, like

```
PROJECT_NAME = "{{var.value.gcp_project}}"
REGION = "{{ var.value.gce_region }}"

...

    create_batch = dataproc.DataprocCreateBatchOperator(
        task_id="create_batch",
        project_id=PROJECT_NAME,
        region=REGION,
```

composer/workflows/s3togcsoperator_tutorial_test.py
dandhlee: Ah that makes sense, apologies for that!

dandhlee (commented):
LGTM, one minor comment then I'll happily stamp 👍

composer/workflows/s3togcsoperator_tutorial.py
kaiyang-code: Done!

dandhlee (approved):
Stampy :3

PR #8145, opened 2022-06-30
fix(appengine): hardcoded PORT -> env var

ace-n:
Fixes b/237096040

### DIFF
--- a/appengine/standard_python3/custom-server/app.yaml
+++ b/appengine/standard_python3/custom-server/app.yaml
@@ -15,6 +15,6 @@
 # [START gae_python38_custom_runtime]
 # [START gae_python3_custom_runtime]
 runtime: python39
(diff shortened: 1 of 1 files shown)


dandhlee (approved):

iennae (approved):
LGTM

iennae:
I'm wondering whether this should have a comment inline with the code as to the expectation for the variable. Thoughts @dandhlee ?

dandhlee:
I think it's better to leave this one as-is, and add comments elsewhere. This code seems to show up embedded in this doc: https://cloud.google.com/appengine/docs/standard/python3/runtime.md, which I think comment there would help more than an inline comment here and then the doc looking slightly weird.

iennae:
Thanks so much @dandhlee :)

PR #8250, opened 2022-08-09
chore(cloudsql): add consistent region tags

jackwotherspoon:
This commit adds region tags that match the other languages. A future
commit will remove the old region tags.


enocom (approved):

dandhlee (changes requested):
Could you file an issue, then add a TODO(link-to-issue) above the older region tags to be deleted?

For example:
```
TODO (https://github.com/GoogleCloudPlatform/python-docs-samples/issues/8000): delete older region tags
# [START cloud_sql_mysql_sqlalchemy_connect_tcp]
```

dandhlee (approved):
LGTM. Thank you!

PR #8290, opened 2022-08-26
data analytics tutorial expansion project

leahecole:
This adds the more polished code from our intern's dataproc focused expansion project. This is an expansion of [this tutorial](https://cloud.google.com/composer/docs/data-analytics-googlecloud) that uses a similar but slightly altered DAG and a very different (more complex) Dataproc job to run. This PR adds a directory with

- A new DAG + test
- A new Dataproc job and test
- A helper script (and test) that converts data to the format it needs to be in to run the tutorial
- The `.txt` file of the data (not yet processed) and the processed `.csv` of it
- A README that will be linked from the original tutorial with a note for users that may be interested in a more complex Dataproc job
It also adds a line to header checker bot to ignore the license check on a text file

## Checklist
- [x] I have followed [Sample Guidelines from AUTHORING_GUIDE.MD](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/main/AUTHORING_GUIDE.md)
- [x] README is updated to include [all relevant information](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/main/AUTHORING_GUIDE.md#readme-file)
- [x] **Tests** pass:   `nox -s py-3.9` (see [Test Environment Setup](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/main/AUTHORING_GUIDE.md#test-environment-setup))
- [x] **Lint** pass:   `nox -s lint` (see [Test Environment Setup](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/main/AUTHORING_GUIDE.md#test-environment-setup))
- [ ] These samples need a new **API enabled** in testing projects to pass (let us know which ones)
- [ ] These samples need a new/updated **env vars** in testing projects set to pass (let us know which ones)
- [x] Please **merge** this PR for me once it is approved.
- [ ] This sample adds a new sample directory, and I updated the [CODEOWNERS file](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/main/.github/CODEOWNERS) with the codeowners for this sample

PR #8296, opened 2022-08-30
Update header checker

leahecole:
in anticipation of #8290

### DIFF
--- a/.github/header-checker-lint.yml
+++ b/.github/header-checker-lint.yml
@@ -15,6 +15,7 @@ ignoreFiles:
   - "**/constraints.txt"
   - "**/constraints-test.txt"
   - "**/apt.txt"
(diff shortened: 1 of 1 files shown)


dandhlee (approved):
LGTM. Does this need to be a separate PR? (just curious)

leahecole:
@dandhlee I think so - it doesn't seem to work as part of https://github.com/GoogleCloudPlatform/python-docs-samples/pull/8290 because I think it runs using whatever is in main

dandhlee:
Gotcha. Yuck :(

Admin merging!

leahecole:
i'm an idiot, the reason it wasn't working was because I made a typo. Thanks for merging this anyway and good on you for correctly being skeptical 😆

PR #8338, opened 2022-09-20
Add sqlserver private IP SSL app engine connection support to cloud sql sample app. 

jsimonweb:


cloud-sql/sql-server/sqlalchemy/connect_connector.py
```diff
@@ -45,6 +51,9 @@ def getconn() -> pytds.Connection:
             user=db_user,
             password=db_pass,
             db=db_name,
+            # [START_EXCLUDE]
+            cafile=db_root_cert,
```
jackwotherspoon, line 55:
```suggestion
            cafile=db_root_cert,
            validate_host=False,
```
does this work without setting `validate_host=False`?

cloud-sql/sql-server/sqlalchemy/connect_connector.py
```diff
@@ -35,6 +35,12 @@ def connect_with_connector() -> sqlalchemy.engine.base.Engine:
     db_name = os.environ["DB_NAME"]  # e.g. 'my-database'
 
     ip_type = IPTypes.PRIVATE if os.environ.get("PRIVATE_IP") else IPTypes.PUBLIC
+    # [START_EXCLUDE]
```
jackwotherspoon, line 38: instead of excludes should we provide a new region tag if this code is to be in a sample/quickstart?

jsimonweb:
Although this currently works as is, we've decided to update the existing `cloud-sql/sql-server/sqlalchemy/connect_tcp.py` file to enable this functionality there instead.

PR #8340, opened 2022-09-21
Add Media CDN Dual Token - Sign Token code snippet

msampathkumar:
## Description

Fixes # http://b/244355824

## Checklist
- [x] README is updated to include [all relevant information](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/main/AUTHORING_GUIDE.md#readme-file)


media_cdn/snippets.py
```diff
@@ -23,8 +23,12 @@
 
 # [START mediacdn_sign_url]
 # [START mediacdn_sign_cookie]
+# [START mediacdn_sign_token]
 import base64
 import datetime
+import hashlib
+import hmac
```
rsamborski, line 30: @msampathkumar These two new imports (hashlib and hmac) are only used by 'mediacd_sign_token' sample. I suggest to exclude them from other region tags to make sure samples published to docs include only necessary imports. WDYT?

media_cdn/snippets.py
```diff
@@ -149,3 +153,38 @@ def sign_cookie(url_prefix: str, key_name: str, base64_key: str, expiration_time
...
+
+    if not expiration_time:
+        expiration_time = datetime.datetime.now() + datetime.timedelta(hours=1)
+    epoch_duration = int(
+        (expiration_time - datetime.datetime.utcfromtimestamp(0)).total_seconds()
+    )
+    output += b"~Expires=" + str(epoch_duration).encode("utf-8")
+    key = base64.standard_b64decode(base64_key)
+    algo = algo.lower()
+    if algo == "sha1":
```
rsamborski, line 182: Should we support SHA1 taking into account it is [no longer considered secure](https://en.wikipedia.org/wiki/SHA-1#:~:text=Since%202005%2C%20SHA%2D1%20has,against%20SHA%2D1%20are%20practical.)? Is Media CDN supporting SHA1?
I would suggest to default to SHA256.

media_cdn/snippets.py
msampathkumar: LGTM. I shall update it ASAP.

media_cdn/snippets.py
msampathkumar: Thanks for sharing the wiki link.

In Dual Token, we are using a long & short token and here we are using SHA1 along with a SHA256 secret and HMAC. I believe this is only to support any a legacy customers softwares. But, let me also check with product team if I have any new information.

msampathkumar: Hi, I have confirmed from the product team.
1. SHA1 is still safe when used along with HMAC
2. SHA1 is supported mainly is for multi-CDN compatibility. Some CDNs only support HMAC-SHA1.

rsamborski: SGTM. Thanks for checking.

media_cdn/snippets.py
```diff
@@ -149,3 +157,38 @@ def sign_cookie(url_prefix: str, key_name: str, base64_key: str, expiration_time
...
```
justin-mp, line 186: signed_tokens can also use the Ed25519 signature algorithm like other signed request modes.  In fact, we encourage people to use Ed25519 algorithm with tokens.  (We only support SHA-1 and SHA-256 for compatibility with other CDNs.)

This sample needs to support Ed25519 signatures.  The differences:
* Ed25519 uses the field `Signature` (instead of `hmac` for HMACs)
* The value is URL-safe base64 encoded (instead of hex direct for HMACs).
You can reuse the digest and signature code from the other code samples.

media_cdn/snippets.py
```diff
@@ -149,3 +157,38 @@ def sign_cookie(url_prefix: str, key_name: str, base64_key: str, expiration_time
...
+    """
+    output = b"URLPrefix=" + base64.standard_b64encode(url_prefix.encode("utf-8"))
+
+    if not expiration_time:
+        expiration_time = datetime.datetime.now() + datetime.timedelta(hours=1)
+    epoch_duration = int(
+        (expiration_time - datetime.datetime.utcfromtimestamp(0)).total_seconds()
+    )
+    output += b"~Expires=" + str(epoch_duration).encode("utf-8")
+    key = base64.standard_b64decode(base64_key)
```
justin-mp, line 184: Not that it matters for our protocol (since this is just for reading the key from the file system), but the other samples use urlsafe_b64decode (e.g., see line 107)

media_cdn/snippets.py
```diff
@@ -149,3 +157,38 @@ def sign_cookie(url_prefix: str, key_name: str, base64_key: str, expiration_time
...
+        url_prefix: URL prefix to sign as a string.
+        base64_key: Secret key as a base64 encoded string.
+        algo: Algorithm can be either `SHA1` or `SHA256`.
+        expiration_time: Expiration time as a UTC datetime object.
+
+    Returns:
+        Returns the Signed URL appended with the query parameters based on the
+        specified URL prefix and configuration.
+    """
+    output = b"URLPrefix=" + base64.standard_b64encode(url_prefix.encode("utf-8"))
```
justin-mp, line 176: This should be base64.urlsafe_b64decode like the other code samples.

justin-mp (commented):
Is there a plan to add other fields?  In particular, `FullPath` might be worth an example.

media_cdn/snippets.py
```diff
@@ -74,7 +82,7 @@ def sign_url(url: str, key_name: str, base64_key: str, expiration_time: datetime
     return signed_url
 
 
-def sign_url_prefix(url: str, url_prefix, key_name: str, base64_key: str, expiration_time: datetime.datetime) -> str:
+def sign_url_prefix(url: str, url_prefix: str, key_name: str, base64_key: str, expiration_time: datetime.datetime) -> str:
```
dandhlee, line 85: beware of line length.
```suggestion
def sign_url_prefix(
    url: str,
    url_prefix: str,
    key_name: str,
    base64_key: str,
    expiration_time: datetime.datetime,
) -> str:
```

media_cdn/snippets.py
```diff
@@ -149,3 +157,38 @@ def sign_cookie(url_prefix: str, key_name: str, base64_key: str, expiration_time
             policy=policy, signature=signature)
     return signed_policy
 # [END mediacdn_sign_cookie]
+
+
+# [START mediacdn_sign_token]
+def sign_token(url_prefix: str, base64_key: bytes, algo: str, expiration_time: datetime.datetime = None) -> bytes:
```
dandhlee, line 163: Ditto:

```suggestion
def sign_token(
    url_prefix: str,
    base64_key: bytes,
    algo: str,
    expiration_time: datetime.datetime = None,
) -> bytes:
```

media_cdn/snippets.py
```diff
@@ -149,3 +157,38 @@ def sign_cookie(url_prefix: str, key_name: str, base64_key: str, expiration_time
...
+    """Gets the Signed URL Suffix string for the Media CDN' Short token URL requests.
+
+    Args:
+        url_prefix: URL prefix to sign as a string.
+        base64_key: Secret key as a base64 encoded string.
+        algo: Algorithm can be either `SHA1` or `SHA256`.
```
dandhlee, line 169: nit: consider `encryption_algorithm` or a similarly more descriptive name.

media_cdn/snippets.py
```diff
@@ -149,3 +157,38 @@ def sign_cookie(url_prefix: str, key_name: str, base64_key: str, expiration_time
...
+        expiration_time: Expiration time as a UTC datetime object.
+
+    Returns:
+        Returns the Signed URL appended with the query parameters based on the
```
dandhlee, line 173:
```suggestion
        The Signed URL appended with the query parameters based on the
```

media_cdn/snippets.py
msampathkumar: Understood! Will update it in the next commit.

media_cdn/snippets.py
```diff
@@ -130,22 +160,109 @@ def sign_cookie(url_prefix: str, key_name: str, base64_key: str, expiration_time
...
+
+    Returns:
+        The Signed URL appended with the query parameters based on the
+        specified URL prefix and configuration.
+    """
+
+    decoded_key = base64.urlsafe_b64decode(base64_key)
+    algo = signature_algorithm.lower()
+
+    to_sign = b""
```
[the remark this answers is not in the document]
msampathkumar: Hi, so `to_sign` specifically defined for `FullPath` & Ed25519 only.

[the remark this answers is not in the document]
justin-mp: I'm still confused.

The parameters in the token and the algorithm are _independent_.  It's possible to use `FullPath` with HMACs just as you could use URLPrefix or PathGlobs with Ed25519.

justin-mp (changes requested):
I think there's something not quite right about the token code.

For tokens, you essentially want to build up two strings: the `output` string and the `to_sign` string.  You then get the signature by running `sign(to_sign)` or `hmac.new(to_sign)` and append that signature value to the end of `output`.

For most fields, what you append to `output` and `to_sign` is the same. `FullPath` and `Headers` have different things in `to_sign` versus `output`.

The most straightforward way to write this is to just to update `output` and `to_sign` for every parameter you support.  It's not the most efficient way, but it's easier to read, which sounds like a good goal for documentation samples.

For example:

```
output = []
to_sign = []
if url_prefix is not None:
    part = f"URLPrefix={url_prefix}"
    output.append(part)
    to_sign.append(part)
if full_path is not None:
    output.append("FullPath")
    to_sign.append("FullPath={full_path}")
...
if expires is not None:
    part = f"Expires={expires}"
    output.append(part)
    to_sign.append(part)
...
to_sign_str = "~".join(to_sign)
signature = # code that generates a signature from to_sign_str
output_str = "~".join(output)
# append signature to output_str
return output_str
```

dandhlee:
@msampathkumar Friendly ping on this PR!

msampathkumar:
Hi @dandhlee, Thank you for the ping

msampathkumar:
Working on a fresh PR for the same branch!

PR #8341, opened 2022-09-21
7491 - people and planet

leahecole:
I'm having a horrible time with timeouts in #7491 and figured I'd try splitting it up by product to find the culprit

### DIFF
--- a/people-and-planet-ai/geospatial-classification/serving_app/Dockerfile
+++ b/people-and-planet-ai/geospatial-classification/serving_app/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.9-slim
+FROM python:3.10-slim
 
 WORKDIR /app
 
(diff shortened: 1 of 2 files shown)


leahecole:
This is the landcover noxfile config - https://github.com/GoogleCloudPlatform/python-docs-samples/blob/main/people-and-planet-ai/land-cover-classification/noxfile_config.py we can see that it's the one that needs to be using 3.7 and somehow this change causes a timeout to happen

leahecole (commented):
@davidcavazos do you have any insights as to why this change would cause the land cover classification sample to timeout? Is it crucial that it be updated to 3.10? If not and if keeping it at 3.9 prevents a mysterious timeout from happening, should we do that?

davidcavazos:
Everything with Beam and Dataflow must run on 3.9 since 3.10 is not officially supported.

leahecole:
Got it! closing this then

PR #8348, opened 2022-09-22
fix: Flaky failures

### LINKED ISSUE #8330: appengine.standard_python3.bundled-services.deferred.flask.main_test: test_upload_and_view failed
Note: #8241 was also for this test, but it was closed more than 10 days ago. So, I didn't mark it flaky.

----

commit: 11e9005835a6f260968ee9ba3198ef530d30af5b
buildURL: [Build Status](https://source.cloud.google.com/results/invocations/c7f48ff9-99d5-43d8-853a-af6af5e5dd0a), [Sponge](http://sponge2/c7f48ff9-99d5-43d8-853a-af6af5e5dd0a)
status: failed
<details><summary>Test output</summary><br><pre>Traceback (most recent call last):
  File "/workspace/appengine/standard_python3/bundled-services/deferred/flask/main_test.py", line 100, in test_upload_and_view
    assert response.text == "30"
AssertionError: assert '20' == '30'
  - 30
  + 20</pre></details>

engelke:
## Description

Fixes #8330 #8347 #8339 

Logs show failures are due to datastore fetch timeouts. That shouldn't be happening, but retries should fix it. Also added an automatic timestamp and will create a policy to regularly delete old data to prevent database from growing too much.

### DIFF
--- a/appengine/standard_python3/bundled-services/deferred/django/main.py
+++ b/appengine/standard_python3/bundled-services/deferred/django/main.py
@@ -28,6 +28,7 @@ my_key = os.environ.get("GAE_VERSION", "Missing")
 
 class Counter(ndb.Model):
     count = ndb.IntegerProperty(indexed=False)
+    updated_at = ndb.DateTimeProperty(auto_now=True)
 
 
 def do_something_later(key, amount):
--- a/appengine/standard_python3/bundled-services/deferred/django/main.py
+++ b/appengine/standard_python3/bundled-services/deferred/django/main.py
@@ -32,9 +32,9 @@ class Counter(ndb.Model):
 
 
 def do_something_later(key, amount):
-    entity = Counter.get_or_insert(key, count=0)
+    entity = Counter.get_or_insert(key, count=0, retries=2)
     entity.count += amount
-    entity.put()
+    entity.put(retries=2)
 
 
 def increment_counter(request):
@@ -51,7 +51,7 @@ def increment_counter(request):
 
 
 def view_counter(request):
-    counter = Counter.get_or_insert(my_key, count=0)
+    counter = Counter.get_or_insert(my_key, count=0, retries=2)
     return HttpResponse(str(counter.count))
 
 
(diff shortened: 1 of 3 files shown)


leahecole (changes requested):
Instead of doing a backoff on a generic `Exception` could you please change it to the specific exception that is thrown? I'm trying to load the issues to see which one it is and I'll add it to suggestion mode when I do but GH is being slow

leahecole (changes requested):
Instead of doing a backoff on a generic `Exception` could you please change it to the specific exception that is thrown? I'm trying to load the issues to see which one it is and I'll add it to suggestion mode when I do but GH is being slow

leahecole (changes requested):
Instead of doing a backoff on a generic `Exception` could you please change it to the specific exception that is thrown? I'm trying to load the issues to see which one it is and I'll add it to suggestion mode when I do but GH is being slow

engelke:
I'd prefer to leave it general, as the rare flaky failures have come from a variety of causes - timeouts, reachability, and so on.

kurtisvg:
https://github.com/GoogleCloudPlatform/python-docs-samples/issues/8330 looks like it was an assertionerror:
```
main_test.test_upload_and_viewcontent_copy | - | 147.208s
-- | -- | --
Traceback (most recent call last):  File "/workspace/appengine/standard_python3/bundled-services/deferred/flask/main_test.py", line 100, in test_upload_and_view    assert response.text == "30"AssertionError: assert '20' == '30'  - 30  + 20
```

Why do we think a retry is going to help? It seems like maybe the sample is wrong.

engelke:
The sample isn't wrong. The tests passes almost every time. (Don't look at the tests for the first commits today, where I was trying various ways to deal with failures; look at the tests that triggered flaky-bot.)

This behavior has been going on for weeks and weeks. No changes to the samples, and suddenly the test fails once.

appengine/standard_python3/bundled-services/deferred/django/main.py
```diff
@@ -28,8 +29,10 @@
 
 class Counter(ndb.Model):
     count = ndb.IntegerProperty(indexed=False)
+    updated_at = ndb.DateTimeProperty(auto_now=True)
```
dandhlee, line 32: Just curious, how does having the timestamp show will remove them from the database? Or is that planned as TODO?

engelke: If there's a timestamp, I can have a policy remove them after a certain time. But I'm not going to move forward with this PR.

engelke:
It's been a while and we haven't had any failures. Failures are very rare, and it's hard to know whether this would help or not with them, but it would make things more complicated. So I'm going to close this unmerged and close the flaky issues for now, too.

PR #8349, opened 2022-09-22
Update Cloud SQL samples for viewing app on linux GCE VM.

jsimonweb:
Update the Cloud SQL sample apps so that when they are run:

`* Running on http://127.0.0.1:8080`

becomes

```
* Running on all addresses (0.0.0.0)
* Running on http://127.0.0.1:8080
```
which enables viewing of the sample app (running on a GCE Linux VM) in a browser on your local computer.

### DIFF
--- a/cloud-sql/mysql/sqlalchemy/app.py
+++ b/cloud-sql/mysql/sqlalchemy/app.py
@@ -158,4 +158,4 @@ def save_vote(db: sqlalchemy.engine.base.Engine, team: str) -> Response:
 
 
 if __name__ == "__main__":
-    app.run(host="127.0.0.1", port=8080, debug=True)
+    app.run(host="0.0.0.0", port=8080, debug=True)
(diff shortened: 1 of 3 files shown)


kurtisvg (changes requested):
Why doesn't 127.0.0.1 work for GCE?

jsimonweb:
>Why doesn't 127.0.0.1 work for GCE?

The Cloud SQL sample app runs as expected on GCE VMs using 127.0.0.1
The issue is that when the sample app is running on a GCE VM using 127.0.01, attempting to load the sample app in a browser using `http://[GCE-VM-external-ip]:8080` on a local computer, fails with the error: ERR_CONNECTION_REFUSED "This site can’t be reached".

After updating the `app.py` file as is done in this PR, the running sample app loads in a browser on a local computer as expected.

An alternative to this PR would be to instruct developers attempting to use their local computer's browser to view the the app running on a GCE VM, to make this PR's change manually as an additional step in the configuration of the sample app.

kurtisvg:
@jbsimon Can we use ssh / port forwarding to avoid this? That would be a more secure pattern: https://cloud.google.com/solutions/connecting-securely#port-forwarding-over-ssh

jsimonweb:
Will update related instructions in product documentation to use port forwarding as suggested.

PR #8353, opened 2022-09-23
Firestore use dict

Mariatta:
Update the snippet to use a dictionary instead of a custom class.
This doc appears in https://cloud.google.com/firestore/docs/samples/firestore-data-set-id-random-collection#firestore_data_set_id_random_collection-php

The sample is using custom City class which was not yet shown to the reader.

Also updated the sample to show how to retrieve the autogenerated id.

Fixes internal issue: 244038329


firestore/cloud-client/snippets.py
```diff
@@ -209,10 +209,14 @@ def add_data_with_id():
 
 
 def add_custom_class_generated_id():
-    db = firestore.Client()
+    db = firestore.Client(project="maratt")
```
kurtisvg, line 212: I think you need to revert this to get the sample to pass

kurtisvg (changes requested):

firestore/cloud-client/snippets.py
Mariatta: Ah yes, mistakenly committed that. I've reverted this. thanks

kurtisvg (approved):

kolea2 (approved):

dandhlee:
No changes to the region tag, ignoring snippet bot check. Admin merging with 2 proper reviews!

PR #8392, opened 2022-10-06
Extend wait time on this test, as it has proven flaky

### LINKED ISSUE #8355: run.image-processing.e2e_test: test_end_to_end failed
Note: #8144 was also for this test, but it was closed more than 10 days ago. So, I didn't mark it flaky.

----

commit: 76822fab361692ad13d8f0164b39c7b63fdc08cc
buildURL: [Build Status](https://source.cloud.google.com/results/invocations/3dfdaa81-bc0a-4a51-9a61-10f69119c09e), [Sponge](http://sponge2/3dfdaa81-bc0a-4a51-9a61-10f69119c09e)
status: failed
<details><summary>Test output</summary><br><pre>Traceback (most recent call last):
  File "/workspace/run/image-processing/e2e_test.py", line 242, in test_end_to_end
    assert len(output_blobs) > 0
AssertionError: assert 0 > 0
 +  where 0 = len([])</pre></details>

muncus:
Fixes #8355


WildSunLove (approved):

muncus:
@GoogleCloudPlatform/python-samples-reviewers - this review has been waiting a week, can someone please take a look?

run/image-processing/e2e_test.py
```diff
@@ -231,7 +231,7 @@ def test_end_to_end(input_bucket, output_bucket):
     # Wait for image processing to complete
     time.sleep(60)
 
-    for x in range(10):
+    for x in range(20):
```
dandhlee, line 234: Instead of just increasing the time, could you try setting a [retry](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/main/AUTHORING_GUIDE.md#retry-rpcs) if it can be used here?

muncus: the RPC does not fail, it just returns empty results until processing completes.
While we could do this with a linear backoff, I think the resulting code would actually be less readable, as it would require an extra function or inlining the check as a predicate.
What do you see as the benefits of using a retry here?

dandhlee: I'd say it scales _somewhat_ better and is more reliable than simply retrying x more times, if we're not sure why it's failing and being flaky. However, in this case it seems like trying with retries may not be the most suitable as this .. surprisingly doesn't fail or throw exceptions.

dandhlee (approved):
LGTM for experimentation.

PR #8395, opened 2022-10-06
test(storagetransfer): Isolate STS Test Credentials

danielbankhead:
Overview:
- Removes shared secret file in favor of secret manager (the secret manager secret now includes the relevant env variables)
- Utilizes `STS_AWS_SECRET_NAME` env variable for STS tests


.kokoro/tests/run_tests.sh
```diff
@@ -96,12 +96,6 @@ fi
 source ./testing/test-env.sh
 export GOOGLE_APPLICATION_CREDENTIALS=$(pwd)/testing/service-account.json
 
-# Import secrets for AWS integration testing. This can be used for products
```
danielbankhead, line 99: We want to use secret store instead.

Undoes this change in this commit: https://github.com/GoogleCloudPlatform/python-docs-samples/pull/7091/files#diff-ff2ac0c2d9eac8a4fc58f24e20895504b9c25a1b6d6c1f6f55a5978c1f5aefc3

storagetransfer/conftest.py
```diff
@@ -32,30 +32,76 @@ def project_id():
...
+
+    # retrieve from secret manager
+    response = client.access_secret_version(request={"name": name})
+
+    # parse secret from secret manager
+    secret = json.loads(response.payload.data.decode("UTF-8"))
+
+    aws_secret_manager_cache = {
+        'aws_access_key_id': secret['AccessKeyId'],
+        'aws_secret_access_key': secret['SecretAccessKey']
```
dandhlee, line 61: nit: The last item in a list of items (in `list`, `dict`, `set`, etc) should end with a comma.
```suggestion
        'aws_secret_access_key': secret['SecretAccessKey'],
```

storagetransfer/conftest.py
```diff
@@ -32,30 +32,76 @@ def project_id():
...
+    aws_access_key_id = os.environ.get("AWS_ACCESS_KEY_ID")
+    aws_secret_access_key = os.environ.get("AWS_SECRET_ACCESS_KEY")
+
+    if sts_aws_secret_name:
+        return aws_retrieve_from_secret_manager(sts_aws_secret_name)
+    else:
+        return {
+            'aws_access_key_id': aws_access_key_id,
+            'aws_secret_access_key': aws_secret_access_key
+        }
```
dandhlee, line 78: `else` in this case is optional, so not having it reduces nesting and still remains straightforward code.
```suggestion
    
    return {
        'aws_access_key_id': aws_access_key_id,
        'aws_secret_access_key': aws_secret_access_key
    }
```

storagetransfer/conftest.py
```diff
@@ -32,30 +32,76 @@ def project_id():
...
+
+
+@ pytest.fixture(scope='module')
```
dandhlee, line 81: Any reason for the extra space here? I couldn't find whether this is new or not.

dandhlee (changes requested):
The secret version ID seems to come in (`projects/*/secrets/*/version/*`) instead of the expected (`projects/*/secrets/*/versions/*`) format, missing `s` on `version` from the tests.

storagetransfer/conftest.py
danielbankhead: Yea, it's a little strange, I'll revert this auto-formatting.

storagetransfer/conftest.py
danielbankhead: Thanks, fixing

dandhlee (approved):
LGTM, all tests for `storagetransfer` are passing. Failing tests are from `.kokoro/tests/run_tests.sh` changes and can be ignored for this PR.

leahecole (approved):

cojenco (approved):

dandhlee:
Had previous tests passing, admin merging (some failures are expected, but they should be unrelated to storagetransfer)

PR #8428, opened 2022-10-18
feat: make trace demo application use OpenTelemetry

minherz:
remove trace demo application that using OpenCensus from the repo. make the removed demo's README file to reference to the active demo. modify the OpenTelemetry based demo so it has functional parity with the removed demo.

## Description

Remove the trace demo app that using OpenCensus in its implementation and make its README redirect users to the demo app that using OpenTelemetry instead.
Modify the OpenTelemetry-based demo to have a functional parity with the removed one. This includes deploying three workloads instead of simulating the nested calls in the single call.
Setup the regional tags, so the Cloud Trace [quickstart documentation][1] can reference the new application.

On the side note, the documentation's main goal is to demonstrate the use of Cloud Trace explorer. However, the demo application can be used for other purposes. This is why it should not shortcut the nested tracing.

DO NOT merge until the changes in the documentation are approved.

[1]: https://cloud.google.com/trace/docs/trace-app-latency


trace/cloud-trace-demo-app-opentelemetry/app/app.py
```diff
@@ -40,20 +47,40 @@ def configure_exporter(exporter):
...
+    if url == "":
+        return keyword, 200
+
+    # Endpoint is the next service to send string to.
+    data = {'body': keyword}
 
-    return "Hello World"
+    # [START trace_context_header]
+    with tracer.start_as_current_span("span_" + keyword):
+        response = requests.get(
```
aabmass, line 70: This call will not properly propagate the trace context as is. It looks like the OC app.py was manually adding the header. Would you prefer to

1. Manually pass the header here as well (I can help you with the propagator code)
2. Use the [`opentelemetry-instrumentation-requests`](https://opentelemetry-python-contrib.readthedocs.io/en/latest/instrumentation/requests/requests.html) package (similar to the `FlaskInstrumentor()`) to get propagation. This will also create a client side span for the request.

I think 2 is the way to go.

aabmass (approved):

trace/cloud-trace-demo-app-opentelemetry/app/app.py
minherz: Thank you. IMO, Option 2 is best here both for implementation and for demonstrating OT instrumentation.

minherz: @aabmass looking at the sample in the [documentation][doc] the instrumentor is initialized.
https://github.com/GoogleCloudPlatform/python-docs-samples/blob/9a7861952f0a42dbe3195b8b2b940fb1d9549c30/trace/cloud-trace-demo-app-opentelemetry/app/app.py#L36-L37
Do you mean that the following call to `tracer.start_as_current_span()` should be removed?
https://github.com/GoogleCloudPlatform/python-docs-samples/blob/9a7861952f0a42dbe3195b8b2b940fb1d9549c30/trace/cloud-trace-demo-app-opentelemetry/app/app.py#L69
Will the instrumentor auto-generate random spans?

[doc]: https://opentelemetry-python-contrib.readthedocs.io/en/latest/instrumentation/requests/requests.html

aabmass: It will generate a span but it won't be random. No harm in having this extra span imo

leahecole:
LGTM python style wise, but I'd love @ymotongpoo to be the final reviewer given his trace expertise

trace/cloud-trace-demo-app-opentelemetry/app/Dockerfile
```diff
@@ -16,12 +16,18 @@
 # https://hub.docker.com/_/python
 FROM python:3.10-slim
 
-# Copy local code to the container image.
+RUN pip3 install --upgrade pip
+
+RUN apt-get update \
```
ymotongpoo, line 21: Is this `apt-get update` necessary?

trace/cloud-trace-demo-app-opentelemetry/app/demo-service-b.yaml.template
```diff
@@ -38,4 +52,4 @@ spec:
   - protocol: TCP
     port: 80
     targetPort: 8080
-  type: LoadBalancer
+  type: LoadBalancer
```
ymotongpoo, line 55: Adding a empty new line is recommended

ymotongpoo (commented):
Added nit comments. Overall LGTM.

trace/cloud-trace-demo-app-opentelemetry/app/Dockerfile
minherz: AFAIK it is a common practice. It was also in the original app. Do you think there is a problem with it?

trace/cloud-trace-demo-app-opentelemetry/app/demo-service-b.yaml.template
minherz: Done

trace/cloud-trace-demo-app-opentelemetry/app/Dockerfile
ymotongpoo: That is the case for the following steps do `apt-get upgrade` or `apt-get install`. In this file, there no such steps after this and this `apt-get update` makes no effect.

trace/cloud-trace-demo-app-opentelemetry/app/Dockerfile
```diff
@@ -16,12 +16,14 @@
...
 
-# Copy local code to the container image.
+RUN pip3 install --upgrade pip
+
 ENV APP_HOME /app
 WORKDIR $APP_HOME
-COPY . ./
-COPY requirements.txt /app/
-RUN pip install Flask gunicorn
+
```
ymotongpoo, line 23: I meant only `apt-get update` isn't necessary, and you can clean up `apt` repository to make the container lighter.

ymotongpoo (approved):
Left a comment on Dockerfile and others LTGM

trace/cloud-trace-demo-app-opentelemetry/app/Dockerfile
minherz: I understand. My thought was that since I did not use it, there is no reason to clean it. :slightly_smiling_face:

trace/cloud-trace-demo-app-opentelemetry/README.md
```diff
@@ -2,73 +2,107 @@
...
+   ```
+
+1. Update GKE cluster credentials and verify the access to the cluster:
+
+   ```bash
+   gcloud container clusters get-credentials cloud-trace-demo --zone $ZONE
+   kubectl get nodes
+   ```
+
+   The output is expected to display 3 work nodes.
```
dandhlee, line 54: Any examples for these 3 work nodes?

trace/cloud-trace-demo-app-opentelemetry/README.md
```diff
@@ -2,73 +2,107 @@
...
+1. Setup GCP project to `YOUR_PROJECT_ID`. Please, replace `YOUR_PROJECT_ID` with your GCP project id.
+
+   ```bash
+   gcloud config set project YOUR_PROJECT_ID
+   ```
+
+1. Setup the location of GKE cluster and create a new cluster named
+"cloud-trace-demo".
+The example below sets the cluster's location to the "us-central1-c" zone.
+You can replace the zone with one which you like. See [zones][] for the full
```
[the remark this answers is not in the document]
minherz: There is. It is in the next sentence started with "See..."

trace/cloud-trace-demo-app-opentelemetry/README.md
minherz: The names of the nodes will be different each time due to auto-generated process. I think that it is sufficient to say that there will be a list of three records (or lines).

dandhlee: If it's consistent output with just different names, something like the example on https://cloud.google.com/kubernetes-engine/docs/tutorials/stateful-workload#configure_the_managed_file_storage_with_using_csi helps quite a lot for users to help verify. Take a look at the example output with the text `The output is similar to the following:`.

minherz: Thank you for the example.

trace/cloud-trace-demo-app-opentelemetry/app/app.py
```diff
@@ -11,49 +11,71 @@
...
+    # If there is an endpoint, send keyword to next service.
+    # Return received input with the keyword
+    keyword = os.getenv("KEYWORD")
+    endpoint = os.getenv("ENDPOINT")
+# [START trace_context_header]
+    if endpoint is not None and endpoint != "":
+        data = {'body': keyword}
+        response = requests.get(
+            endpoint,
+            params=data
```
dandhlee, line 71:
```suggestion
            params=data,
```

trace/cloud-trace-demo-app-opentelemetry/app/app_test.py
```diff
@@ -14,14 +14,25 @@
...
 A sample app demonstrating Google Cloud Trace
 """
+import os
+
 import mock
 
 import app
 
 
+expected_response = '''Hello, I am service A
```
dandhlee, line 24: This doesn't actually seem used.

dandhlee (approved):
LGTM, minor nits.

trace/cloud-trace-demo-app-opentelemetry/app/app.py
minherz: I think it might introduce a confusion to some readers. Especially, if it is done for only one function call.

trace/cloud-trace-demo-app-opentelemetry/app/app_test.py
minherz: removed

trace/cloud-trace-demo-app-opentelemetry/app/app.py
dandhlee: This is a common Python style (https://peps.python.org/pep-0008/#when-to-use-trailing-commas) and is strongly encouraged when you list out items vertically rather than horizontally.

minherz: Understood. Thank you for the link to the guidelines.

PR #8462, opened 2022-11-04
chore: Update license header config to ignore year

kweinmeister:
Temporarily setting this to `true` while repo migrations are underway.


dandhlee (approved):
Thank you! Admin merging.

PR #8492, opened 2022-11-10
lint: ignore class self annotations

davidcavazos:
## Description

Ignore `self` annotations for class methods. Typing them require `typing-extensions` to use the `Self` type, which is redundant and does not provide new information. Otherwise, the class name is not available when defining methods since it hasn't finished defining, so it requires further hacks to annotate.

Furthermore, [PEP 673](https://peps.python.org/pep-0673) has made self type annotations unnecessary, but is only available on Python 3.11+.

## Checklist

### DIFF
--- a/noxfile-template.py
+++ b/noxfile-template.py
@@ -124,6 +124,7 @@ def _determine_local_import_names(start_dir: str) -> List[str]:
 # Linting with flake8.
 #
 # We ignore the following rules:
+#   ANN101: missing type annotation for self in method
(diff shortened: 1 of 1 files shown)


dandhlee (approved):
Thank you! This is great to have, I agree that it is redundant and extremely unnecessary. This can definitely be an optional addition if folks choose to use it, otherwise it should be ignored.

leahecole (approved):
Great catch

PR #8525, opened 2022-11-16
fix: update library version and fix test failure

### LINKED ISSUE #8513: Address logging v3.2.2 issues in samples
Based on #8167, there has been an introduction with a new format of output that's messing with the expected output for the samples in `functions/v2/http_logging`. Few options that Daniel suggested:

```
The easiest solution for testing is probably to use the existing internal flag to disable instrumentation for the tests here: google.cloud.logging_v2._instrumentation_emitted = True. I think changing the tests is also acceptable if that ends up being easier though.
```

engelke:
## Description

Fixes #8513


functions/v2/http_logging/main_test.py
```diff
@@ -45,7 +45,14 @@ def test_functions_log_http_should_print_message(app, capsys):
...
         main.structured_logging(flask.request)
-        _, err = capsys.readouterr()
-        output_json = json.loads(err)
+        out, err = capsys.readouterr()
+
+        # In some situations, the library adds information before the first
+        # intended line. Use the next line in that case.
+        err_lines = err.splitlines()
+        if len(err_lines) > 1:
+            err = err_lines[1]
```
daniel-sanche, line 54: Is this guaranteed to be at index 1? It might be safer to look at the list's content to find the right entry. Maybe using something like this:

```
err = next(log for log in err_lines if "Hello, world" in log)
```

functions/v2/http_logging/requirements.txt
```diff
@@ -1,2 +1,2 @@
-google-cloud-logging==3.1.2
+google-cloud-logging==3.2.2
```
daniel-sanche, line 1: This can probably be bumped to 3.2.5

functions/v2/http_logging/main_test.py
engelke: It seems to be guaranteed to be either the only line, or second if the instrumentation inserted a line before the actual output. I didn't want to go too far in searching for it.

functions/v2/http_logging/requirements.txt
engelke: Done. Let's see if it breaks anything.

functions/v2/http_logging/main_test.py
```diff
@@ -45,7 +45,14 @@ def test_functions_log_http_should_print_message(app, capsys):
...
     with app.test_request_context(headers={'x-cloud-trace-context': f"{mock_trace}/{mock_span};o=1"}):
         main.structured_logging(flask.request)
-        _, err = capsys.readouterr()
-        output_json = json.loads(err)
+        out, err = capsys.readouterr()
+
+        # In some situations, the library adds information before the first
+        # intended line. Use the next line in that case.
+        err_lines = err.splitlines()
+        if len(err_lines) > 1:
```
ace-n, line 53: Nit: could we simplify this by always fetching the _last_ line?

functions/v2/http_logging/main_test.py
```diff
@@ -45,7 +45,14 @@ def test_functions_log_http_should_print_message(app, capsys):
     # Force trace with trace header
     with app.test_request_context(headers={'x-cloud-trace-context': f"{mock_trace}/{mock_span};o=1"}):
         main.structured_logging(flask.request)
-        _, err = capsys.readouterr()
-        output_json = json.loads(err)
+        out, err = capsys.readouterr()
```
dandhlee, line 48: It doesn't seem like `out` is used.

functions/v2/http_logging/main_test.py
engelke: I think it's clear this way.

rogerthatdev (approved):

nicain (approved):

PR #8537, opened 2022-11-18
chore(jobs): add normalized region tags with product prefix

meredithslota:
Adding in standardized region tags with product prefix to enable tracking, sample browser. See companion PR here: https://github.com/GoogleCloudPlatform/java-docs-samples/pull/7442/files


dandhlee (approved):
LGTM. Thank you!

PR #8553, opened 2022-11-23
chore: defrag run sample from GCF dir

ace-n:
**N.B:** this does **not** delete the base sample.

_(That will be addressed in a later PR.)_


ace-n:
This sample is already used in the docs (with the `run_` prefix).

I'd argue the `snippet-bot` warning is best fixed in a follow-up issue.

averikitsch:
There is a similar sample for functions, can we not keep the region tags around that?

ace-n:
Closed in favor of reusing [this sample](https://github.com/GoogleCloudPlatform/python-docs-samples/pull/8555/files#diff-8414ebd47b2df429ee18fb70edce886eb14d88602925c4d6d8c77264072affacR29-R51) in the `functions` directory.

PR #8559, opened 2022-11-28
process: Python asset migration

engelke:
Migrate samples back to python-docs-samples.


asset/snippets/noxfile.py
```diff
@@ -0,0 +1,292 @@
+# Copyright 2019 Google LLC
```
dandhlee, line 1: Please remove.

dandhlee (changes requested):

PR #8839, opened 2022-12-28
chore(functions/tips-scopes): Update sample signature using functions framework

tyampad:
Update [Global versus function scope sample](http://cloud/functions/docs/samples/functions-tips-scopes) function signature using functions framework

Tested by:

Running local tests
Deploying and invoking using curl

### DIFF
--- a/functions/tips-scopes/main.py
+++ b/functions/tips-scopes/main.py
@@ -13,13 +13,13 @@
 # limitations under the License.
 
 
+# [START functions_tips_scopes]
+# [START cloudrun_tips_global_scope]
 import time
 
 import functions_framework
 
 
-# [START functions_tips_scopes]
-# [START cloudrun_tips_global_scope]
 # Placeholder
 def heavy_computation():
     return time.time()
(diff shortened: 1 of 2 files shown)


nicain (approved):

functions/tips-scopes/main.py
```diff
@@ -28,11 +28,15 @@ def light_computation():
 
 # [START functions_tips_scopes]
```
nicain, line 29: WDYT about moving the region tags up to include the definition of heavy_computation, and then putting the import statement above that? This would make sense in terms of being as explicit as possible, and it would pass linting.

functions/tips-scopes/main.py
```diff
@@ -15,7 +15,11 @@
 
 import time
 
+import functions_framework
```
nicain, line 18: Sorry I wasen't clear before, I was typing in a hurry; I think this looks good, except I would put the `import functions_framework` and `import time` inside the region tags too, so that the snippet is a bit more copy-pastable.

nicain (changes requested):

ace-n (approved):

nicain (approved):

PR #8967, opened 2023-01-10
docs: Fix Retail README headers

kweinmeister:
There are missing spaces causing the headers not to render.

See https://github.com/GoogleCloudPlatform/python-docs-samples/tree/main/retail/interactive-tutorials


dandhlee (approved):
Thank you!

soheilazangeneh (approved):

PR #8971, opened 2023-01-11
Revert "chore(deps): update dependency apache-airflow-providers-cncf-kubernetes to v5.1.0"

leahecole:
Reverts GoogleCloudPlatform/python-docs-samples#8867

This was merged without the 3.8 check running, which is the only check that's enabled for Composer - it ended up breaking in the periodic build because it clashes with the constraints. It'll get upraded when we bump the version of Airflow! 😄


leahecole:
These build failures are unrelated to this PR (they actually should skip the tests) and are related to different kokoro issues. 3.8 passed which is what matter

rachael-ds (approved):

leahecole:
Merging because 2.7 is stalled and it is skipped

PR #8992, opened 2023-01-16
docs(samples): Adding regional disk samples

m-strzelczyk:
## Description

Adding samples needed for [this page](https://cloud.google.com/compute/docs/disks/regional-persistent-disk).

## Checklist


compute/client_library/ingredients/disks/attach_disk.py
```diff
@@ -0,0 +1,57 @@
...
+        zone:name of the zone in which the instance you want to use resides.
+        instance_name: name of the compute instance you want to attach a disk to.
+        disk_link: full link to the disk you want to attach. This can be either regional or zonal disk.
+            Expected formats:
+                * https://www.googleapis.com/compute/v1/projects/[project]/zones/[zone]/disks/[disk_name]
+                * /projects/[project]/zones/[zone]/disks/[disk_name]
+                * /projects/[project]/regions/[region]/disks/[disk_name]
+        mode: Specifies in what mode the disk will be attached to the instance. Available options are `READ_ONLY`
+            and `READ_WRITE`. Disk in `READ_ONLY` mode can be attached to multiple instances at once.
+    """
```
savijatv, line 41: WDYT about adding links to IAM permissions required to run this sample? https://cloud.google.com/compute/docs/disks/regional-persistent-disk#expandable-1 in this case.

compute/client_library/ingredients/disks/resize_disk.py
```diff
@@ -0,0 +1,59 @@
...
+import re
+from typing import NoReturn
+
+from google.cloud import compute_v1
+
+
+# <INGREDIENT resize_disk>
+def resize_disk(project_id: str, disk_link: str, new_size_gb: int) -> NoReturn:
+    """
+    Resizes a persistent disk to a new size in GB. After you resize the disk, you must
```
savijatv, line 28: Resizes a persistent disk to a specified size in GB.

compute/client_library/ingredients/disks/attach_disk.py
m-strzelczyk: Sure, good idea.

compute/client_library/snippets/disks/resize_disk.py
```diff
@@ -0,0 +1,120 @@
...
+        request.disks_resize_request_resource = compute_v1.DisksResizeRequest()
+        request.disks_resize_request_resource.size_gb = new_size_gb
+
+    request.disk = search_results["disk_name"]
+    request.project = project_id
+
+    operation = disk_client.resize(request)
+    wait_for_extended_operation(operation, "disk resize")
+
+    return
```
kweinmeister, line 116: Is explicitly including the return statement a Python sample best practice?

Either way, can we move the region tag end statements up to avoid the extra blank lines in the doc sample?

compute/client_library/ingredients/disks/resize_disk.py
```diff
@@ -0,0 +1,59 @@
...
+        request = compute_v1.ResizeDiskRequest()
+        request.zone = search_results['area_name']
+        request.disks_resize_request_resource = compute_v1.DisksResizeRequest()
+        request.disks_resize_request_resource.size_gb = new_size_gb
+
+    request.disk = search_results['disk_name']
+    request.project = project_id
+
+    operation = disk_client.resize(request)
+    wait_for_extended_operation(operation, "disk resize")
```
kweinmeister, line 56: There is a mix of single and double quotes in this sample. Consider making this consistent throughout where possible.

kweinmeister (approved):

compute/client_library/snippets/tests/test_disks.py
```diff
@@ -173,3 +245,26 @@ def test_create_and_delete_regional_disk(test_snapshot):
...
         delete_regional_disk(PROJECT, REGION, disk_name)
+
+
+def test_disk_attachment(autodelete_blank_disk, autodelete_regional_blank_disk, autodelete_compute_instance):
+    instance = get_instance(PROJECT, ZONE, autodelete_compute_instance.name)
+
+    assert len(list(instance.disks)) == 1
+
+    attach_disk(PROJECT, ZONE, instance.name, autodelete_blank_disk.self_link)
+    attach_disk(PROJECT, ZONE, instance.name, autodelete_regional_blank_disk.self_link)
```
rsamborski, line 256: Does it make sense to test read_write mode as well?

compute/client_library/snippets/disks/resize_disk.py
```diff
@@ -0,0 +1,122 @@
...
+# flake8: noqa
+
+
+# This file is automatically generated. Please do not modify it directly.
+# Find the relevant recipe file in the samples/recipes or samples/ingredients
+# directory and apply your changes there.
+
+
+# [START compute_regional_disk_resize]
+# [START compute_disk_resize]
```
rsamborski, line 23: Why do we have the same sample for two region tags?

compute/client_library/snippets/disks/attach_disk.py
```diff
@@ -0,0 +1,120 @@
...
+# flake8: noqa
+
+
+# This file is automatically generated. Please do not modify it directly.
+# Find the relevant recipe file in the samples/recipes or samples/ingredients
+# directory and apply your changes there.
+
+
+# [START compute_regional_disk_attach]
+# [START compute_disk_attach]
```
rsamborski, line 23: Why do we have the same samples for two region tags?

compute/client_library/snippets/disks/resize_disk.py
m-strzelczyk: Because I tend to write more universal samples, while other languages might need to have  separate methods for each. For example with static typing, it wouldn't be possible to have `disk_client` variable holding `compute_v1.RegionDisksClient()` or `compute_v1.DisksClient()` depending on context.

With this sample, the code is not as clean as I hoped, due to different names of used fields, but sometimes this works really nicely.

compute/client_library/snippets/tests/test_disks.py
m-strzelczyk: read_write mode is still in preview: https://cloud.google.com/compute/docs/disks/sharing-disks-between-vms#multi_writer_mode

compute/client_library/snippets/disks/attach_disk.py
m-strzelczyk: See previous comment, it applies to both.

compute/client_library/snippets/disks/resize_disk.py
rsamborski: Makes sense. Thanks for clarifying.

compute/client_library/snippets/tests/test_disks.py
rsamborski: Interesting, so will the sample work if we pass READ_WRITE in mode parameter? If the library doesn't support it yet then probably we should not include this in the comments. If we do support it then I think we should add a test.

compute/client_library/snippets/disks/attach_disk.py
```diff
@@ -0,0 +1,120 @@
...
+        zone:name of the zone in which the instance you want to use resides.
+        instance_name: name of the compute instance you want to attach a disk to.
+        disk_link: full or partial URL to a persistent disk that you want to attach. This can be either
+            regional or zonal disk.
+            Expected formats:
+                * https://www.googleapis.com/compute/v1/projects/[project]/zones/[zone]/disks/[disk_name]
+                * /projects/[project]/zones/[zone]/disks/[disk_name]
+                * /projects/[project]/regions/[region]/disks/[disk_name]
+        mode: Specifies in what mode the disk will be attached to the instance. Available options are `READ_ONLY`
+            and `READ_WRITE`. Disk in `READ_ONLY` mode can be attached to multiple instances at once.
```
dandhlee, line 102: Could you specify why we default to `READ_ONLY` if that's needed? If there's no reason, we should omit having a default value.

compute/client_library/snippets/disks/resize_disk.py
```diff
@@ -0,0 +1,122 @@
...
+import re
+import sys
+from typing import Any, NoReturn
+
+from google.api_core.extended_operation import ExtendedOperation
+from google.cloud import compute_v1
+
+
+def wait_for_extended_operation(
```
dandhlee, line 32: It seems like this code is duplicated in other files as well - could you refactor this to like a utility function instead?

m-strzelczyk: Yes, this code is often repeated, because I always want to provide a demonstration on how to handle the operation object and its results. It would be nice to turn it into an utility function of the Python GAPIC library, but that would be a bigger task that I currently don't even know how to start :)

dandhlee: Ah, just an utility function within `compute/client_library/snippets/util.py` or something would work! Doesn't have to be back in the GAPIC library.

kweinmeister:
```
  File "/workspace/compute/client_library/snippets/tests/test_disks.py", line 255, in test_disk_attachment
    attach_disk(PROJECT, ZONE, instance.name, autodelete_blank_disk.self_link)
TypeError: attach_disk() missing 1 required positional argument: 'mode
```

compute/client_library/snippets/disks/resize_disk.py
m-strzelczyk: But then it wouldn't be part of the sample and people would wonder where is the code for this. I want all the samples to be available as copy-paste pieces that will work, without additional research required from developer to find this one utility function.

kweinmeister:
@m-strzelczyk @rsamborski we merged the region change PR first; can you please resolve this conflict with the right region?

rsamborski (approved):

compute/client_library/snippets/disks/resize_disk.py
dandhlee: Ah, I confused myself thinking that you could import the same region tag over multiple files - you can do it over the same file but not over different files as it'll cause a sync issue. SGTM to keep it as is, the tradeoff for user experience is better to have than maintenance over O(5) occurrences.

dandhlee (approved):
LGTM.

leahecole:
maintainers -leave this PR in the merge queue! It's there so Gh engineers can look at our support ticket!

PR #9022, opened 2023-01-19
fix: Providing better error message for accounts without keys

### LINKED ISSUE #8981: TypeError when try to use physical_security_key_ssh.py to get private key
Thanks for stopping by to let us know something could be better!

**PLEASE READ**: If you have a support contract with Google, please create an issue in the [support console](https://cloud.google.com/support/) instead of filing on GitHub. This will ensure a timely response.

**The issue you're having must be related to a file in this repository.** We are unable to provide assistance for issues unrelated to samples in this repository.

Please include as much information as possible:

[compute/oslogin/physical_security_key_ssh.py](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/HEAD/compute/oslogin/physical_security_key_ssh.py)

## Did you change the file? If so, how?
No I didn't.

## Describe the issue
I ran the comment `python3 SCRIPT_NAME.py --user_key=USER_KEY --ip_address=IP_ADDRESS` [accordingly](https://cloud.google.com/compute/docs/oslogin/security-keys#third-party-tools), however, get the error `TypeError: 'NoneType' object is not iterable`.

I check the code, it seems that the `profile` variable looks fine, but, `profile.get("securityKeys")` gets nothing in return. I think that the format of the `profile` has changed.

Making sure to follow these steps will guarantee the quickest resolution possible.

Thanks!

m-strzelczyk:
## Description

Fixes #8981 

Providing more information in case of accounts without security keys assigned to them.


rsamborski (approved):

nicain (approved):

compute/oslogin/physical_security_key_ssh.py
```diff
@@ -55,15 +55,23 @@ def main(user_key, ip_address, dryrun, directory=None):
         .getLoginProfile(name="users/{}".format(user_key), view="SECURITY_KEY")
         .execute()
     )
-    security_keys = profile.get("securityKeys")
 
     if "posixAccounts" not in profile:
         print("You don't have a POSIX account configured.")
+        print("Did you enable OSLogin on your compute instance?")
```
savijatv, line 61: In this case, are we asking users to enable OS Login for VMs? The answer for the above question could be Yes or No. Can we say something like "Please make sure that you have enabled OS Login for your VM."

compute/oslogin/physical_security_key_ssh.py
```diff
@@ -55,15 +55,23 @@ def main(user_key, ip_address, dryrun, directory=None):
...
         return
 
     username = profile.get("posixAccounts")[0].get("username")
 
     # Write the SSH private key files.
+    security_keys = profile.get("securityKeys")
+
+    if security_keys is None:
+        print("It seems that the account you are using to authenticate does not have any security keys assigned to it.")
```
savijatv, line 70: "The account you are using to authenticate does not ..."

compute/oslogin/physical_security_key_ssh.py
```diff
@@ -55,15 +55,23 @@ def main(user_key, ip_address, dryrun, directory=None):
...
+        print("Please check your Application Default Credentials (https://cloud.google.com/docs/authentication/application-default-credentials).")
+        print("More info about security keys: https://support.google.com/accounts/answer/6103523?visit_id=637673282586358398-2383089289&rd=1")
```
savijatv, line 72: IIUC, this support page describes only about physical security keys. Should we link it to this page instead which talks about the security keys and connecting to VMs with OS Login enabled? - https://cloud.google.com/compute/docs/oslogin/security-keys

ENsu:
Thank you. @m-strzelczyk I did not use the correct account. 

I believe the newly added error message would be really useful if someone encountered the same situation as myself.

PR #9049, opened 2023-01-25
include people-and-planet-ai in auto-label.yaml

rogerthatdev:
This PR updates `auto-label.yaml` to include  `people-and-planet-ai/` which is used in [blunderbuss configuration](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/54f19f06894940230ac1d6cd5e11547cc265c816/.github/blunderbuss.yml#L112).


kweinmeister (approved):

dandhlee (approved):
We _might_ have to fix this with the autolabel bot, but let's give this a try.

PR #9050, opened 2023-01-25
docs(samples): Document AI Samples Updates for `2.10.0`

holtskinner:
- Combine Quickstart Sample into Process Document Sample to reduce maintenence
- Added Native PDF Parsing to OCR Demo (Requires Version `2.10.0`)


telpirion (commented):
I recommend duplicating these samples and updating the duplicates. Here's why: some customers have IT policies that forbid using not-yet-released software, including things labeled as "beta." When we have a GA-quality version of an API alongside a beta "next version" of the product, we try to keep both versions available until the beta features are merged into the GA/stable release track.

PR #9104, opened 2023-02-09
docs(samples): Adding the LRO handling method as a separate file and sample

m-strzelczyk:
Adding this method as a separate file and sample, as I want to be able to link to it from [python-compute README](https://github.com/googleapis/python-compute/blob/main/README.rst) and potentially from the cloud docs page.


leahecole:
for files that previously existed, they should preserve their original copyright year

m-strzelczyk:
@leahecole they do. There are 2 new files with 2023, the two modified only have the replicated copyright header removed, nothing changed apart from that.

kweinmeister (approved):

rsamborski (approved):

m-strzelczyk:
Tests for 3.11 are failing due to misconfiguration, not because of this code. I'm merging as tests for other versions work well.

Issue #8873, opened 2023-01-06
Add support for python 3.11


parthea:
https://www.python.org/downloads/release/python-3111/

parthea:
Re-opening because testing with python 3.11 was rolled back in https://github.com/GoogleCloudPlatform/python-docs-samples/pull/9003

PR #9108, opened 2023-02-10
ci: use prebuilt docker image in presubmits

parthea:
Use a prebuilt docker image in presubmits to solve the issue where presubmits fail during the docker build step. See [this log](https://source.cloud.google.com/results/invocations/ed14226d-09e7-431d-aeaa-2f9d8d3b15d6/log) for an example of the failure where the docker build step timed out.

Towards https://github.com/GoogleCloudPlatform/python-docs-samples/issues/8873

### DIFF
--- a/.kokoro/docker/Dockerfile
+++ b/.kokoro/docker/Dockerfile
@@ -156,7 +156,7 @@ RUN wget --no-check-certificate -O /tmp/get-pip.py 'https://bootstrap.pypa.io/ge
   # https://github.com/docker-library/python/pull/100
   && [ "$(pip list |tac|tac| awk -F '[ ()]+' '$1 == "pip" { print $2; exit }')" = "$PYTHON_PIP_VERSION" ]
 
-# Ensure Pip for all python3 versions
+# Ensure Pip for all python3 versions.
 RUN python3.9 /tmp/get-pip.py
 RUN python3.8 /tmp/get-pip.py
 RUN python3.7 /tmp/get-pip.py
--- a/.trampolinerc
+++ b/.trampolinerc
@@ -47,7 +47,7 @@ if [[ -z "${TRAMPOLINE_IMAGE:-}" ]]; then
 fi
 
 if [[ -z "${TRAMPOLINE_DOCKERFILE:-}" ]]; then
-    TRAMPOLINE_DOCKERFILE=".kokoro/docker/Dockerfile"
+    TRAMPOLINE_DOCKERFILE=""
 fi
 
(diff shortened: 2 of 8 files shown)


kweinmeister (approved):

rsamborski (approved):

.trampolinerc
```diff
@@ -47,7 +47,7 @@ if [[ -z "${TRAMPOLINE_IMAGE:-}" ]]; then
 fi
 
 if [[ -z "${TRAMPOLINE_DOCKERFILE:-}" ]]; then
-    TRAMPOLINE_DOCKERFILE=".kokoro/docker/Dockerfile"
+    TRAMPOLINE_DOCKERFILE=""
```
m-strzelczyk, line 50: I'd love to have a comment explaining what setting this variable does, so we don't have to think too much about this in the future, just have the info on a plate.

m-strzelczyk (approved):

m-strzelczyk:
I think we might have a problem with running **ALL** the tests. @rsamborski recently had a problem with this and solved it by running all the tests manually on his local machine. If the tests fail, I guess we will need to do the same and admin-merge.

rsamborski:
> I think we might have a problem with running **ALL** the tests. @rsamborski recently had a problem with this and solved it by running all the tests manually on his local machine. If the tests fail, I guess we will need to do the same and admin-merge.

I agree. At current repo state it's pretty much impossible to have all tests pass. I think that after admin-merge when we have fixed the docker build issue we can work to fix the flacky/failling tests in individual products to bring repo back to green state.

.kokoro/docker/Dockerfile
```diff
@@ -156,7 +156,7 @@ RUN wget --no-check-certificate -O /tmp/get-pip.py 'https://bootstrap.pypa.io/ge
   # https://github.com/docker-library/python/pull/100
   && [ "$(pip list |tac|tac| awk -F '[ ()]+' '$1 == "pip" { print $2; exit }')" = "$PYTHON_PIP_VERSION" ]
 
-# Ensure Pip for all python3 versions
```
nicain, line 159: This is the change to trigger all the tests, right?

nicain (approved):

.trampolinerc
dandhlee: This tells which dockerfile you can specify to use for the Kokoro job, iirc. We can add a comment in a followup PR.

dandhlee (approved):
Test failures have been looked into, and is unrelated to this PR. LGTM.

PR #8861, opened 2023-01-04
feat: add python 3.11.1 to the samples image

parthea:

PR #9131, opened 2023-02-15
fix: restore python 3.11 in samples docker image

parthea:
Towards https://github.com/GoogleCloudPlatform/python-docs-samples/issues/8873

This change was originally made in https://github.com/GoogleCloudPlatform/python-docs-samples/pull/8861 and reverted in https://github.com/GoogleCloudPlatform/python-docs-samples/pull/9003. The reason for reverting the change is because of an issue where the `docker build` step would time out. The samples docker image is no longer built on the fly in presubmits and we no longer have the `docker build` step.


m-strzelczyk (approved):
LGTM

parthea:
Merging as these changes are not tested during presubmits and this specific changed was previously merged into main in https://github.com/GoogleCloudPlatform/python-docs-samples/pull/8861.

PR #9144, opened 2023-02-17
fix: call DLP API in the region of DataFlow job

minherz:
Modifies the final version of the sample code to call DLP API for the same region region where the DataFlow job is deployed.
Adds placeholder of the change to the boilerplate code.
Updates README with information about the sample files and sets up the link to Neos tutorial.


minherz (commented):
lgtm

arbrown (approved):

kweinmeister (approved):

PR #9157, opened 2023-02-21
chore: remove unused `host` from URL in SQL Server samples

jackwotherspoon:
In latest version of `sqlalchemy-pytds` library #9145, the `host` param has become optional in the SQLAlchemy engine URL, our samples previously had a dummy value of `localhost` set to fulfill the required host param. Now that it is optional, we can remove `localhost` from the URL for simplicity and to avoid confusion.


enocom (approved):

dandhlee (approved):
Will the users know how to use `host` param if required?

jackwotherspoon:
@dandhlee 
> Will the users know how to use `host` param if required?

The users should never have the `host` needed or be required for the samples being updated here. The SQLAlchemy `creator` argument bypasses all URL params. It was only being set to `localhost` as a workaround to a bug in the underlying library.

PR #9171, opened 2023-02-24
feat: add default values to quickstart

JesseLovelace:
Adds default values to the quickstart

this is to make it smoother to run a neos tutorial (user will edit the default values and run the code)


kweinmeister (approved):

storagetransfer/quickstart.py
```diff
@@ -26,8 +26,8 @@
 from google.cloud import storage_transfer
 
 
-def create_one_time_transfer(project_id: str, source_bucket: str,
-                             sink_bucket: str):
+def create_one_time_transfer(project_id: str = "my_project_id", source_bucket: str = "my_source_bucket",
+                             sink_bucket: str = "my_sink_bucket"):
     """Creates a one-time transfer job."""
```
rsamborski, line 31: Can we move the comments explaining each argument to this docstring (similarly to this [example](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/b1d9789da21c13699ca3eb2806946f29f8374a90/compute/client_library/snippets/instances/create.py#L60))?

storagetransfer/quickstart.py
```diff
@@ -26,8 +26,8 @@
 from google.cloud import storage_transfer
 
 
-def create_one_time_transfer(project_id: str, source_bucket: str,
-                             sink_bucket: str):
+def create_one_time_transfer(project_id: str = "my_project_id", source_bucket: str = "my_source_bucket",
```
rsamborski, line 29: I personally don't think adding defaults which won't work if user runs this function without parameters is a good practice. Can you elaborate on how this helps with Neos tutorial?

rsamborski: Can we instruct the user to uncomment lines 36, 39 and 42 to put their values there? Removing these parameters completely will make it unrunnable from shell and break the convention (we use similar parameters in other samples too).

JesseLovelace: The other languages don't have an uncommenting step so I would have to make a special exception for Python in the Neos tutorial code. It's possible, but not elegant in my opinion. 

I understand that it's against convention of normal samples, but this is a quickstart, which tend to work a little differently, like I showed the Storage Quickstart doesn't have any parameters. 

Having default values that need to be edited for the sample to run is also something we do in other quickstarts, for example here it is in the storage node quickstart: https://github.com/googleapis/nodejs-storage/blob/main/samples/quickstart.js#L17

and here in the python bigtable quickstart: https://github.com/googleapis/python-bigtable/blob/HEAD/samples/quickstart/main.py#L23

rsamborski: Thanks for additional context and explanation. I am going to approve this although I still think we should come up with a better way of doing this in the future.

rsamborski (approved):

PR #9209, opened 2023-03-03
chore(deps): Updating black to v23

m-strzelczyk:
## Description

Because #9071 has some weird problems with tests, I will update this dependency manually.


m-strzelczyk:
Since this was done to solve a simple version update and all tests are passing I will admin-merge.

PR #9300, opened 2023-03-17
chore(functions/v2/log): Update log samples to cloud event declarative signature

kenneth-rosario:


functions/v2/log/main.py
```diff
@@ -0,0 +1,48 @@
...
+    """
+    print("Hello, stdout!")
+
+
+# [END functions_cloudevent_log_helloworld]
+
+
+# [START functions_cloudevent_log_stackdriver]
+@functions_framework.cloud_event
+def process_log_entry(event):
```
engelke, line 39: I'm unclear on the context here. A Cloud Function has only one entry point specified, so why does this code seem to have two? Would it be deployed twice, once with each entry point? Should the two entry points be two separate samples?

engelke (changes requested):
Please clarify the use of two entry points in this sample.

functions/v2/log/main.py
kenneth-rosario: I was updating https://github.com/GoogleCloudPlatform/python-docs-samples/blob/main/functions/log/main.py that had both samples together. I will separate them into different samples which is cleaner.

kenneth-rosario: I split the sample into two distinct ones in their respective directories.

engelke (approved):
LGTM

PR #9313, opened 2023-03-20
feat(firestore): Adds OR composite filter

telpirion:


meredithslota:
This region tag doesn't exist in other languages; ok to close out. (ref: firestore_query_composite_filter_or)

PR #9330, opened 2023-03-22
feat(firestore): adds COUNT query sample

telpirion:


firestore/cloud-client/aggregate_query_count_test.py
```diff
@@ -0,0 +1,28 @@
...
+
+os.environ["GOOGLE_CLOUD_PROJECT"] = os.environ["FIRESTORE_PROJECT"]
+PROJECT_ID = os.environ["GOOGLE_CLOUD_PROJECT"]
+
+
+def test_create_count_query(capsys):
+    create_count_query(project_id=PROJECT_ID)
+    out, _ = capsys.readouterr()
+
+    assert "Number of" in out
```
Mariatta, line 28: Can you also assert the alias?

Mariatta:
```suggestion
    assert "Alias of results from query: all" in out
    assert "Number of results from query: <somenumber>" in out
```

Mariatta:
For count query in datastore, we added samples for using it with transaction and snapshot. Would those worth adding as samples in Firestore to? https://github.com/googleapis/python-datastore/pull/383/files

firestore/cloud-client/aggregate_query_count_test.py
telpirion: Done.

Mariatta (approved):

PR #9341, opened 2023-03-24
fix: avoid name collisions

engelke:
## Description

Since multiple instances of the auth test can run simultaneously, use a unique name for the key to avoid interference between them.


engelke:
Well, that didn't help with the flakies. This is a good idea, but if it doesn't help, it's not worth the disruption. Closing unmerged.

PR #9412, opened 2023-03-30
chore(functions/v2/firebase/upper-firestore): Make upper firestore sample gcf gen 2 compatible

kenneth-rosario:


kweinmeister (approved):

functions/v2/firebase/upper-firestore/main.py
```diff
@@ -0,0 +1,48 @@
...
+client = firestore.Client()
+
+
+# Converts strings added to /messages/{pushId}/original to uppercase
+@functions_framework.cloud_event
+def make_upper_case(cloud_event: CloudEvent) -> None:
+    firestore_payload = firestoredata.DocumentEventData()
+    firestore_payload._pb.ParseFromString(cloud_event.data)
+
+    path_parts = firestore_payload.value.name.split("/documents/")[1].split("/")
```
garethgeorge, line 30: This section dealing with path parts is a bit complicated. What are valid inputs for firestore_payload.value.name?  i.e. I suspect this will behave incorrectly if the document path is something like "foo/documents/foo/documents/bar" ?

I wonder if it can be simplified by making the parsing a bit more explicit i.e. something along the lines of:
```
path_parts = firestore_payload.value.name.split("/")
separator_idx = path_parts.index("documents")
collection_path = ...
document_path = ...
```
?

garethgeorge (approved):
LGTM

PR #9414, opened 2023-03-31
fix(issue-9207): Providing better docstrings for KMS-related methods

m-strzelczyk:
## Description

Issue #9207 

Providing better docstrings and hint about solving a problem in the method code itself, as the problem is popping up from time to time. 

The error related in #9207 should be fixed by cl/520842870


dandhlee (approved):
LGTM

PR #9465, opened 2023-04-03
chore: exclude GAE 3.7-earlier samples from later builds

### LINKED ISSUE #9379: GAE "Python 3.7 and earlier" samples are being tested on 3.8+ versions
The samples in this [directory](https://github.com/GoogleCloudPlatform/python-docs-samples/tree/main/appengine/flexible_python37_and_earlier) are being tested on higher versions (see [3.9 build log](https://source.cloud.google.com/results/invocations/54b9f215-3396-40de-8e75-c9e6f9807db0/targets)). We should add`noxfile_config.py` files that exclude 3.8+ ([example](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/main/appengine/standard_python3/bigquery/noxfile_config.py)).

kweinmeister:
Fixes #9379


iennae (approved):
LGTM; failure looks like it's due to the bucket deletion by the vertex team (unclear if automated process is removing the testing bucket)

nicain (approved):

PR #9468, opened 2023-04-03
fix: Add additional logging to _test_body 

nicain:
Hopefully will help identify cause of failures in batch/test_basics.py::test_script_job

REF #9210


engelke (approved):

nicain:
@engelke has observed that this issue always occurs with python3.11; going to try and reproduce with that config

leahecole (approved):
Cautious LGTM - does this need to be here forever? If not, can we open an issue to remove it when we've fixed the issue?

Alternatively, can it stay open as a draft PR for your debugging purposes?

PR #9474, opened 2023-04-04
chore(deps): Updating OSLogin dependencies

m-strzelczyk:
## Description

Fixes #8880 

We need to update couple of things at once, or there will be dependency problems.


msampathkumar (approved):
LGTM

PR #9491, opened 2023-04-04
fix: use same project for all testing

engelke:
## Description

Does not use different projects for different versions of Python any more.

### DIFF
--- a/batch/noxfile_config.py
+++ b/batch/noxfile_config.py
@@ -13,5 +13,5 @@
 #  limitations under the License.
 
 TEST_CONFIG_OVERRIDE = {
-    "gcloud_project_env": "BUILD_SPECIFIC_GCLOUD_PROJECT",
+    "gcloud_project_env": "GOOGLE_CLOUD_PROJECT",
 }
--- a/batch/noxfile_config.py
+++ b/batch/noxfile_config.py
@@ -13,5 +13,6 @@
 #  limitations under the License.
 
 TEST_CONFIG_OVERRIDE = {
-    "gcloud_project_env": "GOOGLE_CLOUD_PROJECT",
+    # "gcloud_project_env": "GOOGLE_CLOUD_PROJECT",
+    "gcloud_project_env": "BUILD_SPECIFIC_GCLOUD_PROJECT",
 }
--- a/data-science-onramp/data-processing/noxfile_config.py
+++ b/data-science-onramp/data-processing/noxfile_config.py
@@ -30,8 +30,8 @@ TEST_CONFIG_OVERRIDE = {
     # to 'BUILD_SPECIFIC_GCLOUD_PROJECT' if you want to opt in using a
     # build specific Cloud project. You can also use your own string
     # to use your own Cloud project.
-    # "gcloud_project_env": "GOOGLE_CLOUD_PROJECT",
-    "gcloud_project_env": "BUILD_SPECIFIC_GCLOUD_PROJECT",
+    "gcloud_project_env": "GOOGLE_CLOUD_PROJECT",
+    # "gcloud_project_env": "BUILD_SPECIFIC_GCLOUD_PROJECT",
     # A dictionary you want to inject into your test. Don't put any
     # secrets here. These values will override predefined values.
     "envs": {},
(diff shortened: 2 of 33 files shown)


nicain (approved):

batch/noxfile_config.py
```diff
@@ -13,5 +13,5 @@
 #  limitations under the License.
 
 TEST_CONFIG_OVERRIDE = {
-    "gcloud_project_env": "BUILD_SPECIFIC_GCLOUD_PROJECT",
+    "gcloud_project_env": "GOOGLE_CLOUD_PROJECT",
```
nicain, line 16: Revert for now, and land in a follow-up CL

nicain (changes requested):

dialogflow/noxfile_config.py
```diff
@@ -30,8 +30,8 @@
     # to 'BUILD_SPECIFIC_GCLOUD_PROJECT' if you want to opt in using a
     # build specific Cloud project. You can also use your own string
     # to use your own Cloud project.
-    # "gcloud_project_env": "GOOGLE_CLOUD_PROJECT",
-    "gcloud_project_env": "BUILD_SPECIFIC_GCLOUD_PROJECT",
+    "gcloud_project_env": "GOOGLE_CLOUD_PROJECT",
```
nicain, line 33: This change, combined with #9489 and upgrading python-docs-samples-tests to Enterprise, should permanently resolve the Dialogflow-related error

batch/noxfile_config.py
m-strzelczyk, line 16: Please don't change this. Batch tests need to spin up VMs and those consume quota (external IP, CPUs etc). With all tests running in single project, this will cause unnecessary quota errors and/or slow down the test execution.

m-strzelczyk (changes requested):
I'm worried that we might face some new problems after squeezing all the tests into single project. Things like quota or tests interfering with each other. 

I assume you have good reasons to change this for the samples you're investigating, so no veto from me in general. I'd only want to keep Batch tests in separate projects, as explained in my comment there.

data-science-onramp/data-processing/noxfile_config.py
```diff
@@ -30,8 +30,8 @@
     # to 'BUILD_SPECIFIC_GCLOUD_PROJECT' if you want to opt in using a
     # build specific Cloud project. You can also use your own string
     # to use your own Cloud project.
-    # "gcloud_project_env": "GOOGLE_CLOUD_PROJECT",
-    "gcloud_project_env": "BUILD_SPECIFIC_GCLOUD_PROJECT",
+    "gcloud_project_env": "GOOGLE_CLOUD_PROJECT",
```
leahecole, line 33: I believe that this is build specific for the VM quota issue as well - dataproc instances count towards the VM quota

dataproc/snippets/noxfile_config.py
```diff
@@ -30,8 +30,8 @@
     # to 'BUILD_SPECIFIC_GCLOUD_PROJECT' if you want to opt in using a
     # build specific Cloud project. You can also use your own string
     # to use your own Cloud project.
-    # "gcloud_project_env": "GOOGLE_CLOUD_PROJECT",
-    "gcloud_project_env": "BUILD_SPECIFIC_GCLOUD_PROJECT",
+    "gcloud_project_env": "GOOGLE_CLOUD_PROJECT",
```
leahecole, line 33: I believe that this is build specific for the VM quota issue as well - dataproc instances count towards the VM quota

engelke:
Small variations between the different test projects have led to a lot of very hard to debug flaky errors. Using a single project will eliminate that problem, while making quota problems more likely. We have managed to resolve most of those problems (future PRs will be coming soon after this one is merged), except for:

- GAE send mail per minute (resolved) and per day quotas. The 100 emails per day quota is a problem when three samples are run in five versions of Python.

- Some compute/client-library tests that currently try a large list of variations. The tests usually don't cause quota issues, but some remain sometimes. Perhaps just testing a smaller representative number of variations will suffice.

Issue #9495, opened 2023-04-05
Remove IoT Core snippets after August 15


leahecole:
IoT Core is being deprecated in August (see [release notes](https://cloud.google.com/iot/docs/release-notes)). Given that the docs and service will no longer be available, we should remove the artifacts related to it.

meredithslota:
This was solved in #10489

meredithslota:
Actually — we should clarify. Is our stance that we'll keep the samples in the folder but note in the README that samples are "archived" and no longer maintained, or should we actually remove/delete them?

meredithslota:
I'd vote that we actually remove them once the docs are also removed, it doesn't serve us to keep them in the repo (IMHO).

leahecole:
Sorry I missed this comment! Given that this service is just not going to be an option anymore, I vote for total removal.

PR #9498, opened 2023-04-05
Skip IoT tests, add deprecation notice to readme

leahecole:
I made an issue to track their upcoming removal #9495, but I want to propose skipping IoT tests until then. Some flakiness with those tests has been blocking merging of dependency PRs, and I don't think our time is best spent adding retries to something that is slated for removal later this year. I believe that they are hanging in 3.11, causing #9437 to timeout

If folks aren't ready for this though and think we should wait on this til August, that's fine, just let me know

### DIFF
--- a/iot/api-client/README.md
+++ b/iot/api-client/README.md
@@ -1,5 +1,9 @@
 # Cloud IoT Core Python Samples
 
+## DEPRECATION NOTICE
+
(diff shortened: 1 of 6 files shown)


engelke (approved):
LGTM

nicain (approved):

PR #9502, opened 2023-04-06
fix: Get project and credentials from the ADC (dialogflow dialogflow_analyze_content_audio_stream)

nicain:
## Description

Include the use of the application default credential to get credentials and project_id to construct the ParticipantsClient

As the sample is currently written, the ParticipantsClient._transport._client is already defined from the ADC. However, the snippet allows the project_id to be passed in, for example from GOOGLE_CLOUD_PROJECT. If the ADC `quota_project_id` and the project_id differ, this can result in really confusing behavior (and nearly impossible for an end-user to debut), when the two projects are on different tiers eg. Enterprise vs. Free.

Note: This fix does not move import statements into region tags, as is the recommended best-practice. The reason is that this results in linting errors that can not be skipped with in-file directives, because there is more than one sample per file. This scope-of-work is tracked by #9505


rsamborski (changes requested):
Tests need fixing: 
```
TypeError: analyze_content_audio_stream() got an unexpected keyword argument 'project_id'
```

And linter is is throwing errors as well:
```
./participant_management.py:114:1: I100 Import statements are in the wrong order. 'import google.auth' should be before 'from google.cloud import dialogflow_v2beta1'
import google.auth
^
./participant_management.py:114:1: E402 module level import not at top of file
import google.auth
^
./participant_management.py:115:1: F811 redefinition of unused 'dialogflow' from line 19
from google.cloud import dialogflow_v2beta1 as dialogflow
^
./participant_management.py:115:1: E402 module level import not at top of file
from google.cloud import dialogflow_v2beta1 as dialogflow
```

m-strzelczyk:
So we lose control over in which project the tests run? They will run in the `python-docs-samples-tests` no matter what, because the SA that runs them is from that project?

dialogflow/participant_management.py
```diff
@@ -127,7 +130,8 @@ def analyze_content_audio_stream(
     # After completing all of your requests, call the "__exit__()" method to safely
     # clean up any remaining background resources. Alternatively, use the client as
     # a context manager.
-    client = dialogflow.ParticipantsClient()
+    credentials, project_id = google.auth.default()
```
rogers140, line 130: In our local run, we usually just use `GOOGLE_APPLICATION_CREDENTIALS` environment variables to get the credentials. Is that something contradicting here?

nicain:
Creating future cleanup issue: #9505

engelke (approved):
LGTM.

dialogflow/participant_management.py
nicain: Nope, that is exactly right. This line will use those credentials. My issue was that the project defined in those credentials was a Free tier project, and not the same project as the Enterprise Tier project I as passing in as project_id. That conflict caused the behavior, so this change is an attempt to prevent that from happening for users in the future.

rsamborski (approved):

PR #9538, opened 2023-04-10
fix: minimize custom tests used

### LINKED ISSUE #9532: Address quota issues for IAM create/delete custom role issue
We are seeing test failures we have seen before in IAM in #9525 and #9526 as a result of #9491. We asked the IAM team about this years ago [on YAQs](https://yaqs.corp.google.com/eng/q/5013374474780672) and never heard back. We ended up choosing to use multiple test projects to solve the issue then and likely will need to do that again now, unless circumstances have changed and someone knows something I don't! (I suspect they haven't though - someone actually pinged me about that YAQs a month or so ago asking what we ended up doing)

engelke:
## Description

Fixes #9532 #9525 #9526 

There is a hard quota of 300 custom roles per project, and deleted custom roles still count against that quota for up to 14 days, so this test now creates and deletes only one custom role when run.

### DIFF
--- a/iam/api-client/custom_roles_test.py
+++ b/iam/api-client/custom_roles_test.py
@@ -14,6 +14,7 @@
 
 import os
 import uuid
+
 import pytest
 
 import custom_roles
--- a/iam/api-client/custom_roles_test.py
+++ b/iam/api-client/custom_roles_test.py
@@ -73,10 +73,11 @@ def test_edit_role(custom_role, capsys):
     custom_roles.edit_role(
         custom_role,
         GCLOUD_PROJECT,
-        CUSTOM_ROLE_TITLE,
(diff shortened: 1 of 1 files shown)


engelke:
The existing test suite creates deletes several tests, but the deleted tests still count against quota for up to 14 days. Now that the quota has been hit, it seems that there will never be enough unused quota for the test suite to create all the roles it wants to before it uses it up again. We should merge this change so the test suite uses only one role, and once the old deleted roles start to age out, it should be able to run to completion.

nicain (approved):

m-strzelczyk (approved):

m-strzelczyk:
I agree to admin-merge this, but first fix the lint problem :)

PR #9541, opened 2023-04-11
docs(samples): Bringing IAM tests back to version-specific projects

### LINKED ISSUE #9532: Address quota issues for IAM create/delete custom role issue

m-strzelczyk:
## Description

Fixes #9532

Bringing back IAM tests to the version-specific projects, to avoid quota issues with number of custom roles.


m-strzelczyk:
Turns out that #9538 is a thing, and should be a better solution.

PR #9625, opened 2023-04-13
chore(functions/spanner): update sample to use declarative signature

kenneth-rosario:


kenneth-rosario:
Sounds good, I have an auto formatter that formatted them. My settings might be opinionated by default to use double quotes

kweinmeister (approved):

PR #9668, opened 2023-04-17
fix: reduce unneeded wait

engelke:
## Description

fix: May avoid flaky "key is expired" errors


engelke:
This may be worth doing in the future for speed, but experiments have shown this is not the cause of the flaky failures.

PR #9720, opened 2023-04-20
chore: cleanup blunderbuss assignments

kweinmeister:


.github/blunderbuss.yml
```diff
@@ -13,218 +13,150 @@
...
+    to:
+      - GoogleCloudPlatform/cloud-retail-team
+
+  # Self-service individuals
+  - labels:
+      - "api: auth"
+    to:
+      - arithmetic1728
 
-assign_prs_by:
```
rsamborski, line 190: Is it intended change to completely remove `assign_prs_by` section?

rsamborski: I've made couple additional changes based on discussions around CODEOWNERS in [this commit](https://github.com/GoogleCloudPlatform/python-docs-samples/pull/9720/commits/fa06f7ae4736a5bef714edd77647c9de575213c9) (included in this PR).

The `assign_prs_by` section might require a longer discussion. I see following options:
 1) We keep the old one
 2) We remove it completely, which means that PRs will be auto assigned to GoogleCloudPlatform/python-samples-owners members only (might not be good for some of the self service teams)
 3) We duplicate the `assign_issues_by` section - hard to keep in sync + we may get people assigned who are unaware of the expectations related to being an assignee

Thoughts?

rsamborski (changes requested):

.github/blunderbuss.yml
```diff
@@ -13,220 +13,145 @@
...
+      - leahecole
+      - rachael-ds
+      - rafalbiegacz
+  - labels:
+      - "api: datascienceonramp"
+    to:
+      - leahecole
+      - bradmiro
+  - labels:
+      - "api: contentwarehouse"
```
rsamborski, line 113: `contentwarehouse` and `documentai` were previously assigned to `data-ai`. Is there a reason for assigning them to an individual?

.github/blunderbuss.yml
kweinmeister: I'm thinking option 3, with comments at the top of both sections (and maybe even the bottom of the file :) ) that both sections need to be updated.

Justification: this is only the assignee. If they see it, great - they have SME knowledge and will be flagged early in the process. If not, a Python docs samples reviewer will. And the process requires 2 reviewers anyway.

.github/blunderbuss.yml
kweinmeister: Document AI ownership is in progress. How about we use this team for now? hhttps://github.com/orgs/GoogleCloudPlatform/teams/document-ai-samples-contributors/members

We can then consider a `document-ai-samples-owners` if needed in the future.

rsamborski: SGTM, please update.

.github/blunderbuss.yml
```diff
@@ -12,221 +12,290 @@
...
+      - "api: auth"
+    to:
+      - arithmetic1728
+
 assign_issues:
   - GoogleCloudPlatform/python-samples-owners
 
 assign_prs:
   - GoogleCloudPlatform/python-samples-owners
+
```
rsamborski, line 297: Minor nit: probably makes sense to move this comment at the end of assign_prs_by section and above the generic assign_issues and assign_prs.

rsamborski (approved):

engelke (approved):

PR #9723, opened 2023-04-20
ci: update default flakybot priority to p2

kweinmeister:
Please review and reject if you would like to keep policy at p1 for new Flakybot issues.

### DIFF
--- a/.github/flakybot.yaml
+++ b/.github/flakybot.yaml
@@ -1,0 +1,15 @@
+# Copyright 2023 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


leahecole:
I would be okay with this as a temporary measure given that we are inundated with failures that are related to maintenance/growing pains and not actual service failures, but I think that once the periodics are stable, it should probably be bumped back to p1. What do other folks think?

engelke:
I agree with leahcole

engelke (approved):

kweinmeister:
Closing for now, thanks for reviewing @leahecole and @engelke

PR #9743, opened 2023-04-20
fixit: Changes to DocAI Samples to simplify maintenence

holtskinner:
- Consolidated process/batch process samples with processor versions
- Update Quickstart to include Processor Creation
- Consolidate Handle Response Samples into a single file to reduce code duplication
  - ~Not sure if I did the region tags completely correct, is there a way to test?~
  - Update: Tested using Devsite staging and it looks correct
- Moved Operations Tests into a single file and removed exception handling in sample
- Add type updates from fixit #9991


documentai/snippets/quickstart_sample_test.py
```diff
@@ -16,25 +16,42 @@
 # flake8: noqa
```
engelke, line 16: Don't disable flake8 entirely without a comment explaining why it was necessary.

engelke (changes requested):
Don't disable flake8 entirely

documentai/snippets/quickstart_sample_test.py
holtskinner: I tried to disable just one error (ANN201 for return values, not sure if there's a way to do this for the whole file instead of adding it to one line) because I'm returning a value outside the displayed sample and I don't want people looking at the docs to get confused.

documentai/snippets/batch_process_documents_sample.py
```diff
@@ -92,11 +98,11 @@ def batch_process_documents(
 
     # Continually polls the operation until it is complete.
     # This could take some time for larger files
-    # Format: projects/PROJECT_NUMBER/locations/LOCATION/operations/OPERATION_ID
+    # Format: projects/{project_id}/locations/{location}/operations/{operation_id}
     try:
         print(f"Waiting for operation {operation.operation.name} to complete...")
         operation.result(timeout=timeout)
-    # Catch exception when operation doesn't finish before timeout
+    # Catch exception when operation doesn"t finish before timeout
```
yil532, line 105: Remove redundant ' character

holtskinner: Done

documentai/snippets/handle_response_sample.py
```diff
@@ -0,0 +1,394 @@
...
+    for page in document.pages:
+        print(f"Page {page.page_number}:")
+        print_page_dimensions(page.dimension)
+        print_detected_langauges(page.detected_languages)
+        print_paragraphs(page.paragraphs, text)
+        print_blocks(page.blocks, text)
+        print_lines(page.lines, text)
+        print_tokens(page.tokens, text)
+
+        # Currently supported in version `pretrained-ocr-v1.1-2022-09-12`
```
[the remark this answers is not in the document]
holtskinner: Currently, this feature is only enabled in that version, but not the one that is newer (unless you send a specific parameter that is only enabled on the v1beta3 endpoint...) But supposedly this feature is going to be added to more versions later this quarter.

I could change this to say "only supported in some processor versions, refer to https://cloud.google.com/document-ai/docs/processors-list#processor_doc-ocr for more information" but I wanted to be more specific.

holtskinner:
I will also make the docs update to update the sample files.

engelke (approved):

PR #9762, opened 2023-04-21
chore: remove alixhami from blunderbuss

alixhami:
This is no longer part of my work (left BigQuery DPE team in 2019)


alixhami:
Nevermind - it looks like this update is covered by another PR: https://github.com/GoogleCloudPlatform/python-docs-samples/pull/9720

PR #9783, opened 2023-04-24
feat: upgrade sample style with pyupgrade

kweinmeister:
Applied [pyupgrade](https://github.com/asottile/pyupgrade) to samples. Used the version parameter corresponding to the minimum supported version of the sample.


leahecole:
@kweinmeister I know this is a draft but is it feasible to break this into smaller PRs when the time comes?

PR #9818, opened 2023-04-26
fix: use common test project

engelke:
## Description

Fixes completes using a common project for all Python version tests


rsamborski:
Compute tests are hitting quota. `test_bulk_create` is trying to spin up 5 instances with 4 VCPUs each (20 VCPUs). With 5 different python versions we need at least 80 VCPUs if all run in parallel. Unfortunately the zone has a quota of 72. I suggest updating [this line](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/edb0f178896b8d55a2f544b5fe6bbf7ed870a2e9/compute/client_library/snippets/tests/test_bulk.py#L45) to:

```
    template.properties.machine_type = "n1-standard-2"
```

This should keep us safe up to 7 parallel Python versions (70 VCPUs). 

The issue with ```test_add_extended_memory_to_instance``` seems to be quota related as well, but the error message is unclear. We could tweak the memory size [in here](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/d302daed2a6740212165379a0145418318b4788f/compute/client_library/snippets/tests/test_custom_types.py#L156) to see if it helps.

leahecole:
> Compute tests are hitting quota. `test_bulk_create` is trying to spin up 5 instances with 4 VCPUs each (20 VCPUs). With 5 different python versions we need at least 80 VCPUs if all run in parallel. Unfortunately the zone has a quota of 72. I suggest updating [this line](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/edb0f178896b8d55a2f544b5fe6bbf7ed870a2e9/compute/client_library/snippets/tests/test_bulk.py#L45) to:
> 
> ```
>     template.properties.machine_type = "n1-standard-2"
> ```
> 
> This should keep us safe up to 7 parallel Python versions (70 VCPUs).
> 
> The issue with `test_add_extended_memory_to_instance` seems to be quota related as well, but the error message is unclear. We could tweak the memory size [in here](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/d302daed2a6740212165379a0145418318b4788f/compute/client_library/snippets/tests/test_custom_types.py#L156) to see if it helps.


just noting that this may cause errors in periodic tests related to quota because some tests that aren't in this PR will also affect this quota

PR #9867, opened 2023-05-02
fix: update demosite to reflect error messages in client

Sita04:
1. Update returned `response` from backend to include `reason` for "bad" actions.
2. Update client messaging to display the bad action messages.
3. Update client response to fetch 'Invalid token' value based on the reason returned.


recaptcha_enterprise/demosite/app/static/demo-6df0841a.js
```diff
@@ -4185,10 +4187,17 @@ class RecaptchaDemo extends s {
...
-          <img alt="Bad" src="${badbad}" />
-        `;
+        if (this.reason) {
+          card = x`
+            <p>Suspicious request. Reason: "${this.reason}".</p>
+            <img alt="Bad" src="${badbad}" />
+          `;
+        } else {
+          card = x`
+            <p>reCAPTCHA is ${percentage || "???"}% confident you're not bad.</p>
```
faube, line 4197: Do we need to update this to support a score of 0 ?

recaptcha_enterprise/demosite/app/urls.py
```diff
@@ -61,19 +74,29 @@ def on_homepage_load() -> Response:
...
         # <!-- ATTENTION: reCAPTCHA Example (Server Part 1/2) Ends -->
 
-        # Return the risk score.
+            # Below code is only used to send response to the client for demo purposes.
+            # DO NOT send scores or other assessment response to the client.
+            if not assessment_response.token_properties.valid:
+                reason = Error.INVALID_TOKEN.value
+            elif assessment_response.token_properties.action != recaptcha_action:
+                reason = Error.ACTION_MISMATCH.value
+            elif assessment_response.risk_analysis.score < SAMPLE_THRESHOLD_SCORE:
```
faube, line 90: <=

Sita04: Thanks for the catch.

recaptcha_enterprise/demosite/app/static/demo-6df0841a.js
Sita04: Actually, I have removed the else case. Because all BAD actions will have an associated reason with it.
Hence, the else case will never get executed.

Kindly review the error message if this is how it should be worded.
`Suspicious request. Reason: `

recaptcha_enterprise/demosite/app/static/global-e680a49614fd8ff8.js
```diff
@@ -41,9 +41,10 @@ function fetchServerResponse({ body, url }) {
 // This code is internal to the demo.
 // It passes the score to the demo to display it.
 function useAssessmentInClient(score) {
-  if (score?.data?.score && score?.data?.verdict) {
+  if (score?.data?.score && score?.data?.label) {
     const demoElement = document.querySelector("recaptcha-demo");
     demoElement.setAttribute("score", score?.data?.score);
-    demoElement.setAttribute("verdict", score?.data?.verdict);
+    demoElement.setAttribute("verdict", score?.data?.label);
```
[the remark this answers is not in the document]
Sita04: @morewry Could you PTAL at this change?

faube (approved):

recaptcha_enterprise/demosite/app/urls.py
```diff
@@ -332,3 +285,26 @@ def on_comment_submit() -> Response:
...
+    if not assessment_response.token_properties.valid:
+        reason = Error.INVALID_TOKEN.value
+        label = Label.BAD.value
+
+    # Check if the returned recaptcha action matches the expected.
+    elif assessment_response.token_properties.action != recaptcha_action:
+        reason = Error.ACTION_MISMATCH.value
+        label = Label.BAD.value
+
+    # Check if the returned score is greater less than or equal to the threshold set.
```
faube, line 305: "greater less than or equal" -> "greater than"

faube (approved):

PR #9902, opened 2023-05-05
fix: wait to get id_token until needed

engelke:
## Description

Fixes failing period test


nicain (approved):

engelke:
Tests were failing with an HTTP 403 Forbidden status code. The token should be valid when created, or an exception would have happened. So it seems it must have expired before use. Expiration takes an hour, but it does wait to build and deploy a service, which might be long enough.

PR #9905, opened 2023-05-05
chore: Add an exception that will break the test

nicain:
As a reminder to re view the workaround from PR #9903 when it is no longer needed


run/idp-sql/e2e_test.py
```diff
@@ -18,6 +18,7 @@
 import json
 import os
 import subprocess
+from datetime import date
```
dandhlee, line 21: Can we use `datetime.datetime` instead? https://github.com/GoogleCloudPlatform/python-docs-samples/blob/main/AUTHORING_GUIDE.md#datetimedatetime-objects

engelke (approved):

PR #9939, opened 2023-05-12
feat: upgrade sample style (Infra)

kweinmeister:
Applied [pyupgrade](https://github.com/asottile/pyupgrade) to use f-strings and newer Python features.

### DIFF
--- a/compute/api/create_instance.py
+++ b/compute/api/create_instance.py
@@ -153,16 +153,16 @@ def main(project, bucket, zone, instance_name, wait=True):
 
     instances = list_instances(compute, project, zone)
 
-    print('Instances in project {} and zone {}:'.format(project, zone))
+    print(f'Instances in project {project} and zone {zone}:')
     for instance in instances:
-        print(' - ' + instance['name'])
+        print(f' - {instance["name"]}')
 
-    print("""
+    print(f"""
 Instance created.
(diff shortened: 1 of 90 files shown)


rsamborski (approved):
LGTM. @m-strzelczyk PTAL and make sure this doesn't break your code generation tooling before we merge.

compute/api/create_instance.py
```diff
@@ -154,7 +153,7 @@ def main(project, bucket, zone, instance_name, wait=True):
 
     instances = list_instances(compute, project, zone)
 
-    print('Instances in project %s and zone %s:' % (project, zone))
+    print('Instances in project {} and zone {}:'.format(project, zone))
```
m-strzelczyk, line 156: I guess if we run this next time, it will transform to f-string :D 

%s -> format() -> f-string can happen only one step at a time ;)

m-strzelczyk (approved):

compute/api/create_instance.py
m-strzelczyk: I fixed this manually.

m-strzelczyk:
I don't know why the test for 3.7 fails. It doesn't seem to be related to the changes at all. Will investigate more tomorrow if it isn't fixed by then.

PR #9942, opened 2023-05-12
feat: upgrade sample style (Cloud SDK)

kweinmeister:
Applied [pyupgrade](https://github.com/asottile/pyupgrade) to use f-strings and newer Python features.


yil532 (approved):
LGTM

dandhlee (approved):
What a great tool! LGTM for Python style.

nit: could you break these into multiple PRs if there's a lot to review? These type of PRs don't take long to review but just want to make sure nothing slips out and we're okay upgrading these - for example, `.format()` syntax is still preferred for logging libraries rather than f-strings.

PR #9952, opened 2023-05-15
fix: add region tag, move code around for docs

iennae:
pair with @nicain; this is not typical usage of if __name__ == .. but
    to make sure documentation and intent of the sample for atlas this
    should work.

## Checklist

### DIFF
--- a/run/pubsub/main.py
+++ b/run/pubsub/main.py
@@ -13,7 +13,6 @@
 # limitations under the License.
 
 # [START cloudrun_pubsub_server_setup]
-# [START run_pubsub_server_setup]
 # [START cloudrun_pubsub_server]
 import base64
 import os
@@ -30,7 +29,6 @@ if __name__ == "__main__":
     # application on Cloud Run. See entrypoint in Dockerfile.
     app.run(host="127.0.0.1", port=PORT, debug=False)
 
-# [END run_pubsub_server_setup]
 # [END cloudrun_pubsub_server_setup]
 # [END cloudrun_pubsub_server]
 
--- a/run/pubsub/main.py
+++ b/run/pubsub/main.py
@@ -12,7 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# [START cloudrun_pubsub_server_setup]
 # [START cloudrun_pubsub_server]
 import base64
 import os
@@ -29,7 +28,6 @@ if __name__ == "__main__":
     # application on Cloud Run. See entrypoint in Dockerfile.
     app.run(host="127.0.0.1", port=PORT, debug=False)
 
-# [END cloudrun_pubsub_server_setup]
 # [END cloudrun_pubsub_server]
 
 
--- a/run/pubsub/main.py
+++ b/run/pubsub/main.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# [START cloudrun_pubsub_server_setup]
 # [START cloudrun_pubsub_server]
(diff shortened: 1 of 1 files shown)


nicain (approved):

run/pubsub/main.py
```diff
@@ -14,15 +14,25 @@
...
 
 
 app = Flask(__name__)
+
+if __name__ == "__main__":
+    PORT = int(os.getenv("PORT")) if os.getenv("PORT") else 8080
+
+    # This is used when running locally. Gunicorn is used to run the
+    # application on Cloud Run. See entrypoint in Dockerfile.
+    app.run(host="127.0.0.1", port=PORT, debug=True)
```
dandhlee, line 31: Just confirming you want this to be `debug=True`?

run/pubsub/main.py
```diff
@@ -14,15 +14,25 @@
 
 # [START cloudrun_pubsub_server_setup]
 # [START run_pubsub_server_setup]
```
dandhlee, line 16: `run_pubsub_server_setup` is not used.
```suggestion
```

run/pubsub/main.py
```diff
@@ -14,15 +14,25 @@
...
 app = Flask(__name__)
+
+if __name__ == "__main__":
+    PORT = int(os.getenv("PORT")) if os.getenv("PORT") else 8080
+
+    # This is used when running locally. Gunicorn is used to run the
+    # application on Cloud Run. See entrypoint in Dockerfile.
+    app.run(host="127.0.0.1", port=PORT, debug=True)
+
 # [END run_pubsub_server_setup]
```
dandhlee, line 33:
```suggestion
```

run/pubsub/main.py
```diff
@@ -14,15 +14,25 @@
 
 # [START cloudrun_pubsub_server_setup]
```
dandhlee, line 15: Is there a difference between `cloudrun_pubsub_server_setup` and `cloudrun_pubsub_server`?

iennae: not in python, but if I don't have both region tags then audits will say cloudrun_pubsub_server is missing.

run/pubsub/main.py
iennae: updated in last commit

run/pubsub/main.py
averikitsch: We only need to use one region tag. We don't always have complete 1:1 matches across languages

run/pubsub/main.py
```diff
@@ -13,16 +13,24 @@
...
 app = Flask(__name__)
-# [END run_pubsub_server_setup]
+
+if __name__ == "__main__":
+    PORT = int(os.getenv("PORT")) if os.getenv("PORT") else 8080
+
+    # This is used when running locally. Gunicorn is used to run the
+    # application on Cloud Run. See entrypoint in Dockerfile.
+    app.run(host="127.0.0.1", port=PORT, debug=False)
+
```
averikitsch, line 31: A python expert may weigh in but it may be more idiomatic to have this at the bottom of the file and duplicate the region tags around it

iennae: I didn't encompass this in the description but that's what I was pairing with Nick on before. Proposed solution was the second best (first was creating a separate file).

engelke: The if __name__ section is not being used in the sample. It's being launched by gunicorn in the Dockerfile and in Cloud Run. It would be better to remove it than move it.

engelke (approved):
LGTM

PR #9953, opened 2023-05-15
fix: install only binary with pip

engelke:
fixes: pip intermittently installing from source, and failing


kweinmeister:
Nice! This might speed up builds, too.

kweinmeister (approved):

engelke:
I'm running the fully nightly tests for each version of Python (the presubmit is only testing the specific file in the PR). So far, so good. This hasn't broken anything yet. Given how long the whole suite takes, we'll look at it in the morning.

PR #9954, opened 2023-05-16
fix: Add missing unit test for tuning.py sample.

nicain:
Add missing unit test for tuning.py sample.

Prompt tuning is initiated from a temporarily cached copy of the training jsonl file (with a UUID name) to guarantee that a separate tuning run is initiated for each time a test job is run. This prevents Vertex Pipeline from looking up a cached version of the tuning job, which would not exercise the functionality.


yil532 (approved):
LGTM

generative_ai/tuning_test.py
```diff
@@ -0,0 +1,96 @@
...
+            for endpoint in aiplatform.Endpoint.list():
+                for _ in endpoint.list_models():
+                    if endpoint.display_name == display_name:
+                        endpoint.undeploy_all()
+                        endpoint.delete()
+            aiplatform.Model(model_registry.model_resource_name).delete()
+
+
+def test_tuning(training_data_filename):
+    """Takes approx. 40 minutes."""
```
rsamborski, line 82: This will substantially extend the time for our test suite to complete (especially for periodic runs). Do we plan to limit execution of these tests to single language or run them less often (i.e. once a week)? WDYT?

kweinmeister: +1. At a minimum, this would be a good situation to test only on endpoints 3.7 and 3.11, by excluding other versions in the noxfile. We should consider going beyond that.

generative_ai/tuning.py
```diff
@@ -44,26 +46,27 @@ def tuning(
...
     model.tune_model(
-      training_data=training_data,
-      # Optional:
-      train_steps=train_steps,
-      tuning_job_location="europe-west4",
-      tuned_model_location="us-central1",
+        training_data=training_data,
+        # Optional:
+        train_steps=train_steps,
+        tuning_job_location="europe-west4",
```
kweinmeister, line 63: Should `tuning_job_location` be pulled out as a constant, or a default parameter?

generative_ai/requirements.txt
```diff
@@ -1,4 +1,7 @@
 pandas==1.1.5; python_version < '3.7'
 pandas==1.3.5; python_version == '3.7'
 pandas==2.0.1; python_version > '3.7'
-google-cloud-aiplatform[pipelines]>=1.25.0
\ No newline at end of file
+shapely<2.0.0
+google-cloud-aiplatform[pipelines]>=1.25.0
+google-auth==2.17.3; python_version < '3.0'
```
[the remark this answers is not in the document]
nicain: Done.

The shapely pinning was necessary for my local test, I will experiment with it a little more until I get a config that works.

kweinmeister (approved):

PR #9957, opened 2023-05-16
fix: Add type hints for secret manager snippets

glasnt:
## Description

Fixes #<ISSUE-NUMBER>

## Checklist

### DIFF
--- a/secretmanager/snippets/add_secret_version.py
+++ b/secretmanager/snippets/add_secret_version.py
@@ -19,8 +19,8 @@ specified payload to an existing secret.
 
 import argparse
 
-import google_crc32c  # type: ignore
 from google.cloud import secretmanager
+import google_crc32c  # type: ignore
 
 
 # [START secretmanager_add_secret_version]
--- a/secretmanager/snippets/add_secret_version.py
+++ b/secretmanager/snippets/add_secret_version.py
@@ -19,11 +19,14 @@ specified payload to an existing secret.
 
 import argparse
 
-import google_crc32c
+import google_crc32c  # type: ignore
+from google.cloud import secretmanager
 
 
 # [START secretmanager_add_secret_version]
-def add_secret_version(project_id, secret_id, payload):
+def add_secret_version(
+    project_id: str, secret_id: str, payload: str
+) -> secretmanager.SecretVersion:
     """
     Add a new secret version to the given secret with the provided payload.
     """
@@ -39,25 +42,27 @@ def add_secret_version(project_id, secret_id, payload):
 
     # Convert the string payload into a bytes. This step can be omitted if you
     # pass in bytes instead of a str for the payload argument.
-    payload = payload.encode("UTF-8")
+    payload_bytes = payload.encode("UTF-8")
 
     # Calculate payload checksum. Passing a checksum in add-version request
     # is optional.
     crc32c = google_crc32c.Checksum()
-    crc32c.update(payload)
+    crc32c.update(payload_bytes)
 
     # Add the secret version.
     response = client.add_secret_version(
         request={
             "parent": parent,
-            "payload": {"data": payload, "data_crc32c": int(crc32c.hexdigest(), 16)},
+            "payload": {
+                "data": payload_bytes,
+                "data_crc32c": int(crc32c.hexdigest(), 16),
+            },
         }
     )
 
     # Print the new secret version name.
     print(f"Added secret version: {response.name}")
     # [END secretmanager_add_secret_version]
-
     return response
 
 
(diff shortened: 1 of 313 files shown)


secretmanager/snippets/add_secret_version.py
```diff
@@ -39,25 +42,27 @@ def add_secret_version(project_id, secret_id, payload):
 
     # Convert the string payload into a bytes. This step can be omitted if you
     # pass in bytes instead of a str for the payload argument.
-    payload = payload.encode("UTF-8")
+    payload_bytes = payload.encode("UTF-8")
```
glasnt, line 45: mypy saw this as a type mismatch because of the self-assigning type change. Renaming the variable helps with clarity.

secretmanager/snippets/quickstart.py
```diff
@@ -65,4 +65,4 @@ def quickstart(_project_id=None, _secret_id=None):
 
 
 if __name__ == "__main__":
-    quickstart()
+    quickstart("YOUR_PROJECT_ID", "YOUR_SECRET_ID")
```
glasnt, line 68: I had issues trying to use `str | None` and have valid mypy tests, so I found just providing defaults here helped. I don't _think_ this quickstart is called directly, but we'll see if tests pass 🤞

engelke: I had something similar with defaults. I had to change some defaults from None to "falsy" values of other types, like "", {}, []... This looks really good (and defaulting them to None doesn't seem to have been a good idea anyway).

secretmanager/snippets/snippets_test.py
```diff
@@ -48,22 +49,24 @@
...
-def iam_user():
+def iam_user() -> str:
     return "serviceAccount:" + os.environ["GCLOUD_SECRETS_SERVICE_ACCOUNT"]
 
 
 @pytest.fixture()
-def secret_id(client, project_id):
+def secret_id(
+    client: secretmanager.SecretManagerServiceClient, project_id: str
+) -> Generator[str, None, None]:
```
engelke, line 69: Is Generator right here? It looks like it's yielding a str to me. What am I missing?

secretmanager/snippets/snippets_test.py
```diff
@@ -79,7 +82,9 @@ def secret_id(client, project_id):
 
 
 @pytest.fixture()
-def secret(client, project_id, secret_id):
+def secret(
+    client: secretmanager.SecretManagerServiceClient, project_id: str, secret_id: str
+) -> Generator[Tuple[str, str, str], None, None]:
```
engelke, line 87: Same question about Generator here.

kweinmeister (approved):

engelke (approved):
Wow, what a PR! Looks good. I had a couple of questions about the Generator type. If they're right, please add a note helping me understand them. Not requesting changes - use your judgment on any comments I made. Thanks!

secretmanager/snippets/snippets_test.py
glasnt: I originally tried `-> str`, but since this method doesn't return anything, that doesn't make sense. 

mypy says `The return type of a generator function should be "Generator" or one of its supertypes`

Adding just `Generator[str]` causes `"Generator" expects 3 type arguments, but 1 given`. Typing documentation for [Generator](https://docs.python.org/3/library/typing.html#typing.Generator) says you can use a triple for `[yield type, send type, return type]`, which is what I used. But that feels overly verbose

Reading further, I could just use `Iterator[str]` which looks cleaner, and mypy is happy.

secretmanager/snippets/snippets_test.py
glasnt: Same as above

secretmanager/snippets/consume_event_notification.py
```diff
@@ -20,7 +20,7 @@
 
 
 # [START secretmanager_consume_event_notification]
-def consume_event_notification(event, unused_context):
+def consume_event_notification(event: dict, unused_context: None) -> str:
```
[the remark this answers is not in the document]
glasnt: Yes, Any would be better here. I used `None` as that's what was passed by the test.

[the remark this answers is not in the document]
glasnt: Ah, it wouldn't be better. From Kokoro:

```
Dynamically typed expressions (typing.Any) are disallowed
```

glasnt:
@engelke comments agreed, tests green. Not sure why I have merge permissions on this repo, but I'll let !me merge

engelke:
Happy you have merge permission, but I'll go ahead now and merge it.

PR #7810, opened 2022-04-21
adding owlbot and notebook ci

loferris:
## Description

Fixes #7809

Note: It's a good idea to open an issue first for discussion.

## Checklist
- [x] Please **merge** this PR for me once it is approved.

PR #10011, opened 2023-05-19
chore: Add Owlbot config to repository

holtskinner:
- Add Owlbot config to repository to automatically format code using `blacken` on Pull Requests.
- Not sure if this was intentionally left out for some reason, feel free to reject if thats the case

### DIFF
--- a/.github/.OwlBot.lock.yaml
+++ b/.github/.OwlBot.lock.yaml
@@ -1,0 +1,17 @@
+# Copyright 2022 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+docker:
+  image: gcr.io/cloud-devrel-public-resources/owlbot-python:latest
+  digest: sha256:b3500c053313dc34e07b1632ba9e4e589f4f77036a7cf39e1fe8906811ae0fce
--- a/.github/.OwlBot.yaml
+++ b/.github/.OwlBot.yaml
@@ -1,0 +1,16 @@
+# Copyright 2022 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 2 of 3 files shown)


engelke (commented):
Can you explain why this is needed? And get all tests to pass, please.

I'm not familiar with the synthtool library, and could not find documentation to answer my question: what is `s.move(templated_files / "noxfile.py")` doing. Is the library overloading the division operator?

kweinmeister:
Just FYI, this was started in an open PR #7810 but not yet completed.

FYI @parthea @loferris

parthea:
> Can you explain why this is needed? 

+1. We should add a note explaining what will happen when the owlbot configuration is added (eg. code formatting will run on every PR). We should also update contributors docs to let folks know that owlbot runs automatically on PRs which branched off `main`, and for branches from forks the `owlbot:run` label needs to be added for owlbot to run.

> I'm not familiar with the synthtool library, and could not find documentation to answer my question: what is 
s.move(templated_files / "noxfile.py") doing. Is the library overloading the division operator?

Regretfully, documentation for synthtool is lacking. I've added some answers below but let me know if I can clarify things further

```
templated_files = gcp.CommonTemplates().py_library()

# Copy the standard noxfile from templated_files
s.move(templated_files / "noxfile.py")
```

The source code for `py_library` is [here](https://github.com/googleapis/synthtool/blob/70ec2609beefb006579939d8231acdfb6c32fe5a/synthtool/gcp/common.py#L241). Regretfully `py_library` doesn't have a docstring that explains what it does. My understanding is that it returns a path to the [googleapis python templates](https://github.com/googleapis/synthtool/tree/master/synthtool/gcp/templates/python_library) which are part of synthtool. 

The source code `s.move` is [here](https://github.com/googleapis/synthtool/blob/70ec2609beefb006579939d8231acdfb6c32fe5a/synthtool/transforms.py#L170).  `s.move()` will move a file specified in `sources` to the current directory.

`sources` can be a `Path`, in that case a [slash operation](https://docs.python.org/3/library/pathlib.html#operators) is supported.

For example, `Path("test") / "testing"` works

```
>>>from pathlib import Path
>>>example_path : Path = Path("test") / "testing"
>>> example_path
PosixPath('test/testing')
```

balajismaniam (approved):
I've reviewed the OwlBot related files. LGTM. The header and snippet-bot checks are failing but seems to be from unrelated from the changes made by OwlBot.

holtskinner:
#10108 fixes the issue with the snippet-bot prefixes for legacy App Engine samples.

kweinmeister (approved):

holtskinner:
Any recommendations on the Kokoro tests? It doesn't look like the failed, just got stuck/never started or completed.

holtskinner:
Not sure what the issues are with the Tests. `generative_ai` seems to fail consistently, but the tests are actually passing, it seems like another error happens after the tests during cleanup (or something along those lines)

Some other tests like `airflow_2022` and cloud run seem to fail intermittently, but not on every run.

I don't think these are due to the owlbot changes, but the tests just haven't been run in a while.

PR #10121, opened 2023-05-29
fix: mock flaky dialogflow test cases

kweinmeister:
Update flaky test cases that are failing the build, with mocks that match the Dialogflow service interface.

### DIFF
--- a/dialogflow/conversation_profile_management_test.py
+++ b/dialogflow/conversation_profile_management_test.py
@@ -15,53 +15,105 @@
 from __future__ import absolute_import
 
 import os
+from unittest import mock
+
+from google.cloud import dialogflow_v2beta1 as dialogflow
+import pytest
 
 import conversation_profile_management
+import test_utils
 
 PROJECT_ID = os.getenv("GOOGLE_CLOUD_PROJECT")
 
 CONVERSATION_PROFILE_DISPLAY_NAME = "fake_conversation_profile_name"
+CONVERSATION_PROFILE_NAME = f"conversationProfiles/{CONVERSATION_PROFILE_DISPLAY_NAME}"
 
 
-def test_create_conversation_profile(capsys):
-    # Check the conversation profile does not yet exists.
-    conversation_profile_management.list_conversation_profiles(PROJECT_ID)
+@pytest.fixture(scope="function")
+def mock_conversation():
+    yield mock.MagicMock(
+        return_value=test_utils.create_mock_conversation(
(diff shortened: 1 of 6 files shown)


dialogflow/conversation_profile_management_test.py
```diff
@@ -15,53 +15,105 @@
...
+        mock_conversation,
+    ):
+        conversation_profile_management.get_conversation_profile(
+            PROJECT_ID, conversation_profile_id
+        )
 
-    out, _ = capsys.readouterr()
-    assert "Display Name: {}".format(CONVERSATION_PROFILE_DISPLAY_NAME) in out
+        out, _ = capsys.readouterr()
+        assert "Display Name: {}".format(CONVERSATION_PROFILE_DISPLAY_NAME) in out
```
andrewferlitsch, line 100: recommend using the f"{}" formatted style

andrewferlitsch (approved):

dialogflow/create_document_test.py
```diff
@@ -15,50 +15,50 @@
...
+def test_create_document(
+    capsys: pytest.CaptureFixture[str],
+    mock_create_document_operation: mock.MagicMock,
+):
+    with mock.patch(
+        "google.cloud.dialogflow_v2beta1.DocumentsClient.create_document",
+        mock_create_document_operation,
+    ):
+        document_management.create_document(
+            PROJECT_ID,
```
nicain, line 56: Since we are mocking out the create_document service, we can replace PROJECT_ID with a MOCK_PROJECT_ID, that is not a real project.

dialogflow/participant_management_test.py
```diff
@@ -13,20 +13,53 @@
...
+        DOCUMENT_DISPLAY_NAME,
+        KNOWLEDGE_BASE_ID,
+        MIME_TYPE,
+        [getattr(dialogflow.Document.KnowledgeType, KNOWLEDGE_TYPE)],
+        CONTENT_URI,
+    )
 
 
-def test_analyze_content_text(capsys):
+def test_analyze_content_text(capsys, mock_create_document_operation, mock_document):
```
nicain, line 62: Does mock_create_document_operation need to be a fixture?

kweinmeister: It doesn't have to be, but it is initializing a shared test resource in another file with some constants. I think it's fine as-is, but please feel free to suggest another approach if needed.

kweinmeister: Update: I changed it to a constant.

balajismaniam (approved):

PR #10201, opened 2023-06-08
Remove torus blunderbuss

rogerthatdev:
This PR will remove blunderbuss configuration that auto assigns issues and PRs to the torus-dpe team.

Reasoning: automatic assignment does not align with current team triage processes

### DIFF
--- a/.github/blunderbuss.yml
+++ b/.github/blunderbuss.yml
@@ -18,18 +18,6 @@
 assign_issues_by:
   # DEE teams
   - labels:
-      - "api: appengine"
-      - "api: cloudbuild"
-      - "api: cloudscheduler"
-      - "api: cloudtasks"
-      - "api: containeranalysis"
-      - "api: eventarc"
-      - "api: functions"
-      - "api: run"
-      - "api: workflows"
-    to:
-      - GoogleCloudPlatform/torus-dpe
-  - labels:
       - "api: batch"
       - "api: compute"
       - "api: cloudkms"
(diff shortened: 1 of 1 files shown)


.github/blunderbuss.yml
```diff
@@ -17,18 +17,6 @@
...
-      - "api: cloudbuild"
-      - "api: cloudscheduler"
-      - "api: cloudtasks"
-      - "api: containeranalysis"
-      - "api: eventarc"
-      - "api: functions"
-      - "api: run"
-      - "api: workflows"
-    to:
-      - GoogleCloudPlatform/torus-dpe
```
dandhlee, line 31: If they're not replaced and only removed this the issues will get lost and not assigned.

We'll need some buy-in from the existing sample owners, but we should at least change these to be owned by the sample owners - or better if you could find another team to take these over.

kweinmeister: Agreed. We should have an internal discussion about this, thinking about both team and language perspectives.

Removing a team here just assigns issues back to the general reviewer pool. Then, there's extra triage for the Python reviewers to find the right person with the product skills needed.

By having a team assigned, the issue will automatically go to a team member who is closer to knowing who can best resolve the issue. It doesn't mean that they must fix the issue themselves, as they can reassign to one of their peers.

rogerthatdev: Thanks for the prompt reviews!

We don't want to abandon these issues, we just don't want blunderbuss to assign them to each and every person on our team. Instead we'd like to be able to triage these issues as they appear in our dashboards.  

I didn't know that leaving them without a label defined blunderbuss would assign the issues to the general reviewer pool.  If that is indeed the case we can have an internal chat as suggested to see how to proceed.

msampathkumar:
Hi, any new updates to share?

rogerthatdev:
We have decided to leave torus blunderbuss configuration in tact for now with a smaller change in the future. Closing this out. 

Thanks!

PR #10234, opened 2023-06-14
[DLP] Implemented deidentifying and reidentifying of table using fpe

Avani-Thakker-Crest:
## Description
Implemented deidentifying and reidentifying of table using fpe. This Pull Request includes 2 sample codes:
1. [dlp_deidentify_table_fpe](https://cloud.google.com/dlp/docs/samples/dlp-deidentify-table-fpe#dlp_deidentify_table_fpe-java)
2. [dlp_reidentify_table_fpe](https://cloud.google.com/dlp/docs/samples/dlp-reidentify-table-fpe#dlp_reidentify_table_fpe-java)

Fixes #<ISSUE-NUMBER>

## Checklist


dlp/snippets/deid.py
```diff
@@ -1938,6 +1938,195 @@ def deidentify_table_with_multiple_crypto_hash(
...
+# [START dlp_deidentify_table_fpe]
+import base64  # noqa: F811, E402, I100
+from typing import Dict, List, Union  # noqa: F811, E402, I100
+
+import google.cloud.dlp  # noqa: F811, E402
+
+
+def deidentify_table_with_fpe(
+    project: str,
+    table_data: Dict[str, Union[List[str], List[List[str]]]],
```
soumya92, line 1950: Can we ask for headers and data in two different parameters? This union type seems hacky, since it's basically asking for two parameters but in a dict form.

dlp/snippets/deid.py
```diff
@@ -1938,6 +1938,195 @@ def deidentify_table_with_multiple_crypto_hash(
...
+        table_data: Dictionary representing table data.
+        deid_content_list: A list of fields in table to de-identify.
+        key_name: The name of the Cloud KMS key used to encrypt ('wrap') the
+            AES-256 key. Example:
+            key_name = 'projects/YOUR_GCLOUD_PROJECT/locations/YOUR_LOCATION/
+            keyRings/YOUR_KEYRING_NAME/cryptoKeys/YOUR_KEY_NAME'
+        wrapped_key: The encrypted ('wrapped') AES-256 key to use. This key
+            should be encrypted using the Cloud KMS key specified by key_name.
+        alphabet: The set of characters to replace sensitive ones with. For
+            more information, see https://cloud.google.com/dlp/docs/reference/
```
soumya92, line 1970: https://cloud.google.com/dlp/docs/reference/rest/v2/projects.deidentifyTemplates#ffxcommonnativealphabet

dlp/snippets/deid.py
```diff
@@ -1938,6 +1938,195 @@ def deidentify_table_with_multiple_crypto_hash(
...
+
+    table = {"headers": headers, "rows": rows}
+
+    # Construct the `item` for table.
+    item = {"table": table}
+
+    # Specify fields to be de-identified.
+    deid_content_list = [{"name": _i} for _i in deid_content_list]
+
+    # The wrapped key is base64-encoded, but the library expects a binary
```
soumya92, line 1992: Can we ask for 'bytes' here instead of a base64 string?

dlp/snippets/deid.py
```diff
@@ -1938,6 +1938,195 @@ def deidentify_table_with_multiple_crypto_hash(
...
+import base64  # noqa: F811, E402, I100
+from typing import Dict, List, Union  # noqa: F811, E402, I100
+
+import google.cloud.dlp  # noqa: F811, E402
+
+
+def deidentify_table_with_fpe(
+    project: str,
+    table_data: Dict[str, Union[List[str], List[List[str]]]],
+    deid_content_list: List[str],
```
soumya92, line 1951: Can this be named `deid_field_names`? I think this is a list of field names, but with "content" it's confusing and could be interpreted as infotypes instead of field names.

soumya92: (same for `reid_` below)

dlp/snippets/deid.py
```diff
@@ -1938,6 +1938,195 @@ def deidentify_table_with_multiple_crypto_hash(
...
+# [START dlp_reidentify_table_fpe]
+import base64  # noqa: F811, E402, I100
+from typing import Dict, List, Union  # noqa: F811, E402, I100
+
+import google.cloud.dlp  # noqa: F811, E402
+
+
+def reidentify_table_with_fpe(
+    project: str,
+    table_data: Dict[str, List[str]],
```
soumya92, line 2045: Same comment on table_data. Here the type is also not correct, since "rows" is a `list[list[str]]`, so you'll need the union again, which is why two params might be better.

dlp/snippets/deid_test.py
```diff
@@ -537,3 +537,52 @@ def test_deidentify_table_with_multiple_crypto_hash(capsys: pytest.CaptureFixtur
...
+        key_name=KEY_NAME,
+    )
+
+    out, _ = capsys.readouterr()
+    assert "11111" not in out
+    assert "22222" not in out
+
+    response = out.split(":")[1:]
+
+    deid_col_id = response.index(' "employee_id"\n}\nheaders {\n  name')
```
soumya92, line 567: This is relying on python's default output formatting? If we need output to be in a specific format it would be better to do that in the sample code.

soumya92 (changes requested):

dlp/snippets/deid.py
Avani-Thakker-Crest: My miss. Updated with two parameters now.

dlp/snippets/deid.py
Avani-Thakker-Crest: Updated.

dlp/snippets/deid.py
Avani-Thakker-Crest: Yes, updated.

dlp/snippets/deid.py
Avani-Thakker-Crest: You meant passing in the raw bytes, already decoded bytes?

soumya92: Yes. On the command line it makes sense to accept base64, but in code we should pass around bytes to make things simpler.

dlp/snippets/deid_test.py
Avani-Thakker-Crest: Thank you for your response. I understand the concern, but to print table in known format, we will have to parse using the same code in sample itself. With that said, if the formatting of proto changes, this will affect the parser written in the sample code. So with either idea, we are dependent on the proto. Please let me know if I am missing anything here. Thank you!

dlp/snippets/deid.py
Avani-Thakker-Crest: I have updated for deid sample for now as per your suggestion, Please check if that is what is expected, If yes, I will update the same for reid sample too. Thank you!

soumya92: Yep, looks good. Thanks!

dlp/snippets/deid.py
```diff
@@ -2040,6 +2040,188 @@ def deidentify_table_with_multiple_crypto_hash(
...
+        })
+
+    # Print out results.
+    print(f"Table after de-identification: {response.item.table}")
+
+
+# [END dlp_deidentify_table_fpe]
+
+
+# [START dlp_reidentify_table_fpe]
```
rsamborski, line 2135: Region tags may be duplicated in a single file and docs will put join them before publication, so you could consider having all imports in the beginning of the file surrounded with all region tags which use them later. This way `noqa` comments could be avoided.

See an example here: https://github.com/GoogleCloudPlatform/python-docs-samples/blob/4687063b99a5906a8fc58d3911660984271a0fb3/media_cdn/snippets.py#L24

rsamborski (approved):

dlp/snippets/deid.py
```diff
@@ -2040,6 +2040,188 @@ def deidentify_table_with_multiple_crypto_hash(
...
+            "item": item,
+        })
+
+    # Print out results.
+    print("Table after re-identification: {}".format(response.item.table))
+
+
+# [END dlp_reidentify_table_fpe]
+
 if __name__ == "__main__":
```
rsamborski, line 2225: Please fix lint:
```
./deid.py:2225:1: C901 'If 2225' is too complex (21)
if __name__ == "__main__":
^
```

Avani-Thakker-Crest: Hi that is because, max-complexity is set to 20, and we have more than 20 samples now  in deid.py file. There are two solutions:
1. increase max_complexity in noxfile.py
2. create two files for deid related samples. 
Please suggest. Thank you

rsamborski: Our sample guidelines recommend using a single file per sample ([see here](https://googlecloudplatform.github.io/samples-style-guide/#one-per-file)). My suggestion is to split it and try to adhere to the above rule in the future. No need to refactor existing samples as there is already a lot of them and that could take long time.

dlp/snippets/deid.py
Avani-Thakker-Crest: Hi, Thank you for reviewing. Imports for every sample was added as part of one of the PRs. https://github.com/GoogleCloudPlatform/python-docs-samples/pull/10051/files Hence every sample is written that way. Please confirm if I should go ahead and make changes the way you suggested. Thank you!

rsamborski: As all samples are written this way we can leave it for now. Nevertheless please consider this in the future.

dlp/snippets/deid.py
Avani-Thakker-Crest: Yes, the entire python sample files are maintained that way, We can go ahead and refactor entire dlp folder with 1 snippet/file file, but that is going to take time. Currently lint check is failing because of this complexity, kindly suggest how should I move ahead. Thank you!

dlp/snippets/deid.py
Avani-Thakker-Crest: Sure got it, thank you

dlp/snippets/deid.py
rsamborski: Please follow @kweinmeister's suggestion (shared in internal chat with @annadata):

> A good starting point, with a low level of effort, would be to analyse the samples for any themes/clusters.  You 
> could then break out the samples into 2+ files, each with a common theme.

I would start with this one and redo any other ones as you extend / modify them. Let me know if you have any further questions.

annadata: @Avani-Thakker-Crest Please create an issue in backlog with the details of this issue and recommendation to break the samples into their own files.
For all new samples, please create new files.

Avani-Thakker-Crest: @annadata @rsamborski I created a new file named: deid_table.py (thinking to put all the table related samples in the new file ). I started adding all the table related samples to the new file, but then felt to confirm things first and wanted avoid conflicts with the main branch, hence for now, I have added these two samples (dlp_deidentify_table_fpe and dlp_reidentify_table_fpe) in the new file deid_table.py to get rid of complexity issues, please check.

rsamborski: LG from my side with a suggestion to decouple tests as well by putting them in `deid_table_test.py`. If there are any common functions used in `deid_test.py` and `deid_table_test.py` you could put it a helper file.

Avani-Thakker: okay got it

Avani-Thakker-Crest: Added new file deid_table_test.py as per the suggestion. Thank you!

PR #10282, opened 2023-06-21
fix: Dialogflow CX detect intent flaky test

yil532:
## Description

Fixes #<ISSUE-NUMBER>

## Checklist


dialogflow-cx/detect_intent_with_sentiment_analysis_test.py
```diff
@@ -39,4 +39,4 @@ def test_detect_intent_positive(text, expected_score_min, expected_score_max):
         text,
         "en-us",
     )
-    assert expected_score_min < score < expected_score_max
+    assert -1 < score < 1
```
dandhlee, line 42: Modify line 32 instead of hardcoding it here

yil532: It looks like hard coded but actually it's not :) It has to be within the [-1, 1] range. I feel it's not necessary to create two more variables... WDYT?

dandhlee: If this is a temporary change then fine as is, otherwise for readability (knowing what these values are) and to allow changing this value based on the parameterized test, we should favor having variables than not.

kweinmeister: I think what Dan's saying is to just change the 0.5 to a -1 in line 32? (I may be mistaken) Then we won't have to create any new variables. We should also consider removing "not happy" parameters if we're not testing for that.

Note that this updated range is now testing whether the output is within the allowable values (-1 to 1), not whether the sentiment itself matches. (Fine by me, just clarifying this)

yil532: Awesome! Got it.

dialogflow-cx/detect_intent_with_sentiment_analysis_test.py
```diff
@@ -29,7 +29,7 @@
 
 @pytest.mark.parametrize(
     "text, expected_score_min, expected_score_max",
-    (["Perfect", 0.5, 1], ["I am not happy", -1, -0.5]),
+    (["Perfect", 0, 1], ["I am not happy", -1, -0]),
```
dandhlee, line 32:
```suggestion
    (["Perfect", 0, 1], ["I am not happy", -1, 0]),
```

dandhlee (approved):
Looks good for Python style.

kweinmeister (approved):

PR #10440, opened 2023-07-20
docs(samples): Improve output in create_migration_workflow.py sample

davidkhala:
After casting response.state from a number type to Enum type cannot make it readable. Better to use name of Enum instead.

Copy from https://github.com/googleapis/python-bigquery-migration/pull/181

### DIFF
--- a/bigquery-migration/snippets/create_migration_workflow.py
+++ b/bigquery-migration/snippets/create_migration_workflow.py
@@ -67,7 +67,7 @@ def create_migration_workflow(
     print("Created workflow:")
     print(response.display_name)
     print("Current state:")
-    print(response.State(response.state))
+    print(response.State(response.state).Name)
 
 
 # [END bigquery_migration_create_workflow]


dandhlee:
In the future, please open an issue first so we can discuss if this change is necessary - corresponding documentation around this sample and unit/integration tests might need to be updated as well, which would be a lot more complex work than just a one line change.

msampathkumar:
@davidkhala - This request is open for a long time. Kindly share if you have any new udpates.

davidkhala:
Hi @msampathkumar , I don't have enough time to resolve those Kokoro CI issue. So please see this as an issue-like PR. I am fine if someone else can include this change by his own sign-off and get it resolved in another PR. If that can happen, I will close this.

msampathkumar:
Error response of Python3.9 test
```
  File "/workspace/bigquery-migration/snippets/create_migration_workflow_test.py", line 62, in test_create_migration_workflow
    create_migration_workflow.create_migration_workflow(path, path, PROJECT_ID)
  File "/workspace/bigquery-migration/snippets/create_migration_workflow.py", line 70, in create_migration_workflow
    print(response.State(response.state).Name)
AttributeError: 'State' object has no attribute 'Name'****
```

Suggestion: To ignore unrequired CI for python versions like 2.7, or other, use a noxfile config. Example: https://github.com/GoogleCloudPlatform/python-docs-samples/blob/main/generative_ai/noxfile_config.py

PR #10441, opened 2023-07-20
test: E2E test on endpoint versions only

kweinmeister:
To reduce testing overhead and costs, update default test matrix to endpoints. Currently these are 3.7 and 3.11, but since Python 3.7 is EOL on 6/27/2003, these endpoints have been set to 3.8 and 3.11.

This ensures that dependency conflicts are detected. The earliest and latest versions are checked.

Samples with non-standard test configurations are not affected by this change.


kweinmeister:
A couple test cases are failing on unrelated issues.

rsamborski (commented):
According to https://devguide.python.org/versions/, Python 3.7 is EOL from 2023-06-27. I suggest to turn off 3.7 testing completely and keep testing on 3.8 and 3.11 by default. I suggest to agree on this change with the language team before proceeding.

kweinmeister:
> According to https://devguide.python.org/versions/, Python 3.7 is EOL from 2023-06-27. I suggest to turn off 3.7 testing completely and keep testing on 3.8 and 3.11 by default. I think it should be agreed with the language team before proceeding.

Updated to 3.8 & 3.11 based on feedback. Python team can review and provide feedback before merging. Thanks!

rsamborski (approved):

dandhlee:
It might require some changes to how we run the tests, but would it be now then better to change `ignored_versions` option to `test_additional_versions`? Now we'd expect less products to run version specific tests other than the endpoint versions - which we can also control much easily when there's endpoint changes - and from a very quick skim it seems like all the changes are mostly for default version changes.

kweinmeister:
@dandhlee I like your proposal. How do you feel about approaching it in 2 steps?
1. Updating the tested versions using the current noxfile config
2. Refactoring noxfile config

We can then verify each step independently, including ensuring that step 2 tests the same versions as 1. Step 2 may also have some edge cases that we haven't thought about yet. I am open either way though, if you feel we should combine those in this PR. Thanks!

rsamborski:
> @rsamborski - I should update this PR to remove 3.7 references after we merge #10453? IIUC, they should just be noise now, since 3.7 won't be tested.

Feel free to leave 3.7 as is (noise). We can remove it together with noxfile config refactoring (step 2 from your list above).

m-strzelczyk (approved):

kweinmeister:
Approvals in place. A couple tests failed due to unrelated issues. Admin-merging.

PR #10442, opened 2023-07-20
Minor Python fix

KillingSeason:
## Description

When following this Python code sample from the Cloud Functions GCP documentation [1], I observed this error message when it was running on Cloud Functions [2].

This situation seems to occur when Python sends, automatically and by default, the self argument to the function even though the function was not expecting any parameter as input when invoked. To fix this issue, adding 'self' to the function 'connect_tcp_socket()' should fix the issue (e.g. cloud_tcp_docket(self)).    

[1]
https://cloud.google.com/sql/docs/postgres/connect-functions#private-ip:~:text=SQL%20Auth%20Proxy.-,Connect%20with%20TCP,-Connect%20using%20the

[2]
TypeError: connect_tcp_socket() takes 0 positional arguments but 1 was
given"

### DIFF
--- a/cloud-sql/postgres/sqlalchemy/connect_tcp.py
+++ b/cloud-sql/postgres/sqlalchemy/connect_tcp.py
@@ -22,7 +22,7 @@ import ssl
 import sqlalchemy
 
 
-def connect_tcp_socket() -> sqlalchemy.engine.base.Engine:
+def connect_tcp_socket(self) -> sqlalchemy.engine.base.Engine:
     """Initializes a TCP connection pool for a Cloud SQL instance of Postgres."""
     # Note: Saving credentials in environment variables is convenient, but not
     # secure - consider a more secure solution such as


jackwotherspoon:
@KillingSeason The `self` argument is only used when a function or method is part of a Python Class. The `connect_tcp_socket` is a standalone function and is not a class method, thus it should not have the `self` argument. 

The code block you are mentioning is meant to be called __from__ a Cloud Function, it is __not__ the Cloud Function itself.

The Cloud Function code that calls `connect_tcp_socket` can be found in `main.py`:
https://github.com/GoogleCloudPlatform/python-docs-samples/blob/main/cloud-sql/postgres/sqlalchemy/main.py

As you can see the Cloud Function takes in a `request` argument as its 1 expected argument, not self.

jackwotherspoon:
Closing as answer above should resolve the user's concerns.

dandhlee:
@KillingSeason in the future, please file an issue first before submitting a pull request, "minor" fixes often require much more than just adding tiny bits into the samples.

PR #10464, opened 2023-07-25
re-add removed file - fixes buganizer bug 288143688

leahecole:
## Description

I removed this file as part of removing the `data-science-onramp` directory but it turns out that this CSV was used in another tutorial and I broke a hyperlink in the process. This re-adds it and adds the info about the source data to fix buganzier bug 288143688


leahecole:
cc @mgtcardenas

rsamborski (approved):

kweinmeister (approved):

PR #10479, opened 2023-07-28
fix(docker): Hardcoding virtualenv version to keep Python 2.7 support

m-strzelczyk:
Since version [20.22 virtualenv no longer supports Python 2.7](https://virtualenv.pypa.io/en/latest/changelog.html#v20-22-0-2023-04-19). We hardcode virtualenv version to keep Python 2.7 while we work on deprecating it from our repositories.

### DIFF
--- a/.kokoro/tests/run_tests.sh
+++ b/.kokoro/tests/run_tests.sh
@@ -90,6 +90,9 @@ export PATH="${HOME}/.local/bin:${PATH}"
 # upgrade pip when needed
 pip install --upgrade pip
 
+# Install virtualenv in version 20.21, since version 20.22 no longer supports Python 2.7
+pip install --user -q virtualenv==20.21
+
 # install nox for testing
 pip install --user -q nox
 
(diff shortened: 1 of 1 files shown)


rsamborski (approved):

m-strzelczyk:
Tests seem to be failing due to quota problems. Should be safe to merge.

.kokoro/tests/run_tests.sh
```diff
@@ -90,6 +90,9 @@ export PATH="${HOME}/.local/bin:${PATH}"
 # upgrade pip when needed
 pip install --upgrade pip
 
+# Install virtualenv in version 20.21, since version 20.22 no longer supports Python 2.7
+pip install --user -q virtualenv==20.21
+
```
dandhlee, line 95: Can we split them for 2.7? There's a runtime specific condition above:
```
if [[ "${RUN_TESTS_SESSION}" == "py-3.8" ]]; then
```

m-strzelczyk: Sure, done!

dandhlee (approved):

PR #10482, opened 2023-07-28
chore: use new team alias in CODEOWNERS

enocom:
We're now the infra-db-sdk team. infra-db-dpes will be removed eventually


enocom:
We'll need to update repo access as well (hence the GitHub errors).

dandhlee (commented):
Could you also update the alias in https://github.com/GoogleCloudPlatform/python-docs-samples/blob/main/.github/blunderbuss.yml

rsamborski (approved):
Approved assuming Dan's comment regarding Blunderbuss is implemented.

meredithslota:
> Could you also update the alias in https://github.com/GoogleCloudPlatform/python-docs-samples/blob/main/.github/blunderbuss.yml

Fixed.

dandhlee (approved):
Thanks!

PR #10576, opened 2023-08-30
fix: lazy init db connection for Cloud Function

jackwotherspoon:
Cloud Function global variables that perform background tasks should be lazy initialized so that the initialization is performed within the request context.

### DIFF
--- a/cloud-sql/sql-server/sqlalchemy/main.py
+++ b/cloud-sql/sql-server/sqlalchemy/main.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from flask import render_template, Response
+from flask import render_template, Request, Response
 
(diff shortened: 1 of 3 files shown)


enocom (approved):

cloud-sql/sql-server/sqlalchemy/main.py
```diff
@@ -12,22 +12,28 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from flask import render_template, Response
+from flask import render_template, Request, Response
```
dandhlee, line 15: nit: Use import statements for packages and modules only, not for individual classes or functions. (https://google.github.io/styleguide/pyguide#22-imports)

Feel free to follow up in a different PR instead.

dandhlee (approved):

PR #10622, opened 2023-09-13
feat(storagetransfer): Adds event driven transfer samples

JesseLovelace:
Adds event driven transfer samples and tests

### DIFF
--- a/storagetransfer/event_driven_aws_transfer.py
+++ b/storagetransfer/event_driven_aws_transfer.py
@@ -1,0 +1,121 @@
+#!/usr/bin/env python
+
+# Copyright 2023 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+Command-line sample that creates aan event driven transfer between two GCS buckets that tracks an AWS SQS queue.
+"""
+
+import argparse
+
+# [START storagetransfer_create_event_driven_gcs_transfer]
+
+from google.cloud import storage_transfer
+
+
+def create_event_driven_aws_transfer(
+        project_id: str,
+        description: str,
+        source_s3_bucket: str,
+        sink_gcs_bucket: str,
+        sqs_queue_arn: str,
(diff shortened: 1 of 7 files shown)


danielbankhead (approved):

storagetransfer/event_driven_aws_transfer.py
```diff
@@ -0,0 +1,121 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+Command-line sample that creates aan event driven transfer between two GCS buckets that tracks an AWS SQS queue.
+"""
+
+import argparse
+
+# [START storagetransfer_create_event_driven_gcs_transfer]
```
msampathkumar, line 23:
```suggestion
# [START storagetransfer_create_event_driven_aws_transfer]
```

storagetransfer/event_driven_aws_transfer.py
```diff
@@ -0,0 +1,121 @@
...
+def create_event_driven_aws_transfer(
+    project_id: str,
+    description: str,
+    source_s3_bucket: str,
+    sink_gcs_bucket: str,
+    sqs_queue_arn: str,
+    aws_access_key_id: str,
+    aws_secret_access_key: str,
+):
+    """Create an event driven transfer between two GCS buckets that tracks an AWS SQS queue"""
```
msampathkumar, line 37: Suggestion: Instead of using body function to explain the arguments(project_id, description, source_s3_bucket, ...), you can do the following.

```
Creates an event drive transfer between GCS bucket using AWS SQS queue

Arguments:
* project_id(str): Google Cloud Project id. Ex: 'my-project-2356'
* description(str): Description for the event driven transfer job
...
...
```

storagetransfer/event_driven_gcs_transfer.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+def create_event_driven_gcs_transfer(
+    project_id: str,
+    description: str,
+    source_bucket: str,
+    sink_bucket: str,
+    pubsub_id: str,
+):
+    """Create an event driven transfer between two GCS buckets that tracks a PubSub subscription"""
```
msampathkumar, line 35: Similar suggestion as above.

storagetransfer/event_driven_aws_transfer.py
JesseLovelace: It's done the way it is here on purpose--it's so that the user can simply copy the function body of the code including the comments, then uncomment the lines that set the variables. It's how all the other transfer service samples are set up, so I'd prefer to keep it this way

msampathkumar: I understand. Thank you.

msampathkumar (approved):
LGTM.

msampathkumar:
@JesseLovelace You can merged this PR now. If any issues arise, let us know.

PR #10629, opened 2023-09-16
fix: refactor error reporting code sample

minherz:
align code sample for error reporting with code in other languages (b/275730159):

- use the "same" exception type and message
- refactor the method name

### DIFF
--- a/error_reporting/snippets/report_exception.py
+++ b/error_reporting/snippets/report_exception.py
@@ -17,13 +17,13 @@
 
 # [START error_reporting_quickstart]
 # [START error_reporting_setup_python]
-def simulate_error():
+def report_exception():
     from google.cloud import error_reporting
 
     client = error_reporting.Client()
     try:
         # simulate calling a method that's not defined
-        raise NameError
+        raise Exception('Something went wrong')
     except Exception:
         client.report_exception()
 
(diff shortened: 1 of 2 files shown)


error_reporting/snippets/report_exception_test.py
```diff
@@ -18,7 +18,7 @@
 
 
 def test_error_sends():
-    report_exception.simulate_error()
+    report_exception.report_exception()
```
rsamborski, line 21: suggestion: This test doesn't seem to `test` anything besides running the code. Probably not in scope of this change, but could we log an issue to extend this test to check if error_reporting actually worked?

rsamborski (approved):
Approved with a suggestion for future improvement.

error_reporting/snippets/report_exception.py
```diff
@@ -17,13 +17,13 @@
 
 # [START error_reporting_quickstart]
 # [START error_reporting_setup_python]
-def simulate_error():
+def report_exception():
     from google.cloud import error_reporting
 
     client = error_reporting.Client()
     try:
         # simulate calling a method that's not defined
```
muncus, line 25: nit: update this comment, since we're using a generic exception now.

muncus (approved):

error_reporting/snippets/report_exception.py
minherz: Reported as #10631

minherz: remove the comment because the code is simple and self-explanatory.

PR #10633, opened 2023-09-18
fix: add constraint to the age of import log range

### LINKED ISSUE #10632: Limit range dates for log import to be less than 30 days from the current date
### Is your feature request related to a problem? Please describe.

The current version of the import logs solution allows to import logs of any age.
However, the Logging API requires ingested logs not to be older than 30 days.
Logs with older timestamps are not stored and no error is returned.

### Describe the solution you'd like.

To prevent no error results when logs are not imported, the import range should be limited to the interval between now() and now()-30 days. 

### Describe alternatives you've considered.

Support for any import range requires storing the original timestamp in a user label and using now() for the timestamp.
This solution means the data of the imported log is modified which is not desired.

### Context
The issue is related to the code stored under `/logging/import-logs`

minherz:
Ensure that neither start nor end dates of the range are older than 29 days from now.

Fixes #10632

### DIFF
--- a/logging/import-logs/main.py
+++ b/logging/import-logs/main.py
@@ -79,6 +79,18 @@ def _day(blob_name: str) -> int:
     return int(blob_name[offset : offset + 2])
 
 
+def _is_valid_import_range() -> bool:
+    if START_DATE > END_DATE:
+        eprint("Start date of the import time range should be earlier than end date")
+        return False
+    # avoid importing time range older than 29 days
+    today = date.today()
+    if (today - START_DATE).days > 29 or (today - END_DATE).days > 29:
+        eprint("Date range includes days older than 29 days from today.")
+        return False
+    return True
+
+
 def calc_import_range() -> Tuple[date, date]:
     """Calculate import range for the task based on full import range and number of tasks"""
     if TASK_COUNT == 1:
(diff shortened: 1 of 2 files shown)


logging/import-logs/main.py
```diff
@@ -79,6 +79,18 @@ def _day(blob_name: str) -> int:
     return int(blob_name[offset : offset + 2])
 
 
+def _is_valid_import_range() -> bool:
```
rsamborski, line 82: nit: for consistency with other functions in this file please add a docstring explaining what this function is for.

logging/import-logs/main.py
```diff
@@ -79,6 +79,18 @@ def _day(blob_name: str) -> int:
...
+    if START_DATE > END_DATE:
+        eprint("Start date of the import time range should be earlier than end date")
+        return False
+    # avoid importing time range older than 29 days
+    today = date.today()
+    if (today - START_DATE).days > 29 or (today - END_DATE).days > 29:
```
rsamborski, line 88: nit: the second part of this statement (`(today - END_DATE).days > 29`) will never be true, because START_DATE > END_DATE, so it could be skipped.

rsamborski (approved):
Approved with minor nits.

NimJay (approved):

logging/import-logs/main.py
minherz: I will move the comment to the docstring. Can you confirm that private methods with few lines of code still require a docstring?

logging/import-logs/main.py
minherz: nit: I agree but it does not influence a performance and the statement is more verbose this way.

logging/import-logs/main.py
rsamborski: I don't see a strict rule in sample guide, but taking into account that all other functions had such docstring (even private ones) I have noticed it.

logging/import-logs/main.py
rsamborski: SG. Thanks for taking a look.

PR #10639, opened 2023-09-21
Additional code samples

valentinhuerta1996:
Code sample to list documents under a folder and remove a document from a folder

## Description

Fixes #<ISSUE-NUMBER>

## Checklist

### DIFF
--- a/contentwarehouse/snippets/delete_folder_document_link_sample.py
+++ b/contentwarehouse/snippets/delete_folder_document_link_sample.py
@@ -1,0 +1,53 @@
+# Copyright 2023 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+
+# [START delete_folder_document_link_sample]
+import sys
+from google.cloud import contentwarehouse
+# TODO(developer): Uncomment these variables before running the sample.
+# document_folder_link_name = "YOUR_DOCUMENT_LINK_NAME" # The name of the document-link to be deleted Format: projects/{project_number}/locations/{location}/documents/{source_document_id}/documentLinks/{document_link_id}
+# document_name = "YOUR_DOCUMENT_NAME" # The name of the document to be deleted Format: projects/{project_number}/locations/{location}/documents/{target_document_id
+# user_id = "user:xxxx@example.com" # Format is "user:xxxx@example.com"
+
+def delete_folder_document_link(document_folder_link_name: str, document_name: str, user_id: str) -> None:
+
+    # Create a Link Service client
+    link_client = contentwarehouse.DocumentLinkServiceClient()
+
+    # Initialize delete document linked request
+    delete_link_request = contentwarehouse.DeleteDocumentLinkRequest(
+        name=document_folder_link_name,
+        request_metadata=contentwarehouse.RequestMetadata(
+            user_info=contentwarehouse.UserInfo(id=user_id))
+    )
+
+    # Make the delete request
+    link_client.delete_document_link(request=delete_link_request)
+
+     # Initialize list linked targets request to verify deletion
+    linked_targets_request = contentwarehouse.ListLinkedTargetsRequest(
+        parent=document_name,
+        request_metadata=contentwarehouse.RequestMetadata(
+            user_info=contentwarehouse.UserInfo(id=user_id))
+    )
+
+    # Make the request
+    linked_targets_response = link_client.list_linked_targets(request=linked_targets_request)
+
+    #linked_targets_response returns False if link does not exist
+    if bool(linked_targets_response) is False:
+        print(f"Document Link to Folder Deleted Successfully")
+# [END delete_folder_document_link_sample]
--- a/contentwarehouse/snippets/delete_folder_document_link_sample_test.py
+++ b/contentwarehouse/snippets/delete_folder_document_link_sample_test.py
@@ -1,0 +1,33 @@
+# # Copyright 2023 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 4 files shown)


contentwarehouse/snippets/delete_folder_document_link_sample.py
```diff
@@ -0,0 +1,53 @@
...
+#
+
+
+# [START delete_folder_document_link_sample]
+import sys
+from google.cloud import contentwarehouse
+# TODO(developer): Uncomment these variables before running the sample.
+# document_folder_link_name = "YOUR_DOCUMENT_LINK_NAME" # The name of the document-link to be deleted Format: projects/{project_number}/locations/{location}/documents/{source_document_id}/documentLinks/{document_link_id}
+# document_name = "YOUR_DOCUMENT_NAME" # The name of the document to be deleted Format: projects/{project_number}/locations/{location}/documents/{target_document_id
+# user_id = "user:xxxx@example.com" # Format is "user:xxxx@example.com"
```
dandhlee, line 23: If the purpose was to use this by defining and initializing the variables, this should be done inside the caller method. Otherwise users would have to define and initialize the variables as a global variable then make use of it again inside the caller method.

If the purpose was to document them, prefer using docstrings instead.

contentwarehouse/snippets/delete_folder_document_link_sample.py
```diff
@@ -0,0 +1,53 @@
...
+        parent=document_name,
+        request_metadata=contentwarehouse.RequestMetadata(
+            user_info=contentwarehouse.UserInfo(id=user_id))
+    )
+
+    # Make the request
+    linked_targets_response = link_client.list_linked_targets(request=linked_targets_request)
+
+    #linked_targets_response returns False if link does not exist
+    if bool(linked_targets_response) is False:
```
dandhlee, line 51: This doesn't seem right but I could be wrong.

If `linked_targets_response` does return False then you shouldn't have to cast to a boolean, and can simply do `if not linked_targets_response:`. If it doesn't return False then the comment should be updated to explain so.

contentwarehouse/snippets/delete_folder_document_link_sample.py
```diff
@@ -0,0 +1,53 @@
...
+        print(f"Document Link to Folder Deleted Successfully")
```
dandhlee, line 52: f-string not necessary if there's nothing to evaluate.

contentwarehouse/snippets/delete_folder_document_link_sample_test.py
```diff
@@ -0,0 +1,33 @@
...
+#
+
+import os
+
+from contentwarehouse.snippets import delete_folder_document_link_sample
+import pytest
+
+document_folder_link_name = "YOUR_DOCUMENT_LINK_NAME" # The name of the document-link to be deleted Format: projects/{project_number}/locations/{location}/documents/{source_document_id}/documentLinks/{document_link_id}
+document_name = "YOUR_DOCUMENT_NAME" # The name of the document to be deleted Format: projects/{project_number}/locations/{location}/documents/{target_document_id
+user_id = "user:xxxx@example.com" # Format is "user:xxxx@example.com"
```
dandhlee, line 23: If these variables are used only once then prefer using them directly in the tests, especially in the tests. Avoid logic in tests.

contentwarehouse/snippets/delete_folder_document_link_sample.py
```diff
@@ -0,0 +1,53 @@
...
+import sys
+from google.cloud import contentwarehouse
+# TODO(developer): Uncomment these variables before running the sample.
+# document_folder_link_name = "YOUR_DOCUMENT_LINK_NAME" # The name of the document-link to be deleted Format: projects/{project_number}/locations/{location}/documents/{source_document_id}/documentLinks/{document_link_id}
+# document_name = "YOUR_DOCUMENT_NAME" # The name of the document to be deleted Format: projects/{project_number}/locations/{location}/documents/{target_document_id
+# user_id = "user:xxxx@example.com" # Format is "user:xxxx@example.com"
+
+def delete_folder_document_link(document_folder_link_name: str, document_name: str, user_id: str) -> None:
+
+    # Create a Link Service client
```
dandhlee, line 27: Use complete sentences for comments rather than short forms.
```suggestion
    # Create a Link Service client.
```

contentwarehouse/snippets/list_documents_folder_sample.py
```diff
@@ -0,0 +1,194 @@
...
+        display_name="My Test Folder",
+        document_schema_name=folder_schema.name,
+    )
+
+    # Define Request to create Folder
+    create_folder_request=contentwarehouse.CreateDocumentRequest(
+        parent=parent,
+        document=folder,
+        request_metadata=contentwarehouse.RequestMetadata(
+            user_info=contentwarehouse.UserInfo(id=user_id))
```
dandhlee, line 64: each list item laid out vertically should always end with a comma - some languages don't do this but this is preferred style for Python.
```suggestion
            user_info=contentwarehouse.UserInfo(id=user_id)),
```

contentwarehouse/snippets/list_documents_folder_sample.py
```diff
@@ -0,0 +1,194 @@
...
+    )
+
+    # Create a Folder for the given schema
+    folder_response=folder_client.create_document(request=create_folder_request)
+
+    # Read the output
```
dandhlee, line 70: What output should the user expect? If it's very self explanatory then fine to leave as-is, otherwise it might be good to put expected responses.

contentwarehouse/snippets/list_documents_folder_sample.py
```diff
@@ -0,0 +1,194 @@
...
+    print(f"Rule Engine Output: {folder_response.rule_engine_output}")
+    print(f"Folder Created: {folder_response.document}")
+
+    return folder_response
+
+def create_document(project_number: str, location: str, user_id: str) -> contentwarehouse.Document:
```
dandhlee, line 76: I can see that these snippets don't exist in the contentwarehouse directory unfortuantely, but they shouldn't belong only in `list_documents` sample unless that's the only use, which I'd be very surprised if it was.

Could you break these up into separate modules and then import them and use them instead? Otherwise users seeing the snippet for `list` would also have to see `create folder` and `create document` samples as well which might be irrelevant.

dandhlee (changes requested):
Reviewing for sample style.

contentwarehouse/snippets/delete_folder_document_link_sample.py
```diff
@@ -0,0 +1,53 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+
+# [START delete_folder_document_link_sample]
```
holtskinner, line 17:
```suggestion
# [START contentwarehouse_delete_folder_document_link_sample]
```

holtskinner (changes requested):

holtskinner:
Also, change the PR message to be more descriptive. Something like

`docs(contentwarehouse): Added List Documents and Delete Folder Samples`

PR #10652, opened 2023-09-22
chore(documentai): Add Premium Features to Document AI OCR sample

holtskinner:
- Adds samples for 
  - https://cloud.google.com/document-ai/docs/enterprise-document-ocr
  - https://cloud.google.com/document-ai/docs/ocr-add-ons

- Won't work until v1 endpoint enables premium features for python client library


documentai/snippets/handle_response_sample_test.py
```diff
@@ -42,6 +42,30 @@ def test_process_document_ocr(capsys):
...
     assert "FakeDoc" in out
 
+    # Document Quality
+    assert "Quality score" in out
+
+    # Font Detection
+    assert "Font Size" in out
+    assert "Handwritten" in out
+
+    # Checkbox Detection
```
rsamborski, line 52: suggestion: As this is another call to the service I suggest moving it out to a separate test.

rsamborski (approved):
Approved with one suggestion

documentai/snippets/handle_response_sample.py
```diff
@@ -162,17 +172,39 @@ def print_image_quality_scores(
...
+    print(
+        f"           Text Color (RGBa): {style_info.text_color.red}, {style_info.text_color.green}, {style_info.text_color.blue}, {style_info.text_color.alpha}"
+    )
+
+
+def print_visual_elements(
+    visual_elements: Sequence[documentai.Document.Page.VisualElement], text: str
+) -> None:
+    """
+    Only supported in version `pretrained-ocr-v2.0-2023-06-02`
```
kweinmeister, line 194: Will it continue to only be supported in this one specific version? Or should we generalize this to say something like "Only supported in version `pretrained-ocr-v2.0-2023-06-02` and later"?

kweinmeister (approved):

documentai/snippets/handle_response_sample.py
holtskinner: The product team hasn't confirmed one way or another. There's no guarantee it'll be supported in later versions.

PR #10730, opened 2023-10-04
chore(discoveryengine): Update Search Sample with more options

holtskinner:
- Include Content Search Spec, Query expansion, and spell correction
- Context: 
  - https://cloud.google.com/generative-ai-app-builder/docs/snippets
  - https://cloud.google.com/generative-ai-app-builder/docs/get-search-summaries

### DIFF
--- a/discoveryengine/search_sample.py
+++ b/discoveryengine/search_sample.py
@@ -17,7 +17,7 @@
 from typing import List
 
 from google.api_core.client_options import ClientOptions
-from google.cloud import discoveryengine
+from google.cloud import discoveryengine_v1 as discoveryengine
 
 # TODO(developer): Uncomment these variables before running the sample.
 # project_id = "YOUR_PROJECT_ID"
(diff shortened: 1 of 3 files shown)


discoveryengine/search_sample.py
```diff
@@ -52,16 +52,44 @@ def search_sample(
...
+        ),
     )
+
     response = client.search(request)
-    for result in response.results:
-        print(result)
+    print(response)
 
-    return response.results
+    return response
```
rsamborski, line 92: suggestion: please consider updating [the tests](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/00be56fecb3a8405b735a4e40910dfad115aeaca/discoveryengine/search_sample_test.py#L24), because it seems they only check if there is a response and even return type change didn't trigger failure.

rsamborski (approved):
Approved with a suggestion.

discoveryengine/search_sample.py
holtskinner: Done.

PR #10731, opened 2023-10-04
revert dependabot Airflow changes

leahecole:
As noted in chat, dependabot made some Airflow changes we actually didn't want!

### DIFF
--- a/composer/blog/gcp-tech-blog/data-orchestration-with-composer/requirements.txt
+++ b/composer/blog/gcp-tech-blog/data-orchestration-with-composer/requirements.txt
@@ -1,6 +1,6 @@
 # be sure to update the constraints file to match
 # see https://airflow.apache.org/docs/apache-airflow/stable/installation/installing-from-pypi.html#constraints-files
-apache-airflow[google]==2.7.1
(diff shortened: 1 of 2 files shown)


rsamborski (approved):

msampathkumar:
@leahecole To prevent this in future, why not move all info in requirements.txt to constrains.txt ?

leahecole:
@msampathkumar this is a very Airflow-y reason but the way that Airflow works, we have regular requirements like other python packages do - the actual Airflow package is one of them, but sometimes users need regular direct dependencies, like scipy, numpy, etc. They may want to install those in their Composer environments via a requirements.txt file. However, the constraints file is something that we need on the sample maintenance side to ensure that we are reproducing the same conditions and using the same indirect every time during testing. More in the [Airflow docs](https://airflow.apache.org/docs/apache-airflow/stable/installation/installing-from-pypi.html#why-we-need-constraints). If we had everything in one file yes, it would consistently work, but if users are looking to use our requirements files in their Composer environment, the files would be very bloated.

msampathkumar:
@leahecole  Thank you.

PR #10765, opened 2023-10-16
docs(discoveryengine): Create Multi-Turn Search Sample

holtskinner:
## Description

- Add Sample for Multi-Turn Search (Conversational Search)
- https://cloud.google.com/generative-ai-app-builder/docs/multi-turn-search

### DIFF
--- a/discoveryengine/multi_turn_search_sample_test.py
+++ b/discoveryengine/multi_turn_search_sample_test.py
@@ -15,43 +15,28 @@
 
 import os
 
-from discoveryengine import search_sample
+from discoveryengine import multi_turn_search_sample
 
 project_id = os.environ["GOOGLE_CLOUD_PROJECT"]
-search_query = "Google"
+search_queries = ["What is Google?", "What was their revenue in 2021?"]
 
 
-def test_search():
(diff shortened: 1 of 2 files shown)


discoveryengine/multi_turn_search_sample_test.py
```diff
@@ -0,0 +1,42 @@
...
+
+    assert responses
+
+    for response in responses:
+        assert response.reply
+        assert response.conversation
+        assert response.search_results
+
+        for result in response.search_results:
+            assert result.document.name
```
leahecole, line 42: Just double checking because I'm unfamiliar with the product - Is the reason you're just asserting truthiness that the response can vary? As long as these variables are defined, it's considered good?

kweinmeister (approved):

discoveryengine/multi_turn_search_sample_test.py
holtskinner: Correct, the response can vary depending on the data and how the search ranking orders the responses.

m-strzelczyk (approved):

PR #10767, opened 2023-10-17
feature: samples for Cloud Tasks http queues

engelke:


rsamborski (commented):
Is there a reason for putting these samples in `cloud_tasks` directory instead of `cloud_tasks/snippets` where others are already located?

iennae:
@rsamborski the snippets subdirectory is inherited from migration. I don't know if we have an explicit plan to put all snippets within a subdirectory of product.

cloud_tasks/http_queues.py
```diff
@@ -0,0 +1,229 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+# [START cloud_tasks_create_http_queue]
+import urllib
+
+import google.protobuf
+from google.cloud import tasks_v2beta3 as tasks
```
iennae, line 20: @engelke might want to flag or share why this is pulling from beta for now.

engelke:
Placing the new samples is tricky. Want to keep the HTTP queue ones clearly together, but since they all ended up in one file, that might be good enough. What do people think?

Will add comment about the beta version import (and use in HTTP endpoint URL).

Question: should imports be inside each region tag, or not? Lint's not happy, and needs pragmas to override it.

cloud_tasks/http_queues.py
```diff
@@ -0,0 +1,229 @@
...
+            http_target["uri_override"]["path_override"] = {"path": parsedUri.path}
+        if parsedUri.query:
+            http_target["uri_override"]["query_override"] = {
+                "query_params": parsedUri.query
+            }
+
+        queue.http_target = http_target
+        update_mask.paths.append("http_target")
+
+    if max_per_second != 0.0:
```
glasnt, line 134: Since these values are in the default variables, is there any reason why these have to be conditional assignments? Is there a reason why max_burst can't be send as it's default of 0, for example?

cloud_tasks/http_queues.py
engelke: Will do.

engelke:
> > Placing the new samples is tricky. Want to keep the HTTP queue ones clearly together, but since they all ended up in one file, that might be good enough. What do people think?
> 
> If we'd like to adhere to sample guidelines we should have a single region tag per file ([see here](https://googlecloudplatform.github.io/samples-style-guide/#one-per-file))
> 
> > Will add comment about the beta version import (and use in HTTP endpoint URL).
> > Question: should imports be inside each region tag, or not? Lint's not happy, and needs pragmas to override it.
> 
> Per [our guidelines](https://googlecloudplatform.github.io/samples-style-guide/#imports) imports should be part of the region tag. Lint is complaining, because the imports are not located in the top of the file and they're duplicated. I can suggest one of following approaches:
> 
> 1. Splitting samples to separate files
> 2. Duplicating region_tags to cover imports and different functions, see `mediacdn_dualtoken_sign_path_component` [as an example](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/b61f8bc8f4fcfa706c9489b7c4fb664ea2639a3b/media_cdn/dualtoken.py#L18C10-L18C48) for an example

Will do, once tests pass. Should pass now, with current library version in requirements.txt.

cloud_tasks/http_queues.py
engelke: I think at least one in the API treats 0 slightly different from the default.

cloud_tasks/http_queues/create_http_queue.py
```diff
@@ -0,0 +1,76 @@
...
+    parsedUri = urllib.parse.urlparse(uri)
+
+    http_target = {
+        "uri_override": {
+            "host": parsedUri.hostname,
+            "uri_override_enforce_mode": 2,  # ALWAYS use this endpoint
+        }
+    }
+    if parsedUri.scheme == "http":  # defaults to https
+        http_target["uri_override"]["scheme"] = 1
```
rsamborski, line 52: nit: can we substitute this by a constant (if one exists) to be more descriptive? Also if I read the proto file correctly `2` is used for HTTPS.

cloud_tasks/http_queues/delete_http_queue_test.py
```diff
@@ -0,0 +1,71 @@
...
+            },
+        )
+    )
+
+    yield queue
+
+    try:
+        client.delete_queue(name=queue.name)
+    except Exception as e:
+        print(f"Tried my best to clean up, but could not: {e}")
```
rsamborski, line 61: suggestion: if the test passes correctly and queue is deleted then we will see this error message. I suggest to not display this text if exception is "NotFound" which is expected at this point.

cloud_tasks/http_queues/create_http_queue_test.py
```diff
@@ -0,0 +1,42 @@
...
+# HTTP Queues are currently in public beta
+from google.cloud import tasks_v2beta3 as tasks
+
+import create_http_queue
+
+
+def test_create() -> None:
+    # Use the default project and a random name for the test queue
+    _, project = google.auth.default()
+    name = uuid.uuid4().hex
```
rsamborski, line 28: nit: can we add a prefix to all resources created during these tests so we can identify any leftoves in the future? For example by using `tests-tasks-v2beta3-` as a prefix.

cloud_tasks/http_queues/delete_http_queue_test.py
```diff
@@ -0,0 +1,71 @@
...
+
+HOST = "example.com"
+LOCATION = "us-central1"
+
+
+@pytest.fixture
+def q():
+    # Use the default project and a random name for the test queue
+    _, project = google.auth.default()
+    name = uuid.uuid4().hex
```
rsamborski, line 35: nit: ditto on the naming prefix

cloud_tasks/http_queues/get_http_queue_test.py
```diff
@@ -0,0 +1,68 @@
...
+
+HOST = "example.com"
+LOCATION = "us-central1"
+
+
+@pytest.fixture
+def q():
+    # Use the default project and a random name for the test queue
+    _, project = google.auth.default()
+    name = uuid.uuid4().hex
```
rsamborski, line 34: nit: ditto on the naming prefix

cloud_tasks/http_queues/update_http_queue.py
```diff
@@ -0,0 +1,89 @@
...
+        parsedUri = urllib.parse.urlparse(uri)
+
+        http_target = {
+            "uri_override": {
+                "host": parsedUri.hostname,
+                "uri_override_enforce_mode": 2,  # ALWAYS use this endpoint
+            }
+        }
+        if parsedUri.scheme == "http":  # defaults to https
+            http_target["uri_override"]["scheme"] = 1
```
rsamborski, line 59: nit: ditto on using a constant

cloud_tasks/http_queues/create_http_queue.py
```diff
@@ -0,0 +1,76 @@
...
+    # for an HTTP queue. For legacy reasons, this property is an object
+    # with multiple fields instead of a string containing a URI.
+
+    # Extract the various components of the URI provided by the caller
+    parsedUri = urllib.parse.urlparse(uri)
+
+    http_target = {
+        "uri_override": {
+            "host": parsedUri.hostname,
+            "uri_override_enforce_mode": 2,  # ALWAYS use this endpoint
```
rsamborski, line 48: nit: can we substitute this with a constant (if one exists)?

cloud_tasks/http_queues/update_http_queue.py
```diff
@@ -0,0 +1,89 @@
...
+    # Track which fields need to be updated
+    update_mask = {"paths": []}
+
+    if uri:
+        parsedUri = urllib.parse.urlparse(uri)
+
+        http_target = {
+            "uri_override": {
+                "host": parsedUri.hostname,
+                "uri_override_enforce_mode": 2,  # ALWAYS use this endpoint
```
rsamborski, line 55: nit: ditto on using a constant

rsamborski (approved):
Approved with couple non-blocking suggestions for consideration.

cloud_tasks/http_queues.py
engelke: Done.

cloud_tasks/http_queues.py
engelke: Looking at it again, the real reason is that if the default value is given, we don't want to change the queue's existing value, which may not be zero.

cloud_tasks/http_queues/create_http_queue.py
engelke: Since it defaults to HTTPS, I only set it if the user wants HTTP instead. I replaced it with the constant tasks.types.UriOverride.Scheme.HTTP, which is a long way to say 1.

cloud_tasks/http_queues/delete_http_queue_test.py
engelke: Okay.

cloud_tasks/http_queues/create_http_queue_test.py
engelke: Okay, but I think just test-tasks- is enough for this purpose.

glasnt (approved):

glasnt (approved):
Approved, pending nits from Remi

PR #10904, opened 2023-11-02
Pin MySQL sample to Python 3.11

shreejad:
The current base image 'python:3' takes the latest image available, which is 3.12 (recently released). All dependencies of the mySql sample are not yet compatible with 3.12. Thus, pinning the base version to 3.11 like all other samples in this repository.

### DIFF
--- a/cloud-sql/mysql/sqlalchemy/Dockerfile
+++ b/cloud-sql/mysql/sqlalchemy/Dockerfile
@@ -14,7 +14,7 @@
 
 # Use the official Python image.
 # https://hub.docker.com/_/python
-FROM python:3
+FROM python:3.11
 
 # Copy application dependency manifests to the container image.
 # Copying this separately prevents re-running pip install on every code change.


sriram-mahavadi (approved):

cloud-sql/mysql/sqlalchemy/Dockerfile
```diff
@@ -14,7 +14,7 @@
 
 # Use the official Python image.
 # https://hub.docker.com/_/python
-FROM python:3
+FROM python:3.11
```
dandhlee, line 17: Can you verify if 3.11 is available in the image you're installing?

shreejad: Yes, it is available. I've checked by ensuring the `docker build .` works. It is currently broken with `python:3`.

Another sample which uses this image - https://github.com/GoogleCloudPlatform/python-docs-samples/blob/68fa1ecd8027a216ffc2e5cd0fa3956b6c852b9f/run/idp-sql/Dockerfile#L7

jinglundong (approved):

dandhlee (approved):

jackwotherspoon (approved):

PR #10916, opened 2023-11-09
docs: update regions tags for service extension samples

### LINKED ISSUE #10788: Service extension callouts add_header sample - code cleanup
This is a request for code clean-up service extensions plugins

Major issues/concerns
* Service extension sample's private and publish keys (both) are shared in Github. This is not the general practise followed in this Github Repo and this private key file sharing need to be re-evaluated. (preferred options is No).
* Similar to above, a proto file is shared in this sample. This also need to be re-evaluated. (preferred options is No).

Minor issues
* Service Extension python samples are using region tags like `serviceextension_add_header` for callouts. Similar headers is used for service extensions plugs in GoogleCloudPlatform/service-extensions-samples.  This needs to be updated to use prefix like `serviceextension_callouts_`
* The docker file needs to be updated to update port-mappring. Use 8443 instead of 443, similaly 8000 instead of 80.  (Docker file, try `EXPOSE 8443 433` instead of `EXPOSE 433`)

msampathkumar:
## Description

Fixes #10788

## Checklist

### DIFF
--- a/service_extensions/callouts/add_header/Dockerfile
+++ b/service_extensions/callouts/add_header/Dockerfile
@@ -11,7 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-# [START serviceextensions_add_header_docker]
+# [START serviceextensions_callouts_add_header_docker]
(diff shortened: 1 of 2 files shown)


glasnt (approved):

rsamborski (approved):
Should be safe to merge as non of these tags is used in docs yet.

PR #10917, opened 2023-11-09
feat: Makefile for consistent python testing interface

muncus:
## Description

This is a work in progress.

Create a Makefile that can be used to provide a consistent interface to running
samples, regardless of language.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,7 @@ repo_root = $(shell pwd)
 export GOOGLE_CLOUD_PROJECT ?= ${GOOGLE_SAMPLE_PROJECT}
 export BUILD_SPECIFIC_GCLOUD_PROJECT ?= ${GOOGLE_SAMPLE_PROJECT}
 
-build:
+build: check-env
 	pip install nox
 	cd ${dir}
 	pip install -r requirements.txt
@@ -29,7 +29,7 @@ test: check-env build noxfile.py
 	cd ${dir}
 	nox $(NOXFLAGS) -s py-$(py)
 
-lint: noxfile.py
+lint: check-env noxfile.py
 	pip install nox black
 	cd ${dir}
 	nox $(NOXFLAGS) -s blacken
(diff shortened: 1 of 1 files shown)


m-strzelczyk:
Would it be possible to add some kind of `init` command that would:

* create a virtualenv
* install `nox` and `black` in this virtualenv
* Use this virtualenv for other commands?

Right now, if I tried to use this makefile on a fresh git clone, I might get errors about missing nox binary etc.

At least, there should be some instructions that for this to work, there has to be `nox` in `$PATH` and `black` installed for the `nox -s blacken` command

muncus:
We can certainly ensure that nox and black are pip-installed before other actions are called.

I'm wary of creating and using a virtualenv in the Makefile itself, because we would need to be  aware of the user's OS and shell - Would it be sufficient to surface a warning if run outside of a virtualenv?

m-strzelczyk:
Yeah, a warning, with a link to some information on how to set up a virtualenv would be great.

Makefile
```diff
@@ -0,0 +1,53 @@
...
+lint: noxfile.py
+	pip install nox black
+	cd ${dir}
+	nox $(NOXFLAGS) -s blacken
+	nox $(NOXFLAGS) -s lint
+
+# if no noxfile is present, we create one from the toplevel noxfile-template.py
+noxfile.py:
+	cd ${dir}
+	cp ${repo_root}/noxfile-template.py noxfile.py
```
m-strzelczyk, line 41: I think this will just overwrite the existing noxfile.py, no matter if it's there or not. I think you want to add `-n` flag here.

m-strzelczyk (approved):

Makefile
muncus: I *think* Make will do the right thing when a noxfile.py exists already, but there's no harm in the -n, so i've added it anyway just to be safe

muncus:
Without a python virtualenv, the user will now receive the following warning:

```
$ make lint dir=translate/samples/snippets/
Makefile:48: Use of a Python Virtual Environment is recommended. See README.md for details.
```

PR #10991, opened 2023-12-01
chore(docs): update import log README

minherz:
## Description

add instructions about importing logs older than 30 days to the sample's README file.


msampathkumar (approved):
LGTM

PR #11035, opened 2023-12-15
fix(airflow_db_cleanup): sessions cleanup

mokshasoul:
This fixes an issue where the python operator gets killed if the ssession table is large enough by using COUNT(*) instead of loading everything into memory


apilaskowski (approved):
Nice catch. Thank you for updating.

iennae (approved):
LGTM

Thanks for the contribution and improving how the sample works.

PR #11277, opened 2024-03-05
Add clarifying comments to the Dockerfile.

tvalentyn:


tvalentyn:
R: @Polber

kweinmeister (approved):

Polber (approved):

tvalentyn:
Can this be merged? Thanks!

rsamborski:
@davidcavazos It seems the failing tests are not related to changes, but PTAL. I can admin-merge if you're OK with that.

rsamborski (approved):

davidcavazos:
It looks like it's failing on 3.12 while building requirements, I think when building `apache-beam`. @tvalentyn does Apache Beam support Python 3.12? If not, we should disable testing on 3.12 for this sample.

tvalentyn:
> we should disable testing on 3.12 for this sample.

Done

dataflow/flex-templates/streaming_beam/noxfile_config.py
```diff
@@ -25,7 +25,7 @@
     # > ℹ️ We're opting out of all Python versions except 3.8.
     # > The Python version used is defined by the Dockerfile, so it's redundant
     # > to run multiple tests since they would all be running the same Dockerfile.
```
davidcavazos, line 27: Can you add a comment here saying that gcr.io/dataflow-templates-base/python3-template-launcher-base uses Python 3.8 so that's the only version we're testing?

Or if it uses a different version, let's make that match here.

davidcavazos (approved):

dataflow/flex-templates/streaming_beam/noxfile_config.py
tvalentyn: Does it matter which version is used by template launcher? I think the template environment should be self-contained. I wonder why py3.12 tests were failing, should have looked at the logs.

tvalentyn: Here are the relevant test logs from https://btx-internal.corp.google.com/invocations/9f2054ec-9594-453e-86b5-0689b623b21b/targets/cloud-devrel%2Fpython-docs-samples%2Fpython3.12%2Fpresubmit/log:

```
******************** TESTING PROJECTS ********************
btlr run --max-cmd-duration=60m **/requirements.txt --git-diff origin/main... . -- /workspace/.kokoro/tests/run_single_test.sh
Collecting directories that match pattern...337 collected.
Checking for changes with "git diff"... [0 of 337 complete].
Running command(s)... [0 of 1 complete].
#
# dataflow/flex-templates/streaming_beam
#

------------------------------------------------------------
- testing /workspace/dataflow/flex-templates/streaming_beam
------------------------------------------------------------

 Using noxfile-template from parent folder (/workspace). 

nox > Running session py-3.12
nox > Creating virtual environment (virtualenv) using python3.12 in .nox/py-3-12
nox > python -m pip install --upgrade pip
nox > python -m pip install -r requirements.txt --only-binary :all
nox > Command python -m pip install -r requirements.txt --only-binary :all failed with exit code 1:
Collecting apache-beam==2.41.0 (from apache-beam[gcp]==2.41.0->-r requirements.txt (line 1))
  Downloading apache-beam-2.41.0.zip (2.9 MB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 2.9/2.9 MB 11.7 MB/s eta 0:00:00
```

tvalentyn: nox is trying to install requirements.txt. Not sure where this is configured, but this is not necessary and AFAICT this environment is not used to build or submit the pipeline.

tvalentyn: i think current comment is fine.

davidcavazos: It's a global configuration in the root directory of the repo, but we shouldn't edit that. Currently it will always install requirements automatically. So when it tried to install `apache-beam` on Python 3.12 it failed.

PR #11290, opened 2024-03-13
chore: update embedding sample to be more developer friendly

jackwotherspoon:
Note: I am a Googler on a different team that noticed this sample was slightly confusing when I tried to use it so putting this PR up as a potential improvement. 

As a developer wanting to copy/paste this sample into my workload I would expect it to take in an input string and then get the embedding for it? This is what the notebook example for this already does... https://github.com/GoogleCloudPlatform/vertex-ai-samples/blob/main/notebooks/official/generative_ai/text_embedding_new_api.ipynb

### DIFF
--- a/generative_ai/embedding.py
+++ b/generative_ai/embedding.py
@@ -16,13 +16,12 @@
 from vertexai.language_models import TextEmbeddingModel
 
 
-def text_embedding() -> list:
+def text_embedding(text: str = "What is life?") -> list:
     """Text embedding with a Large Language Model."""
     model = TextEmbeddingModel.from_pretrained("textembedding-gecko@001")
-    embeddings = model.get_embeddings(["What is life?"])
-    for embedding in embeddings:
-        vector = embedding.values
-        print(f"Length of Embedding Vector: {len(vector)}")
+    embeddings = model.get_embeddings([text])
+    vector = embeddings[0].values
+    print(f"Length of Embedding Vector: {len(vector)}")
     return vector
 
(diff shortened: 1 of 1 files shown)


generative_ai/embedding.py
```diff
@@ -16,13 +16,12 @@
...
+def text_embedding(text: str = "What is life?") -> list:
     """Text embedding with a Large Language Model."""
     model = TextEmbeddingModel.from_pretrained("textembedding-gecko@001")
-    embeddings = model.get_embeddings(["What is life?"])
-    for embedding in embeddings:
-        vector = embedding.values
-        print(f"Length of Embedding Vector: {len(vector)}")
+    embeddings = model.get_embeddings([text])
+    vector = embeddings[0].values
+    print(f"Length of Embedding Vector: {len(vector)}")
```
jackwotherspoon, line 24: removed loop here, really no need to loop if we are only returning 1 embedding...

jackwotherspoon:
@yil532 curious what you think on this? (saw you as a previous contributor to this file) 😄

jackwotherspoon:
FYI: Test that is failing has nothing to do with code touched in this PR

gericdong (approved):

m-strzelczyk (approved):

jackwotherspoon:
Looks like test that was blocking this PR was fixed as part of https://github.com/GoogleCloudPlatform/python-docs-samples/pull/11313

PR #11406, opened 2024-04-04
chore: remove 3.12 tests from python 3.7 samples

kweinmeister:
3.12 tests are running on samples that are only supposed to run on 3.7.


rsamborski (changes requested):
I left couple minor comments, but my biggest concern is that this change completely turns off any tests. We're currently running tests in 2.7, 3.8-3.12 while the changes exclude all of these versions.

Is this intended?

kweinmeister:
> I left couple minor comments, but my biggest concern is that this change completely turns off any tests. We're currently running tests in 2.7, 3.8-3.12 while the changes exclude all of these versions.
> 
> Is this intended?

These samples are a duplicated set of the appengine samples used in the docs that only support Python <= 3.7. They have lower dependency versions than the other set of appengine samples supporting 3.8-3.12. Unless we bring back Python 3.7 testing, these will not be tested, which is a separate issue from this PR. We shouldn't be testing these on 3.12 for certain.

rsamborski (approved):

PR #11412, opened 2024-04-04
feat: update region tags for existing Imagen samples

irataxy:
## Description

Fixes b:329291575

### DIFF
--- a/generative_ai/imagen/edit_image_mask.py
+++ b/generative_ai/imagen/edit_image_mask.py
@@ -20,6 +20,7 @@ Example usage:
         --output_file <filepath> --prompt <text>
 """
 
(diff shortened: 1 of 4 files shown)


gericdong (approved):

m-strzelczyk (approved):
To get rid of the snippet bot complaints, just make it double-region tagged. It's OK to have two region tags overlap. Then, when the docs are updated to use the new tag, you'll be able to remove the old tags from here.

irataxy:
> To get rid of the snippet bot complaints, just make it double-region tagged. It's OK to have two region tags overlap. Then, when the docs are updated to use the new tag, you'll be able to remove the old tags from here.

Done

PR #11413, opened 2024-04-05
fix: Tidy up api_endpoint processing for Text-Embedding Model Tuning demo.

lee1premium:
## Text embedding model tuning

Fixes: b/320696662

## Checklist
- [ X] I have followed [Sample Guidelines from AUTHORING_GUIDE.MD](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/main/AUTHORING_GUIDE.md)

### DIFF
--- a/generative_ai/embedding_model_tuning.py
+++ b/generative_ai/embedding_model_tuning.py
@@ -34,7 +34,7 @@ def tune_embedding_model(
     batch_size: int = 50,
     iterations: int = 300,
 ) -> pipeline_jobs.PipelineJob:
-    match = re.search(r"([\w]+-[\w]+).*-aiplatform.+", api_endpoint)
+    match = re.search(r"^([\w]+-[\w]+)", api_endpoint)
     location = match.group(1) if match else "us-central1"
     job = aiplatform.PipelineJob(
         display_name=pipeline_job_display_name,
--- a/generative_ai/embedding_model_tuning.py
+++ b/generative_ai/embedding_model_tuning.py
@@ -34,7 +34,7 @@ def tune_embedding_model(
     batch_size: int = 50,
     iterations: int = 300,
 ) -> pipeline_jobs.PipelineJob:
-    match = re.search(r"(.+)(-autopush|-staging)?-aiplatform.+", api_endpoint)
+    match = re.search(r"([\w]+-[\w]+).*-aiplatform.+", api_endpoint)
     location = match.group(1) if match else "us-central1"
     job = aiplatform.PipelineJob(
         display_name=pipeline_job_display_name,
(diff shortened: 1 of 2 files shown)


generative_ai/embedding_model_tuning.py
```diff
@@ -34,7 +34,7 @@ def tune_embedding_model(
     batch_size: int = 50,
     iterations: int = 300,
 ) -> pipeline_jobs.PipelineJob:
-    match = re.search(r"(.+)(-autopush|-staging)?-aiplatform.+", api_endpoint)
+    match = re.search(r"([\w]+-[\w]+).*-aiplatform.+", api_endpoint)
     location = match.group(1) if match else "us-central1"
```
dandhlee, line 38: nit: consider a bit more global friendly location (unless your API is location restricted) for non-US folks as well

generative_ai/embedding_model_tuning.py
```diff
@@ -34,7 +34,7 @@ def tune_embedding_model(
     batch_size: int = 50,
     iterations: int = 300,
 ) -> pipeline_jobs.PipelineJob:
-    match = re.search(r"(.+)(-autopush|-staging)?-aiplatform.+", api_endpoint)
+    match = re.search(r"([\w]+-[\w]+).*-aiplatform.+", api_endpoint)
```
dandhlee, line 37: Wouldn't this be easier with a string split instead of regex search?

dandhlee (commented):
Please use a more descriptive PR title and appropriate type. This seems more like a bug fix than a feature.

generative_ai/embedding_model_tuning.py
lee1premium: I think re.search is cleaner than two splits and a join:

* api_endpoint = "us-central1-xxx-yyy.aiplatform.googleapis.com"
* location1 = re.search(r"^([\w]+-[\w]+)", api_endpoint)[1]
* location2 = "-".join(api_endpoint.split(".")[0].split("-")[:2])  # doesn't look simpler to me.

generative_ai/embedding_model_tuning.py
lee1premium: I guess that 99.9% of the Vertex AI samples would use "us-central1" as the baseline.

dandhlee: Just a friendly reminder that not everyone needs to stick to one region, and there is a cost/environmental concerns over sticking only to one location. Fine to keep as-is, but if there's no restriction consider using different locations in the future.

generative_ai/embedding_model_tuning.py
lee1premium: It doesn't work for this case:
* api_endpoint = "us-central1-xxx-yyy.zzzz.aiplatform.googleapis.com"

You have to do two splits and a join to handle other funny cases.

dandhlee: Ah. That's icky. Got it.

dandhlee (approved):
Reviewing for Python samples.

lee1premium:
You should disable FAILED test_gemini_examples.py::test_gemini_safety_config_example

PR #11422, opened 2024-04-08
fix: moving imports inside functions

### LINKED ISSUE #9505: Refactor dialogflow participant_management.py to have one sample per file
In the process, move import statements into the region tags

BigBlackWolf:
## Description

Fixes #9505 
Samples haven't reflect, which exact import is needed to make it working. Refactoring to make it more precise


m-strzelczyk (approved):

m-strzelczyk:
To unlock the 3.12 build, we need to update `pyaudio` to `0.2.14`. Please do that and we should be good.

PR #11358, opened 2024-03-23
build: Add setuptools to Dockerfile

kweinmeister:
Python 3.12 removed distutils ([details](https://docs.python.org/3/whatsnew/3.12.html#distutils)).

As a result, several 3.12 sample tests are failing. Any time a dependency wheel is built (like `appengine/flexible_python37_and_earlier/scipy` [here](https://btx-internal.corp.google.com/invocations/09d254c0-1f60-4041-96fb-d0f9c9dcf6f0/targets/cloud-devrel%2Fpython-docs-samples%2Fpython3.12%2Fpresubmit/log)), we get an error.

The release note mentions distutils is still available through setuptools. This PR adds setuptools to the build image.

PR #11534, opened 2024-04-17
Update Dockerfile to fix Python 3.12 tests

holtskinner:
Add explicit python3.12 pip install for setuptools

## Description

Follow-up to #11358


holtskinner:
Failing tests are Video Intelligence due to 403 Permission Error and a Timeout Error for Long-running speech

kweinmeister (approved):

PR #11542, opened 2024-04-18
chore: update cloud-sql/sqlserver deps

jackwotherspoon:
These two deps are already part of `cloud-sql-python-connector` and not needed separately in requirements.

This is causing test failures in #11465 as it is overriding proper versions of `google-auth`


jackwotherspoon:
@kweinmeister this will fix the broken test in mentioned PR

holtskinner (approved):

PR #11444, opened 2024-04-12
feat: Service account snippet

BigBlackWolf:
## Description

Fixes #<ISSUE-NUMBER>

Note: Before submitting a pull request, please open an issue for discussion if you are not associated with Google.

## Checklist
- [ ] I have followed [Sample Guidelines from AUTHORING_GUIDE.MD](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/main/AUTHORING_GUIDE.md)
- [ ] README is updated to include [all relevant information](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/main/AUTHORING_GUIDE.md#readme-file)
- [ ] **Tests** pass:   `nox -s py-3.9` (see [Test Environment Setup](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/main/AUTHORING_GUIDE.md#test-environment-setup))
- [ ] **Lint** pass:   `nox -s lint` (see [Test Environment Setup](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/main/AUTHORING_GUIDE.md#test-environment-setup))
- [ ] These samples need a new **API enabled** in testing projects to pass (let us know which ones)
- [ ] These samples need a new/updated **env vars** in testing projects set to pass (let us know which ones)
- [ ] This sample adds a new sample directory, and I updated the [CODEOWNERS file](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/main/.github/CODEOWNERS) with the codeowners for this sample
- [ ] This sample adds a new **Product API**, and I updated the [Blunderbuss issue/PR auto-assigner](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/main/.github/blunderbuss.yml) with the codeowners for this sample
- [ ] Please **merge** this PR for me once it is approved

PR #11567, opened 2024-04-23
feat: project iam policy snippet

BigBlackWolf:
## Description

Fixes #<ISSUE-NUMBER>

## Checklist


iam/cloud-client/snippets/modify_policy_remove_member.py
```diff
@@ -0,0 +1,60 @@
...
+    Possible format for member:
+        * user:{emailid}
+        * serviceAccount:{emailid}
+        * group:{emailid}
+        * deleted:user:{emailid}?uid={uniqueid}
+        * deleted:serviceAccount:{emailid}?uid={uniqueid}
+        * deleted:group:{emailid}?uid={uniqueid}
+        * domain:{domain}
+    """
+    for bind in bindings:
```
BigBlackWolf, line 38: @m-strzelczyk I've done 2 snippets in a different way and would like to align, which approach will meet expectations:
- **iam_modify_policy_add_member** is rewritten using e2e approach requesting and modifying the policies
- **iam_modify_policy_remove_member** is rewritten in the same was it was done in apiary

m-strzelczyk: OK, thanks for providing the options.

I think the best format would be to have structure similar to the `add_member` version, but with a twist. Perhaps make it `iam_modify_policy_add_member(project_id, role, member)` and have it fetch the current policy and update it. Similar with remove member.

Also, would it be possible to add `etag` handling? In case of IAM policies it's quite important not to overwrite changes that others might be doing in parallel and etags are the way to do it. If you send a policy update with etag of the previous policy, it will be checked if the policy in the cloud didn't change. If it did, you will need to repeat the whole operation basically with the new base for your change.

m-strzelczyk:
This one is also depending on #11444 - won't work without it.

iam/cloud-client/snippets/modify_policy_remove_member.py
BigBlackWolf: Makes sense, since the entire communication is supposed to be in a way _read->modify->write_. I've updated it accordingly.

Speaking of `etag` - that's a good point, I've missed that. I've added a comment for `set_policy`, with encourage to add backoff to be sure changes applied.

Regarding `etag` in tests - it is tricky, cause we want to restore policy to default no matter what. Because of that `etag` has to be omitted. As a consequence `etag` can't be tracked, since tests are running in parallel, hitting the same project =(

m-strzelczyk (approved):

PR #11589, opened 2024-04-25
samples(vertex-ai): Add missing parameter for RAG Upload File

holtskinner:
Update to 1.49.0

### DIFF
--- a/generative_ai/rag_test.py
+++ b/generative_ai/rag_test.py
@@ -21,7 +21,7 @@ import pytest
 import rag
 import vertexai
 
-# TODO: Remove once Allowlist is removed
+# TODO(https://github.com/GoogleCloudPlatform/python-docs-samples/issues/11557): Remove once Allowlist is removed
 pytest.skip(allow_module_level=True)
 
 PROJECT_ID = os.getenv("GOOGLE_CLOUD_PROJECT")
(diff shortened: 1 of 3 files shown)


holtskinner:
Failing test is unrelated to this sample

generative_ai/rag_test.py
```diff
@@ -21,13 +21,14 @@
 import rag
 import vertexai
 
-# TODO: Remove once SDK is updated
+# TODO: Remove once Allowlist is removed
```
parthea, line 24: Please file an issue in this repo
```suggestion
# TODO(<replace me with link to Github issue>: Remove once Allowlist is removed
```

generative_ai/rag_test.py
```diff
@@ -83,12 +84,14 @@ def test_upload_file(test_corpus, test_file):
 def test_import_files(test_corpus):
     response = rag.import_files(PROJECT_ID, test_corpus.name, [GCS_FILE])
     assert response.imported_rag_files_count > 0
+    rag.delete_file(PROJECT_ID, response.imported_rag_files[0].name)
```
parthea, line 87: `try/finally` could be used to ensure that resources are cleaned up

parthea (approved):

generative_ai/rag_test.py
holtskinner: I'm not sure for this example. If an exception occured, then response might not be populated, so it might not make sense to run the deletion if there's a failure.

PR #11750, opened 2024-05-10
Upload a sample code of Gemini

482170765:
## Description
Upload a sample code of Gemini with vertexai SDK.


dandhlee:
Hi there, we do not accept code samples from third party :)

PR #11793, opened 2024-05-24
remove extra comment for adding project id

keithballinger:
## Description

This PR removes an extraneous commented line.

### DIFF
--- a/generative_ai/gemini_guide_example.py
+++ b/generative_ai/gemini_guide_example.py
@@ -19,9 +19,6 @@ def generate_text(project_id: str) -> str:
 
     from vertexai.generative_models import GenerativeModel, Part
 
-    # TODO(developer): Update and un-comment below line
-    # project_id = "PROJECT_ID"
-
     vertexai.init(project=project_id, location="us-central1")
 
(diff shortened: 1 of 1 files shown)


msampathkumar:
Hi Keith, thank you for the PR. These TODO comments are by-design.

PTAL. https://cloud.google.com/vertex-ai/generative-ai/docs/reference/libraries#use 

When the sample is displayed in doc pages, these TODO let the user know what action item they need to do before running the script.

msampathkumar (commented):
Please excuse. This is by design.

keithballinger:
> Hi Keith, thank you for the PR. These TODO's comments are by-design.
> 
> PTAL. https://cloud.google.com/vertex-ai/generative-ai/docs/reference/libraries#use
> 
> When the sample is displayed in cloud doc pages, these TODO let the user know what action item they need to do before running the script.

In this case, though, it's not needed, as the function itself takes the project id as a parameter. So, when using the sample they'd call the generate_text method with the project_id. So either the function parameter should be changed to not take a project_id, or the commented out line should be removed. I selected the later.

PR #11808, opened 2024-05-31
fix(compute): renaming region name

m-strzelczyk:
Replacing the region tag name, as the previous one was incorrect and colliding with existing one.


msampathkumar (approved):
LGTM

No code changes.

PR #11844, opened 2024-06-07
Update example to update deprecated modules

bjankie1:
The sample codes use the deprecated modules. For example,

```
from airflow.operators import bash_operator
from airflow.operators import python_operator
```

Those are [deprecated](https://airflow.apache.org/docs/apache-airflow/2.3.2/_api/airflow/operators/bash_operator/index.html).

### DIFF
--- a/composer/workflows/simple.py
+++ b/composer/workflows/simple.py
@@ -23,8 +23,8 @@ from airflow import models
 
 # [END composer_simple_define_dag]
 # [START composer_simple_operators]
-from airflow.operators import bash_operator
-from airflow.operators import python_operator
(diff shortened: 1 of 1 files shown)


bjankie1 (commented):
The sample codes use the deprecated modules. For example,

```
from airflow.operators import bash_operator
from airflow.operators import python_operator
```

Those are [deprecated](https://airflow.apache.org/docs/apache-airflow/2.3.2/_api/airflow/operators/bash_operator/index.html).

filipknapik (approved):
looks good to me

leahecole (approved):

PR #11847, opened 2024-06-09
add STT v2 API infinite streaming example Issue ##11596

### LINKED ISSUE #11596: Example of Infinite Streaming for Google Speech-to-Text API v2
I have not found any examples of Infinite Streaming with the Google Speech-to-Text API v2, and it's unclear whether there is a mechanism in version 2 of the API that facilitates this functionality. Could you provide an example of implementation or confirm if this capability is supported?

Thanks!

gangchen03:
## Description

Fixes #11596

This PR provides an example of Infinite Streaming with the Google Speech-to-Text API v2.

### DIFF
--- a/speech/microphone/transcribe_streaming_infinite_v2.py
+++ b/speech/microphone/transcribe_streaming_infinite_v2.py
@@ -26,10 +26,10 @@ Example usage:
 
 # [START speech_transcribe_infinite_streaming_v2]
 
+import argparse
 import queue
 import re
 import sys
-import argparse
 import time
 
 from google.cloud.speech_v2 import SpeechClient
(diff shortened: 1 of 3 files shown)


speech/microphone/transcribe_streaming_infinite_v2.py
```diff
@@ -0,0 +1,377 @@
...
+To install using pip:
+
+    pip install pyaudio
+    pip install termcolor
+
+Example usage:
+    python transcribe_streaming_infinite_v2.py
+"""
+
+# [START speech_transcribe_infinite_streaming]
```
holtskinner, line 27: You should probably use a different region tag than the original.

Maybe `speech_transcribe_infinite_streaming_v2`?

holtskinner (approved):

speech/microphone/transcribe_streaming_infinite_v2.py
gangchen03: added. thx.

PR #11871, opened 2024-06-18
feat(generative-ai): Add context caching samples (1)

gericdong:
## Description

Fixes b346529113

Add context caching samples:
- Create context cache

PR #11878, opened 2024-06-19
feat(generative-ai): Add context caching samples (2)

gericdong:
## Description

Fixes b346529113

Add context caching samples part 2:
- use
- get
- update
- delete

Part 1 is in #11871

### DIFF
--- a/generative_ai/context_caching/use_context_cache.py
+++ b/generative_ai/context_caching/use_context_cache.py
@@ -1,0 +1,38 @@
+# Copyright 2024 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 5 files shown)


m-strzelczyk (approved):

generative_ai/context_caching/use_context_cache.py
```diff
@@ -0,0 +1,38 @@
...
+    from vertexai.preview.generative_models import GenerativeModel
+    from vertexai.preview import caching
+
+    # TODO(developer): Update and un-comment below lines
+    # project_id = "PROJECT_ID"
+    # cache_id = "CACHE_ID"
+
+    vertexai.init(project=project_id, location="us-central1")
+
+    cached_content = caching.CachedContent(cached_content_name=cache_id)
```
m-strzelczyk, line 29: In the update sample you set the TTL - what happens when we call this line and the `cache_id` cached content is outdated? What kind of error we get and how do we handle it?

gericdong: Using a expired cached content will be given an error message. Users will have to create a new cached content. I will check with docs to see if we have already documented this scenario. Thanks.

Issue #4609, opened 2020-09-01
testing(run): timeout in run/logging-manual


tmatsuo:
I saw weird timeout in our py-3.8 builds.

[Build](https://source.cloud.google.com/results/invocations/4253288f-9bca-4cfc-b8cc-a31b2a4bb8de/targets)

```
nox > Running session py-3.8
nox > Creating virtual environment (virtualenv) using python3.8 in .nox/py-3-8
nox > pip install -r requirements.txt
nox > pip install -r requirements-test.txt
nox > pytest --junitxml=sponge_log.xml
============================= test session starts ==============================
platform linux -- Python 3.8.3, pytest-6.0.1, py-1.9.0, pluggy-0.13.1 -- /workspace/run/logging-manual/.nox/py-3-8/bin/python
cachedir: .pytest_cache
rootdir: /workspace, configfile: pytest.ini
collecting ... collected 3 items

e2e_test.py::test_end_to_end PASSED                                      [ 33%]
main_test.py::test_no_trace_in_headers PASSED                            [ 66%]
main_test.py::test_with_cloud_headers PASSED                             [100%]

------- generated xml file: /workspace/run/logging-manual/sponge_log.xml -------
======================== 3 passed in 2103.89s (0:35:03) ========================
nox > Session py-3.8 was successful.


err: signal: killed
```

It seems like the test finishes rather quick, but btlr sent the signal.

kurtisvg:
@tmatsuo It looks like the runtime was 2103.89s - 35 minutes. Is it possible that the job was hung on a different part - after nox was run?

tmatsuo:
Oh you're right, so this is expected because now we limit each job to 30m:
https://github.com/GoogleCloudPlatform/python-docs-samples/blob/5df67f0af433fadd28ad29fc7fb10c37b26fc236/.kokoro/tests/run_tests.sh#L149

Let's close this for now and if the timeout continues, I'll do something about it.

tmatsuo:
I actually ran the test, and the e2e test is taking a long time, way longer than expected. I think we should debug this further.

tmatsuo:
@grant I think the test might be hanging indefinitely. Probably we should set timeout on log retrieval.

The test in question is in:
https://github.com/GoogleCloudPlatform/python-docs-samples/tree/master/run/logging-manual

tmatsuo:
[...]

PR #11881, opened 2024-06-20
Add Gemma Flex Template example

jrmccluskey:
## Description

Adds a Gemma Flex Template example and an e2e test running on Dataflow. This code example is similar to #11284, but using a Pytorch model and deploying as a flex template. The e2e test will need model weights staged to GCS like the streaming Gemma example.

## Checklist

### DIFF
--- a/.kokoro/tests/run_tests.sh
+++ b/.kokoro/tests/run_tests.sh
@@ -168,7 +168,7 @@ test_prog="${PROJECT_ROOT}/.kokoro/tests/run_single_test.sh"
 
 btlr_args=(
     "run"
-    "--max-cmd-duration=60m"
+    "--max-cmd-duration=70m"
     "**/requirements.txt"
 )
 
--- a/.kokoro/tests/run_tests.sh
+++ b/.kokoro/tests/run_tests.sh
@@ -168,7 +168,7 @@ test_prog="${PROJECT_ROOT}/.kokoro/tests/run_single_test.sh"
 
 btlr_args=(
     "run"
-    "--max-cmd-duration=70m"
+    "--max-cmd-duration=60m"
     "**/requirements.txt"
 )
 
--- a/dataflow/conftest.py
+++ b/dataflow/conftest.py
@@ -164,34 +164,7 @@ def pubsub_wait_for_messages(subscription_path: str) -> list[str]:
     subscriber = pubsub.SubscriberClient()
     with subscriber:
         response = subscriber.pull(subscription=subscription_path, max_messages=10)
-
         messages = [m.message.data.decode("utf-8") for m in response.received_messages]
-        if not messages:
-            raise ValueError("pubsub_wait_for_messages no messages received")
-
-        print(f"pubsub_receive got {len(messages)} message(s)")
-        for msg in messages:
-            print(f"- {repr(msg)}")
-
-        ack_ids = [m.ack_id for m in response.received_messages]
-        subscriber.acknowledge(subscription=subscription_path, ack_ids=ack_ids)
-        print(f"pubsub_receive ack messages")
-    return messages
-
-
-
-@retry.Retry(retry.if_exception_type(ValueError), timeout=TIMEOUT_SEC)
-def pubsub_wait_for_messages_raw(subscription_path: str) -> list[str]:
-    # Let the test decode content itself
-    from google.cloud import pubsub
-
-    subscriber = pubsub.SubscriberClient()
-    with subscriber:
-        print(f"pulling from subscription_path")
-        response = subscriber.pull(subscription=subscription_path, max_messages=10)
-
-        # Do not decode, let calling test do this.
-        messages = [m.message.data for m in response.received_messages]
         if not messages:
             raise ValueError("pubsub_wait_for_messages no messages received")
 
--- a/dataflow/conftest.py
+++ b/dataflow/conftest.py
@@ -794,6 +794,7 @@ class Utils:
         parameters: dict[str, str] = {},
         project: str = PROJECT,
         region: str = REGION,
+        additional_experiments: dict[str,str] = {},
     ) -> str:
         import yaml
 
@@ -814,6 +815,11 @@ class Utils:
             f"--parameters={name}={value}"
             for name, value in {
                 **parameters,
+            }.items()
+        ] + [
+            f"--additional-experiments={name}={value}"
+            for name, value in {
+                **additional_experiments,
             }.items()
         ]
         logging.info(f"{cmd}")
--- a/dataflow/conftest.py
+++ b/dataflow/conftest.py
@@ -831,7 +831,7 @@ class Utils:
         logging.info(f">> {Utils.dataflow_job_url(job_id, project, region)}")
         yield job_id
 
-        Utils.dataflow_jobs_cancel(job_id)
+        Utils.dataflow_jobs_cancel(job_id, region=region)
 
     @staticmethod
     def dataflow_extensible_template_run(
--- a/dataflow/conftest.py
+++ b/dataflow/conftest.py
@@ -850,7 +850,7 @@ class Utils:
             f"--project={project}",
             f"--region={region}",
             f"--staging-location=gs://{bucket_name}/staging",
-            f"--temp-location'gs://{bucket_name}/staging",
+            f"--temp-location=gs://{bucket_name}/staging",
         ] + [
             f"--parameters={name}={value}"
             for name, value in {
--- a/dataflow/conftest.py
+++ b/dataflow/conftest.py
@@ -163,7 +163,12 @@ def pubsub_wait_for_messages(subscription_path: str) -> list[str]:
 
     subscriber = pubsub.SubscriberClient()
     with subscriber:
+        print(f"pulling from subscription_path")
         response = subscriber.pull(subscription=subscription_path, max_messages=10)
+
+        print("Raw data received:")
+        print([str(m.message.data) for m in response.received_messages])
+
         messages = [m.message.data.decode("utf-8") for m in response.received_messages]
         if not messages:
             raise ValueError("pubsub_wait_for_messages no messages received")
--- a/dataflow/conftest.py
+++ b/dataflow/conftest.py
@@ -850,6 +850,7 @@ class Utils:
             f"--project={project}",
             f"--region={region}",
             f"--staging-location=gs://{bucket_name}/staging",
+            f"--temp-location'gs://{bucket_name}/staging",
         ] + [
             f"--parameters={name}={value}"
             for name, value in {
--- a/dataflow/conftest.py
+++ b/dataflow/conftest.py
@@ -166,10 +166,33 @@ def pubsub_wait_for_messages(subscription_path: str) -> list[str]:
         print(f"pulling from subscription_path")
         response = subscriber.pull(subscription=subscription_path, max_messages=10)
 
-        print("Raw data received:")
-        print([str(m.message.data) for m in response.received_messages])
-
         messages = [m.message.data.decode("utf-8") for m in response.received_messages]
+        if not messages:
+            raise ValueError("pubsub_wait_for_messages no messages received")
+
+        print(f"pubsub_receive got {len(messages)} message(s)")
+        for msg in messages:
+            print(f"- {repr(msg)}")
+
+        ack_ids = [m.ack_id for m in response.received_messages]
+        subscriber.acknowledge(subscription=subscription_path, ack_ids=ack_ids)
+        print(f"pubsub_receive ack messages")
+    return messages
+
+
+
+@retry.Retry(retry.if_exception_type(ValueError), timeout=TIMEOUT_SEC)
+def pubsub_wait_for_messages_raw(subscription_path: str) -> list[str]:
+    # Let the test decode content itself
+    from google.cloud import pubsub
+
+    subscriber = pubsub.SubscriberClient()
+    with subscriber:
+        print(f"pulling from subscription_path")
+        response = subscriber.pull(subscription=subscription_path, max_messages=10)
+
+        # Do not decode, let calling test do this.
+        messages = [m.message.data for m in response.received_messages]
         if not messages:
             raise ValueError("pubsub_wait_for_messages no messages received")
 
--- a/dataflow/conftest.py
+++ b/dataflow/conftest.py
@@ -163,7 +163,6 @@ def pubsub_wait_for_messages(subscription_path: str) -> list[str]:
 
     subscriber = pubsub.SubscriberClient()
     with subscriber:
-        print(f"pulling from subscription_path")
         response = subscriber.pull(subscription=subscription_path, max_messages=10)
 
         messages = [m.message.data.decode("utf-8") for m in response.received_messages]
--- a/dataflow/conftest.py
+++ b/dataflow/conftest.py
@@ -855,7 +855,6 @@ class Utils:
             f"--gcs-location={template_path}",
             f"--project={project}",
             f"--region={region}",
-            f"--staging-location=gs://{bucket_name}/staging",
             f"--temp-location=gs://{bucket_name}/staging",
         ] + [
             f"--parameters={name}={value}"
--- a/dataflow/gemma-flex-template/Dockerfile
+++ b/dataflow/gemma-flex-template/Dockerfile
@@ -26,14 +26,14 @@ RUN apt-get update \
     && rm -rf /var/lib/apt/lists/*
 
 COPY requirements.txt requirements.txt
-RUN pip install --upgrade pip \
-    && pip install --no-cache-dir -r requirements.txt 
-
-RUN git clone https://github.com/google/gemma_pytorch.git
-RUN pip install ./gemma_pytorch
+RUN pip install --no-cache-dir --upgrade pip \
+    && pip install --no-cache-dir -r requirements.txt \
+    && git clone https://github.com/google/gemma_pytorch.git \
+    && pip install --no-cache-dir ./gemma_pytorch
 
 # Copy SDK entrypoint binary from Apache Beam image, which makes it possible to
-# use the image as SDK container image.
+# use the image as SDK container image. The Beam version should match the 
+# version specified in requirements.txt
 COPY --from=apache/beam_python3.10_sdk:2.56.0 /opt/apache/beam /opt/apache/beam
 
 # Copy Flex Template launcher binary from the launcher image, which makes it
--- a/dataflow/gemma-flex-template/Dockerfile
+++ b/dataflow/gemma-flex-template/Dockerfile
@@ -23,7 +23,7 @@ RUN apt-get update \
     && apt-get install -y --no-install-recommends curl \
     && apt-get install -y --no-install-recommends wget \
     && apt-get install -y --no-install-recommends git \
-    && && rm -rf /var/lib/apt/lists/*
+    && rm -rf /var/lib/apt/lists/*
 
 COPY requirements.txt requirements.txt
 RUN pip install --upgrade pip \
--- a/dataflow/gemma-flex-template/Dockerfile
+++ b/dataflow/gemma-flex-template/Dockerfile
@@ -18,11 +18,12 @@ FROM ${SERVING_BUILD_IMAGE}
 
 WORKDIR /workspace
 
-RUN apt-get update
-RUN apt-get install -y --no-install-recommends apt-utils
-RUN apt-get install -y --no-install-recommends curl
-RUN apt-get install -y --no-install-recommends wget
-RUN apt-get install -y --no-install-recommends git
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends apt-utils \
+    && apt-get install -y --no-install-recommends curl \
+    && apt-get install -y --no-install-recommends wget \
+    && apt-get install -y --no-install-recommends git \
+    && && rm -rf /var/lib/apt/lists/*
 
 COPY requirements.txt requirements.txt
 RUN pip install --upgrade pip \
@@ -31,8 +32,12 @@ RUN pip install --upgrade pip \
 RUN git clone https://github.com/google/gemma_pytorch.git
 RUN pip install ./gemma_pytorch
 
-# Copy files from official SDK image, including script/dependencies.
+# Copy SDK entrypoint binary from Apache Beam image, which makes it possible to
+# use the image as SDK container image.
 COPY --from=apache/beam_python3.10_sdk:2.56.0 /opt/apache/beam /opt/apache/beam
+
+# Copy Flex Template launcher binary from the launcher image, which makes it
+# possible to use the image as a Flex Template base image.
 COPY --from=gcr.io/dataflow-templates-base/python310-template-launcher-base:latest /opt/google/dataflow/python_template_launcher /opt/google/dataflow/python_template_launcher
 
 # Copy the model directory downloaded from Kaggle and the pipeline code.
--- a/dataflow/gemma-flex-template/Dockerfile
+++ b/dataflow/gemma-flex-template/Dockerfile
@@ -25,7 +25,7 @@ RUN apt-get update \
 COPY requirements.txt requirements.txt
 RUN pip install --no-cache-dir --upgrade pip \
     && git clone https://github.com/google/gemma_pytorch.git \
-    && pip install --no-cache-dir -r requirements.txt
+    && pip install --no-cache-dir -r requirements.txt ./pytorch_gemma
 
 # Copy SDK entrypoint binary from Apache Beam image, which makes it possible to
 # use the image as SDK container image.
--- a/dataflow/gemma-flex-template/Dockerfile
+++ b/dataflow/gemma-flex-template/Dockerfile
@@ -19,21 +19,17 @@ FROM ${SERVING_BUILD_IMAGE}
 WORKDIR /workspace
 
 RUN apt-get update \
-    && apt-get install -y --no-install-recommends apt-utils \
-    && apt-get install -y --no-install-recommends curl \
-    && apt-get install -y --no-install-recommends wget \
-    && apt-get install -y --no-install-recommends git \
+    && apt-get install -y --no-install-recommends apt-utils curl wget git \
     && rm -rf /var/lib/apt/lists/*
 
 COPY requirements.txt requirements.txt
 RUN pip install --no-cache-dir --upgrade pip \
-    && pip install --no-cache-dir -r requirements.txt \
     && git clone https://github.com/google/gemma_pytorch.git \
-    && pip install --no-cache-dir ./gemma_pytorch
+    && pip install --no-cache-dir -r requirements.txt
 
 # Copy SDK entrypoint binary from Apache Beam image, which makes it possible to
-# use the image as SDK container image. The Beam version should match the 
-# version specified in requirements.txt
+# use the image as SDK container image.
+# The Beam version should match the version specified in requirements.txt
 COPY --from=apache/beam_python3.10_sdk:2.56.0 /opt/apache/beam /opt/apache/beam
 
 # Copy Flex Template launcher binary from the launcher image, which makes it
--- a/dataflow/gemma-flex-template/Dockerfile
+++ b/dataflow/gemma-flex-template/Dockerfile
@@ -12,7 +12,7 @@
 
 # This uses Ubuntu with Python 3.10 and comes with CUDA drivers for
 # GPU use.
-ARG SERVING_BUILD_IMAGE=pytorch/pytorch:2.1.2-cuda11.8-cudnn8-runtime
+ARG SERVING_BUILD_IMAGE=pytorch/pytorch:2.3.1-cuda11.8-cudnn8-runtime
 
 FROM ${SERVING_BUILD_IMAGE}
 
--- a/dataflow/gemma-flex-template/Dockerfile
+++ b/dataflow/gemma-flex-template/Dockerfile
@@ -25,7 +25,7 @@ RUN apt-get update \
 COPY requirements.txt requirements.txt
 RUN pip install --no-cache-dir --upgrade pip \
     && git clone https://github.com/google/gemma_pytorch.git \
-    && pip install --no-cache-dir -r requirements.txt ./pytorch_gemma
+    && pip install --no-cache-dir -r requirements.txt ./gemma_pytorch
 
 # Copy SDK entrypoint binary from Apache Beam image, which makes it possible to
 # use the image as SDK container image.
--- a/dataflow/gemma-flex-template/Dockerfile
+++ b/dataflow/gemma-flex-template/Dockerfile
@@ -10,7 +10,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# This uses Ubuntu with Python 3.10
+# This uses Ubuntu with Python 3.10 and comes with CUDA drivers for
+# GPU use.
 ARG SERVING_BUILD_IMAGE=pytorch/pytorch:2.1.2-cuda11.8-cudnn8-runtime
 
 FROM ${SERVING_BUILD_IMAGE}
--- a/dataflow/gemma-flex-template/Dockerfile
+++ b/dataflow/gemma-flex-template/Dockerfile
@@ -1,0 +1,44 @@
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This uses Ubuntu with Python 3.10
+ARG SERVING_BUILD_IMAGE=pytorch/pytorch:2.1.2-cuda11.8-cudnn8-runtime
+
+FROM ${SERVING_BUILD_IMAGE}
+
+WORKDIR /workspace
+
+RUN apt-get update
+RUN apt-get install -y --no-install-recommends apt-utils
+RUN apt-get install -y --no-install-recommends curl
+RUN apt-get install -y --no-install-recommends wget
+RUN apt-get install -y --no-install-recommends git
+
+COPY requirements.txt requirements.txt
+RUN pip install --upgrade pip \
+    && pip install --no-cache-dir -r requirements.txt 
+
+RUN git clone https://github.com/google/gemma_pytorch.git
+RUN pip install ./gemma_pytorch
+
+# Copy files from official SDK image, including script/dependencies.
+COPY --from=apache/beam_python3.10_sdk:2.56.0 /opt/apache/beam /opt/apache/beam
+COPY --from=gcr.io/dataflow-templates-base/python310-template-launcher-base:latest /opt/google/dataflow/python_template_launcher /opt/google/dataflow/python_template_launcher
+
+# Copy the model directory downloaded from Kaggle and the pipeline code.
+COPY pytorch_model pytorch_model
+COPY custom_model_gemma.py custom_model_gemma.py
+
+ENV FLEX_TEMPLATE_PYTHON_PY_FILE="custom_model_gemma.py"
+
+# Set the entrypoint to Apache Beam SDK launcher.
+ENTRYPOINT ["/opt/apache/beam/boot"]
--- a/dataflow/gemma-flex-template/README.md
+++ b/dataflow/gemma-flex-template/README.md
@@ -1,10 +1,10 @@
-# RunInference on Dataflow Streaming with Gemma and Flex Templates
+# RunInference on Dataflow streaming with Gemma and Flex Templates
 
 Gemma is a family of lightweight, state-of-the art open models built from research and technology used to create the Gemini models.
 You can use Gemma models in your Apache Beam inference pipelines with the `RunInference` transform.
 
 This example demonstrates how to use a Gemma model running on Pytorch in a streaming Dataflow pipeline that has Pub/Sub sources and sinks. This pipeline
-is deployed using flex templates.
+is deployed by using Flex Templates.
 
 For more information about using RunInference, see [Get started with AI/ML pipelines](https://beam.apache.org/documentation/ml/overview/) in the Apache Beam documentation.
 
@@ -14,7 +14,7 @@ Follow the steps in this section to create the necessary environment to run this
 
 ### Enable Google Cloud services
 
-This workflow uses multiple Google Cloud Platform products, including Dataflow, Pub/Sub, Google Cloud Storage, and Artifact Registry. Before you start the workflow, create a Google Cloud project that has the following services enabled:
+This workflow uses multiple Google Cloud products, including Dataflow, Pub/Sub, Google Cloud Storage, and Artifact Registry. Before you start the workflow, create a Google Cloud project that has the following services enabled:
 
 * Dataflow
 * Pub/Sub
@@ -27,7 +27,7 @@ Your Google Cloud project also needs to have Nvidia L4 GPU quota. For more infor
 
 ### Create a custom container
 
-To build a custom container, use Docker. This repository contains a Dockerfile that you can use to build your custom container. To build and push a container to Artifact Registry by using DockerFollow, follow the instructions in the [Build and push the image](https://cloud.google.com/dataflow/docs/guides/build-container-image#build_and_push_the_image) section of the "Build custom container images for Dataflow" page in the Google Cloud documentation.
+To build a custom container, use Docker. This repository contains a Dockerfile that you can use to build your custom container. To build and push a container to Artifact Registry by using DockerFollow, follow the instructions in the [Build and push the image](https://cloud.google.com/dataflow/docs/guides/build-container-image#build_and_push_the_image) section of "Build custom container images for Dataflow" in the Google Cloud documentation.
 
 ### Create Pub/Sub topics for input and output
 
@@ -56,23 +56,23 @@ Install Apache Beam and the dependencies required to run the pipeline in your lo
 pip install -U -r requirements.txt
 ```
 
-This example also requires having access to the base class for the Pytorch implementation of Gemma.
+This example also requires access to the base class for the PyTorch implementation of Gemma.
 ```
 git clone https://github.com/google/gemma_pytorch.git
 pip install ./gemma_pytorch
 ```
 
-## Code Overview
+## Code overview
 
 This section provides details about the custom model handler and the formatting `DoFn` used in this example.
 
 ### Custom model handler
-This notebook defines a custom model handler that loads the model by constructing a configuration object and loading the model's checkpoint from the local filesystem.
-This approach differs from Pytorch model loading process followed in the Beam Pytorch model handler, which is why a custom implementation is necessary. 
+This notebook defines a custom model handler that loads the model. The model handler constructs a configuration object and loads the model's checkpoint from the local filesystem.
+Because this approach differs from the PyTorch model loading process followed in the Beam PyTorch model handler, a custom implementation is necessary. 
 
 To customize the behavior of the handler, implement the following methods: `load_model`, `validate_inference_args`, and `share_model_across_processes`.
 
-The Pytorch implementation of the Gemma models has a `generate` method
+The PyTorch implementation of the Gemma models has a `generate` method
 that generates text based on a prompt. To route the prompts correctly, use this function in the `run_inference` function.
 
 ```py
@@ -159,7 +159,11 @@ class FormatOutput(beam.DoFn):
 ```
 
 ## Build the Flex Template
-Run the following code from the directory to build the Dataflow flex template. Replace `$TEMPLATE_FILE` with a GCS path written as a `.json` file. Set `SDK_CONTAINER_IMAGE` to the name Docker image created earlier and `$PROJECT` is the GCP project you created earlier. 
+Run the following code from the directory to build the Dataflow flex template.
+
+- Replace `$TEMPLATE_FILE` with a Google Cloud Storage path written as a JSON file.
+- Set `SDK_CONTAINER_IMAGE` to the name of the Docker image created previously.
+- `$PROJECT` is the Google Cloud project that you created previously. 
 
 ```sh
 gcloud dataflow flex-template build $TEMPLATE_FILE \
@@ -170,7 +174,7 @@ gcloud dataflow flex-template build $TEMPLATE_FILE \
 ```
 
 ## Start the pipeline
-Run the following code from the directory to start the Dataflow streaming pipeline. Replace `$TEMPLATE_FILE`, `$REGION`, `$GCS_BUCKET`, `$INPUT_SUBSCRIPTION`, `$OUTPUT_TOPIC`, `$SDK_CONTAINER_IMAGE`, and `$PROJECT` with the Google Cloud Project resources you created earlier. It may take around 15 minutes for the worker to start up and begin accepting messages from the input Pub/Sub topic. 
+To start the Dataflow streaming job, run the following code from the directory. Replace `$TEMPLATE_FILE`, `$REGION`, `$GCS_BUCKET`, `$INPUT_SUBSCRIPTION`, `$OUTPUT_TOPIC`, `$SDK_CONTAINER_IMAGE`, and `$PROJECT` with the Google Cloud project resources you created previously. It might take as much as 15 minutes for the worker to start up and to begin accepting messages from the input Pub/Sub topic. 
 
 ```sh
 gcloud dataflow flex-template run "gemma-flex-`date +%Y%m%d-%H%M%S`" \
@@ -188,7 +192,7 @@ gcloud dataflow flex-template run "gemma-flex-`date +%Y%m%d-%H%M%S`" \
 
 ## Send a prompt to the model and check the response
 
-In the Google Cloud console, navigate to the Pub/Sub topics page, and then select your input topic. On the **Messages** tab, click **Publish Message**. Add a message into the pipeline for the Dataflow job to pick up and pass through the model. The Dataflow job outputs the response to the Pub/Sub sink topic. To check the response from the model, you can manually pull messages from the destination topic. For more information, see [Publish messages](https://cloud.google.com/pubsub/docs/publisher#publish-messages) in the Google Cloud documentation.
+In the Google Cloud console, navigate to the Pub/Sub topics page, and then select your input topic. On the **Messages** tab, click **Publish Message**. Add a message for the Dataflow job to pick up and pass through the model. The Dataflow job outputs the response to the Pub/Sub sink topic. To check the response from the model, you can manually pull messages from the destination topic. For more information, see [Publish messages](https://cloud.google.com/pubsub/docs/publisher#publish-messages) in the Google Cloud documentation.
 
 ## Clean up resources
 
--- a/dataflow/gemma-flex-template/README.md
+++ b/dataflow/gemma-flex-template/README.md
@@ -1,6 +1,6 @@
 # RunInference on Dataflow streaming with Gemma and Flex Templates
 
-Gemma is a family of lightweight, state-of-the art open models built from research and technology used to create the Gemini models.
+Gemma is a family of lightweight, state-of-the-art open models built from research and technology used to create the Gemini models.
 You can use Gemma models in your Apache Beam inference pipelines with the `RunInference` transform.
 
 This example demonstrates how to use a Gemma model running on Pytorch in a streaming Dataflow pipeline that has Pub/Sub sources and sinks. This pipeline
--- a/dataflow/gemma-flex-template/README.md
+++ b/dataflow/gemma-flex-template/README.md
@@ -10,6 +10,8 @@ For more information about using RunInference, see [Get started with AI/ML pipel
 
 ## Before you begin
 
+Make sure you have followed the [Dataflow setup instructions](../../README.md).
+
 Follow the steps in this section to create the necessary environment to run this workflow.
 
 ### Enable Google Cloud services
@@ -18,12 +20,33 @@ This workflow uses multiple Google Cloud products, including Dataflow, Pub/Sub, 
 
 * Dataflow
 * Pub/Sub
+* Compute Engine
+* Cloud Logging
 * Google Cloud Storage
+* Google Cloud Storage JSON
+* Cloud Build
+* Datastore
+* Cloud Resource Manager
 * Artifact Registry
 
 Using these services incurs billing charges.
 
 Your Google Cloud project also needs to have Nvidia L4 GPU quota. For more information, see [GPU quota](https://cloud.google.com/compute/resource-usage#gpu_quota) in the Google Cloud documentation.
+
+### Download and save the model
+
+Save a version of the Gemma 2B model. Downloaded the model from [Kaggle](https://www.kaggle.com/models/google/gemma/pyTorch/1.1-2b-it). This download is a `.tar.gz` archive. Extract the archive into a directory and name it `pytorch_model`. 
+
+### Create a cloud storage bucket
+
+Click [here to create a GCS bucket](https://console.cloud.google.com/storage/create-bucket) or run the following commands:
+
+```sh
+export GCS_BUCKET="your--bucket"
+gsutil mb gs://$GCS_BUCKET
+```
+
+Make sure your GCS bucket name does __not__ include the `gs://` prefix
 
 ### Create a custom container
 
@@ -31,36 +54,7 @@ To build a custom container, use Docker. This repository contains a Dockerfile t
 
 ### Create Pub/Sub topics for input and output
 
-To create your Pub/Sub source and sink, follow the instructions in [Create a Pub/Sub topic](https://cloud.google.com/pubsub/docs/create-topic#pubsub_create_topic-Console) in the Google Cloud documentation. For this example, create two topics, one input topic and one output topic. For input, the topic must have a subscription that you can provide to the Gemma model. 
-
-### Download and save the model
-
-Save a version of the Gemma 2B model. Downloaded the model from [Kaggle](https://www.kaggle.com/models/google/gemma/pyTorch/1.1-2b-it). This download is a `.tar.gz` archive. Extract the archive into a directory and name it `pytorch_model`. 
-
-### Optional: Create a new virtual environment
-
-The Python major and minor version contained in the custom container must match the environment used for job submission. For this example, use Python version 3.10.
-
-```
-python3.10 -m venv /tmp/venv
-source /tmp/venv/bin/activate
-```
-
-For more information, see [venv — Creation of virtual environments](https://docs.python.org/3/library/venv.html).
-
-### Install dependencies
-
-Install Apache Beam and the dependencies required to run the pipeline in your local environment. 
-
-```
-pip install -U -r requirements.txt
-```
-
-This example also requires access to the base class for the PyTorch implementation of Gemma.
-```
-git clone https://github.com/google/gemma_pytorch.git
-pip install ./gemma_pytorch
-```
+To create your Pub/Sub source and sink, follow the instructions in [Create a Pub/Sub topic](https://cloud.google.com/pubsub/docs/create-topic#pubsub_create_topic-Console) in the Google Cloud documentation. For this example, create two topics, one input topic and one output topic. Follow the instructions in [Create pull subscriptions](https://cloud.google.com/pubsub/docs/create-subscription) to create a pull subscription for each of the two topics you jsut created. The input subscription allows you to provide input to the pipeline, while the output subscription will allow you to see the output from the pipeline during and after execution.
 
 ## Code overview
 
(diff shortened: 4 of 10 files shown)


jrmccluskey:
Not sure what happened on the kokoro test run, the test target passed but the test execution as a whole was killed right after the test passed

dataflow/gemma-flex-template/README.md
```diff
@@ -0,0 +1,201 @@
+# RunInference on Dataflow Streaming with Gemma and Flex Templates
```
rszper, line 1:
```suggestion
# RunInference on Dataflow streaming with Gemma and Flex Templates
```

dataflow/gemma-flex-template/README.md
```diff
@@ -0,0 +1,201 @@
...
+
+Gemma is a family of lightweight, state-of-the art open models built from research and technology used to create the Gemini models.
+You can use Gemma models in your Apache Beam inference pipelines with the `RunInference` transform.
+
+This example demonstrates how to use a Gemma model running on Pytorch in a streaming Dataflow pipeline that has Pub/Sub sources and sinks. This pipeline
+is deployed using flex templates.
```
rszper, line 7:
```suggestion
is deployed by using Flex Templates.
```

dataflow/gemma-flex-template/README.md
```diff
@@ -0,0 +1,201 @@
...
+
+For more information about using RunInference, see [Get started with AI/ML pipelines](https://beam.apache.org/documentation/ml/overview/) in the Apache Beam documentation.
+
+## Before you begin
+
+Follow the steps in this section to create the necessary environment to run this workflow.
+
+### Enable Google Cloud services
+
+This workflow uses multiple Google Cloud Platform products, including Dataflow, Pub/Sub, Google Cloud Storage, and Artifact Registry. Before you start the workflow, create a Google Cloud project that has the following services enabled:
```
rszper, line 17:
```suggestion
This workflow uses multiple Google Cloud products, including Dataflow, Pub/Sub, Google Cloud Storage, and Artifact Registry. Before you start the workflow, create a Google Cloud project that has the following services enabled:
```

dataflow/gemma-flex-template/README.md
```diff
@@ -0,0 +1,201 @@
...
+* Google Cloud Storage
+* Artifact Registry
+
+Using these services incurs billing charges.
+
+Your Google Cloud project also needs to have Nvidia L4 GPU quota. For more information, see [GPU quota](https://cloud.google.com/compute/resource-usage#gpu_quota) in the Google Cloud documentation.
+
+### Create a custom container
+
+To build a custom container, use Docker. This repository contains a Dockerfile that you can use to build your custom container. To build and push a container to Artifact Registry by using DockerFollow, follow the instructions in the [Build and push the image](https://cloud.google.com/dataflow/docs/guides/build-container-image#build_and_push_the_image) section of the "Build custom container images for Dataflow" page in the Google Cloud documentation.
```
rszper, line 30:
```suggestion
To build a custom container, use Docker. This repository contains a Dockerfile that you can use to build your custom container. To build and push a container to Artifact Registry by using DockerFollow, follow the instructions in the [Build and push the image](https://cloud.google.com/dataflow/docs/guides/build-container-image#build_and_push_the_image) section of "Build custom container images for Dataflow" in the Google Cloud documentation.
```

dataflow/gemma-flex-template/README.md
```diff
@@ -0,0 +1,201 @@
...
+
+### Install dependencies
+
+Install Apache Beam and the dependencies required to run the pipeline in your local environment. 
+
+```
+pip install -U -r requirements.txt
+```
+
+This example also requires having access to the base class for the Pytorch implementation of Gemma.
```
rszper, line 59:
```suggestion
This example also requires access to the base class for the PyTorch implementation of Gemma.
```

dataflow/gemma-flex-template/README.md
```diff
@@ -0,0 +1,201 @@
...
+```
+git clone https://github.com/google/gemma_pytorch.git
+pip install ./gemma_pytorch
+```
+
+## Code Overview
```
rszper, line 65:
```suggestion
## Code overview
```

dataflow/gemma-flex-template/README.md
```diff
@@ -0,0 +1,201 @@
...
+
+This section provides details about the custom model handler and the formatting `DoFn` used in this example.
+
+### Custom model handler
+This notebook defines a custom model handler that loads the model by constructing a configuration object and loading the model's checkpoint from the local filesystem.
```
rszper, line 70:
```suggestion
This notebook defines a custom model handler that loads the model. The model handler constructs a configuration object and loads the model's checkpoint from the local filesystem.
```

dataflow/gemma-flex-template/README.md
```diff
@@ -0,0 +1,201 @@
...
+This approach differs from Pytorch model loading process followed in the Beam Pytorch model handler, which is why a custom implementation is necessary.
```
rszper, line 71:
```suggestion
Because this approach differs from the PyTorch model loading process followed in the Beam PyTorch model handler, a custom implementation is necessary. 
```

dataflow/gemma-flex-template/README.md
```diff
@@ -0,0 +1,201 @@
...
+
+This section provides details about the custom model handler and the formatting `DoFn` used in this example.
+
+### Custom model handler
+This notebook defines a custom model handler that loads the model by constructing a configuration object and loading the model's checkpoint from the local filesystem.
+This approach differs from Pytorch model loading process followed in the Beam Pytorch model handler, which is why a custom implementation is necessary. 
+
+To customize the behavior of the handler, implement the following methods: `load_model`, `validate_inference_args`, and `share_model_across_processes`.
+
+The Pytorch implementation of the Gemma models has a `generate` method
```
rszper, line 75:
```suggestion
The PyTorch implementation of the Gemma models has a `generate` method
```

dataflow/gemma-flex-template/README.md
```diff
@@ -0,0 +1,201 @@
...
+The output from a keyed model handler is a tuple of the form `(key, PredictionResult)`. To format that output into a string before sending it to the answer Pub/Sub topic, use an extra `DoFn`.
+
+```py
+class FormatOutput(beam.DoFn):
+  def process(self, element, *args, **kwargs):
+    yield "Key : {key}, Input: {input}, Output: {output}".format(key=element[0], input=element[1].example, output=element[1].inference)
+```
+
+## Build the Flex Template
+Run the following code from the directory to build the Dataflow flex template. Replace `$TEMPLATE_FILE` with a GCS path written as a `.json` file. Set `SDK_CONTAINER_IMAGE` to the name Docker image created earlier and `$PROJECT` is the GCP project you created earlier.
```
rszper, line 162:
```suggestion
Run the following code from the directory to build the Dataflow flex template.

- Replace `$TEMPLATE_FILE` with a Google Cloud Storage path written as a JSON file.
- Set `SDK_CONTAINER_IMAGE` to the name of the Docker image created previously.
- `$PROJECT` is the Google Cloud project that you created previously. 
```

dataflow/gemma-flex-template/README.md
```diff
@@ -0,0 +1,201 @@
...
+```sh
+gcloud dataflow flex-template build $TEMPLATE_FILE \
+  --image $SDK_CONTAINER_IMAGE \
+  --sdk-language "PYTHON" \
+  --metadata-file metadata.json \
+  --project $PROJECT
+```
+
+## Start the pipeline
+Run the following code from the directory to start the Dataflow streaming pipeline. Replace `$TEMPLATE_FILE`, `$REGION`, `$GCS_BUCKET`, `$INPUT_SUBSCRIPTION`, `$OUTPUT_TOPIC`, `$SDK_CONTAINER_IMAGE`, and `$PROJECT` with the Google Cloud Project resources you created earlier. It may take around 15 minutes for the worker to start up and begin accepting messages from the input Pub/Sub topic.
```
rszper, line 173:
```suggestion
To start the Dataflow streaming job, run the following code from the directory. Replace `$TEMPLATE_FILE`, `$REGION`, `$GCS_BUCKET`, `$INPUT_SUBSCRIPTION`, `$OUTPUT_TOPIC`, `$SDK_CONTAINER_IMAGE`, and `$PROJECT` with the Google Cloud project resources you created previously. It might take as much as 15 minutes for the worker to start up and to begin accepting messages from the input Pub/Sub topic. 
```

dataflow/gemma-flex-template/README.md
```diff
@@ -0,0 +1,201 @@
...
+  --parameters device="GPU" \
+  --parameters sdk_container_image=$SDK_CONTAINER_IMAGE \
+  --parameters dataflow_service_options="worker_accelerator=type:nvidia-l4;count:1;install-nvidia-driver:5xx" \
+  --project $PROJECT \
+  --worker-machine-type "g2-standard-4"
+```
+
+## Send a prompt to the model and check the response
+
+In the Google Cloud console, navigate to the Pub/Sub topics page, and then select your input topic. On the **Messages** tab, click **Publish Message**. Add a message into the pipeline for the Dataflow job to pick up and pass through the model. The Dataflow job outputs the response to the Pub/Sub sink topic. To check the response from the model, you can manually pull messages from the destination topic. For more information, see [Publish messages](https://cloud.google.com/pubsub/docs/publisher#publish-messages) in the Google Cloud documentation.
```
rszper, line 191:
```suggestion
In the Google Cloud console, navigate to the Pub/Sub topics page, and then select your input topic. On the **Messages** tab, click **Publish Message**. Add a message for the Dataflow job to pick up and pass through the model. The Dataflow job outputs the response to the Pub/Sub sink topic. To check the response from the model, you can manually pull messages from the destination topic. For more information, see [Publish messages](https://cloud.google.com/pubsub/docs/publisher#publish-messages) in the Google Cloud documentation.
```

jrmccluskey:
Not sure what's happening here, test passes but the test session gets killed consistently. @engelke can you take a look?

image

jrmccluskey:
blunderbuss: assign

dataflow/gemma-flex-template/README.md
```diff
@@ -0,0 +1,205 @@
...
+        self._checkpoint_path = checkpoint_path
+        if device == 'GPU':
+            logging.info("Device is set to CUDA")
+            self._device = torch.device('cuda')
+        else:
+            logging.info("Device is set to CPU")
+            self._device = torch.device('cpu')
+        self._env_vars = {}
+
+    def share_model_across_processes(self) -> bool:
```
tvalentyn, line 116: Do you want to explain why we do this and what are implications? perhaps:

```
# This allows us to load a large model only once per worker VM, which decreases the pipeline memory requirements.
```

dataflow/gemma-flex-template/README.md
```diff
@@ -0,0 +1,205 @@
...
+This notebook defines a custom model handler that loads the model. The model handler constructs a configuration object and loads the model's checkpoint from the local filesystem.
+Because this approach differs from the PyTorch model loading process followed in the Beam PyTorch model handler, a custom implementation is necessary. 
+
+To customize the behavior of the handler, implement the following methods: `load_model`, `validate_inference_args`, and `share_model_across_processes`.
+
+The PyTorch implementation of the Gemma models has a `generate` method
+that generates text based on a prompt. To route the prompts correctly, use this function in the `run_inference` function.
+
+```py
+class GemmaPytorchModelHandler(ModelHandler[str, PredictionResult,
```
tvalentyn, line 79: having large number of code in the md. might cause an inconsistency down the road if we modify the source but not the MD. Maybe we only highlight most important points here, and send user to the source instead? We can and should have comments in the source anyway.

dataflow/gemma-flex-template/README.md
```diff
@@ -0,0 +1,205 @@
...
+
+* Dataflow
+* Pub/Sub
+* Google Cloud Storage
+* Artifact Registry
+
+Using these services incurs billing charges.
+
+Your Google Cloud project also needs to have Nvidia L4 GPU quota. For more information, see [GPU quota](https://cloud.google.com/compute/resource-usage#gpu_quota) in the Google Cloud documentation.
+
```
tvalentyn, line 27: should this have instructions where to get gemma weights? the pytorch_model folder

dataflow/gemma-flex-template/custom_model_gemma.py
```diff
@@ -0,0 +1,179 @@
...
+    parser.add_argument(
+        "--checkpoint_path",
+        required=False,
+        default="pytorch_model/gemma-2b-it.ckpt",
+        help="path to the Gemma model weights in the custom worker container",
+    )
+    parser.add_argument(
+        "--tokenizer_path",
+        required=False,
+        default="pytorch_model/tokenizer.model",
```
tvalentyn, line 143: is this part of gemma distribution?

dataflow/gemma-flex-template/requirements.txt
```diff
@@ -0,0 +1,3 @@
+apache_beam[gcp]==2.56.0
```
tvalentyn, line 1: for reproducible builds, it is better to have transitive dependencies as well:

https://github.com/GoogleCloudPlatform/python-docs-samples/blob/c93accadf3bd29e9c3166676abb2c95564579c5e/dataflow/flex-templates/pipeline_with_dependencies/requirements.txt#L22

You could do something similar or add a comment:

```
# For reproducible builds, it is better to also include transitive dependencies: https://github.com/GoogleCloudPlatform/python-docs-samples/blob/c93accadf3bd29e9c3166676abb2c95564579c5e/dataflow/flex-templates/pipeline_with_dependencies/requirements.txt#L22 , but for simplicii of this example, we are only including the top-level dependencies.
```

tvalentyn (approved):

dataflow/gemma-flex-template/custom_model_gemma.py
jrmccluskey: Yes the tokenizer is included when you download the pytorch version of gemma

dataflow/gemma-flex-template/README.md
jrmccluskey: Explained below

dataflow/gemma-flex-template/README.md
```diff
@@ -0,0 +1,205 @@
...
+### Create a custom container
+
+To build a custom container, use Docker. This repository contains a Dockerfile that you can use to build your custom container. To build and push a container to Artifact Registry by using DockerFollow, follow the instructions in the [Build and push the image](https://cloud.google.com/dataflow/docs/guides/build-container-image#build_and_push_the_image) section of "Build custom container images for Dataflow" in the Google Cloud documentation.
```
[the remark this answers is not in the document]
jrmccluskey: fixed

dataflow/gemma-flex-template/Dockerfile
```diff
@@ -0,0 +1,45 @@
...
+COPY requirements.txt requirements.txt
+RUN pip install --upgrade pip \
+    && pip install --no-cache-dir -r requirements.txt 
+
+RUN git clone https://github.com/google/gemma_pytorch.git
+RUN pip install ./gemma_pytorch
+
+# Copy files from official SDK image, including script/dependencies.
+COPY --from=apache/beam_python3.10_sdk:2.56.0 /opt/apache/beam /opt/apache/beam
+COPY --from=gcr.io/dataflow-templates-base/python310-template-launcher-base:latest /opt/google/dataflow/python_template_launcher /opt/google/dataflow/python_template_launcher
```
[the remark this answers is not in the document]
jrmccluskey: fixed

dataflow/gemma-flex-template/Dockerfile
```diff
@@ -0,0 +1,45 @@
...
+RUN apt-get install -y --no-install-recommends git
+
+COPY requirements.txt requirements.txt
+RUN pip install --upgrade pip \
+    && pip install --no-cache-dir -r requirements.txt 
+
+RUN git clone https://github.com/google/gemma_pytorch.git
+RUN pip install ./gemma_pytorch
+
+# Copy files from official SDK image, including script/dependencies.
```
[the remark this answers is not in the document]
jrmccluskey: updated comment

dataflow/gemma-flex-template/README.md
jrmccluskey: added

dataflow/gemma-flex-template/requirements.txt
jrmccluskey: done

dataflow/gemma-flex-template/README.md
jrmccluskey: This is a fair point, I think the emphasis in the text is on model loading being different so I think cutting the large model function and the `run_inference` call itself is fine for simplicity's sake

dataflow/gemma-flex-template/Dockerfile
```diff
@@ -0,0 +1,45 @@
...
+
+# This uses Ubuntu with Python 3.10 and comes with CUDA drivers for
+# GPU use.
+ARG SERVING_BUILD_IMAGE=pytorch/pytorch:2.1.2-cuda11.8-cudnn8-runtime
+
+FROM ${SERVING_BUILD_IMAGE}
+
+WORKDIR /workspace
+
+RUN apt-get update
```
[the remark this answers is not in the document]
jrmccluskey: updated

dataflow/gemma-flex-template/README.md
```diff
@@ -0,0 +1,205 @@
...
+git clone https://github.com/google/gemma_pytorch.git
+pip install ./gemma_pytorch
+```
+
+## Code overview
+
+This section provides details about the custom model handler and the formatting `DoFn` used in this example.
+
+### Custom model handler
+This notebook defines a custom model handler that loads the model. The model handler constructs a configuration object and loads the model's checkpoint from the local filesystem.
```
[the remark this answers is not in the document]
jrmccluskey: not sure what you're asking here

jrmccluskey:
@kweinmeister Still need an approving review from someone if you could take a look

dataflow/gemma-flex-template/README.md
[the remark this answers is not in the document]
iennae: @jrmccluskey it seems like you mean something else besides notebook here. Is this the first time notebook is being used? In other places this describes this as an example and springing the notebook description here seems confusing.

iennae (changes requested):
I'm flagging this for additional review and I've raised this with the DevRel Engineering team to provide additional feedback. 

There is a lot of code in the README and that will make this a difficult sample to maintain.

dataflow/gemma-flex-template/README.md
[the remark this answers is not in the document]
jrmccluskey: ah, fixed

jrmccluskey:
If the complaint is with the model handler code I don't think it's too much of a change to cut that code in favor of linking to the source instead.

dataflow/gemma-flex-template/README.md
```diff
@@ -0,0 +1,179 @@
...
+
+### Create a custom container
+
+To build a custom container, use Docker. This repository contains a Dockerfile that you can use to build your custom container. To build and push a container to Artifact Registry by using Docker or Cloud Build, follow the instructions in the [Build and push the image](https://cloud.google.com/dataflow/docs/guides/build-container-image#build_and_push_the_image) section of "Build custom container images for Dataflow" in the Google Cloud documentation.
+
+### Create Pub/Sub topics for input and output
+
+To create your Pub/Sub source and sink, follow the instructions in [Create a Pub/Sub topic](https://cloud.google.com/pubsub/docs/create-topic#pubsub_create_topic-Console) in the Google Cloud documentation. For this example, create two topics, one input topic and one output topic. For input, the topic must have a subscription that you can provide to the Gemma model. 
+
+### Download and save the model
```
glasnt, line 36: This section needs to come before the "Create custom container" section

dataflow/gemma-flex-template/README.md
```diff
@@ -0,0 +1,179 @@
...
+
+Save a version of the Gemma 2B model. Downloaded the model from [Kaggle](https://www.kaggle.com/models/google/gemma/pyTorch/1.1-2b-it). This download is a `.tar.gz` archive. Extract the archive into a directory and name it `pytorch_model`. 
+
+### Optional: Create a new virtual environment
```
glasnt, line 40: I think this step is "optional" because you don't need to create a virtualenv to run the next step, however: there's no use for a local dev environment detailed in this README, so I think these two sections can be removed.

dataflow/gemma-flex-template/README.md
```diff
@@ -0,0 +1,179 @@
...
+```py
+class FormatOutput(beam.DoFn):
+  def process(self, element, *args, **kwargs):
+    yield "Key : {key}, Input: {input}, Output: {output}".format(key=element[0], input=element[1].example, output=element[1].inference)
+```
+
+## Build the Flex Template
+Run the following code from the directory to build the Dataflow flex template.
+
+- Replace `$TEMPLATE_FILE` with a Google Cloud Storage path written as a JSON file.
```
glasnt, line 138: This was unclear to me when going through these instructuons. 

The $TEMPLATE_FILE needs to be a GCS storage path, that will be used as the output for this command. The "written as a JSON file" denotes it should be a `.json` file, but given this description and the section description, I wasn't sure if this was a file I had to create with contents relating to the GCS bucket. 

A way to make this more clear might be to specify "Create a bucket" then use that value as part of the variable: `gs://$CONFIG_BUCKET/config.json` (for example).

glasnt: You also never explicitly say to create any buckets. This is possibly one of many you need (one for the config, possibly a separate one for staging later)

dataflow/gemma-flex-template/README.md
```diff
@@ -0,0 +1,179 @@
...
+## Start the pipeline
+To start the Dataflow streaming job, run the following code from the directory. Replace `$TEMPLATE_FILE`, `$REGION`, `$GCS_BUCKET`, `$INPUT_SUBSCRIPTION`, `$OUTPUT_TOPIC`, `$SDK_CONTAINER_IMAGE`, and `$PROJECT` with the Google Cloud project resources you created previously. It might take as much as 15 minutes for the worker to start up and to begin accepting messages from the input Pub/Sub topic. 
+
+```sh
+gcloud dataflow flex-template run "gemma-flex-`date +%Y%m%d-%H%M%S`" \
+  --template-file-gcs-location $TEMPLATE_FILE \
+  --region $REGION \
+  --staging-location $GCS_BUCKET \
+  --parameters messages_subscription=$INPUT_SUBSCRIPTION \
+  --parameters responses_topic=$OUTPUT_TOPIC \
```
glasnt, line 159: These need to be the fully qualified pubsub subscription and pubsub topic name. I received errors trying to use just the topic/subscription name.

dataflow/gemma-flex-template/README.md
```diff
@@ -0,0 +1,179 @@
...
+Follow the steps in this section to create the necessary environment to run this workflow.
+
+### Enable Google Cloud services
+
+This workflow uses multiple Google Cloud products, including Dataflow, Pub/Sub, Google Cloud Storage, and Artifact Registry. Before you start the workflow, create a Google Cloud project that has the following services enabled:
+
+* Dataflow
+* Pub/Sub
+* Google Cloud Storage
+* Artifact Registry
```
glasnt, line 22: This is not the complete list of services that need to be enabled. I also had to enable Compute and Cloud Build. There may also be other APIs, as I ended up referencing the Quickstart Python walkthrough which had many more APIs required that hadn't been enabled in my project yet.

dataflow/gemma-flex-template/README.md
```diff
@@ -0,0 +1,179 @@
...
+
+Your Google Cloud project also needs to have Nvidia L4 GPU quota. For more information, see [GPU quota](https://cloud.google.com/compute/resource-usage#gpu_quota) in the Google Cloud documentation.
+
+### Create a custom container
+
+To build a custom container, use Docker. This repository contains a Dockerfile that you can use to build your custom container. To build and push a container to Artifact Registry by using Docker or Cloud Build, follow the instructions in the [Build and push the image](https://cloud.google.com/dataflow/docs/guides/build-container-image#build_and_push_the_image) section of "Build custom container images for Dataflow" in the Google Cloud documentation.
+
+### Create Pub/Sub topics for input and output
+
+To create your Pub/Sub source and sink, follow the instructions in [Create a Pub/Sub topic](https://cloud.google.com/pubsub/docs/create-topic#pubsub_create_topic-Console) in the Google Cloud documentation. For this example, create two topics, one input topic and one output topic. For input, the topic must have a subscription that you can provide to the Gemma model.
```
glasnt, line 34: This was unclear for me, and I didn't realise I needed two topics, both with linked subscriptions (I wasn't able to get messages out of the sink topic without adding a retention to the topic, which created a pull subscription for me). It might be useful to add the gcloud commands required, which means you can use the topic/subscription variables later in the dataflow invocation

dataflow/gemma-flex-template/README.md
```diff
@@ -0,0 +1,179 @@
...
+```sh
+gcloud dataflow flex-template run "gemma-flex-`date +%Y%m%d-%H%M%S`" \
+  --template-file-gcs-location $TEMPLATE_FILE \
+  --region $REGION \
+  --staging-location $GCS_BUCKET \
+  --parameters messages_subscription=$INPUT_SUBSCRIPTION \
+  --parameters responses_topic=$OUTPUT_TOPIC \
+  --parameters device="GPU" \
+  --parameters sdk_container_image=$SDK_CONTAINER_IMAGE \
+  --parameters dataflow_service_options="worker_accelerator=type:nvidia-l4;count:1;install-nvidia-driver:5xx" \
```
glasnt, line 162: I experienced issues with this parameter ("Unrecognized parameter"). I tried declaring this in the metadata.json to bypass this issue, and this allowed this command to work, and this config showed up in the pipeline options of the job in the Cloud Console, but I'm not a Dataflow expert so I don't know if this is the correct solution. All other parameters in this command apart from sdk_container_image are declared in the metadata.json, so I'm not sure which are default accepted parameters and could not find any documentation to list such things (if they exist)

dataflow/gemma-flex-template/README.md
```diff
@@ -0,0 +1,179 @@
...
+  --project $PROJECT
+```
+
+## Start the pipeline
+To start the Dataflow streaming job, run the following code from the directory. Replace `$TEMPLATE_FILE`, `$REGION`, `$GCS_BUCKET`, `$INPUT_SUBSCRIPTION`, `$OUTPUT_TOPIC`, `$SDK_CONTAINER_IMAGE`, and `$PROJECT` with the Google Cloud project resources you created previously. It might take as much as 15 minutes for the worker to start up and to begin accepting messages from the input Pub/Sub topic. 
+
+```sh
+gcloud dataflow flex-template run "gemma-flex-`date +%Y%m%d-%H%M%S`" \
+  --template-file-gcs-location $TEMPLATE_FILE \
+  --region $REGION \
```
glasnt, line 156: I could not use us-central1 due to capacity constraints, but trying asia-southeast1 caused knockon effects, because this invocation isn't specifying a temporary location, and a default bucket was created for me ("default_gcs_bucket_name" method in [apache_beam.io.gcp.gcsio](https://beam.apache.org/releases/pydoc/2.29.0/_modules/apache_beam/io/gcp/gcsio.html#GcsIO) according to the log files) that was invalid because the bucket name was too long due to the length of the region name. I was able to use asia-east1 to get around this issue, but I think the higher level solution is ensuring that this invocation specifies a `--temp-location` (unsure if this needs to be a parameter of gcloud or an additional `--parameters`.

dataflow/gemma-flex-template/README.md
```diff
@@ -0,0 +1,179 @@
...
+  --parameters device="GPU" \
+  --parameters sdk_container_image=$SDK_CONTAINER_IMAGE \
+  --parameters dataflow_service_options="worker_accelerator=type:nvidia-l4;count:1;install-nvidia-driver:5xx" \
+  --project $PROJECT \
+  --worker-machine-type "g2-standard-4"
+```
+
+## Send a prompt to the model and check the response
+
+In the Google Cloud console, navigate to the Pub/Sub topics page, and then select your input topic. On the **Messages** tab, click **Publish Message**. Add a message for the Dataflow job to pick up and pass through the model. The Dataflow job outputs the response to the Pub/Sub sink topic. To check the response from the model, you can manually pull messages from the destination topic. For more information, see [Publish messages](https://cloud.google.com/pubsub/docs/publisher#publish-messages) in the Google Cloud documentation.
```
glasnt, line 169: I did not know what to send as a message. I would suggest some sample input. 

I also had to add retention to the sink to generate a pull subscription to be able to pull responses to the job. I understand that in practice there might be follow on uses of the sink topic, but for the purposes of this example, there were additional steps needed to check for success.

dataflow/gemma-flex-template/README.md
```diff
@@ -0,0 +1,179 @@
...
+
+## Clean up resources
+
+To avoid incurring charges to your Google Cloud account for the resources used in this example, clean up the resources that you created.
+
+
+*   Cancel the streaming Dataflow job.
+*   Delete the Pub/Sub topic and subscriptions.
+*   Delete the custom container from Artifact Registry.
+*   Empty the `tmp` directory of your Google Cloud Storage bucket.
```
glasnt, line 179: This was never specified to be created in the first place. Possibly could be updated to delete created buckets.

dataflow/gemma-flex-template/README.md
```diff
@@ -0,0 +1,179 @@
...
+## Send a prompt to the model and check the response
+
+In the Google Cloud console, navigate to the Pub/Sub topics page, and then select your input topic. On the **Messages** tab, click **Publish Message**. Add a message for the Dataflow job to pick up and pass through the model. The Dataflow job outputs the response to the Pub/Sub sink topic. To check the response from the model, you can manually pull messages from the destination topic. For more information, see [Publish messages](https://cloud.google.com/pubsub/docs/publisher#publish-messages) in the Google Cloud documentation.
+
+## Clean up resources
+
+To avoid incurring charges to your Google Cloud account for the resources used in this example, clean up the resources that you created.
+
+
+*   Cancel the streaming Dataflow job.
```
glasnt, line 176: Unsure if it should also be archived to complete cleanup.

dataflow/gemma-flex-template/custom_model_gemma.py
```diff
@@ -0,0 +1,182 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import logging
+
+from typing import Any
+from typing import Dict
+from typing import Iterable
+from typing import Optional
+from typing import Sequence
```
glasnt, line 21: These imports are overly verbose and can be reduced to a single line (this occurs elsewhere in this file, but was also pointed out in the previous Gemma PR)

dataflow/gemma-flex-template/custom_model_gemma.py
```diff
@@ -0,0 +1,182 @@
...
+    with beam.Pipeline(options=beam_options) as p:
+        _ = (
+            p | "Read Topic" >>
+            beam.io.ReadFromPubSub(subscription=args.messages_subscription)
+            | "Parse" >> beam.Map(lambda x: x.decode("utf-8"))
+            | "RunInference-Gemma" >> RunInference(
+                handler)  # Send the prompts to the model and get responses.
+            |
+            "Format Output" >> beam.ParDo(FormatOutput())  # Format the output.
+            | "Publish Result" >> beam.io.gcp.pubsub.WriteStringsToPubSub(
```
glasnt, line 181: According to my logs, this method is deprecated since 2.7.0 and should be `WriteToPubSub`

dataflow/gemma-flex-template/metadata.json
```diff
@@ -0,0 +1,21 @@
...
+      {
+        "name": "responses_topic",
+        "label": "Output messages topic",
+        "helpText": "The pub/sub topic to push outputs to"
+      },
+      {
+        "name": "device",
+        "label": "Device",
+        "helpText": "The device to run inference on. Can be 'CPU' or 'GPU'."
+      }
```
glasnt, line 19: Per other comments, this file may need to be extended, but unclear.

dataflow/gemma-flex-template/requirements-test.txt
```diff
@@ -0,0 +1,7 @@
+google-cloud-aiplatform==1.49.0
+google-cloud-dataflow-client==0.8.10
+google-cloud-pubsub==2.21.4
+google-cloud-storage==2.16.0
+pytest==7.4.0
```
glasnt, line 5: Pytest is now up to 8.x. Unless this and other dependencies need to be backpinned, please pin the latest version. Otherwise, as soon as this PR is merged and renovate runs, it will create a PR to bump these anyway.

glasnt (commented):
I tried to run this README and experienced some friction. I was able to eventually get it working, but I'm not sure if my workarounds are best practice for Dataflow (first time using this product)

dataflow/gemma-flex-template/README.md
glasnt: I also think this is the cause of the [failing tests](https://github.com/GoogleCloudPlatform/python-docs-samples/pull/11881#issuecomment-2189254462). In the e2e testing, you're defaulting to us-central1, which I was having contention issues with, and checking the testing project logs, it's also having issues (I can't isolate if the logs I'm seeing are from this test or the other gemma sample, but I think it's the same issue.)

glasnt:
Debugging the tests: the output shows it's a timeout, but the tests are successful(?)

```
collecting ... collected 1 item

e2e_test.py::test_pipeline_dataflow PASSED                               [100%]

-- generated xml file: /workspace/dataflow/gemma-flex-template/sponge_log.xml --
======================== 1 passed in 3642.73s (1:00:42) ========================
nox > Session py-3.10 was successful.


err: signal: killed
```

The kokoro config is set to a max of 60 min ([config](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/main/.kokoro/tests/run_tests.sh#L171)). And you've [configured the test to have a 5400s (2 hour) timeout](https://github.com/GoogleCloudPlatform/python-docs-samples/pull/11881/files#diff-4afb597afe4ed3c3e1fe1eb73b509b86ee3f27d0326e53cbfd4713a484eceb8dR127). 

A similar issue reported in https://github.com/GoogleCloudPlatform/python-docs-samples/issues/4609. 

At a guess, while the image is created each time (~20 mins) and it takes time for the job to start (~20 mins), the success message isn't being received, and this the system has 20 minutes of wait before it times out. 

How long is this entire e2e test expected to be, and is the 2h wait there intentionally? Something else will need to be updated for that decorator to be respected.

dataflow/gemma-flex-template/README.md
jrmccluskey: expanded the list

dataflow/gemma-flex-template/README.md
jrmccluskey: Made the two topic-two subscription requirement more clear, added a link to subscription creation requirements

dataflow/gemma-flex-template/README.md
jrmccluskey: moved

dataflow/gemma-flex-template/README.md
jrmccluskey: removed

dataflow/gemma-flex-template/README.md
jrmccluskey: added instructions for creating a bucket, altered instructions for the template file path

dataflow/gemma-flex-template/README.md
jrmccluskey: the temp location change is relatively easy, just an extra flag that can share the same bucket as the staging location

dataflow/gemma-flex-template/README.md
jrmccluskey: clarified

dataflow/gemma-flex-template/README.md
jrmccluskey: yeah that was a holdover while I experimented, dataflow parameters (sdk_container_image, dataflow_service_options) don't actually need to be passed as parameters. weirdly enough some of them can be passed and parsed as parameters by the dataflow service, but not all of them. cleaned up

dataflow/gemma-flex-template/README.md
jrmccluskey: added a sample

dataflow/gemma-flex-template/README.md
jrmccluskey: should be okay to leave unarchived, we can add that as an optional step though

dataflow/gemma-flex-template/README.md
jrmccluskey: updated

dataflow/gemma-flex-template/requirements-test.txt
jrmccluskey: done

dataflow/gemma-flex-template/metadata.json
jrmccluskey: should be okay to leave as-is as explained above

dataflow/gemma-flex-template/custom_model_gemma.py
jrmccluskey: Updated

dataflow/gemma-flex-template/custom_model_gemma.py
jrmccluskey: Beam style in-repo enforces single line per-import, so I've built a bit of a habit from that. fixed

jrmccluskey:
As far as the E2E test timeout, in early testing we were dancing around the hour-mark as far as runs (some slightly under, some slightly over) so it definitely needs to be over an hour. Building the container + running the job as an invocation from a flex template takes substantial time, so we may need a little longer on the kokoro timeout

dataflow/gemma-flex-template/Dockerfile
```diff
@@ -18,11 +18,12 @@ FROM ${SERVING_BUILD_IMAGE}
 
 WORKDIR /workspace
 
-RUN apt-get update
-RUN apt-get install -y --no-install-recommends apt-utils
-RUN apt-get install -y --no-install-recommends curl
-RUN apt-get install -y --no-install-recommends wget
-RUN apt-get install -y --no-install-recommends git
+RUN apt-get update \
```
tvalentyn, line 21: This can be: 

```
RUN apt-get update \
    && apt-get install -y --no-install-recommends apt-utils curl wget git
    && rm -rf /var/lib/apt/lists/*
```

tvalentyn:
> Building the container + running the job as an invocation from a flex template takes substantial time.

You can significantly bring this down by not including the model and the GPU software into the flex template image. This is a scenario where having two separate images would be better.

tvalentyn:
Also, we can speed up launch time by download the model into the SDK worker container from GCS during container startup, instead of shipping it inside the contanier.  Currently, this could be done by using a custom entrypoint like https://github.com/liferoad/beamllm/blob/main/containers/ollama/entrypoint.sh, eventually we will have a Beam API for that.

tvalentyn:
including in the container proper would be more reliable  but definitely slower, and the model would definitely be a dead weight for the flex template image.

dataflow/conftest.py
```diff
@@ -850,6 +850,7 @@ def dataflow_extensible_template_run(
             f"--project={project}",
             f"--region={region}",
             f"--staging-location=gs://{bucket_name}/staging",
+            f"--temp-location'gs://{bucket_name}/staging",
```
glasnt, line 853: Formatting error
```suggestion
            f"--temp-location=gs://{bucket_name}/staging",
```

dataflow/gemma-flex-template/README.md
```diff
@@ -0,0 +1,177 @@
...
+You can use Gemma models in your Apache Beam inference pipelines with the `RunInference` transform.
+
+This example demonstrates how to use a Gemma model running on Pytorch in a streaming Dataflow pipeline that has Pub/Sub sources and sinks. This pipeline
+is deployed by using Flex Templates.
+
+For more information about using RunInference, see [Get started with AI/ML pipelines](https://beam.apache.org/documentation/ml/overview/) in the Apache Beam documentation.
+
+## Before you begin
+
+Make sure you have followed the [Dataflow setup instructions](../../README.md).
```
glasnt, line 13: This relative path points to the main repo README, not the dataflow README. Suggest using absolute path
```suggestion
Make sure you have followed the [Dataflow setup instructions](/dataflow/README.md).
```

dataflow/gemma-flex-template/README.md
```diff
@@ -135,38 +129,41 @@ class FormatOutput(beam.DoFn):
...
-  --staging-location $GCS_BUCKET \
+  --temp-location gs://$GCS_BUCKET/tmp \
+  --staging-location gs://$GCS_BUCKET \
   --parameters messages_subscription=$INPUT_SUBSCRIPTION \
   --parameters responses_topic=$OUTPUT_TOPIC \
   --parameters device="GPU" \
-  --parameters sdk_container_image=$SDK_CONTAINER_IMAGE \
-  --parameters dataflow_service_options="worker_accelerator=type:nvidia-l4;count:1;install-nvidia-driver:5xx" \
+  --sdk_container_image=$SDK_CONTAINER_IMAGE \
+  --dataflow_service_options="worker_accelerator=type:nvidia-l4;count:1;install-nvidia-driver:5xx" \
```
glasnt, line 157: These parameters aren't recognised by gcloud, and aren't described in https://cloud.google.com/sdk/gcloud/reference/dataflow/flex-template/run

```

ERROR: (gcloud.dataflow.flex-template.run) unrecognized arguments:
  --sdk_container_image=us-central1-docker.pkg.dev/$PROJECT_ID/docker/dataflow/gemma-flex:1
  --dataflow_service_options=worker_accelerator=type:nvidia-l4;count:1;install-nvidia-driver:5xx
  To search the help text of gcloud commands, run:
  gcloud help -- SEARCH_TERMS
```

tvalentyn: > Interesting, we don't expose two pretty important service options for running GPU pipelines via flex templates then. I should be able to update the parameters to pass them through, but that's surprising

* According to https://github.com/GoogleCloudPlatform/python-docs-samples/tree/main/dataflow/flex-templates/pipeline_with_dependencies#run-the-template, --parameters sdk_container_image works
* According to https://cloud.google.com/dataflow/docs/reference/service-options , `--additional-experiments` should be used for service options.

jrmccluskey: the former is true, the latter doesn't seem to work? dataflow gets the correct information in the UI but this error is thrown on startup:

image

This looks like a known issue based on https://github.com/google/dataflow-ml-starter/blob/main/flex/metadata.json

tvalentyn: Discussed offline;
Can we document the reason for using 5xx? Is it a requirement for L4?

This should be Dataflow docs, but  I am not seeing this configuration option documented in https://cloud.google.com/dataflow/docs/gpu/use-gpus#drivers  or in https://cloud.google.com/dataflow/docs/gpu/gpu-support#availability

jrmccluskey: IIRC back when I was first implementing Gemma-on-Dataflow the 5xx driver was necessary for L4s. Strange to not see any mention of 4xx vs 5xx on https://cloud.google.com/dataflow/docs/reference/service-options, that feels like a gap

rszper: I think the 5xx drivers are now the default, so we don't need to specify to use them.

jrmccluskey: I'll do a test run locally to see if that's the case

jrmccluskey: Confirmed that 5xx is no longer necessary

tvalentyn:
What are the next steps on this PR?

tvalentyn:
@jrmccluskey are you still working on changes or you are waiting for a review (if so, please say explicitly that you have addressed previous comments, since current PR status seems to be Changes requested). I see that tests are still failing -- did we figure out how to increase the time limits?

jrmccluskey:
To be clear, comments have been addressed and I am waiting for a review. As far as the timeout, the kokoro config linked above can be updated to run longer; however, I was holding off on updating that since it's a repo-wide timeout. I suppose I should go ahead and update that just for the sake of having a green run on the PR.

tvalentyn:
@glasnt it appears you might have to submit an ICLA for the github user id you've been using here, see: https://github.com/GoogleCloudPlatform/python-docs-samples/pull/11881/checks?check_run_id=27283317137

glasnt:
> @glasnt it appears you might have to submit a CLA for the github user id you've been using here, see: https://github.com/GoogleCloudPlatform/python-docs-samples/pull/11881/checks?check_run_id=27283317137

Resolved.

dataflow/gemma-flex-template/Dockerfile
tvalentyn: @jrmccluskey could you update this please?

.kokoro/tests/run_tests.sh
```diff
@@ -168,7 +168,7 @@ test_prog="${PROJECT_ROOT}/.kokoro/tests/run_single_test.sh"
 
 btlr_args=(
     "run"
-    "--max-cmd-duration=60m"
+    "--max-cmd-duration=70m"
```
glasnt, line 171: I think updating this is causing every test to run, which is the cause of many errors unrelated to this PR

dataflow/conftest.py
```diff
@@ -850,6 +856,7 @@ def dataflow_extensible_template_run(
             f"--project={project}",
             f"--region={region}",
             f"--staging-location=gs://{bucket_name}/staging",
+            f"--temp-location=gs://{bucket_name}/staging",
```
glasnt, line 859: Some environments are reporting this parameter doesn't exist.

glasnt:
The Python 3.10 CI checks might have succeeded, apart from cleanup: 

```
Job has terminated in state FAILED: Workflow job: 2024-07-10_11_09_14-9934820475476190981 failed. Please ensure you have permission to access the job and the `--region` flag, us-central1, matches the job's region.
```

https://btx.cloud.google.com/invocations/0be77b16-c664-4214-b5bf-b70c2588330c/log

dataflow/conftest.py
jrmccluskey: some but not all? these should both be available as per https://cloud.google.com/dataflow/docs/guides/templates/configuring-flex-templates#locations

.kokoro/tests/run_tests.sh
jrmccluskey: that makes sense

jrmccluskey:
Same problem we've seen with some runs here: 

image

Not too much of a surprise, if every test ran at around the same time quota would be tight

.kokoro/tests/run_tests.sh
davidcavazos: Let's try reverting this to 60m, tests shouldn't run for that long anyways and we'll get rid of the "run all the tests" issue.

dataflow/gemma-flex-template/Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+    && apt-get install -y --no-install-recommends wget \
+    && apt-get install -y --no-install-recommends git \
+    && rm -rf /var/lib/apt/lists/*
+
+COPY requirements.txt requirements.txt
+RUN pip install --upgrade pip \
+    && pip install --no-cache-dir -r requirements.txt 
+
+RUN git clone https://github.com/google/gemma_pytorch.git
+RUN pip install ./gemma_pytorch
```
davidcavazos, line 33: This should be `pip install --no-cache-dir ./gemma_pytorch` to reduce image size.

Also I noticed that this installs `torch-2.3.1` while this container is using `pytorch/pytorch:2.1.2-cuda11.8-cudnn8-runtime`, so I'm not sure if there are any incompatibilities. This might require pinning some versions or some other way to make sure the versions match.

It might be a good idea to have some sort of assert somewhere to make sure the versions match.

dataflow/gemma-flex-template/Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+
+# Copy SDK entrypoint binary from Apache Beam image, which makes it possible to
+# use the image as SDK container image.
+COPY --from=apache/beam_python3.10_sdk:2.56.0 /opt/apache/beam /opt/apache/beam
```
davidcavazos, line 37: Please add a comment that the Beam image version should match the one on the requirements.txt file

dataflow/gemma-flex-template/custom_model_gemma.py
```diff
@@ -0,0 +1,178 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import logging
+
+from typing import Any, Dict, Iterable, Optional, Sequence
```
davidcavazos, line 17: As per [PEP-585](https://peps.python.org/pep-0585)
- `Dict` should be `dict`, so that import could be removed
- `Iterable` and `Sequence` should be imported from `collections.abc`

You might need to add a `from __future__ import annotations` to make it work with Beam.

dataflow/gemma-flex-template/custom_model_gemma.py
```diff
@@ -0,0 +1,178 @@
...
+          model: The Gemma model being used.
+          inference_args: Any additional arguments for an inference.
+
+        Returns:
+          An Iterable of type PredictionResult.
+        """
+        # Loop each text string, and use a tuple to store the inference results.
+        predictions = []
+        result = model.generate(prompts=batch, device=self._device)
+        predictions.append(result)
```
davidcavazos, line 103: Initializing `predictions` to an empty list and then mutating it seems a bit weird. How about just initializing in place?

```py
predictions = [result]
```

dataflow/gemma-flex-template/custom_model_gemma.py
```diff
@@ -0,0 +1,178 @@
...
+        return [PredictionResult(x, y) for x, y in zip(batch, predictions)]
+
+
+class FormatOutput(beam.DoFn):
```
davidcavazos, line 107: This could be a simple function rather than an entire class.

```py
def format_output(element):
  return "Input: {input}, Output: {output}".format(
      input=element.example, output=element.inference)
```

And then use `beam.Map` instead of `beam.ParDo`:

```py
    | "Format Output" >> beam.Map(format_output)
```

dataflow/gemma-flex-template/custom_model_gemma.py
```diff
@@ -0,0 +1,178 @@
...
+    args, beam_args = parser.parse_known_args()
+
+    config = get_config_for_2b()
+
+    logging.getLogger().setLevel(logging.INFO)
+    beam_options = PipelineOptions(
+        beam_args,
+        streaming=True,
+    )
+    beam_options.view_as(SetupOptions).save_main_session = True
```
davidcavazos, line 161: This could be set inline rather than mutating the options.

```py
    beam_options = PipelineOptions(
        beam_args,
        save_main_session=True,
        streaming=True,
    )
```

dataflow/gemma-flex-template/custom_model_gemma.py
```diff
@@ -0,0 +1,178 @@
...
+
+    with beam.Pipeline(options=beam_options) as p:
+        _ = (
+            p | "Read Topic" >>
+            beam.io.ReadFromPubSub(subscription=args.messages_subscription)
+            | "Parse" >> beam.Map(lambda x: x.decode("utf-8"))
+            | "RunInference-Gemma" >> RunInference(
+                handler)  # Send the prompts to the model and get responses.
+            |
+            "Format Output" >> beam.ParDo(FormatOutput())  # Format the output.
```
davidcavazos, line 176: This comment could be omitted, both the step name and function name already state this.

davidcavazos:
After looking at the system logs for the Dataflow workers in one of the earlier runs, it looks like the workers don't have enough disk space to load the model.

`Handler for GET /v1.41/images/get returned error: write /var/lib/docker/tmp/docker-export-3281602760/0e6537f85f3ebad7a4b5af8385d234950c2861657142f4f53123b65c153127fe/layer.tar: no space left on device`

dataflow/gemma-flex-template/custom_model_gemma.py
jrmccluskey: removed

dataflow/gemma-flex-template/custom_model_gemma.py
jrmccluskey: ah this was a holdover from the keras example that wasn't able to batch requests, fixed

dataflow/gemma-flex-template/custom_model_gemma.py
jrmccluskey: done

dataflow/gemma-flex-template/Dockerfile
jrmccluskey: done

dataflow/gemma-flex-template/Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends apt-utils \
+    && apt-get install -y --no-install-recommends curl \
+    && apt-get install -y --no-install-recommends wget \
+    && apt-get install -y --no-install-recommends git \
+    && rm -rf /var/lib/apt/lists/*
+
+COPY requirements.txt requirements.txt
+RUN pip install --upgrade pip \
```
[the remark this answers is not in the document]
jrmccluskey: done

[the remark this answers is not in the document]
jrmccluskey: done

dataflow/conftest.py
```diff
@@ -850,6 +856,7 @@ def dataflow_extensible_template_run(
             f"--project={project}",
             f"--region={region}",
             f"--staging-location=gs://{bucket_name}/staging",
```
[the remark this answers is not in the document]
jrmccluskey: done

davidcavazos:
`Startup of the worker pool in zone us-central1-a failed to bring up any of the desired 1 workers. Please refer to https://cloud.google.com/dataflow/docs/guides/common-errors#worker-pool-failure for help troubleshooting. ZONE_RESOURCE_POOL_EXHAUSTED_WITH_DETAILS: Instance 'dataflow-gemma-flex-templ-07171029-mk9a-harness-7qn5' creation failed: The zone 'projects/python-docs-samples-tests/zones/us-central1-a' does not have enough resources available to fulfill the request.  '(resource type:compute)'.`

Looks like quota issue?

jrmccluskey:
What are next steps here? This PR has been hanging for a while

davidcavazos:
We need the tests to pass in order to merge.

tvalentyn:
Is it still quota issues or something else?

tvalentyn:
```
=================================== FAILURES ===================================
____________________________ test_pipeline_dataflow ____________________________
  File "/workspace/dataflow/gemma-flex-template/.nox/py-3-10/lib/python3.10/site-packages/google/api_core/retry/retry_unary.py", line 144, in retry_target
    result = target()
  File "/workspace/dataflow/conftest.py", line 167, in pubsub_wait_for_messages
    messages = [m.message.data.decode("utf-8") for m in response.received_messages]
  File "/workspace/dataflow/conftest.py", line 167, in 
    messages = [m.message.data.decode("utf-8") for m in response.received_messages]
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xb5 in position 3: invalid start byte

```

dataflow/gemma-flex-template/e2e_test.py
```diff
@@ -0,0 +1,154 @@
...
+        flex_template_path: str,
+        messages_subscription: str,
+        responses_topic: str,
+) -> str:
+    yield from utils.dataflow_flex_template_run(
+        job_name=NAME,
+        template_path=flex_template_path,
+        bucket_name=bucket_name,
+        project=project,
+        region=location,
```
glasnt, line 116: Other dataflow examples default to the conftest.py region of us-central1. I think changing this to (presumably) us-west1 here will make this sample work, because there's availability in that region.

Attempting to change this value here to see if it will work (will revert if required)

glasnt:
I've tried some experiments on the region changing, and the decode error. These may need to be updated/reverted etc.

dataflow/conftest.py
```diff
@@ -163,7 +163,12 @@ def pubsub_wait_for_messages(subscription_path: str) -> list[str]:
 
     subscriber = pubsub.SubscriberClient()
     with subscriber:
+        print(f"pulling from subscription_path")
         response = subscriber.pull(subscription=subscription_path, max_messages=10)
+
+        print("Raw data received:")
+        print([str(m.message.data) for m in response.received_messages])
```
glasnt, line 170: @jrmccluskey based on this debugging, your messages are coming back base64 encoded. 

From the debugging out of the last execution: 

```
Raw data received:
['b\'"zn\\xb58b\\xb2+\\x1a\\xb5\\xeb-~\\x8a\\xda?+a\\xa2{\\x1a\\x9a\\x99^:\\xebi\\xba\\xd9\\xe7Z\\xb8\\xady\\xa3\\xf2\\xb6\\x1a\\\'\\xa6\\xba \\xad\\xa9\\xad\\xa2)\\xa9\\x95\\xe9\\x9e\\x9e\\xd6\\x9b\\x8av\\xab\\xca\\xc7\\x9a\\xad\\xc8Z\\x96\\n+\\x8a\\xd8f\\xb6\\x87\\xe2\\x9d\\xdbaz\\x9a,\\x8a\\xd8\\xa8\\x9e\\x87\\xda\\xb2\\x97\\x9c\\x89\\xf8\\x9eu\\xe9^\\x99\\xe9\\xed\\x8av\\xa5\\x8a\\xcbg\\x9e\\x9c\\xad\\x86\\x89\\xe7u\\xe7\\xdb\\x8av\\xab\\xcb\\xfb\\x1ej\\xb7!\\x96+-\\xd6\\xd6\\xab\\x81\\xebg\\x96\\x8c\'']
```

With some slight manipulation to remove repr escapes (`s/\\/\`, `s/\'/'`), this noise looks like data. 

```
>>> data = b'"zn\xb58b\xb2+\x1a\xb5\xeb-~\x8a\xda?+a\xa2{\x1a\x9a\x99^:\xebi\xba\xd9\xe7Z\xb8\xady\xa3\xf2\xb6\x1a\"\xa6\xba \xad\xa9\xad\xa2)\xa9\x95\xe9\x9e\x9e\xd6\x9b\x8av\xab\xca\xc7\x9a\xad\xc8Z\x96\n+\x8a\xd8f\xb6\x87\xe2\x9d\xdbaz\x9a,\x8a\xd8\xa8\x9e\x87\xda\xb2\x97\x9c\x89\xf8\x9eu\xe9^\x99\xe9\xed\x8av\xa5\x8a\xcbg\x9e\x9c\xad\x86\x89\xe7u\xe7\xdb\x8av\xab\xcb\xfb\x1ej\xb7!\x96+-\xd6\xd6\xab\x81\xebg\x96\x8c"'
>>> 
>>> base64.b64encode(data)
b'InputThisisatestforaPythonsampleOutputnnWriteaPythoiprogramtoimplementabinarysearchalgorithmtofindthepositionofaspecifiedelementinalistnnpythonndefbinary/searchlist1targetnlowi'
```

My guess here there's a missing de/serialisation step somewhere.

Comparing with the existing gemma example, removing the differences with the implied `run` when using `with beam.Pipeline` ([source](https://stackoverflow.com/a/68718420/124019)), can you think of any other reason why the data would be half de/serialized like this? 

(Self note: The debugging in this particular part of the file needs to be removed before merging)

jrmccluskey: This is strange. The write we do to Pub/Sub converts our string to bytes and writes to the pub sub topic. I don't think we're missing anything as far as getting the data out. The implicit run shouldn't matter here either

glasnt:
This should now run green with the encoding issues resolved. Since I worked so much on this one, I'd want someone !me to approve this PR

jrmccluskey:
Sure enough, actually green! Thank you for the debugging work @glasnt !

davidcavazos:
I made some code simplifications. I think the issue with the byte64 was that Pub/Sub expects bytes and it was being passed a string. The local runner failed, but I suspect the Dataflow runner was implicitly converting it into base64. I made some changes to try to simplify the code a bit, and hopefully the tests stay green.

jrmccluskey:
Looks like the workflow was running during the github outage last night, failed on the git clone step

dataflow/gemma-flex-template/Dockerfile
davidcavazos: Done

davidcavazos (approved):
LGTM, tests are passing but take ~55m to run. We can merge, but it would be nice to optimize this further.

glasnt:
> LGTM, tests are passing but take 50+ minutes to run. We can merge, but it would be nice to optimize this further.

A majority of this is the image build, blocking on network; but as long as it's <60m, we're ok.

dataflow/gemma-flex-template/README.md
```diff
@@ -0,0 +1,179 @@
...
+```sh
+export GCS_BUCKET="your--bucket"
+gsutil mb gs://$GCS_BUCKET
+```
+
+Make sure your GCS bucket name does __not__ include the `gs://` prefix
+
+### Create a custom container
+
+To build a custom container, use Docker. This repository contains a Dockerfile that you can use to build your custom container. To build and push a container to Artifact Registry by using Docker or Cloud Build, follow the instructions in the [Build and push the image](https://cloud.google.com/dataflow/docs/guides/build-container-image#build_and_push_the_image) section of "Build custom container images for Dataflow" in the Google Cloud documentation.
```
iennae, line 53: question: just curious if any supported OCI image would do. Artifact Registry supports them. Then, this is more of a We've provided a sample Dockerfile to showcase building a custom container. There are other tools folks use for building containers and it's better to not imply the specific tool choice if it's not necessary. I might recommend dropping the first sentence of "To build a custom container, use Docker"

iennae (approved):
Thanks everyone. Really appreciate the team effort to reduce the time and complexity of this sample.

PR #11885, opened 2024-06-20
samples(discoveryengine): Add Website Import

holtskinner:
For page https://cloud.google.com/generative-ai-app-builder/docs/create-data-store-es

### DIFF
--- a/discoveryengine/site_search_engine_sample.py
+++ b/discoveryengine/site_search_engine_sample.py
@@ -14,24 +14,23 @@
 #
 
 
-# [START genappbuilder_create_target_site]
-from google.api_core.client_options import ClientOptions
-
-from google.cloud import discoveryengine_v1 as discoveryengine
-
-# TODO(developer): Uncomment these variables before running the sample.
-# project_id = "YOUR_PROJECT_ID"
-# location = "YOUR_LOCATION" # Values: "global"
-# data_store_id = "YOUR_DATA_STORE_ID"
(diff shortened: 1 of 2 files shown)


discoveryengine/site_search_engine_sample_test.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+def test_create_target_site():
+    response = site_search_engine_sample.create_target_site(
+        project_id,
+        location,
+        data_store_id,
+        uri_pattern="https://cloud.google.com/generative-ai-app-builder/docs/*",
+    )
+    assert response
```
davidcavazos, line 33: Nit: `assert response, response` to aid debugging

discoveryengine/site_search_engine_sample.py
```diff
@@ -0,0 +1,132 @@
...
+# limitations under the License.
+#
+
+
+# [START genappbuilder_create_target_site]
+from google.api_core.client_options import ClientOptions
+
+from google.cloud import discoveryengine_v1 as discoveryengine
+
+# TODO(developer): Uncomment these variables before running the sample.
```
davidcavazos, line 22: Do we want to have these TODOs inside the sample function? Otherwise wouldn't they be overridden by the function arguments when called?

davidcavazos (approved):

discoveryengine/site_search_engine_sample.py
holtskinner: I'm following the typical style used by most python code samples.

PR #11951, opened 2024-06-27
Updated forecast sample

ey00:
The default model response is a bit confusing since some fields aren't being included (b/349685260).


iennae (approved):
Before this gets merged, the doc response output CL should also be queued so that the sample shown doesn't conflict with the response I think. https://cloud.google.com/vertex-ai/generative-ai/docs/multimodal/control-generated-output#forecast

gericdong (approved):

PR #12062, opened 2024-07-23
soft-delete-cost-analysis: fixing return type

marcinlimanski:
## Description
Fixing run time error with return type

### DIFF
--- a/storage/cost-analysis/storage_soft_delete_relative_cost_analyzer.py
+++ b/storage/cost-analysis/storage_soft_delete_relative_cost_analyzer.py
@@ -28,6 +28,7 @@ Relative cost of each bucket = deleted_bytes / total_byte_seconds
 """
 # [START storage_soft_delete_relative_cost]
 from __future__ import annotations
+
 import argparse
 import json
 import google.cloud.monitoring_v3 as monitoring_client
@@ -58,7 +59,7 @@ def get_soft_delete_cost(
     soft_delete_window: float,
     agg_days: int,
     lookback_days: int,
-) -> Dict[str, List[Dict[str, float]]]:
+) -> dict[str, list[dict[str, float]]]:
     """Calculates soft delete costs for buckets in a Google Cloud project.
 
     Args:
@@ -97,10 +98,10 @@ def calculate_soft_delete_costs(
     project_name: str,
     query_client: monitoring_client.QueryServiceClient,
     soft_delete_window: float,
-    storage_ratios_by_bucket: Dict[str, float],
+    storage_ratios_by_bucket: dict[str, float],
     agg_days: int,
     lookback_days: int,
-) -> Dict[str, List[Dict[str, float]]]:
+) -> dict[str, list[dict[str, float]]]:
     """Calculates the relative cost of enabling soft delete for each bucket in a
        project for certain time frame in secs.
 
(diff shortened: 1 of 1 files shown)


storage/cost-analysis/storage_soft_delete_relative_cost_analyzer.py
```diff
@@ -238,7 +238,7 @@ def soft_delete_relative_cost_analyzer(
     agg_days: int = 30,
     lookback_days: int = 360,
     list_buckets: bool = False,
-) -> str | Dict[str, float]:  # Note potential string output
+    ) -> Union[str, Dict[str, float]]: # Note potential string output
```
msampathkumar, line 241: Thanks for the PR. As per https://docs.python.org/3/library/typing.html#special-forms, `Union` does the same job as `|`. Any particular reason why change this ?

marcinlimanski: Thanks for quick replay! Running the script with python 3.9 I got this issue:

python-docs-samples/storage/cost-analysis/storage_soft_delete_relative_cost_analyzer.py", line 241, in <module>
    ) -> str | Dict[str, float]:  # Note potential string output
TypeError: unsupported operand type(s) for |: 'type' and '_GenericAlias'

The above change fixes this.

marcinlimanski: The shorthand '|' for union is only supported in python 3.10 and up.

iennae:
@marcinlimanski thanks for the PR and your first time contribution to the repository. It looks like you need to sign the CLA before we can accept this request.

marcinlimanski:
@iennae Thanks for the reminder, CLA is signed. PR ready to go.

storage/cost-analysis/storage_soft_delete_relative_cost_analyzer.py
davidcavazos: We can still use `|` in Python 3.9 by adding `from __future__ import annotations`, just confirmed in my local setup.

storage/cost-analysis/storage_soft_delete_relative_cost_analyzer.py
```diff
@@ -31,7 +31,7 @@
 
 import argparse
 import json
-from typing import Dict, List
+from typing import Dict, List, Union
```
davidcavazos, line 34: As per [PEP-585](https://peps.python.org/pep-0585/), we can use `dict` and `list` directly (like `dict[str, float]`), so we can get rid of this whole import.

storage/cost-analysis/storage_soft_delete_relative_cost_analyzer.py
```diff
@@ -238,7 +236,7 @@ def soft_delete_relative_cost_analyzer(
     agg_days: int = 30,
     lookback_days: int = 360,
     list_buckets: bool = False,
-) -> str | Dict[str, float]:  # Note potential string output
+    ) -> str | Dict[str, float]: # Note potential string output
```
davidcavazos, line 239: `Dict[str, float]` -> `dict[str, float]`

davidcavazos (approved):

PR #12063, opened 2024-07-23
Remove execute/assessment from homepage load.

faube:
Remove execute/assessment from homepage load.

This is no longer something we want to showcase in reCAPTCHA demosite.


msampathkumar:
@Sita04 you might be the best person to address this. PTAL when possible.

Cory-Kramer (approved):

Sita04 (approved):

PR #12089, opened 2024-07-31
fix: Text-Embedding Model Serving demo.

kewent:
## Description

Fixes: http://b/349444341

## Checklist

### DIFF
--- a/generative_ai/embedding_preview.py
+++ b/generative_ai/embedding_preview.py
@@ -19,21 +19,24 @@ from vertexai.language_models import TextEmbeddingInput, TextEmbeddingModel
 
 
 def embed_text(
-    texts: List[str] = ["banana muffins? ", "banana bread? banana muffins?"],
+    texts: List[str] = [
+        "Retrieve a function that adds two numbers",
+        "def func(a, b): return a + b",
+    ],
     task: str = "CODE_RETRIEVAL_QUERY",
     model_name: str = "text-embedding-preview-0815",
     dimensionality: Optional[int] = 256,
 ) -> List[List[float]]:
-    """Embeds texts with a pre-trained, foundational model."""
-    model = TextEmbeddingModel.from_pretrained(model_name)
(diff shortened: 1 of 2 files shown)


holtskinner (approved):

generative_ai/embedding_preview.py
```diff
@@ -18,13 +18,31 @@
 from vertexai.language_models import TextEmbeddingInput, TextEmbeddingModel
 
 
-def embed_text(
-    texts: List[str] = ["banana muffins? ", "banana bread? banana muffins?"],
-    task: str = "QUESTION_ANSWERING",
-    model_name: str = "text-embedding-preview-0409",
+def embed_code_query(
+    texts: List[str] = ["Retrieve a function that adds two numbers"],
```
davidcavazos, line 22: Can we inline all the function arguments into the code samples? It makes the samples easier to read.

generative_ai/embedding_preview.py
```diff
@@ -18,13 +18,31 @@
...
+    task: str = "CODE_RETRIEVAL_QUERY",
+    model_name: str = "text-embedding-preview-0815",
     dimensionality: Optional[int] = 256,
 ) -> List[List[float]]:
```
davidcavazos, line 26: As per [PEP-585](https://peps.python.org/pep-0585/), we can use `list` directly, so this can be changed to `list[list[float]]`, and the typing import can be removed.

generative_ai/embedding_preview.py
kewent: Done.

generative_ai/embedding_preview.py
kewent: Since we're passing the python 3.8 build, I think we may have to use List instead of list. WDYT?

davidcavazos: For older Python versions, you can include `from __future__ import annotations` and that should work.

davidcavazos (approved):

PR #12105, opened 2024-08-06
chore: Remove quota project from API key auth sample.

clundin25:
Quota should be derived from the API key.

ETA:
based on documentation for the underlying API, the call authenticates with the API key and doesn't do anything associated with a passed in project ID. 

This will update https://cloud.google.com/docs/authentication/api-keys


iennae:
@clundin25 based on the test results, you'll need to update the tests as well in api_key_test.py.

PR #12158, opened 2024-08-09
Simplified next set of Speech Samples (set-3)

Thoughtseize1:
## Description

Fixes #<ISSUE-NUMBER>

## Checklist

### DIFF
--- a/speech/snippets/transcribe_feature_in_recognizer.py
+++ b/speech/snippets/transcribe_feature_in_recognizer.py
@@ -20,7 +20,8 @@ from google.api_core.exceptions import NotFound
 from google.cloud.speech_v2 import SpeechClient
 from google.cloud.speech_v2.types import cloud_speech
 
-
+# TODO (Developer): Update the PROJECT_ID to the value of your project
+# PROJECT_ID = "your-project-id"
 PROJECT_ID = os.getenv("GOOGLE_CLOUD_PROJECT")
 
 
@@ -43,7 +44,6 @@ def transcribe_feature_in_recognizer(
     recognizer_name = (
         f"projects/{PROJECT_ID}/locations/global/recognizers/{recognizer_id}"
     )
-
     try:
         # Use an existing recognizer
         recognizer = client.get_recognizer(name=recognizer_name)
--- a/speech/snippets/transcribe_feature_in_recognizer.py
+++ b/speech/snippets/transcribe_feature_in_recognizer.py
@@ -12,50 +12,69 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-
-import argparse
-
 # [START speech_transcribe_feature_in_recognizer]
+import os
+
+from google.api_core.exceptions import NotFound
+
 from google.cloud.speech_v2 import SpeechClient
 from google.cloud.speech_v2.types import cloud_speech
 
 
+PROJECT_ID = os.getenv("GOOGLE_CLOUD_PROJECT")
+
+
+# noinspection PyTypeChecker
 def transcribe_feature_in_recognizer(
-    project_id: str,
-    recognizer_id: str,
     audio_file: str,
+    recognizer_id: str,
 ) -> cloud_speech.RecognizeResponse:
-    """Transcribe an audio file using an existing recognizer."""
+    """Use an existing recognizer or create a new one to transcribe an audio file.
+    Args:
+        audio_file (str): The path to the audio file to be transcribed.
+            Example: "resources/audio.wav"
+        recognizer_id (str): The ID of the recognizer to be used or created. ID should be unique
+            within the project and location.
+    Returns:
+        cloud_speech.RecognizeResponse: The response containing the transcription results.
+    """
     # Instantiates a client
     client = SpeechClient()
-
-    request = cloud_speech.CreateRecognizerRequest(
-        parent=f"projects/{project_id}/locations/global",
-        recognizer_id=recognizer_id,
-        recognizer=cloud_speech.Recognizer(
-            default_recognition_config=cloud_speech.RecognitionConfig(
-                auto_decoding_config=cloud_speech.AutoDetectDecodingConfig(),
-                language_codes=["en-US"],
-                model="latest_long",
-                features=cloud_speech.RecognitionFeatures(
-                    enable_automatic_punctuation=True,
-                ),
-            ),
-        ),
+    recognizer_name = (
+        f"projects/{PROJECT_ID}/locations/global/recognizers/{recognizer_id}"
     )
 
-    operation = client.create_recognizer(request=request)
-    recognizer = operation.result()
-
-    print("Created Recognizer:", recognizer.name)
+    try:
+        # Try to fetch the existing recognizer
+        recognizer = client.get_recognizer(name=recognizer_name)
+        print("Using existing Recognizer:", recognizer.name)
+    except NotFound:
+        # Create a new recognizer if it does not exist
+        request = cloud_speech.CreateRecognizerRequest(
+            parent=f"projects/{PROJECT_ID}/locations/global",
+            recognizer_id=recognizer_id,
+            recognizer=cloud_speech.Recognizer(
+                default_recognition_config=cloud_speech.RecognitionConfig(
+                    auto_decoding_config=cloud_speech.AutoDetectDecodingConfig(),
+                    language_codes=["en-US"],
+                    model="latest_long",
+                    features=cloud_speech.RecognitionFeatures(
+                        enable_automatic_punctuation=True,
+                    ),
+                ),
+            ),
+        )
+        operation = client.create_recognizer(request=request)
+        recognizer = operation.result()
+        print("Created Recognizer:", recognizer.name)
 
     # Reads a file as bytes
     with open(audio_file, "rb") as f:
-        content = f.read()
+        audio_content = f.read()
 
     request = cloud_speech.RecognizeRequest(
-        recognizer=f"projects/{project_id}/locations/global/recognizers/{recognizer_id}",
-        content=content,
+        recognizer=f"projects/{PROJECT_ID}/locations/global/recognizers/{recognizer_id}",
+        content=audio_content,
     )
 
     # Transcribes the audio into text
@@ -71,11 +90,6 @@ def transcribe_feature_in_recognizer(
 
 
 if __name__ == "__main__":
-    parser = argparse.ArgumentParser(
-        description=__doc__, formatter_class=argparse.RawDescriptionHelpFormatter
+    transcribe_feature_in_recognizer(
+        audio_file="resources/audio.wav", recognizer_id="id-recognizer"
     )
-    parser.add_argument("project_id", help="GCP Project ID")
-    parser.add_argument("recognizer_id", help="Recognizer ID to use for recogniition")
-    parser.add_argument("audio_file", help="Audio file to stream")
-    args = parser.parse_args()
-    transcribe_feature_in_recognizer(args.project_id, args.audio_file)
--- a/speech/snippets/transcribe_feature_in_recognizer.py
+++ b/speech/snippets/transcribe_feature_in_recognizer.py
@@ -11,8 +11,6 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
-# [START speech_transcribe_feature_in_recognizer]
 import os
 
 from google.api_core.exceptions import NotFound
@@ -20,8 +18,6 @@ from google.api_core.exceptions import NotFound
 from google.cloud.speech_v2 import SpeechClient
 from google.cloud.speech_v2.types import cloud_speech
 
-# TODO (Developer): Update the PROJECT_ID to the value of your project
-# PROJECT_ID = "your-project-id"
 PROJECT_ID = os.getenv("GOOGLE_CLOUD_PROJECT")
 
 
@@ -39,8 +35,14 @@ def transcribe_feature_in_recognizer(
     Returns:
         cloud_speech.RecognizeResponse: The response containing the transcription results.
     """
+    # [START speech_transcribe_feature_in_recognizer]
+
     # Instantiates a client
     client = SpeechClient()
+
+    # TODO (Developer): uncomment and update the PROJECT_ID and recognizer_id
+    # PROJECT_ID = "your-project-id"
+    # recognizer_id = "id-recognizer"
     recognizer_name = (
         f"projects/{PROJECT_ID}/locations/global/recognizers/{recognizer_id}"
     )
@@ -83,10 +85,9 @@ def transcribe_feature_in_recognizer(
     for result in response.results:
         print(f"Transcript: {result.alternatives[0].transcript}")
 
+    # [END speech_transcribe_feature_in_recognizer]
+
     return response
-
-
-# [END speech_transcribe_feature_in_recognizer]
 
 
 if __name__ == "__main__":
--- a/speech/snippets/transcribe_feature_in_recognizer.py
+++ b/speech/snippets/transcribe_feature_in_recognizer.py
@@ -45,11 +45,11 @@ def transcribe_feature_in_recognizer(
     )
 
     try:
-        # Try to fetch the existing recognizer
+        # Use an existing recognizer
         recognizer = client.get_recognizer(name=recognizer_name)
         print("Using existing Recognizer:", recognizer.name)
     except NotFound:
-        # Create a new recognizer if it does not exist
+        # Create a new recognizer
         request = cloud_speech.CreateRecognizerRequest(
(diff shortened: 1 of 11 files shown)


speech/snippets/transcribe_feature_in_recognizer.py
```diff
@@ -12,50 +12,69 @@
...
-        recognizer_id=recognizer_id,
-        recognizer=cloud_speech.Recognizer(
-            default_recognition_config=cloud_speech.RecognitionConfig(
-                auto_decoding_config=cloud_speech.AutoDetectDecodingConfig(),
-                language_codes=["en-US"],
-                model="latest_long",
-                features=cloud_speech.RecognitionFeatures(
-                    enable_automatic_punctuation=True,
+    try:
+        # Try to fetch the existing recognizer
```
msampathkumar, line 48:
```suggestion
        # use existing recognizer
```

speech/snippets/transcribe_feature_in_recognizer.py
```diff
@@ -12,50 +12,69 @@
...
+        recognizer = client.get_recognizer(name=recognizer_name)
+        print("Using existing Recognizer:", recognizer.name)
+    except NotFound:
+        # Create a new recognizer if it does not exist
```
msampathkumar, line 52:
```suggestion
        # create a new recognizer
```

msampathkumar (commented):
Overall LGTM. Lets wait for the other teams to respond.

speech/snippets/transcribe_feature_in_recognizer.py
Thoughtseize1: Rewrite comment

speech/snippets/transcribe_feature_in_recognizer.py
```diff
@@ -12,50 +12,69 @@
...
 
-import argparse
+from google.api_core.exceptions import NotFound
 
-# [START speech_transcribe_feature_in_recognizer]
 from google.cloud.speech_v2 import SpeechClient
 from google.cloud.speech_v2.types import cloud_speech
 
 
+PROJECT_ID = os.getenv("GOOGLE_CLOUD_PROJECT")
```
davidcavazos, line 24: Can we move this outside the region tag with a comment with explicit instructions on setting this manually?

Same with other cases

speech/snippets/transcribe_feature_in_recognizer_test.py
```diff
@@ -24,29 +24,29 @@
...
-    project_id = os.getenv("GOOGLE_CLOUD_PROJECT")
     recognizer_id = "recognizer-" + str(uuid4())
 
     def cleanup() -> None:
-        delete_recognizer(project_id, recognizer_id)
+        delete_recognizer(recognizer_id)
 
     request.addfinalizer(cleanup)
 
     response = transcribe_feature_in_recognizer.transcribe_feature_in_recognizer(
```
davidcavazos, line 48: From a quick glance it looks like this test is only testing the case when the recognizer doesn't exist and a new one is created. Since a new branch of code was added, this should also test the case when the recognizer exists.

speech/snippets/transcribe_model_selection.py
```diff
@@ -14,38 +14,39 @@
...
-
-    audio = speech.RecognitionAudio(content=content)
+    audio = speech.RecognitionAudio(content=audio_content)
 
     config = speech.RecognitionConfig(
         encoding=speech.RecognitionConfig.AudioEncoding.LINEAR16,
         sample_rate_hertz=16000,
         language_code="en-US",
-        model=model,
+        model=model,  # Chosen model
```
davidcavazos, line 49: To make this sample copy-paste runnable, can we hard-code a specific model for the sample? It would also be nice to have comment with a link to the documentation to see the possible options.

speech/snippets/transcribe_model_selection_v2.py
```diff
@@ -12,37 +12,45 @@
...
     # Reads a file as bytes
     with open(audio_file, "rb") as f:
-        content = f.read()
+        audio_content = f.read()
 
     config = cloud_speech.RecognitionConfig(
         auto_decoding_config=cloud_speech.AutoDetectDecodingConfig(),
         language_codes=["en-US"],
-        model=model,
+        model=model,  # Chosen model
```
davidcavazos, line 47: Same here, we should have a specific model for the sample and a link to see the possible options available.

speech/snippets/transcribe_feature_in_recognizer_test.py
Thoughtseize1: I've added a new `try/except` block in `transcribe_feature_in_recognizer` function [here](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/8eebcab04db893edcad9ecd86565b9f501d3697d/speech/snippets/transcribe_feature_in_recognizer.py#L47) 

Before my changes, this function always created a new recognizer and did not attempt to use an existing one. In test case, we create a recognizer that should be deleted.

I added an additional function call to the test to see if the function uses an existing recognizer.
Also added additional asserts

speech/snippets/transcribe_feature_in_recognizer.py
Thoughtseize1: I guess i can add some comment to help users understand how they can update PROJECT_ID. 

If I move PROJECT_ID outside of the region tag, we will lose the sense to simplify Sample and will not pass lint checks.

speech/snippets/transcribe_model_selection.py
Thoughtseize1: I think it's a good practice to let users send model into function arguments. 
We could use if statement inside `RecognitionConfig` but I think it will complicate the code.

```
  config = speech.RecognitionConfig(
        encoding=speech.RecognitionConfig.AudioEncoding.LINEAR16,
        sample_rate_hertz=16000,
        language_code="en-US",
        model=model if model else "phone_call",
    )
```
Moreover, a, the list of possible models is already presented in the docstrings of each function with examples. And also in the main block 👀

I’d leave it as is.

speech/snippets/transcribe_model_selection_v2.py
Thoughtseize1: Similar to the previous comment, I would leave it as is.
All possible models are presented via a link in the docstring
IMHO

speech/snippets/transcribe_feature_in_recognizer_test.py
```diff
@@ -24,33 +24,48 @@
...
+            os.path.join(_RESOURCES, "audio.wav"), recognizer_id
+        )
+    )
+    # Call function one more time to test work with the existing recognizer
+    response_with_existing_recognizer = (
+        transcribe_feature_in_recognizer.transcribe_feature_in_recognizer(
+            os.path.join(_RESOURCES, "audio.wav"), recognizer_id
+        )
+    )
+    out, _ = capsys.readouterr()
```
davidcavazos, line 59: Testing on stdout can be brittle and break if the underlying implementation changes the output messages. The recommended approach is to return the response from the sample function and test that directly here.

speech/snippets/transcribe_feature_in_recognizer.py
```diff
@@ -12,50 +12,69 @@
...
+# PROJECT_ID = "your-project-id"
+PROJECT_ID = os.getenv("GOOGLE_CLOUD_PROJECT")
+
 
+# noinspection PyTypeChecker
 def transcribe_feature_in_recognizer(
-    project_id: str,
-    recognizer_id: str,
     audio_file: str,
+    recognizer_id: str,
```
davidcavazos, line 31: For a sample, it's best to have hard coded values to make it copy-paste-runnable. Can we inline these arguments into the sample instead of relying on opaque function arguments? 

From below, it looks like `audio_file="resources/audio.wav", recognizer_id="id-recognizer"` would work for these?

speech/snippets/transcribe_feature_in_recognizer.py
```diff
@@ -71,11 +90,6 @@ def transcribe_feature_in_recognizer(
 
 
 if __name__ == "__main__":
-    parser = argparse.ArgumentParser(
-        description=__doc__, formatter_class=argparse.RawDescriptionHelpFormatter
+    transcribe_feature_in_recognizer(
```
davidcavazos, line 93: As per the [style guide](https://googlecloudplatform.github.io/samples-style-guide/#no-cli) it's discouraged to provide a CLI interface on code samples.

speech/snippets/transcribe_file_v2.py
```diff
@@ -12,24 +12,33 @@
...
 from google.cloud.speech_v2.types import cloud_speech
 
+# TODO (Developer): Update the PROJECT_ID to the value of your project
+# PROJECT_ID = "your-project-id"
+PROJECT_ID = os.getenv("GOOGLE_CLOUD_PROJECT")
+
 
 def transcribe_file_v2(
-    project_id: str,
     audio_file: str,
```
davidcavazos, line 27: Can we inline a specific value for `audio_file` into the code sample?

speech/snippets/transcribe_file_v2.py
```diff
@@ -56,10 +65,4 @@ def transcribe_file_v2(
...
 
 if __name__ == "__main__":
-    parser = argparse.ArgumentParser(
-        description=__doc__, formatter_class=argparse.RawDescriptionHelpFormatter
-    )
-    parser.add_argument("project_id", help="GCP Project ID")
-    parser.add_argument("audio_file", help="Audio file to stream")
-    args = parser.parse_args()
-    transcribe_file_v2(args.project_id, args.audio_file)
+    transcribe_file_v2("resources/audio.wav")
```
davidcavazos, line 68: We should remove CLIs unless they're strictly necessary for the code sample within the region tags to work.

speech/snippets/transcribe_gcs_v2.py
```diff
@@ -12,26 +12,26 @@
...
 
+# TODO (Developer): Update the PROJECT_ID to the value of your project
+# PROJECT_ID = "your-project-id"
+PROJECT_ID = os.getenv("GOOGLE_CLOUD_PROJECT")
+
 
 def transcribe_gcs_v2(
-    project_id: str,
-    gcs_uri: str,
+    audio_uri: str,
```
davidcavazos, line 27: Can we inline a specific value for `audio_uri` into the sample?

speech/snippets/transcribe_gcs_v2.py
```diff
@@ -61,10 +61,4 @@ def transcribe_gcs_v2(
...
 
 if __name__ == "__main__":
-    parser = argparse.ArgumentParser(
-        description=__doc__, formatter_class=argparse.RawDescriptionHelpFormatter
-    )
-    parser.add_argument("project_id", help="GCP Project ID")
-    parser.add_argument("gcs_uri", help="URI to GCS file")
-    args = parser.parse_args()
-    transcribe_gcs_v2(args.project_id, args.gcs_uri)
+    transcribe_gcs_v2("gs://cloud-samples-data/speech/audio.flac")
```
davidcavazos, line 64: Can we remove the CLI interface?

speech/snippets/transcribe_model_selection.py
```diff
@@ -14,38 +14,39 @@
...
-import argparse
-
+# [START speech_transcribe_model_selection]
 from google.cloud import speech
 
 
-# [START speech_transcribe_model_selection]
 def transcribe_model_selection(
-    speech_file: str,
+    audio_file: str,
```
davidcavazos, line 24: Can we inline specific values for `audio_file` and `model` into the sample?

speech/snippets/transcribe_model_selection.py
```diff
@@ -63,17 +64,27 @@ def transcribe_model_selection(
 
 
 # [START speech_transcribe_model_selection_gcs]
+
+
 def transcribe_model_selection_gcs(
-    gcs_uri: str,
+    audio_uri: str,
```
davidcavazos, line 70: Can we inline specific values for `audio_uri` and `model` into the sample?

speech/snippets/transcribe_model_selection.py
```diff
@@ -100,20 +111,11 @@ def transcribe_model_selection_gcs(
 
 
 if __name__ == "__main__":
```
davidcavazos, line 113: Can we get rid of this entire CLI interface?

speech/snippets/transcribe_model_selection_v2.py
```diff
@@ -12,37 +12,47 @@
...
 
+# TODO (Developer): Update the PROJECT_ID to the value of your project
+# PROJECT_ID = "your-project-id"
+PROJECT_ID = os.getenv("GOOGLE_CLOUD_PROJECT")
+
 
 def transcribe_model_selection_v2(
-    project_id: str,
-    model: str,
     audio_file: str,
```
davidcavazos, line 27: Can we inline specific values for `audio_file` and `model` into the sample?

speech/snippets/transcribe_model_selection_v2.py
```diff
@@ -58,11 +68,4 @@ def transcribe_model_selection_v2(
...
 if __name__ == "__main__":
-    parser = argparse.ArgumentParser(
-        description=__doc__, formatter_class=argparse.RawDescriptionHelpFormatter
-    )
-    parser.add_argument("project_id", help="GCP Project ID")
-    parser.add_argument("model", help="Speech-to-Text model with which to transcribe")
-    parser.add_argument("audio_file", help="Audio file to stream")
-    args = parser.parse_args()
-    transcribe_model_selection_v2(args.project_id, args.model, args.audio_file)
+    transcribe_model_selection_v2("resources/audio.wav", "short")
```
davidcavazos, line 71: Can we remove this CLI interface?

speech/snippets/transcribe_feature_in_recognizer.py
msampathkumar: Pinging @davidcavazos in pvt.

Thoughtseize1: I replaced the region tags and left hints for our users. It looks much better🙌

speech/snippets/transcribe_feature_in_recognizer.py
Thoughtseize1: Yes, that will work. As I mentioned earlier, I added a `try/except` block that creates a `recognizer` if it doesn't exist and uses the existing one if it does. Additionally, I updated the `region tag`, so developers will only need to replace the variable values with their own.

speech/snippets/transcribe_feature_in_recognizer.py
Thoughtseize1: This issue solved by replacing the `region_tag` 🙌 Users will be able to see the complete example in the full version of Sample.

speech/snippets/transcribe_feature_in_recognizer_test.py
Thoughtseize1: Thanks! I removed the check for stdout and kept the double function call to verify that the function can work with an existing recognizer.

speech/snippets/transcribe_file_v2.py
Thoughtseize1: Yes! A good solution would be to replace the region_tag. 🙌
PTAL

Thoughtseize1:
@davidcavazos @msampathkumar Please approve my PR or provide feedback🙂

m-strzelczyk (changes requested):
In some region tags, you include the neccessary imports, while in some you don't. Please include the imports in all the region tags, so the users don't have to view the full files to have a working code.

speech/snippets/transcribe_model_selection.py
```diff
@@ -14,38 +14,32 @@
...
-Example usage:
-    python transcribe_model_selection.py \
-        resources/Google_Gnome.wav --model video
-    python transcribe_model_selection.py \
-        gs://cloud-samples-tests/speech/Google_Gnome.wav --model video
 """
-
-import argparse
-
 from google.cloud import speech
```
msampathkumar, line 18: please move the imports to be inside region tags

speech/snippets/transcribe_file_v2.py
```diff
@@ -12,24 +12,30 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import os
```
msampathkumar, line 15: Except for `os` module, please move other imports to be inside region tags

speech/snippets/transcribe_feature_in_recognizer.py
```diff
@@ -11,51 +11,72 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import os
```
msampathkumar, line 14: Except for `os` module, please move other imports to be inside region tags

m-strzelczyk (approved):

PR #12160, opened 2024-08-09
Skip initial key version creation in create_key_for_import.py

tdbhacks:
The samples in this section skip the initial key version creation, except for the python one: https://cloud.google.com/kms/docs/importing-a-key#create_targets

## Description

## Checklist

### DIFF
--- a/kms/snippets/create_key_for_import.py
+++ b/kms/snippets/create_key_for_import.py
@@ -45,6 +45,7 @@ def create_key_for_import(
             "algorithm": algorithm,
             "protection_level": protection_level,
         },
+        "skip_initial_version_creation": True,
     }
 
     # Build the parent key ring name.


iennae:
hey @tdbhacks thanks for filing this PR. I do see that for the other languages for this sample it does do the skip for this specific sample. It seems like there is something up with the implementation of this language API and might be something missing in the sample based on the test failing. (it's only succeeding for the other runtimes because it's not actually running I think. 

the error is 

"  File "/workspace/kms/snippets/.nox/py-3-12/lib/python3.12/site-packages/proto/marshal/rules/message.py", line 36, in to_proto
    return self._descriptor(**value)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: Protocol message CryptoKey has no "skip_initial_version_creation" field.

When I looked at https://googleapis.dev/python/cloudkms/latest/_modules/google/cloud/kms_v1/types/resources.html (based on the comment) it wasn't super helpful, so I started digging more. 

I found https://cloud.google.com/python/docs/reference/cloudkms/2.11.0/google.cloud.kms_v1.types.CreateCryptoKeyRequest which does provide the field name that you mentioned. But, we aren't calling CreateCryptoKeyRequest directly in this sample. 

I checked the generated samples for the API to see whether there was any good reference material there but I didn't see anything.  I do think this should be consistent across languages, I'm just not entirely sure how to go about it. I'm hoping this gives you enough information about the problem to see how to fix this.

iennae:
I think I figured it out see #12161

tdbhacks:
> I think I figured it out see #12161

Ah indeed, I added it to the wrong block by mistake! Thanks for spinning up that PR so quickly, much appreciated :)

PR #12544, opened 2024-08-30
Update poll.js to include XML Header in fetch request

riathakkar:
## Description

Fixes b/362304568

Note: Before submitting a pull request, please open an issue for discussion if you are not associated with Google.

## Checklist
- [ ] I have followed [Sample Guidelines from AUTHORING_GUIDE.MD](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/main/AUTHORING_GUIDE.md)
- [ ] README is updated to include [all relevant information](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/main/AUTHORING_GUIDE.md#readme-file)
- [ ] **Tests** pass:   `nox -s py-3.9` (see [Test Environment Setup](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/main/AUTHORING_GUIDE.md#test-environment-setup))
- [ ] **Lint** pass:   `nox -s lint` (see [Test Environment Setup](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/main/AUTHORING_GUIDE.md#test-environment-setup))
- [ ] These samples need a new **API enabled** in testing projects to pass (let us know which ones)
- [ ] These samples need a new/updated **env vars** in testing projects set to pass (let us know which ones)
- [ ] This sample adds a new sample directory, and I updated the [CODEOWNERS file](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/main/.github/CODEOWNERS) with the codeowners for this sample
- [ ] This sample adds a new **Product API**, and I updated the [Blunderbuss issue/PR auto-assigner](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/main/.github/blunderbuss.yml) with the codeowners for this sample
- [ ] Please **merge** this PR for me once it is approved

PR #12557, opened 2024-09-03
fix(gae): resolve linting issues in main_test.py

iennae:
## Description
unblocks merging of #12544

### DIFF
--- a/appengine/standard/iap/main_test.py
+++ b/appengine/standard/iap/main_test.py
@@ -12,9 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import main
 import webtest
 
+import main
 
 def test_index(testbed, login):
     app = webtest.TestApp(main.app)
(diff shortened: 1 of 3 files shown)


briandorsey (approved):
LGTM

glasnt:
Tests appear to be failing because there's no `pytest` defined for Python 3+

Suggest removing the `python_version` for pytest in requirements-test.txt https://github.com/GoogleCloudPlatform/python-docs-samples/blob/2d6db5de16f38e09a0bce9674bebf705a6c2a364/appengine/standard/iap/requirements-test.txt#L2, which should pass the tests

(why were these tests green earlier? they appear to be kokoro skipped last time this sample was run)

iennae:
thank you @glasnt ! I was wondering what was up

glasnt:
This will be the sort of thing you'll need to update to:

https://github.com/GoogleCloudPlatform/python-docs-samples/pull/12561/files#diff-72885259355e2caa839d2ffcc71fad928d4af9bc5dae781a408a468f0068d885R3

davidcavazos:
I reproduced this locally. I think `six` is not an indirect dependency anymore. I think we can simply add it to the dependencies.

PR #12592, opened 2024-09-13
Update generate_self_signed_jwt.py

riathakkar:
This is a bug fix in response to a listener report. The datetime object needs to return the current time in epoch

## Description

Fixes b/366503744

## Checklist
- [X ] I have followed [Sample Guidelines from AUTHORING_GUIDE.MD](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/main/AUTHORING_GUIDE.md)
- [X] README is updated to include [all relevant information](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/main/AUTHORING_GUIDE.md#readme-file)
- [X ] Please **merge** this PR for me once it is approved

### DIFF
--- a/iap/generate_self_signed_jwt.py
+++ b/iap/generate_self_signed_jwt.py
@@ -12,8 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import datetime
 import json
+import time
 
 import google.auth
 from google.cloud import iam_credentials_v1
@@ -35,15 +35,15 @@ def generate_jwt_payload(service_account_email: str, resource_url: str) -> str:
       Access the application with the JWT in the Authorization Header.
       curl --verbose --header 'Authorization: Bearer SIGNED_JWT' URL
     """
-    iat = datetime.datetime.now(tz=datetime.timezone.utc).timestamp()
-    exp = iat + 3600
+    now = int(time.time())
+
     return json.dumps(
         {
             "iss": service_account_email,
             "sub": service_account_email,
             "aud": resource_url,
-            "iat": iat,
-            "exp": exp,
+            "iat": now,
+            "exp": now + 3600,
         }
     )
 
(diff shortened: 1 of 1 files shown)


iennae (changes requested):
I dug into this since it to better understand what is happening here. 

Based on https://cloud.google.com/iam/docs/reference/credentials/rest/v1/projects.serviceAccounts/signJwt it seems like the required context is 

string

Required. The JWT payload to sign. Must be a serialized JSON object that contains a JWT Claims Set. For example: {"sub": "user@example.com", "iat": 313435}

If the JWT Claims Set contains an expiration time (exp) claim, it must be an integer timestamp that is not in the past and no more than 12 hours in the future.

So then I looked at what would a valid claims set look like. 

IAT - "This is a timestamp formatted as whole seconds since January 01, 1970, which is a standard UNIX representation of time. The timestamp must be an integer (no decimals) and it must be in UTC."

I checked to see whether your code would work: 

>>> iat=datetime.datetime.now(tz=datetime.timezone.utc).timestamp
>>> exp = iat + 3600
  File "<stdin>", line 1, in <module>
TypeError: unsupported operand type(s) for +: 'builtin_function_or_method' and 'int'
 
 
 I suspect what is needed is something like 
 
 now_utc = datetime.datetime.now(tz=datetime.timezone.utc)
 iat = int(now_utc.timestamp())
 
 Which I checked : 
 >>> now_utc = datetime.datetime.now(tz=datetime.timezone.utc)
>>> iat = int(now_utc.timestamp())
>>> iat
1726289956 

it's helpful to share a reference to the bug with a b/ which might provide additional context I don't have here. 

Finally, I think a test might be missing here if we're not catching this

riathakkar:
Thanks for the catch. Did something silly when copying this over. I think tests would be helpful, but it would be a non trivial effort to do so. Id love to pick it up as another task and get the code sample fixed now for our customers.

iennae:
@riathakkar can you update the code so that it works using time or int()?

iennae (approved):

PR #12601, opened 2024-09-17
fix: (datastore) add imports to snippets

daniel-sanche:
Some of the datastore snippets are missing required imports. This change pulls the necessary imports into each sample in snippets.py

### DIFF
--- a/datastore/cloud-client/snippets.py
+++ b/datastore/cloud-client/snippets.py
@@ -684,6 +684,8 @@ def inequality_sort(client):
 
 
 def inequality_sort_invalid_not_same(client):
+    import google.cloud.exceptions
+
     try:
         # [START datastore_inequality_sort_invalid_not_same]
         query = client.query(kind="Task")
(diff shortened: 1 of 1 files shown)


datastore/cloud-client/snippets.py
```diff
@@ -887,6 +913,9 @@ def property_by_kind_run_query(client):
 
 def regional_endpoint():
     # [START datastore_regional_endpoints]
+    from google.cloud import datastore
```
gkevinzheng, line 916: Why is this the only sample that has this import?

daniel-sanche: All the other ones also include [the preamble](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/ec9128da73d7832cb970cd0442562a54be1a1ad1/datastore/cloud-client/snippets.py#L17), which imports datastore and creates the client

gkevinzheng (approved):
LGTM

iennae (approved):
recognizing that this is a pre-existing issue where there are a bunch of samples in one file rather than one sample per file, (hopefully this can get refactored and fixed in the future) but also wanting to help unblock getting this fixed so that samples are copy pastable where they are showing up in c.g.c approving this.

PR #12598, opened 2024-09-17
feat(GenAI): Gemma2 samples for Model Garden deployments to Vertex AI endpoints

rsamborski:
## Description

Fixes b/365590892

Gemma2 samples for Model Garden deployments to Vertex AI endpoints
Tests are mocked for now until we make a decision if we want to have these models permanently deployed for testing purposes.

## Checklist
- [X] I have followed [Sample Guidelines from AUTHORING_GUIDE.MD](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/main/AUTHORING_GUIDE.md)
- [X] README is updated to include [all relevant information](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/main/AUTHORING_GUIDE.md#readme-file)
- [X] **Tests** pass:   `nox -s py-3.9` (see [Test Environment Setup](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/main/AUTHORING_GUIDE.md#test-environment-setup))
- [X] **Lint** pass:   `nox -s lint` (see [Test Environment Setup](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/main/AUTHORING_GUIDE.md#test-environment-setup))
- [X] This sample adds a new sample directory, and I updated the [CODEOWNERS file](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/main/.github/CODEOWNERS) with the codeowners for this sample
- [X] Please **merge** this PR for me once it is approved

PR #12617, opened 2024-09-20
feat: demonstrating testing with mocks using actual interfaces

telpirion:
## Description

For some resource-intensive APIs, in this case `Endpoint.predict()` which requires a deployed endpoint, use a `Mock` or `MagicMock` to ensure that the client library interface is observed.

### DIFF
--- a/generative_ai/gemma2_predict_gpu.py
+++ b/generative_ai/gemma2_predict_gpu.py
@@ -1,0 +1,27 @@
+# Copyright 2024 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


generative_ai/gemma2_predict_gpu.py
```diff
@@ -0,0 +1,27 @@
...
+def endpoint_predict_sample(
+    project: str, location: str, instances: list, endpoint: str
+):
+    aiplatform.init(project=project, location=location)
+
+    endpoint = aiplatform.Endpoint(endpoint)
+
+    prediction = endpoint.predict(instances=instances)
+    print(prediction)
+    return prediction
```
glasnt, line 27: Should there be a region tag for this block? Or is it already being captured in another sample

rsamborski: This is a PoC PR showing a recommended way of handling mocks - this is the [actual sample PR](https://github.com/GoogleCloudPlatform/python-docs-samples/pull/12598)

telpirion:
Oops! I intended to mark this as a "draft." @glasnt -- I'll follow up with you over chat about this PR; my apologies that I didn't do it sooner.

telpirion:
Converting this to draft; will follow up directly.

telpirion:
PR no longer needed. Closing.

PR #12671, opened 2024-10-03
feat(securitycenter): add sample for creating client with regional endpoint

hegemonic:
## Description

This sample shows how to create a Security Command Center client that uses a regional endpoint, along with a client that uses the default endpoint. That's because Security Command Center's regional endpoints support only a limited subset of Security Command Center resources that are subject to data residency controls: https://cloud.google.com/security-command-center/docs/data-residency-support#scc-regional-urls-rest

As a result, it's likely that customers will need to create two clients: one for their regional endpoint, to call the methods it supports, and another for the default endpoint, to call all other methods.

This sample will be used here: https://cloud.google.com/security-command-center/docs/data-residency-support#regional-urls

## Checklist


hegemonic:
I think the lint failure is caused by a file that I'm not changing in this PR. Do I also need to fix the unrelated linter errors before you merge this PR? If so, should I fix the linter errors in this PR, or in a separate PR?

Also, I'm not sure why the tests are failing in Python 3.8 and 3.12, but not other versions. Any ideas?

hegemonic:
@Sita04 Friendly ping.

hegemonic:
@Sita04 Another friendly ping.

securitycenter/snippets_v2/snippets_create_client_v2.py
```diff
@@ -0,0 +1,40 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Examples of working with clients for Security Command Center."""
+from typing import Dict
+
+
+# [START securitycenter_set_client_endpoint_v2]
```
Sita04, line 19: Please move the region tag to line 15 to include the import statement as well.

Sita04 (approved):
Thanks for your patience. Sample looks good to me. One nit.

Regarding the lint and build error questions, I think python-reviewers would know more info. On the outset, it is unrelated to this sample, but looks like an env variable is missing.

hegemonic:
> Regarding the lint and build error questions, I think python-reviewers would know more info. On the outset, it is unrelated to this sample, and looks like an env variable is missing.

Thanks, @Sita04! I addressed the nit; please merge if everything looks okay.

I'm just here to add this one sample, so I'll let python-reviewers sort out the lint errors and the missing environment variable.

Sita04:
@hegemonic The lint fails on the samples in this PR. PTAL

hegemonic:
> * nox -s py-3.9

hegemonic:
> @hegemonic The lint fails on the samples in this PR. PTAL

Sorry about that—should be fixed now.

I also simplified the sample so that it only creates a regional client. I did that because when I sent a PR for the Java equivalent of this sample, the Java reviewer wouldn't accept a sample that created both regional and non-regional clients. I still think it would be useful to show the creation of both clients, but I think it's more important to make the samples do approximately the same thing in each language.

PR #12706, opened 2024-10-17
docs(samples): VAIS Answer - Add comment describing that all options are included by default

holtskinner:
## Description

Fixes b/369766709

### DIFF
--- a/discoveryengine/answer_query_sample.py
+++ b/discoveryengine/answer_query_sample.py
@@ -20,10 +20,6 @@ from google.api_core.client_options import ClientOptions
 from google.cloud import discoveryengine_v1 as discoveryengine
 
 # TODO(developer): Uncomment these variables before running the sample.
-# The `answer_generation_spec` below includes all available answer phase options.
-# Modify or remove any options that are not relevant to your use case to optimize
-# the answer generation process.
-
 # project_id = "YOUR_PROJECT_ID"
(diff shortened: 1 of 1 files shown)


discoveryengine/answer_query_sample.py
```diff
@@ -20,6 +20,8 @@
 from google.cloud import discoveryengine_v1 as discoveryengine
 
 # TODO(developer): Uncomment these variables before running the sample.
+# NOTE: This sample shows all available answer phase options. Remove any options that you don't plan to use.
```
iennae, line 23: nit: would recommend something a little more clear and specific for this comment. Suggestion: 

# The `answer_generation_spec` below includes all available answer phase options. 
# Modify or remove any options that are not relevant to your use case to optimize 
# the answer generation process.

iennae (approved):
I think the comment could be a little more clear and I gave an alternative, but it's up to you.

msampathkumar (approved):
Only docs updates. No code changes.

discoveryengine/answer_query_sample.py
```diff
@@ -20,6 +20,10 @@
 from google.cloud import discoveryengine_v1 as discoveryengine
 
 # TODO(developer): Uncomment these variables before running the sample.
+# The `answer_generation_spec` below includes all available answer phase options.
```
msampathkumar, line 23: Suggestion: Can you move the #TODO comment to be above project_id, location variables definition ?

PR #12711, opened 2024-10-19
chore: patch imported log entries to handle dict to protobuf parsing

minherz:
## Description

Add a patch to address a problem with the Google logging library that fails to parse a dictionary to protobuf object when the dictionary has a key (`serviceData`) matching the probuf's deprecated field (`service_data`).
See https://github.com/googleapis/python-logging/issues/945 for more details about the problem.
This patch has to be removed once the issue with the client library is resolved.

Patching code is added to the `_patch_entry()` method since it already "patches" fields in the log dictionionary.


iennae:
I don't think this is completely deprecated yet. https://cloud.google.com/logging/docs/audit/understanding-audit-logs#servicedata-services There are still services listed that do use "the older format". 

I'm wondering if this is trying to back patch older APIs that there are larger issues. Is there a newer version of a library that might need to be used for BigQuery/Audit Logs? I couldn't find a lot of documentation on this subject but I did find https://github.com/GoogleCloudPlatform/bigquery-utils/blob/master/views/audit/README.md so it feels like something might be missing to help make this more visible.

PR #12768, opened 2024-11-13
Update Maintenance DAG 

apilaskowski:
Removing DagRuns sometimes failed in case of environments where a lot of Dags were triggered by schedule, and had the same execution_date.

Part of changes were required by linter.

Since we are using `keep_last` only in case of DagRuns I propose to remove this complicated logic that was used previously, and simplify it like in this PR.

If there exist a good reason to maintain previous approach we can easily restore it while still fixing aforementioned issue.

## Description

Fixes #<ISSUE-NUMBER>

## Checklist


rafalbiegacz (approved):

composer/workflows/airflow_db_cleanup.py
```diff
@@ -319,37 +315,27 @@ def build_query(
...
     query = session.query(airflow_db_model)
 
     logging.info("INITIAL QUERY : " + str(query))
 
-    if not keep_last:
-        query = query.filter(
-            age_check_column <= max_date,
-        )
+    if airflow_db_model == DagRun:
+        newest_dagrun = session.query(airflow_db_model) \
```
kosteev, line 325: Can you, please, avoid using "\" to wrap statement (here and below).

From pep8 guidelines:
"""
The preferred way of wrapping long lines is by using Python’s implied line continuation inside parentheses, brackets and braces. Long lines can be broken over multiple lines by wrapping expressions in parentheses. These should be used in preference to using a backslash for line continuation.
"""

composer/workflows/airflow_db_cleanup.py
```diff
@@ -319,37 +315,27 @@ def build_query(
     airflow_db_model,
     age_check_column,
     max_date,
-    keep_last,
-    keep_last_filters=None,
-    keep_last_group_by=None,
+    dag_id=None
```
kosteev, line 318: It took me a while to understand that `dag_id` is passed/used only for DagRun model. It is worth to have docstring for this method with some explanation

composer/workflows/airflow_db_cleanup.py
```diff
@@ -319,37 +315,27 @@ def build_query(
...
     logging.info("INITIAL QUERY : " + str(query))
 
-    if not keep_last:
-        query = query.filter(
-            age_check_column <= max_date,
-        )
+    if airflow_db_model == DagRun:
+        newest_dagrun = session.query(airflow_db_model) \
+            .filter(airflow_db_model.dag_id == dag_id) \
+            .order_by(airflow_db_model.execution_date).first()
```
kosteev, line 327: Should it be actually .last()?

composer/workflows/airflow_db_cleanup.py
```diff
@@ -410,13 +396,11 @@ def cleanup_function(**context):
...
-                age_check_column,
-                max_date,
-                keep_last,
-                keep_last_filters,
-                keep_last_group_by,
+                session=session,
+                airflow_db_model=airflow_db_model,
+                age_check_column=age_check_column,
+                max_date=max_date,
+                dag_id=None,
```
kosteev, line 403: IMHO passing here `dag_id` parameter seems a bit unreadable.
We anyway have it None as default in signature of the method, and since this parameter shouldn't be used for other tables than DagRun then it can be misleading to pass it?

composer/workflows/airflow_db_cleanup.py
```diff
@@ -429,17 +413,14 @@ def cleanup_function(**context):
...
-                    keep_last,
-                    keep_last_filters,
-                    keep_last_group_by,
+                    session=session,
+                    airflow_db_model=airflow_db_model,
+                    age_check_column=age_check_column,
+                    max_date=max_date,
+                    dag_id=dag_id,
                 )
-                query = query.filter(airflow_db_model.dag_id == dag)
```
kosteev, line 442: I am confused.
If we do not have this filter by dag_id here, then for all tables != DagRun we will not have filter on dag_id?
Should we have filter inside build_query method?

composer/workflows/airflow_db_cleanup.py
apilaskowski: Makes sense. Updating.

composer/workflows/airflow_db_cleanup.py
apilaskowski: You meant backslashes, right? Probably formatting hidden that. I agree with you, thank you for noticing that.

composer/workflows/airflow_db_cleanup.py
apilaskowski: Yes, my intention was to have this inside `build_query` block, but I missed that later.
This way it still works, but it makes sense to filter by dag. 
PTAL (when dag_id is present we filter by it).

composer/workflows/airflow_db_cleanup.py
apilaskowski: Agree.

composer/workflows/airflow_db_cleanup.py
apilaskowski: Indeed I had 'desc' function here, but it got lost at some point when refactoring. Thank you!

composer/workflows/airflow_db_cleanup.py
apilaskowski: PTAL on the docstring

apilaskowski: Btw it is not used only for DagRuns, fixed! Thanks!

kosteev (approved):
Did you run this DAG to check it works as expected?
I wonder if we have any unit tests or integration tests for this?

apilaskowski:
> Did you run this DAG to check it works as expected? I wonder if we have any unit tests or integration tests for this?

I executed this DAG on my environment where I had backfilled with a lot of data.
Before you mentioned that I missed `desc` I missed that in logs.

This DAG didn't have any additional tests, and we are working on other solution on Composer site so it was decided that we are not going to invest more time than needed on this one.

PR #12776, opened 2024-11-19
feat(texttospeech): Create sample for Synthesize speech multiple speakers

holtskinner:
- https://cloud.google.com/text-to-speech/docs/create-dialogue-with-multispeakers

### DIFF
--- a/texttospeech/snippets/synthesize_speech_multiple_speakers.py
+++ b/texttospeech/snippets/synthesize_speech_multiple_speakers.py
@@ -1,0 +1,84 @@
+#!/usr/bin/env python
+# Copyright 2014 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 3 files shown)


texttospeech/snippets/synthesize_speech_multiple_speakers.py
```diff
@@ -0,0 +1,84 @@
...
+    Make sure to be working in a virtual environment.
+    """
+    from google.cloud import texttospeech_v1beta1 as texttospeech
+
+    # Instantiates a client
+    client = texttospeech.TextToSpeechClient()
+
+    multi_speaker_markup = texttospeech.MultiSpeakerMarkup(
+        turns=[
+            texttospeech.MultiSpeakerMarkup.Turn(
```
msampathkumar, line 36: Just a thought! 

Why not do `from google.cloud import texttospeech_v1beta1.MultiSpeakerMarkup.Turn as Turn`
texttospeech. ?

holtskinner: Because our official styles say to not import individual classes. (Even though I disagree for samples)

PR #12988, opened 2024-12-11
docs(generative_ai): Add Vertex AI Express Mode Sample

holtskinner:
- https://cloud.google.com/vertex-ai/generative-ai/docs/start/express-mode/overview
- NOTE: This cannot be tested in our current infrastructure because Express Mode requires a new project set up with a Gmail account that has never used Google Cloud before


generative_ai/express_mode/api_key_example.py
msampathkumar, line 1: Hi, this does not look any different from the other samples.

As I understand from https://cloud.google.com/vertex-ai/generative-ai/docs/start/express-mode/overview#different, the major differences is the use of an API keys. If we create an API keys & store in the CICD env variables, will it be enough for creating a test-case?

holtskinner: The problem is creating the API Keys. They must be created with a gmail account that has never used Google Cloud before, and they will only work for 90 days.

generative_ai/express_mode/api_key_example.py
```diff
@@ -0,0 +1,34 @@
...
+    vertexai.init(api_key="YOUR_API_KEY")
+
+    model = GenerativeModel("gemini-1.5-flash")
+
+    response = model.generate_content("Explain bubble sort to me")
+
+    print(response.text)
+    # Example response:
+    # Bubble Sort is a simple sorting algorithm that repeatedly steps through the list
+
```
msampathkumar, line 32:
```suggestion
```

generative_ai/express_mode/api_key_example.py
```diff
@@ -0,0 +1,34 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+def generate_content() -> None:
+
```
msampathkumar, line 17:
```suggestion
```

msampathkumar (approved):
LGTM. Thank you for your work.

msampathkumar:
CICD error does not related to this PR. Reaching out to the admins.

```
nox -s py-3.13
nox > Error while collecting sessions.

```

PR #13049, opened 2025-01-10
feat: imagen 3 inpainting and outpainting samples

katiemn:
New samples for Imagen 3 editing


glasnt (approved):
Pending confirmation from product expert

PR #13053, opened 2025-01-14
feat(dialogflow-cx): Example implementation of Streaming Detect Intent with continuous microphone input and audio output

### LINKED ISSUE #13041: Streaming Detect Intent with continuous microphone input
## Is your feature request related to a problem? Please describe.

Many customers are seeking an example of how to use Dialogflow CX's streaming detect intent capability. However, the provided example for detect intent streaming uses a local audio file for simulation, which doesn't provide a clear path for integrating with actual streaming audio input.

A real use case with microphone streaming input would be helpful to customers.

## Describe the solution you'd like.

Provide an example using microphone streaming audio input to streaming_detect_intent, which communicates with DFCX conversational agents and returns an audio response for playback.

ffeldhaus:
## Description

Fixes #13041

## Checklist
- [X] I have followed [Sample Guidelines from AUTHORING_GUIDE.MD](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/main/AUTHORING_GUIDE.md)

README.rst is not updated as it seems to be auto generated.


glasnt:
Internal b/389902797

dialogflow-cx/streaming_detect_intent_infinite.py
```diff
@@ -0,0 +1,720 @@
...
+```bash
+python streaming_detect_intent_infinite.py --help
+```
+
+The script will start capturing audio from the default microphone, stream it
+to Dialogflow CX, and play back the audio responses. Press Ctrl+C to exit the
+program gracefully. Say "Hello" to trigger the Default Intent.
+"""
+
+# [START streaming_detect_intent_infinite]
```
glasnt, line 63: snippetbot is reporting errors here: the region tag needs to start with the product 
```suggestion
# [START dialogflow_streaming_detect_intent_infinite]
```

dialogflow-cx/streaming_detect_intent_infinite.py
```diff
@@ -0,0 +1,720 @@
...
+            args.model,
+            args.language_code,
+            args.dialogflow_timeout,
+            args.sample_rate,
+            args.voice,
+            args.single_utterance,
+        )
+    )
+
+# [END streaming_detect_intent_infinite]
```
glasnt, line 720:
```suggestion
# [END dialogflow_streaming_detect_intent_infinite]
```

dialogflow-cx/streaming_detect_intent_infinite.py
```diff
@@ -0,0 +1,720 @@
...
+    """Audio Input / Output"""
+
+    def __init__(
+        self,
+        rate: int,
+        chunk_size: int,
+        audio_encoding: str = "LINEAR16",
+        output_audio_encoding: str = "LINEAR16",
+    ) -> None:
+        self._rate: int = rate
```
davidcavazos, line 112: The purpose of a code sample is to provide a simple example on how to use the APIs or show a common pattern, not to provide a fully customizable tool. The amount of options here makes the sample look a lot more intimidating than what it actually is.

Can we inline as many of these options into where they're used? That makes the code sample easier to read, since you can directly see all the values rather than having to scroll up to see which configurations are important and which can be left to None or some other empty value.

Let's just leave here the variables needed to keep state on the object. Any configuration should be inlined.

dialogflow-cx/streaming_detect_intent_infinite.py
```diff
@@ -0,0 +1,720 @@
...
+
+    audioIO: AudioIO = AudioIO(sample_rate, chunk_size)
+    dialogflow_streaming: DialogflowCXStreaming = DialogflowCXStreaming(
+        agent_name,
+        language_code,
+        model,
+        dialogflow_timeout,
+        debug,
+        sample_rate,
+        "LINEAR16",
```
davidcavazos, line 629: Since these LINEAR16 are hardcoded and not configurable, can we remove them from `DialogCXStreaming` altogether and inline them directly where they're used?

dialogflow-cx/streaming_detect_intent_infinite.py
```diff
@@ -0,0 +1,720 @@
...
+class DialogflowCXStreaming:
+    """Manages the interaction with the Dialogflow CX Streaming API."""
+
+    def __init__(
+        self,
+        agent_name: str,
+        language_code: str,
+        model: str,
+        dialogflow_timeout: float,
+        debug: bool = False,
```
davidcavazos, line 299: Same here, can we only keep the required arguments? Please inline any "constant" configuration to reduce complexity.

dialogflow-cx/streaming_detect_intent_infinite.py
```diff
@@ -0,0 +1,720 @@
...
+            client_options: ClientOptions = ClientOptions(
+                api_endpoint=f'{location}-dialogflow.googleapis.com',
+                quota_project_id=project,
+            )
+        else:
+            client_options = ClientOptions(quota_project_id=project)
+        self.client: dialogflowcx_v3.SessionsAsyncClient = (
+            dialogflowcx_v3.SessionsAsyncClient(client_options=client_options)
+        )
+        self.agent_name: str = agent_name
```
davidcavazos, line 321: Same here, please only keep whatever is necessary to keep state. Configurations should be inlined.

davidcavazos (changes requested):
This PR also needs tests. Since it's an "infinite" streaming CLI app, there's no need to test it end-to-end, but unit tests around the most important functions would be nice.

It's a little tricky since most functions are methods depending on the internal state of an object, so each test might need to create its own object.

I'm okay with more minimal testing around `AudioIO` since that's what streams from the microphone and can get tricky. However the Dialogflow code should be tested, it seems it could simply use a bytes generator with some fixed data, or maybe it could be read from a file?

ffeldhaus:
> This PR also needs tests. Since it's an "infinite" streaming CLI app, there's no need to test it end-to-end, but unit tests around the most important functions would be nice.
> 
> It's a little tricky since most functions are methods depending on the internal state of an object, so each test might need to create its own object.
> 
> I'm okay with more minimal testing around `AudioIO` since that's what streams from the microphone and can get tricky. However the Dialogflow code should be tested, it seems it could simply use a bytes generator with some fixed data, or maybe it could be read from a file?

I added a simple test similar to the existing test in [streaming_detect_intent_partial_response_test.py](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/main/dialogflow-cx/streaming_detect_intent_partial_response_test.py) reusing the some hello.wav resource file as input and the same existing Dialogflow Agent.

I reused and extended the MockPyAudio from [transcribe_streaming_infinite_v2_test.py](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/main/speech/microphone/transcribe_streaming_infinite.py).

Please review again, all comments so far should be addressed.

davidcavazos (approved):

PR #13066, opened 2025-01-16
ci: make Python 3.9 minimum required build

jackwotherspoon:
Python 3.8 is EOL: https://devguide.python.org/versions/

Making Python 3.9 the minimum required build to pass.


davidcavazos (approved):

iennae (approved):
there are probably other areas that need to get fixed by this change because I think there are configurations that say ignore Python 3.9 testing.

parthea (approved):

jackwotherspoon:
> there are probably other areas that need to get fixed by this change because I think there are configurations that say ignore Python 3.9 testing.

@iennae Do you have any examples? I can update them if would you like as part of this PR if need be.

davidcavazos:
@iennae that's a great point. I guess we could grep over noxfile_config files to look for ignored versions including 3.9, and then update them accordingly

jackwotherspoon:
Also tried updating the README badges to reflect current versions but seems 3.12 and 3.13 badges maybe don't exist...

image

jackwotherspoon:
@davidcavazos @iennae I went through all the `noxfile_config` files and updated all files who were running Python 3.8 but ignoring Python 3.9 and swapped them to ignore Python 3.8 and run Python 3.9 now 

Let me know if that approach seems okay to you  👍

ShuoyuM (approved):

davidcavazos (approved):

AUTHORING_GUIDE.md
```diff
@@ -923,7 +923,7 @@ Add the new environment variables to the `envs` dictionary.
 ```py
 TEST_CONFIG_OVERRIDE = {
     # You can opt out from the test for specific Python versions.
-    "ignored_versions": ["2.7", "3.7", "3.9", "3.10", "3.11"],
+    "ignored_versions": ["2.7", "3.8", "3.10", "3.11", "3.12"],
```
iennae, line 926: did this need to match 3.13 from above?

jackwotherspoon: Don't believe so, we want samples to support all those versions but we purely test lowest and highest supported version by default. Does that make sense?

.github/sync-repo-settings.yaml
```diff
@@ -44,7 +44,7 @@ branchProtectionRules:
   requiredStatusCheckContexts:
     - "Kokoro CI - Lint"
     - "Kokoro CI - Python 2.7 (App Engine Standard Only)"
```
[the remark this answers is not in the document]
jackwotherspoon: I'll do that as a follow-up PR, want to make this one purely Python 3.9 --> Python 3.9

PR #13081, opened 2025-01-21
feat(securitycenter): Add Resource SCC Org Mgt API ETD Custom Modules (GetEff, ListEff, ListDesc, Validate)

vijaykanthm:
## Description

Fixes [b/347347769](http://b/347347769), [b/347347958](http://b/347347958), [b/347348171](http://b/347348171), [b/347348172](http://b/347348172)
This PR adds SCC Org Managament API ETD Custom Module Samples to GetEff, ListEff, ListDesc & Validate.

## Checklist


anissa-abu (approved):

davidcavazos (changes requested):
The code samples themselves look good, however I think the tests could be simplified. Is there a reason that tests get a random module? This seems like it could cause flaky non-deterministic tests

From what I see, it seems like multiple modules are created, then tests run, and then there's a cleanup step that deletes them. This could be done with "yield" fixtures to allow for cleanup. https://docs.pytest.org/en/6.2.x/fixture.html#yield-fixtures-recommended

Instead of getting a random module, I would create a fixture around a module, and pass that into each test. If the tests mutate the modules or there's any risk of race conditions, I would create a separate module fixture for each test that needs its own.

If you do end up needing multiple modules, there can be a helper function to create and destroy them. And then each fixture can just do "create, yield, destroy".

vijaykanthm:
> The code samples themselves look good, however I think the tests could be simplified. Is there a reason that tests get a random module? This seems like it could cause flaky non-deterministic tests
> 
> From what I see, it seems like multiple modules are created, then tests run, and then there's a cleanup step that deletes them. This could be done with "yield" fixtures to allow for cleanup. https://docs.pytest.org/en/6.2.x/fixture.html#yield-fixtures-recommended
> 
> Instead of getting a random module, I would create a fixture around a module, and pass that into each test. If the tests mutate the modules or there's any risk of race conditions, I would create a separate module fixture for each test that needs its own.
> 
> If you do end up needing multiple modules, there can be a helper function to create and destroy them. And then each fixture can just do "create, yield, destroy".

Hi David, thank you for the feedback and the helpful suggestion regarding fixtures. I initially tried creating a separate module for each test method, but that approach caused race conditions and ran into limitations with multiple write/update operations on an organization.

To address this, I opted to create a small number of modules (2-3) at the start of the tests, collect them into a list, and reuse them by assigning a random module to each test. This approach helps avoid race conditions while staying within the constraints.

For cleanup, I’m currently using the `addfinalizer` approach, which aligns with the second method outlined in the pytest documentation you shared, ensuring that all modules are deleted at the end of the tests.

glasnt:
With #13004 merged  this PR will need to be rebased from main, and edited to only include the new files and any adjustments to dependencies

davidcavazos (approved):

PR #13109, opened 2025-01-29
controlled generation(genai): Add new code samples

msampathkumar:
## Description

Adding new code using GenAI SDK

## Checklist


genai/controlled_generation/genai_ctrlgen_with_class_schema.py
```diff
@@ -0,0 +1,62 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+def generate_content() -> str:
+    # [START genai_ctrlgen_with_class_schema]
```
Valeriy-Burlaka, line 17: Should we use the `googlegenaisdk_` prefix in region tags?

msampathkumar: Yes. You are right.

genai/controlled_generation/ctrlgen_with_class_schema.py
```diff
@@ -0,0 +1,62 @@
...
+        ingredients: list[str]
+
+    client = genai.Client()
+    response = client.models.generate_content(
+        model="gemini-2.0-flash-001",
+        contents="List a few popular cookie recipes.",
+        config={
+            "response_mime_type": "application/json",
+            "response_schema": list[Recipe],
+        },
```
holtskinner, line 32: Can you use the `GenerateContentConfig` class?

genai/controlled_generation/ctrlgen_with_class_schema.py
```diff
@@ -0,0 +1,62 @@
...
+    #       "1 teaspoon vanilla extract",
+    #       "2 large eggs",
+    #       "2 cups chocolate chips"
+    #     ],
+    #     "recipe_name": "Classic Chocolate Chip Cookies"
+    #   }, ... ]
+
+    # To convert the response.text to Recipe class objects.
+    my_recipes = TypeAdapter(list[Recipe]).validate_json(response.text)
+    assert all(isinstance(r, Recipe) for r in my_recipes)
```
holtskinner, line 56: You don't need to do this, can just use `response.parsed`

genai/controlled_generation/ctrlgen_with_enum_schema.py
```diff
@@ -0,0 +1,42 @@
...
+    client = genai.Client()
+    response = client.models.generate_content(
+        model="gemini-2.0-flash-001",
+        contents="What type of instrument is an oboe?",
+        config={
+            "response_mime_type": "text/x.enum",
+            "response_schema": {
+                "type": "STRING",
+                "enum": ["Percussion", "String", "Woodwind", "Brass", "Keyboard"],
+            },
```
holtskinner, line 29: Same here, use class.

genai/controlled_generation/ctrlgen_with_enum_schema.py
```diff
@@ -0,0 +1,42 @@
...
+
+    client = genai.Client()
+    response = client.models.generate_content(
+        model="gemini-2.0-flash-001",
+        contents="What type of instrument is an oboe?",
+        config={
+            "response_mime_type": "text/x.enum",
+            "response_schema": {
+                "type": "STRING",
+                "enum": ["Percussion", "String", "Woodwind", "Brass", "Keyboard"],
```
holtskinner, line 28: Should we recommend using an `Enum` class instead of this list of Strings?

genai/controlled_generation/ctrlgen_with_nested_class_schema.py
```diff
@@ -0,0 +1,54 @@
...
+
+    class Recipe(BaseModel):
+        recipe_name: str
+        rating: Grade
+
+    client = genai.Client()
+    response = client.models.generate_content(
+        model="gemini-2.0-flash-001",
+        contents="List about 10 home-baked cookies and give them grades based on tastiness.",
+        config={
```
holtskinner, line 39: Same here, config class.

genai/controlled_generation/noxfile_config.py
```diff
@@ -0,0 +1,42 @@
...
+
+# You can copy this file into your directory, then it will be imported from
+# the noxfile.py.
+
+# The source of truth:
+# https://github.com/GoogleCloudPlatform/python-docs-samples/blob/main/noxfile_config.py
+
+TEST_CONFIG_OVERRIDE = {
+    # You can opt out from the test for specific Python versions.
+    "ignored_versions": ["2.7", "3.7", "3.8", "3.10", "3.11"],
```
holtskinner, line 25: Should this be updated to include other versions, like 3.12 or 3.13?

genai/controlled_generation/requirements-test.txt
```diff
@@ -0,0 +1,4 @@
+backoff==2.2.1
+google-api-core==2.19.0
```
holtskinner, line 2: There are newer versions of this

genai/controlled_generation/requirements-test.txt
holtskinner, line 1: Should there be a regular `requirements.txt`?

holtskinner (changes requested):

genai/controlled_generation/ctrlgen_with_class_schema.py
```diff
@@ -0,0 +1,62 @@
...
+def generate_content() -> str:
+    # [START googlegenaisdk_ctrlgen_with_class_schema]
+    from google import genai
+    from pydantic import BaseModel, TypeAdapter
+
+    class Recipe(BaseModel):
+        recipe_name: str
+        ingredients: list[str]
+
+    client = genai.Client()
```
holtskinner, line 25: Should these clients have `vertexai=True` and project/location set? Or are we recommending those be set at the environment variable level?

genai/controlled_generation/requirements-test.txt
msampathkumar: When automated bots raise pkg version for `requirements-text.txt`, they can be managed by other reviewers. So best to have both.

genai/controlled_generation/requirements-test.txt
msampathkumar: Not a major issues. We have automated bots that can check and updates these packages.

genai/controlled_generation/noxfile_config.py
msampathkumar: Good point. I will update them.

msampathkumar: Updating the config run tests for 3.9 & 3.13 version

genai/controlled_generation/ctrlgen_with_class_schema.py
msampathkumar: We are going for the simplest option here, `client = genai.Client()`

genai/controlled_generation/ctrlgen_with_class_schema.py
msampathkumar: We will have another example to focus on `GenerateContentConfig`. So being straight forward here (less jargons).

genai/controlled_generation/ctrlgen_with_enum_schema.py
msampathkumar: Soon we will be adding another example to highlight `Enum`

genai/controlled_generation/ctrlgen_with_class_schema.py
holtskinner: Can you show how to access response.parsed in the sample?

msampathkumar: It is possible to do it but why?

msampathkumar: `response.text` is the common pattern we follow, so I did not see the need to change it.

holtskinner: It would be useful to show both, so that users know it's an option.

holtskinner (approved):

genai/controlled_generation/ctrlgen_with_class_schema.py
msampathkumar: Good point but this is like going down a rabbit hole which I leave it to the user's to work upon. Also useful* can be relative to the user's perspective.

msampathkumar: Simple is good.

PR #13143, opened 2025-02-06
feat(modelarmor-samples): add model armor template and prompt samples

Sita04:
## Description

Adds model armor samples for CRUD operation on templates and user prompt sanitization.

## Checklist
- [x] These samples need a new **API enabled** in testing projects to pass (modelarmor.googleapis.com)


glasnt:
Looks like after your linting changes, you'll need `pytest` in a `requirements.txt` file. If you still have more changes to make, do you want to make this PR draft until you're ready for review?

model_armor/create_template.py
```diff
@@ -0,0 +1,63 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+def create_model_armor_template(project_id: str, location: str, template_id: str):
+    # [START modelarmor_create_template]
+
+    from google.cloud import modelarmor_v1
+    client = modelarmor_v1.ModelArmorClient(transport="rest", client_options={
```
holtskinner, line 19: Can you use the `ClientOptions` class? Similar to this sample https://github.com/GoogleCloudPlatform/python-docs-samples/blob/main/discoveryengine/search_sample.py

model_armor/create_template.py
```diff
@@ -0,0 +1,63 @@
...
+        "api_endpoint": "modelarmor.us-central1.rep.googleapis.com"})
+
+    # TODO(Developer): Uncomment these variables and initialize
+    # project_id = "your-google-cloud-project-id"
+    # location = "us-central1"
+    # template_id = "template_id"
+
+    template = {
+        "name": f"projects/{project_id}/locations/{location}/templates/{template_id}",
```
holtskinner, line 28: There should be a method to get this path on the client object. Like `template_path()`

model_armor/create_template.py
```diff
@@ -0,0 +1,63 @@
...
+        "filter_config": {
```
holtskinner, line 29: Can you use the `Template` class instead of a dict? Works better for IDEs/autocompletion.

holtskinner (changes requested):

model_armor/create_template.py
Sita04: The template class unfortunately doesn't provide this now.

Sita04:
@glasnt This PR is now ready for your review. Thanks for the patience!

model_armor/create_template.py
```diff
@@ -0,0 +1,63 @@
...
+    )
+
+    # Make the request
+    response = client.create_template(request=request)
+
+    # Response
+    return response
+
+
+# [END modelarmor_create_template]
```
glasnt, line 63: https://googlecloudplatform.github.io/samples-style-guide/ "Process the result" > "show the user how to interact with a returned object programmatically by printing example attributes to the console."

I would suggest changing this to have the template printed, then the `[END` region tag, then the return statement to support testing. 

I'm not sure how to interact with the template so I don't know if you can just `print(response)` or if it requires more steps, so I can't make a suggested change. 

This comment applies to all other samples that return before the `[END`.

model_armor/delete_template.py
```diff
@@ -0,0 +1,40 @@
...
+
+def delete_model_armor_template(project_id: str, location: str, template_id: str) -> None:
+    # [START modelarmor_delete_template]
+
+    from google.api_core.client_options import ClientOptions
+    from google.cloud import modelarmor_v1
+
+    client = modelarmor_v1.ModelArmorClient(
+        transport="rest",
+        client_options=ClientOptions(api_endpoint=f"modelarmor.{location}.rep.googleapis.com"),
```
glasnt, line 24: Is this a temporary configuration? I'm not familiar with `.rep.googleapis.com` endpoints, but overriding `api_endpoint` reads to me as a beta workaround.

model_armor/create_template.py
glasnt: Also, having a `return` outside of a Python method is invalid so this sample would not be copy/pastable, even after replacing the project_id/location/template_id variables.

Sita04: Thanks for the catch! I've updated to print the response. And added the region tag before return.

glasnt (approved):
All of my comments are addressed, unsure if @holtskinner had others

model_armor/update_template.py
```diff
@@ -0,0 +1,60 @@
...
+    from google.cloud.modelarmor_v1 import Template, DetectionConfidenceLevel, FilterConfig, PiAndJailbreakFilterSettings, \
+        MaliciousUriFilterSettings
+
+    client = modelarmor_v1.ModelArmorClient(
+        transport="rest",
+        client_options=ClientOptions(api_endpoint=f"modelarmor.{location}.rep.googleapis.com"),
+    )
+
+    # TODO(Developer): Uncomment these variables and initialize
+    # project_id = "your-google-cloud-project-id"
```
holtskinner, line 32: I think there's a standard format for how to make placeholders. (It makes it possible to add editable placeholder variables in the documentation)

This or something like it?
```py
# project_id = "YOUR_PROJECT_ID"
# location = "us-central1"
```

model_armor/create_template.py
```diff
@@ -0,0 +1,63 @@
...
+        client_options=ClientOptions(api_endpoint=f"modelarmor.{location}.rep.googleapis.com"),
+    )
+
+    # TODO(Developer): Uncomment these variables and initialize
+    # project_id = "your-google-cloud-project-id"
+    # location = "us-central1"
+    # template_id = "template_id"
+
+    template = Template(
+        name=f"projects/{project_id}/locations/{location}/templates/{template_id}",
```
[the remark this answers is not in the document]
Sita04: Currently, we do not have an equivalent client method to construct this path.

model_armor/sanitize_user_prompt.py
```diff
@@ -0,0 +1,56 @@
...
+from google.cloud.modelarmor_v1 import SanitizeUserPromptResponse
+
+
+def sanitize_user_prompt(
+    project_id: str, location: str, template_id: str
+) -> SanitizeUserPromptResponse:
+    # [START modelarmor_sanitize_user_prompt]
+
+    from google.api_core.client_options import ClientOptions
+    from google.cloud import modelarmor_v1
```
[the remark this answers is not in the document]
Sita04: Updated. Thanks

model_armor/sanitize_user_prompt.py
```diff
@@ -0,0 +1,56 @@
...
+
+    # Define the prompt
+    user_prompt = "Can you describe this link? https://testsafebrowsing.appspot.com/s/malware.html"
+
+    # Initialize request argument(s)
+    user_prompt_data = modelarmor_v1.DataItem()
+    user_prompt_data.text = user_prompt
+
+    request = modelarmor_v1.SanitizeUserPromptRequest(
+        name=f"projects/{project_id}/locations/{location}/templates/{template_id}",
```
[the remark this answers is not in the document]
Sita04: ditto

model_armor/list_templates.py
```diff
@@ -0,0 +1,45 @@
...
+        client_options=ClientOptions(api_endpoint=f"modelarmor.{location}.rep.googleapis.com"),
+    )
+
+    # TODO(Developer): Uncomment these variables and initialize
+    # project_id = "your-google-cloud-project-id"
+    # location = "us-central1"
+
+    # Initialize request argument(s)
+    request = modelarmor_v1.ListTemplatesRequest(
+        parent=f"projects/{project_id}/locations/{location}"
```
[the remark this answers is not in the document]
Sita04: ditto

model_armor/get_template.py
```diff
@@ -0,0 +1,46 @@
...
+        client_options=ClientOptions(api_endpoint=f"modelarmor.{location}.rep.googleapis.com"),
+    )
+
+    # TODO(Developer): Uncomment these variables and initialize
+    # project_id = "your-google-cloud-project-id"
+    # location = "us-central1"
+    # template_id = "template_id"
+
+    # Initialize request arguments
+    request = modelarmor_v1.GetTemplateRequest(
```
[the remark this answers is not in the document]
Sita04: ditto

model_armor/delete_template.py
```diff
@@ -0,0 +1,40 @@
...
+    )
+
+    # TODO(Developer): Uncomment these variables and initialize
+    # project_id = "your-google-cloud-project-id"
+    # location = "us-central1"
+    # template_id = "template_id"
+
+    request = modelarmor_v1.DeleteTemplateRequest(
+        name=f"projects/{project_id}/locations/{location}/templates/{template_id}",
```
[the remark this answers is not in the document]
Sita04: ditto

model_armor/create_template.py
```diff
@@ -0,0 +1,63 @@
...
+            ),
+            malicious_uri_filter_settings=MaliciousUriFilterSettings(
+                filter_enforcement=MaliciousUriFilterSettings.MaliciousUriFilterEnforcement.ENABLED,
+            )
+        ),
+    )
+
+    # Initialize request arguments
+    request = modelarmor_v1.CreateTemplateRequest(
+        parent=f"projects/{project_id}/locations/{location}",
```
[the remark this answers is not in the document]
Sita04: ditto

litos661 (approved):

PR #13146, opened 2025-02-10
feat(cloud-sql): enable lazy refresh for postgres, mysql and sql server connectors

kgala2:
Enabled lazy refresh for the Python Cloud SQL connector (Postgres, MySQL, and SQL Server).

### DIFF
--- a/cloud-sql/mysql/sqlalchemy/connect_connector.py
+++ b/cloud-sql/mysql/sqlalchemy/connect_connector.py
@@ -42,7 +42,7 @@ def connect_with_connector() -> sqlalchemy.engine.base.Engine:
     ip_type = IPTypes.PRIVATE if os.environ.get("PRIVATE_IP") else IPTypes.PUBLIC
     refresh_strategy = RefreshStrategy.LAZY
 
-    connector = Connector(ip_type = ip_type, refresh_strategy = refresh_strategy)
+    connector = Connector(ip_type=ip_type, refresh_strategy=refresh_strategy)
 
     def getconn() -> pymysql.connections.Connection:
         conn: pymysql.connections.Connection = connector.connect(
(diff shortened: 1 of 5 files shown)


cloud-sql/mysql/sqlalchemy/connect_connector.py
```diff
@@ -40,8 +40,9 @@ def connect_with_connector() -> sqlalchemy.engine.base.Engine:
     db_name = os.environ["DB_NAME"]  # e.g. 'my-database'
 
     ip_type = IPTypes.PRIVATE if os.environ.get("PRIVATE_IP") else IPTypes.PUBLIC
+    refresh_strategy = RefreshStrategy.LAZY
 
-    connector = Connector(ip_type)
+    connector = Connector(ip_type=ip_type, refresh_strategy=refresh_strategy)
```
jackwotherspoon, line 45: nit: there is no need to set `refresh_strategy` as it is only used in one place.

The Cloud SQL Python Connector also allows taking a `str` for `refresh_strategy` instead of the custom type.
If we use `"lazy"` it is a bit more user friendly for the sample as customers will not need to import the custom type of `RefreshStrategy` 😄 

Let's also add to the existing comment to explain why we are setting it. 

Something along the lines of "Setting 'refresh_strategy' argument to "lazy" enables lazy refresh mode which is recommended for serverless environments to avoid background refreshes from throttling CPU."
```suggestion
    connector = Connector(ip_type=ip_type, refresh_strategy="lazy")
```

jackwotherspoon: Same applies to other engine types.

jackwotherspoon (changes requested):

iennae (approved):
nit - I think the comment might be a little misleading because it comes across that it's only recommended for serverless environments; AFAIK, it's less useful in non-serverless environments because those instances are long lived so the performance gains may not be as pronounced but it's perfectly ok to use in non-serverless environments as well.

jackwotherspoon:
> AFAIK, it's less useful in non-serverless environments because those instances are long lived so the performance gains may not be as pronounced but it's perfectly ok to use in non-serverless environments as well.

This is correct, it is perfectly acceptable to use `refresh_strategy="lazy"` in non-serverless environments. It just means that the first connection attempt after the "cache" has expired will be slightly less performant as it blocks on a "refresh".

I am fine with removing the comment to make the sample more robust and less confusing. 😄

jackwotherspoon (approved):

PR #13154, opened 2025-02-12
feat(genai): Add new Controlled Generation samples

holtskinner:
- CtrlGen with Enum Class
- CtrlGen with Resp Schema

Added requirements.txt


msampathkumar (approved):
LGTM

PR #13160, opened 2025-02-12
feat(genai): Update Text Generation with PDF Sample

holtskinner:
- Changed to a basic Doc Summarization example instead of reading code from PDF.

### DIFF
--- a/genai/text_generation/textgen_with_pdf.py
+++ b/genai/text_generation/textgen_with_pdf.py
@@ -24,8 +24,7 @@ def generate_content() -> str:
     model_id = "gemini-2.0-flash-001"
 
     prompt = """
-    You are a very professional document summarization specialist.
-    Please summarize the given document.
+    You are a highly skilled document summarization specialist. Your task is to provide a concise executive summary of no more than 300 words. Please summarize the given document for a general audience.
     """
 
     pdf_file = Part.from_uri(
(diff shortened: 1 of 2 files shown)


genai/text_generation/textgen_with_pdf.py
```diff
@@ -16,32 +16,34 @@
...
     from google import genai
     from google.genai.types import HttpOptions, Part
 
     client = genai.Client(http_options=HttpOptions(api_version="v1"))
     model_id = "gemini-2.0-flash-001"
 
-    python_code = Part.from_uri(
-        file_uri="https://storage.googleapis.com/cloud-samples-data/generative-ai/text/inefficient_fibonacci_series_python_code.pdf",
+    prompt = """
+    You are a very professional document summarization specialist.
```
iennae, line 27: @holtskinner I know this prompt is used in other places but I'm wondering if it should actually showcase a recommendation here as this feels very basic without any additional context or specificity. The "very" is what triggered this. 

Thoughts on something like the following

"You are a highly skilled document summarization specialist. Your task is to provide a concise executive summary of no more than 300 words. Please summarize the given document for a general audience." 

This feels like an area of sample style specifically around prompting that we haven't set a standard for but I'd like to encourage some thoughtfulness in this area so we set a bar of expectation around prompting that follows the recommendation we share. This is entirely not my specialty so I look to you to identify what this might look like. Thanks so much!

holtskinner: That sounds good to me

msampathkumar (approved):
LGTM

PR #13161, opened 2025-02-12
fix: update suggested video for summarization

glasnt:
Fixes test failures for generative_ai/video (e.g. https://github.com/GoogleCloudPlatform/python-docs-samples/pull/13152)

Updates the sample video being used for summarization. The original video was marked private, this is another video of a similar type from the same channel (Made with Google). 


iennae (approved):
Thanks @glasnt

PR #13188, opened 2025-02-25
chore(genai): Update Chat and Code Exec Samples for 1.3.0 SDK Update

holtskinner:
## Description

https://github.com/googleapis/python-genai/releases/tag/v1.3.0

Update `genai/text_generation/textgen_chat_with_txt.py` to use `UserContent` and `ModelContent`

Update `genai/tools/tools_code_exec_with_txt.py` and `genai/tools/tools_code_exec_with_txt_local_img.py` to use `response.executable_code` and `response.code_execution_result`

### DIFF
--- a/genai/tools/tools_code_exec_with_txt.py
+++ b/genai/tools/tools_code_exec_with_txt.py
@@ -35,19 +35,32 @@ def generate_content() -> str:
             temperature=0,
         ),
     )
-    for part in response.candidates[0].content.parts:
-        if part.executable_code:
-            print(part.executable_code)
-        if part.code_execution_result:
-            print(part.code_execution_result)
+    print("# Code:")
+    print(response.executable_code)
+    print("# Outcome:")
+    print(response.code_execution_result)
+
     # Example response:
-    # code='...' language='PYTHON'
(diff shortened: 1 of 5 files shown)


genai/tools/tools_code_exec_with_txt.py
```diff
@@ -35,19 +35,32 @@ def generate_content() -> str:
             temperature=0,
         ),
     )
-    for part in response.candidates[0].content.parts:
-        if part.executable_code:
-            print(part.executable_code)
-        if part.code_execution_result:
-            print(part.code_execution_result)
+    print("# Code:")
```
iennae, line 38: based on the update, it seems like there might still be multiple candidates but only one is returned. Is there a need to provide some way of parsing that in the default (just looking at https://github.com/googleapis/python-genai/commit/3725ddf44a0df04f6fac0bd5295f45ed20b4a8fe and seeing this)

holtskinner: From my understanding, there should only be one candidate generated for Code Execution queries.

iennae (approved):

PR #13231, opened 2025-03-13
fix(bigquery): fix typos and comments in cloud-client samples

eapl-gemugami:
## Description

Fixes b/403272595

- Fix comments (for example 'grant' instead of 'revoke', and found typos)
- Remove unnecessary comments (for example in imports at the start of the sample)
- Unify comments style

## Checklist

### DIFF
--- a/bigquery/cloud-client/revoke_dataset_access.py
+++ b/bigquery/cloud-client/revoke_dataset_access.py
@@ -60,7 +60,7 @@ def revoke_dataset_access(dataset_id: str, entity_id: str) -> list[AccessEntry]:
             ["access_entries"],
         )
 
-        # Show a success message.
+        # Notify user that the API call was successful.
         full_dataset_id = f"{dataset.project}.{dataset.dataset_id}"
         print(f"Revoked dataset access for '{entity_id}' to ' dataset '{full_dataset_id}.'")
     except PreconditionFailed:  # A read-modify-write error.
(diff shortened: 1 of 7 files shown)


bigquery/cloud-client/revoke_dataset_access.py
```diff
@@ -61,6 +60,7 @@ def revoke_dataset_access(dataset_id: str, entity_id: str) -> list[AccessEntry]:
             ["access_entries"],
         )
 
+        # Show a success message.
```
iennae, line 63: This is successful completion of the requested action but it feels like the wrong comment. Maybe something more like following a pattern of 

// # Print the XXX response to verify the API call was successful 

where XXX is the type of response.

eapl-gemugami: For this case I wasn't printing the response, or verifying it. Only assuming that the API call worked, and printing some message to the CLI with info about which entity lost access to the dataset.

```
dataset = client.update_dataset(
    dataset,
    ["access_entries"],
)

full_dataset_id = f"{dataset.project}.{dataset.dataset_id}"
print(f"Revoked dataset access for '{entity_id}' to ' dataset '{full_dataset_id}.'")
```

What about something like:
`# Inform the user the API call was successful.`

eapl-gemugami: I've changed it to `# Notify user that the API call was successful.` in the latest commit. Please let me know if there is a better approach.

iennae (approved):

PR #13255, opened 2025-03-25
feat(translate): add region tag 'import_client_library' and refactor sample

eapl-gemugami:
## Description

Fixes Internal:
b/405384416

- Add region tag 'translate_v3_import_client_library'
- Remove remaining Markdown files from previous repos [1](https://github.com/googleapis/python-translate) and [2](https://github.com/googleapis/google-cloud-python/tree/main/packages/google-cloud-translate/samples)
- Fix styling of the sample and removed unnecessary dependencies
- Refactor to improve sample objective of showing how to translate a simple text.

## Checklist

### DIFF
--- a/translate/samples/snippets/translate_v3_translate_text.py
+++ b/translate/samples/snippets/translate_v3_translate_text.py
@@ -17,7 +17,7 @@ import os
 
 # Import the Google Cloud Translation library.
 # [START translate_v3_import_client_library]
-from google.cloud import translate_v3beta1 as translate
+from google.cloud import translate_v3
 # [END translate_v3_import_client_library]
 
 PROJECT_ID = os.environ.get("GOOGLE_CLOUD_PROJECT")
@@ -27,7 +27,7 @@ def translate_text(
     text: str = "YOUR_TEXT_TO_TRANSLATE",
     source_language_code: str = "en-US",
     target_language_code: str = "fr",
-) -> translate.TranslationServiceClient:
+) -> translate_v3.TranslationServiceClient:
     """Translate Text from a Source language to a Target language.
     Args:
         text: The content to translate.
@@ -39,7 +39,7 @@ def translate_text(
     """
 
     # Initialize Translation client.
-    client = translate.TranslationServiceClient()
+    client = translate_v3.TranslationServiceClient()
     parent = f"projects/{PROJECT_ID}/locations/global"
 
     # MIME type of the content to translate.
--- a/translate/samples/snippets/translate_v3_translate_text.py
+++ b/translate/samples/snippets/translate_v3_translate_text.py
@@ -25,14 +25,16 @@ PROJECT_ID = os.environ.get("GOOGLE_CLOUD_PROJECT")
 
 def translate_text(
     text: str = "YOUR_TEXT_TO_TRANSLATE",
-    language_code: str = "fr",
+    source_language_code: str = "en-US",
+    target_language_code: str = "fr",
 ) -> translate_v3.TranslationServiceClient:
-    """Translating Text from English.
+    """Translate Text from a Source language to a Target language.
     Args:
         text: The content to translate.
-        language_code: The language code for the translation.
-            E.g. "fr" for French, "es" for Spanish, etc.
-            Available languages:
+        source_language_code: The code of the source language.
+        target_language_code: The code of the target language.
+            For example: "fr" for French, "es" for Spanish, etc.
+            Find available languages and codes here:
             https://cloud.google.com/translate/docs/languages#neural_machine_translation_model
     """
 
@@ -40,26 +42,18 @@ def translate_text(
     client = translate_v3.TranslationServiceClient()
     parent = f"projects/{PROJECT_ID}/locations/global"
 
-    # Translate text from English to chosen language.
-    # Supported MIME types:
-    # https://cloud.google.com/translate/docs/supported-formats
+    # Translate text from the source to the target language.
     response = client.translate_text(
         contents=[text],
-        target_language_code=language_code,
         parent=parent,
-        mime_type="text/plain",
-        source_language_code="en-US",
+        source_language_code=source_language_code,
+        target_language_code=target_language_code,
     )
 
-    # Display the translation for each input text provided.
-    for translation in response.translations:
-        print(f"Translated text: {translation.translated_text}")
-
-    # Example response:
+    # Display the translation for the text.
+    # For example, for "Hello! How are you doing today?":
     # Translated text: Bonjour comment vas-tu aujourd'hui?
+    print(f"Translated text: {response.translations[0].translated_text[0]}")
+
     return response
 # [END translate_v3_translate_text]
-
-
-if __name__ == "__main__":
-    translate_text(text="Hello! How are you doing today?", language_code="fr")
--- a/translate/samples/snippets/translate_v3_translate_text.py
+++ b/translate/samples/snippets/translate_v3_translate_text.py
@@ -17,7 +17,7 @@ import os
 
 # Import the Google Cloud Translation library.
 # [START translate_v3_import_client_library]
-from google.cloud import translate_v3
+from google.cloud import translate_v3beta1 as translate
 # [END translate_v3_import_client_library]
 
 PROJECT_ID = os.environ.get("GOOGLE_CLOUD_PROJECT")
@@ -27,7 +27,7 @@ def translate_text(
     text: str = "YOUR_TEXT_TO_TRANSLATE",
     source_language_code: str = "en-US",
     target_language_code: str = "fr",
-) -> translate_v3.TranslationServiceClient:
+) -> translate.TranslationServiceClient:
     """Translate Text from a Source language to a Target language.
     Args:
         text: The content to translate.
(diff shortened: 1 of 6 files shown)


translate/samples/snippets/translate_v3_translate_text.py
```diff
@@ -31,13 +32,17 @@ def translate_text(
...
 
+    # Initialize Translation client.
     client = translate_v3.TranslationServiceClient()
     parent = f"projects/{PROJECT_ID}/locations/global"
-    # Translate text from English to chosen language
-    # Supported mime types: # https://cloud.google.com/translate/docs/supported-formats
+
+    # Translate text from English to chosen language.
+    # Supported MIME types:
+    # https://cloud.google.com/translate/docs/supported-formats
```
iennae, line 45: I know this is not your original comment, but I'm wondering whether the link to supported types is a useful comment here, or oh maybe it's missing a comment. There is the chosen language and then there is the contents and specifying the type

eapl-gemugami: I moved it from the in-line mime parameter to a few lines above, but I agree, the place is not the best, I'll rework it

iennae: Yeah, now that I'm looking at this I think updating the "translate text form english to chose language to 

# Call the Translation API to translate the input text. 
# Supported MIME types: 


something about the flow of the comments didn't work well when looking at it fresh to me.

translate/samples/snippets/translate_v3_translate_text.py
```diff
@@ -31,13 +32,17 @@ def translate_text(
         text: The content to translate.
         language_code: The language code for the translation.
```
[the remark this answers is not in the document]
eapl-gemugami: I agree with the suggestion, although [documentation](https://cloud.google.com/translate/docs/languages#neural_machine_translation_model) says:
> Most language code parameters conform to [ISO-639](https://wikipedia.org/wiki/ISO_639) identifiers, **except where noted**.

Leaving "The language code", and the link to the language models should be enough as there are many different variations and exceptions.

It's funny, the `en-US` used in `source_language_code` line 51 below, is not in the supported list, only `en`, although it works.

translate/samples/snippets/translate_v3_translate_text.py
eapl-gemugami: I'm watching that the MIME type is not required for this sample. I was confused with another sample which uploads a PDF.

As it passes tests without a MIME type, which defaults to `text/html`.

eapl-gemugami:
@iennae looking into your feedback I seized the opportunity to refactor the sample as David told me for previous cases.

I think the original sample was inconsistent on its purpose of 'translating text', for example only translating from English to another language, so I added the concept of source and target languages as parameters, which is like usually the translating platforms work.

I also removed the MIME type, not strictly required for this example, and cleaned it a bit the sample and the test.
As it is not a minor change in the sample, could you review it again please? Thanks!

translate/samples/snippets/translate_v3_translate_text.py
```diff
@@ -25,41 +25,35 @@
...
             https://cloud.google.com/translate/docs/languages#neural_machine_translation_model
     """
 
     # Initialize Translation client.
     client = translate_v3.TranslationServiceClient()
     parent = f"projects/{PROJECT_ID}/locations/global"
 
-    # Translate text from English to chosen language.
-    # Supported MIME types:
-    # https://cloud.google.com/translate/docs/supported-formats
```
glasnt: mime_type defaults to `text/plain` ([source](https://github.com/googleapis/google-cloud-node/blob/main/packages/google-cloud-translate/samples/generated/v3/translation_service.translate_text.js#L38-L39)), but I think it's important to still highlight that this API supports different mime types.

translate/samples/snippets/translate_v3_translate_text.py
```diff
@@ -25,41 +25,35 @@
...
         parent=parent,
-        mime_type="text/plain",
-        source_language_code="en-US",
+        source_language_code=source_language_code,
+        target_language_code=target_language_code,
     )
 
-    # Display the translation for each input text provided.
-    for translation in response.translations:
-        print(f"Translated text: {translation.translated_text}")
```
glasnt, line 51: Translations is plural here, is there a reason this was changed to return only the first result?

iennae: Oh interesting, looks like the test only tested the first result so maybe pulling that intent in?

translate/samples/snippets/translate_v3_translate_text.py
eapl-gemugami: That link says
> If left blank, the MIME type defaults to "text/html". Perhaps that Doc is outdated related to the client library? I'll take a look into it

eapl-gemugami:
@iennae @glasnt Thanks for your review! My thought process for this refactor starts at this page (sample appears in other pages as well): https://cloud.google.com/translate/docs/samples/translate-v3-translate-text

Based on the title and context, I noticed the following
- I'd expect for this sample to receive a plain text and translate it from one language into another.
- It shows how to translate a text from English to another (defaulting to French). As the source language is not autodetected, it has to be supplied manually, I think it's better for the developer to supply the text, any source language and the target language. The API request these, so it makes more senso to me that way.
- A single text is supplied, and then converted into a list of 1 element, it makes no sense to iterate over a list of results if a single one is expected -> I see two options, supplying a text and reading the first result, or supplying two texts in the sample and print all the results.
Personally I'd vote for the second option, usually in i18n is better to prepare the user for multiple "things" instead of the fist one.
- Although I agree on showing that the API supports many MIME types, I think that for this sampl the default one could be enough, and perhaps in another sample it could be supplied as an HTML input to show that the default value has to be changed when working with some other kind of content to translate.
That based on the simplicity principle.

That said, these are suggestions based on my first sight on the sample. I'd need to check all the translate samples in context, or know the intent behind this sample.

Please let me know if my suggestion would work, otherwise I can revert it.

translate/samples/snippets/translate_v3_translate_text.py
eapl-gemugami: Thanks @glasnt, I share my thought process about this, in a comment in this PR (as it may require more context)

Basically, the sample only supplies 1 text, so it makes no sense to me to iterate a list in this case.

glasnt (commented):
Going by the [source](https://cloud.google.com/translate/docs/advanced/translating-text-v3#translate_v3_translate_text_with_model-drest) you linked, I've made some comments to address the additional changes made, while trying to keep in line with the sample style guide and similarity with other implementations (e.g. not all implementations take a source language in the function definition, but may use a default value).

translate/samples/snippets/translate_v3_translate_text.py
```diff
@@ -13,49 +13,47 @@
...
+            For example: "fr" for French, "es" for Spanish, etc.
+            Find available languages and codes here:
+            https://cloud.google.com/translate/docs/languages#neural_machine_translation_model
     """
 
+    # Initialize Translation client.
     client = translate_v3.TranslationServiceClient()
     parent = f"projects/{PROJECT_ID}/locations/global"
-    # Translate text from English to chosen language
-    # Supported mime types: # https://cloud.google.com/translate/docs/supported-formats
```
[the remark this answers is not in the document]
eapl-gemugami: I think just restoring the line could be confusing for the developer, as there are many MIME types, and it could not be obvious that we are doing a "Text translation on v3".

What about something like:
```
    # MIME type of the content to translate.
    # Find supported types for `Text translation - Advanced (v3)` here:
    # https://cloud.google.com/translate/docs/supported-formats
    mime_type = "text/plain"

    # Translate text from the source to the target language.
    response = client.translate_text(
        contents=[text],
        parent=parent,
        mime_type=mime_type,
```

Another option could be using this comment from other samples, although I like it more the approach of linking to that `supported-formats` URL, as it gives more context.
```
// Go sample
MimeType:           "text/plain", // Mime types: "text/plain", "text/html"

// Node sample
mimeType: 'text/plain', // mime types: text/plain, text/html
```

eapl-gemugami:
@glasnt I have applied the feedback from your [latest comment](https://github.com/GoogleCloudPlatform/python-docs-samples/pull/13255#pullrequestreview-2719422441).

I only have a question on [this approach](https://github.com/GoogleCloudPlatform/python-docs-samples/pull/13255#discussion_r2016903429) for the MIME type.

If that looks correct, it's ready to be reviewed again. Thanks!

eapl-gemugami:
cc @alarconesparza @OremGLG @hivanalejandro (Team colleagues)

OremGLG:
LGTM 👍

translate/samples/snippets/translate_v3_translate_text.py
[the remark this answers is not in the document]
glasnt: A "better" solution would be for the docs link to be part of the docs page where this sample is used, and for the linked page itself to have better sections so one could link to the specific element, rather than duplicating content and potentially coming out of date (e.g. if a mimetype is added/removed, then the comments on Go/Node would be incorrect).

Because "Text translation - Advanced (v3)" is not currently a section in that doc, just keep the original text to the link.

glasnt (approved):
Pending doclink update

translate/samples/snippets/translate_v3_translate_text.py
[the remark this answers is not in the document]
eapl-gemugami: Agreed, I have restored the following lines
```
# Supported mime types:
# https://cloud.google.com/translate/docs/supported-formats
```

And I'll create a new issue to fix the docs, and to link the sample to the right section.

glasnt (approved):

PR #13292, opened 2025-04-07
chore(cloudrun): refactor to sample 'cloudrun_service_to_service_receive' and its test

eapl-gemugami:
## Description

Fixes Internal: b/405225642

- Add region tag as requested in b/405225642#comment6
- Update dependencies to latest versions
- Increase coverage of unit tests for anonymous requests and submitting invalid token
- Remove `noxfile_config.py` as it was outdated for Python 3.8 and it isn't necessary anymore
- Apply fixes for latest Style Guide
- Add type hints
- Rename variables to be consistent with their values
- Replace ints for HTTP status with `http.HTTPStatus` enum
- Rewrite comments for accuracy, as some were incomplete or misleading

## Checklist


glasnt:
_incorrectly marked this PR as fixed by another, please ignore)_

run/service-auth/receive_test.py
```diff
@@ -25,33 +25,55 @@
...
+    HTTP_STATUS_SERVICE_UNAVAILABLE,
+    HTTP_STATUS_GATEWAY_TIMEOUT,
+],
+
+
+@pytest.fixture(scope="module")
+def service() -> tuple[str, str]:
+    """Deploys a Cloud Run service and returns its URL and a valid token."""
+    # Add a unique suffix to create distinct service names.
+    service_name = f"receive-{uuid.uuid4()}"
```
glasnt, line 56: You dropped the `.hex` from this call.
```suggestion
    service_name = f"receive-{uuid.uuid4().hex}"
```

run/service-auth/receive_test.py
```diff
@@ -25,33 +25,55 @@
...
+STATUS_FORCELIST = [
+    HTTP_STATUS_BAD_REQUEST,
+    HTTP_STATUS_UNAUTHORIZED,
+    HTTP_STATUS_FORBIDDEN,
+    HTTP_STATUS_NOT_FOUND,
+    HTTP_STATUS_INTERNAL_SERVER_ERROR,
+    HTTP_STATUS_BAD_GATEWAY,
+    HTTP_STATUS_SERVICE_UNAVAILABLE,
+    HTTP_STATUS_GATEWAY_TIMEOUT,
+],
```
glasnt, line 49: HTTP status codes aren't really "magic numbers", but you're also creating a bunch of manual variables for things that are mostly only directly used in your "STATUS_FORCELIST"

Instead of re-creating your own magic variables, you can use the existing lists from either the core standard library [`HTTPStatus`](https://docs.python.org/3/library/http.html#http.HTTPStatus), or from [`requests.codes`](https://requests.readthedocs.io/en/latest/api/#requests.codes), since that'a already imported. Use these to declare your list (as it is called multiple times)

run/service-auth/receive_test.py
```diff
@@ -101,19 +124,75 @@ def services():
...
+
+
+def test_anonymous_request_on_cloud_run(service: tuple[str, str]) -> None:
+    endpoint_url = service[0]
+
+    req = request.Request(endpoint_url)
+    try:
+        _ = request.urlopen(req)
+    except error.HTTPError as e:
+        assert e.code == HTTP_STATUS_FORBIDDEN
```
glasnt, line 163: You could use `requests.codes.forbidden` or `HTTPStatus.FORBIDDEN` here. 

The comparison on the HTTPStatus check will give correct equality checks with numbers.

run/service-auth/receive_test.py
```diff
@@ -101,19 +124,75 @@ def services():
...
+
+    retry_strategy = Retry(
+        total=3,
+        status_forcelist=STATUS_FORCELIST,
```
glasnt, line 167: From earlier in the comments: this is the only time you're using most of the http status codes.

Instead of re-creating your own magic variables, you can use the existing lists from either the core standard library [`HTTPStatus`](https://docs.python.org/3/library/http.html#http.HTTPStatus), or from [`requests.codes`](https://requests.readthedocs.io/en/latest/api/#requests.codes), since that'a already imported.

run/service-auth/receive_test.py
```diff
@@ -101,19 +124,75 @@ def services():
...
+    )
+    adapter = HTTPAdapter(max_retries=retry_strategy)
+
+    client = requests.session()
+    client.mount("https://", adapter)
+
+    response = client.get(endpoint_url, headers={"Authorization": f"Bearer {token}"})
+    response_content = response.content.decode("UTF-8")
+
+    assert response.status_code == HTTP_STATUS_OK
```
glasnt, line 151: requests.codes.ok or HTTPStatus.OK here

run/service-auth/receive_test.py
```diff
@@ -101,19 +124,75 @@ def services():
...
+    )
+    adapter = HTTPAdapter(max_retries=retry_strategy)
+
+    client = requests.session()
+    client.mount("https://", adapter)
+
+    response = client.get(endpoint_url)
+    response_content = response.content.decode("UTF-8")
+
+    assert response.status_code == HTTP_STATUS_OK
```
glasnt, line 179: requests.codes.ok or HTTPStatus.OK here

run/service-auth/receive_test.py
```diff
@@ -101,19 +124,75 @@ def services():
...
+def test_invalid_token(service: tuple[str, str]) -> None:
+    endpoint_url = service[0]
 
-    req = request.Request(url)
+    req = request.Request(endpoint_url)
     try:
         _ = request.urlopen(req)
     except error.HTTPError as e:
-        assert e.code == 403
+        assert e.code == HTTP_STATUS_FORBIDDEN
```
glasnt, line 191: requests.codes.ok or HTTPStatus.OK here

run/service-auth/receive_test.py
```diff
@@ -101,19 +124,75 @@ def services():
...
+    retry_strategy = Retry(
+        total=3,
+        status_forcelist=STATUS_FORCELIST,
+        allowed_methods=["GET", "POST"],
+        backoff_factor=3,
+    )
+    adapter = HTTPAdapter(max_retries=retry_strategy)
+
+    client = requests.session()
+    client.mount("https://", adapter)
```
glasnt, line 146: This entire section is repeated for each test you're declaring. You should make this a method that can reuse for each of your test methods.

run/service-auth/receive_test.py
```diff
@@ -101,19 +124,75 @@ def services():
     )
 
 
-def test_auth(services):
-    url = services[0]
-    token = services[1]
+def test_authentication_on_cloud_run(service: tuple[str, str]) -> None:
+    endpoint_url = service[0]
```
glasnt, line 128: You're re-declaring this at the start of each of your tests. You could create a fixture that returns this endpoint for you. (Particularly because the `0` here could be a 'magic number', as Cloud Run services now have multiple URLs

run/service-auth/receive_test.py
```diff
@@ -122,8 +201,9 @@ def test_auth(services):
     client = requests.session()
     client.mount("https://", adapter)
 
-    response = client.get(url, headers={"Authorization": f"Bearer {token}"})
+    # Sample token from https://jwt.io for John Doe.
+    token = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiaWF0IjoxNTE2MjM5MDIyfQ.SflKxwRJSMeKKF2QT4fwpMeJf36POk6yJV_adQssw5c"
```
glasnt, line 205: This is a magic token, and should be generated directly if you want to use such functionality. 

This is even used in other samples in this repo: https://github.com/GoogleCloudPlatform/python-docs-samples/blob/main/appengine/standard_python3/pubsub/main_test.py#L52 (the exact fixture might be too complex for your use case, but it's an example to borrow from)

glasnt (changes requested):
Initial feedback, mostly around working with 'magic' HTTP values and reducing duplicate code.

run/service-auth/data/privatekey.pem
```diff
@@ -0,0 +1,27 @@
+-----BEGIN RSA PRIVATE KEY-----
```
glasnt, line 1: My apologies for not realising my advise would get you to create this sort of file. This is another example of a magic value, and we should never be committing keys to our repo. I'll investigate a simpler way of generating a JWT to fail that specific test.

run/service-auth/receive_test.py
```diff
@@ -15,53 +15,79 @@
...
+    HTTPStatus.FORBIDDEN,
+    HTTPStatus.NOT_FOUND,
+    HTTPStatus.INTERNAL_SERVER_ERROR,
+    HTTPStatus.BAD_GATEWAY,
+    HTTPStatus.SERVICE_UNAVAILABLE,
+    HTTPStatus.GATEWAY_TIMEOUT,
 ],
 
 
+DATA_DIR = os.path.join(os.path.dirname(__file__), "data")
```
glasnt, line 49: If this is the process you need to generate a token that will cause a failure, it might be useful to copy the example from jwt.io

run/service-auth/receive_test.py
```diff
@@ -145,65 +179,36 @@ def test_authentication_on_cloud_run(service: tuple[str, str]) -> None:
...
-        assert e.code == HTTP_STATUS_FORBIDDEN
-
-    retry_strategy = Retry(
-        total=3,
-        status_forcelist=STATUS_FORCELIST,
-        allowed_methods=["GET", "POST"],
-        backoff_factor=3,
+def test_invalid_token(client: Session, endpoint_url: str, fake_token: str) -> None:
+    response = client.get(
+        endpoint_url, headers={"Authorization": f"Bearer {fake_token}"}
```
glasnt, line 209: Further: if you want to test a bad token, maybe just test if "Bearer i-am-not-a-real-token" fails?

run/service-auth/data/privatekey.pem
eapl-gemugami: No worries! I thought the same.

About creating a fake token, I was also thinking of creating an RSA private key on the fly, to avoid relying on an external file, although it would require another dependency (cryptography, OpenSSH, pycryptodome, etc.).

For example:
```
key = rsa.generate_private_key(
    backend=crypto_default_backend(),
    public_exponent=65537,
    key_size=2048
)

private_key = key.private_bytes(
    crypto_serialization.Encoding.PEM,
    crypto_serialization.PrivateFormat.PKCS8,
    crypto_serialization.NoEncryption()
)
```

Source: https://stackoverflow.com/a/39126754

run/service-auth/receive_test.py
eapl-gemugami: It should fail, I'll try that approach.

eapl-gemugami:
@glasnt I've just committed the approach of using `Bearer i-am-not-a-real-token`.
I don't think that for this sample we need to cover a huge range of real issues found while validating a token, although I wanted to be a bit more explicit in the sample to cover [samples-style-guide/#errors](https://googlecloudplatform.github.io/samples-style-guide/#errors) with a simple try/except in [receive.py#L50-L54](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/3c454d76bcf8b02b07e68f3a9faa6aabc5d40016/run/service-auth/receive.py#L50-L54)

[app.py#L31-L37](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/3c454d76bcf8b02b07e68f3a9faa6aabc5d40016/run/service-auth/app.py#L31-L37) (flask view outside the sample) was adjusted to support this case.

glasnt (approved):
I believe this is now all correct

PR #13358, opened 2025-05-08
Fix type annotations

GameRoMan:


iennae (approved):
Thanks for contributing to improving our samples! We appreciate it.

iennae:
The Python 3.9 failure is expected, the Python 3.13 build failure has 

TypeError: unsupported operand type(s) for |: 'type' and 'tuple'

so it looks like this needs some improvements.

PR #13473, opened 2025-06-30
docs: add region tag around run/mcp-server/test_server.py

jackwotherspoon:
Add missing region tag around test code (needed for sample in docs)
and update Dockerfile region tag to not include newline.

### DIFF
--- a/run/mcp-server/test_server.py
+++ b/run/mcp-server/test_server.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# [START cloudrun_test_mcpserver]
 import asyncio
 
 from fastmcp import Client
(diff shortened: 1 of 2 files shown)


run/mcp-server/test_server.py
```diff
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# [START cloudrun_test_mcpserver]
```
glasnt: I would suggest since this is a new tag and a simple change, we take the time to make sure the region tag matches the other tag prefix for this sample. 

```suggestion
# [START cloudrun_mcpserver_test]
```

run/mcp-server/test_server.py
```diff
@@ -35,3 +36,4 @@ async def test_server():
 
 if __name__ == "__main__":
     asyncio.run(test_server())
+# [END cloudrun_test_mcpserver]
```
glasnt:
```suggestion
# [END cloudrun_mcpserver_test]
```

PR #13481, opened 2025-07-01
Add Github Runner with Cloud Run Worker Pools to Cloud Run samples

sagarrandive:
Add Github Runner with Cloud Run Worker Pools to Cloud Run samples. This sample also includes autoscaler for worker pool GH runner using Cloud Run Function


glasnt:
Hi @sagarrandive. This sample might be a bit too complex for this repository. I'll ping you internally on specifics.

run/github-runner/Dockerfile
```diff
@@ -0,0 +1,13 @@
+# syntax=docker/dockerfile:1
+# Best practices: https://docs.docker.com/build/building/best-practices/
+
+FROM ghcr.io/actions/actions-runner:2.322.0
+
+# Add scripts with right permissions.
+USER root
```
steren, line 7: Seems overkill to run as root just to set a file as executable, I don't think you need to. Can you ask around how to solve that without root?

run/github-runner/github-runner-autoscaler/main.py
```diff
@@ -0,0 +1,254 @@
...
+from google.auth.transport.requests import Request as GoogleRequest
+import google.auth
+from flask import Request
+from google.cloud import secretmanager
+from github import Github
+
+
+# --- Configuration ---
+PROJECT_ID = os.environ.get('GCP_PROJECT')
+LOCATION = 'us-central1' # Or your desired region
```
steren, line 31: Can you retrieve this from metadata server?
See https://github.com/GoogleCloudPlatform/cloud-run-hello/blob/master/hello.go#L121

run/github-runner/github-runner-autoscaler/main.py
```diff
@@ -0,0 +1,254 @@
...
+from google.oauth2 import service_account
+from google.auth.transport.requests import Request as GoogleRequest
+import google.auth
+from flask import Request
+from google.cloud import secretmanager
+from github import Github
+
+
+# --- Configuration ---
+PROJECT_ID = os.environ.get('GCP_PROJECT')
```
steren, line 30: Same here, retrieve from metadata server: https://github.com/GoogleCloudPlatform/cloud-run-hello/blob/master/hello.go#L105

run/github-runner/github-runner-autoscaler/main.py
```diff
@@ -0,0 +1,254 @@
...
+   Retrieves the current manualInstanceCount of the Cloud Run worker pool.
+   Returns the instance count as an integer, or -1 if retrieval fails.
+   """
+   auth_req, access_token = get_authenticated_request()
+   if not access_token:
+       logging.error("Failed to retrieve Google Cloud access token to get current instance count.")
+       return -1
+
+
+   url = f"https://run.googleapis.com/v2/projects/{PROJECT_ID}/locations/{LOCATION}/workerPools/{CLOUD_RUN_WORKER_POOL_NAME}"
```
steren, line 86: Have you considered using the Cloud Run Python client library instead?
This works, but client library might be more robust.

glasnt:
b/431731056

run/github-runner/github-runner-autoscaler/main.py
glasnt: It doesn't look like this is supported in the SDK yet. The replacement PR will address this further

PR #13672, opened 2025-12-17
Update quickstart samples for python 3.14

ShellyAgg:
## Description

Updating samples for python 3.14


glasnt:
The `appengine/standard_python3/building-an-app/building-an-app-1` sample is failing with the following error: 

```
ERROR main_test.py - AttributeError: module 'ast' has no attribute 'Str'
```

This is probably due to using an older version of PyTest (7.0.1) that does not support Python 3.14. Since this PR is based on a branch, I will update this. 

If there are further failures, we will need to confirm which versions of Python are being used in this teset, and ensure we have a valid version of PyTest for each tested version (it might be 3.9 and 3.14, it depends which noxfile covers this sample)

glasnt:
The Python 2.7 check failure is unrelated to this check, as it is configured to be skipped for these samples.

glasnt (approved):

PR #14512, opened 2026-08-14
 feat(genai/embeddings): migrate lower dimension embeddings to genai

XrossFox:
## Description

 Migrates the sample for generating multimodal embeddings with lower dimensionality to use the new `google-genai` SDK and adds the
 corresponding test.

Fixes b/546128628

### Testing
- [ ] **I have tested this change on a live environment and verified it works as intended.**

### Compliance & Style
- [ ] I have followed the [Sample Guidelines](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/main/AUTHORING_GUIDE.md).
- [ ] The README is updated with [all relevant information](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/main/AUTHORING_GUIDE.md#readme-file) (if applicable).
- [ ] **If this PR adds a new sample directory:** I updated the [CODEOWNERS file](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/main/.github/CODEOWNERS).

---

## Post-Approval Actions


XrossFox:
Tested locally:

image

image

genai/embeddings/generate_embeddings_with_lower_dimension.py
```diff
@@ -0,0 +1,67 @@
...
+        project=PROJECT_ID,
+        location=LOCATION_ID,
+    )
+
+    image_part = genai.types.Part.from_uri(
+        file_uri=IMAGE_URI,
+        mime_type="image/png",
+    )
+
+    text_part = genai.types.Part.from_text(text=CONTEXTUAL_TEXT)
```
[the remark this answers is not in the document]
XrossFox: Tested locally, it does.

genai/embeddings/generate_embeddings_with_lower_dimension.py
```diff
@@ -0,0 +1,67 @@
...
+    """Generates multimodal embeddings (image + text) with custom lower dimensionality
+
+    using the modern google-genai SDK.
+    """
+
+    client = genai.Client(
+        enterprise=True,
+        project=PROJECT_ID,
+        location=LOCATION_ID,
+    )
```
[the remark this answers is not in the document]
XrossFox: It does not, tested locally.

genai/embeddings/generate_embeddings_with_lower_dimension.py
```diff
@@ -0,0 +1,67 @@
...
+
+    config = genai.types.EmbedContentConfig(output_dimensionality=EMBEDDING_DIMENSION)
+
+    response = client.models.embed_content(
+        model=EMBEDDING_MODEL,
+        contents=[contents],
+        config=config,
+    )
+
+    embeddings = response.embeddings[0].values
```
Kef131: Here we assumes array length > 0. If the API returns a successful but empty structure, this will raise an unhandled `IndexError`.
Feasible to add a guard clause?` if not response.embeddings:`?

XrossFox: As i understand, the proto file for this endpoints defines that the response has to be at least a ContentEmbedding object, and it shouldnt be possible to be empty:

```
// The response to an `EmbedContentRequest`.
message EmbedContentResponse {
  // Output only. The embedding generated from the input content.
  ContentEmbedding embedding = 1 [(google.api.field_behavior) = OUTPUT_ONLY];
}
```

Kef131: The [proto](google3/third_party/googleapis/stable/google/ai/generativelanguage/v1beta/generative_service.proto) file as you referenced, define this as a singular object however, 
```
class EmbedContentResponse(_common.BaseModel):
  """Response for the embed_content method."""

  sdk_http_response: Optional[HttpResponse] = Field(
      default=None, description="""Used to retain the full HTTP response."""
  )
  embeddings: Optional[list[ContentEmbedding]] = Field(
      default=None,
      description="""The embeddings for each request, in the same order as provided in
      the batch request.
      """,
  )
  metadata: Optional[EmbedContentMetadata] = Field(
      default=None,
      description="""Vertex API only. Metadata about the request.
      """,
  )
```

- because it is list, this translates the Singular Proto field into a  List (embeddings: list[...]). If we try to grab the first item [0] of an empty list [], will immediately crash with an `IndexError.`
- it is an Optional, which can be `None`, if the API has an issue, response.embeddings might be `None`. 

As mentioned before, this is a defensive coding comment, optional to add and if it is not worth it, is ok to leave it like that.

Kef131 (approved):

genai/embeddings/generate_embeddings_with_lower_dimension.py
XrossFox: Done. Just remember, that if there is an error on the server side, it should never return a 200 ok, which would raise an exception via the genai SDK.

amcolin (approved):

